PR #52, opened 2015-01-05
Support for psi games and the future perfect.

adam-higuera:
Important additions:

Accessing has been reworked so that access effects (like psi games) that require input from one or both players will cause the cards remaining to be accessed to be pushed into a queue until all of the necessary input has been secured.  To facilitate this, two new keys have been added to the game state, `:pending-accesses` and `cards-accessed`.  The former holds a list of cards that will be accessed after the current access is resolved, the latter is used to correctly print a message stating which card (#1, #2, etc) the runner is currently accessing when processing of the queue begins again.

These kinds of effects are created with the macro `effect-with-pending-input`, which works the same way as `effect`, except that it causes `(:pending-input @state)` (which the new access system checks in order to ensure that it's OK to proceed with accessing) to return true when the ability resolves without a target (setting the relevant user's prompt and setting the other user's prompt to indicate waiting for input) and causing it to return nil when the ability resolves with a target (which happens when the relevant user clicks an option in his/her prompt).

Note that access abilities are responsible for restarting the access queue, so any access ability created with `effect-with-pending-input` must end with a call to

`(access-queue (:pending-accesses @state)
                               (+ 1 (:cards-accessed @state)))`

to ensure that accessing cards resumes after the input-dependent access effect is resolved.

Psi games have been implemented using the new `effect-with-pending-input` infrastructure.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -14,6 +14,8 @@
 
   :source-paths ["src/node" "src/cljs"]
 
+  :java-cmd "/Library/Java/JavaVirtualMachines/jdk1.8.0_25.jdk/Contents/Home/bin/java"
+
   :cljsbuild {
     :builds [
       {:id "node"
--- a/src/node/game/cards.cljs
+++ b/src/node/game/cards.cljs
@@ -1183,15 +1183,15 @@
              (psi-game
               {:ability-corp-win
                {:msg "The Future Perfect's ability prevents it from being stolen."
-                :effect (effect
-                 (deregister-pending-input card nil)
-                 (access-queue (:pending-accesses @state)))}
+                :effect (effect                 
+                         (access-queue (:pending-accesses @state)
+                                       (+ 1 (:cards-accessed @state))))}
                :ability-runner-win
                {:msg "The Future Perfect's ability does not prevent it from being stolen."
                 :effect (effect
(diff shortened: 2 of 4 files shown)


adam-higuera:
By the way, I've also implemented The Future Perfect, Cerebral Cast, Caprice Nisei, and Psychic Field.

mtgred:
I was already working on Psi implementation so unfortunately it will be difficult to merge your changes. We can perhaps have a talk about tasks in the pipeline to avoid work duplication in the future.

A few remarks:
- Please add `npm-debug.log` your global `.gitignore`
- Please remove debug `console.log` before committing
- I think your commits are somehow messed up. Eg. In the commit "small change to future perfect" there is no code related to Future Perfect, it looks more like a merge from the master branch.

adam-higuera:
Sorry, I didn't see this and assumed (wrongly) that github would notify me of activity on this pull request.

I doubt it makes sense to try and merge these changes.  However, it wasn't wasted effort; I understand the architecture pretty well at this point, and I believe I could be helpful in implementing some game mechanics or other features.

I'm not really a github pro (as you can tell by the cruft I accidentally added to your repo), so I don't know if there's a PM system or something, but we should get in touch via e-mail.

PR #76, opened 2015-01-26
Added Priority Requisition

Smashman:


mtgred:
Hi,

With the current state of functionalities I think it's not very user friendly to implement Priority Requisition. The issue is when there are multiple unrezzed ICEs with the same name on the board. There is no way to distinguish them in the menu at the moment.

I'm currently working on a targeting functionality to select cards with the mouse. It would be better to wait for that functionality to be ready before implementing Priority Requisition.

Smashman:
I understand. Perhaps I could add the zone to the selection box as a stop gap?

mtgred:
The targeting functionality has been implemented and Priority Requisition has been automated.

PR #110, opened 2015-02-19
Elizabeth Mills

JoelCFC25:
Attempt to piece together based on Reversed Accounts, Foxfire, Ronin, etc

### DIFF
--- a/src/node/game/cards.cljs
+++ b/src/node/game/cards.cljs
@@ -424,12 +424,12 @@
     :leave-play #(remove-watch % :ekomind)}
    
    "Elizabeth Mills"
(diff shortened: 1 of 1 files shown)


mtgred:
The code is almost ok.

1. Can you can replace `(system-msg ...)` with `:msg "remove 1 bad publicity` ?
2. The Click cost has be at the same level as the rest of the ability.

JoelCFC25:
Does it look OK now? At least people might conceivably use this card as opposed to my other feeble attempt, heh.

mtgred:
1. The `:msg` part is used to generate the log message "<username> uses <cardname> to <msg>". It means `:msg` should be "remove 1 bad publicity", that is without "s" at the end of "remove".

2. I think there are extra white space on line 425.

PR #120, opened 2015-02-22
Fetch card images directly to where they'll be referenced by local server

rkbodenner:
All in the name of a seamless setup process.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,6 @@ Launch MongoDB and fetch card data:
 ```
 $ mongod
 $ cd data
-$ mkdir img
 $ coffee fetch.coffee
 ```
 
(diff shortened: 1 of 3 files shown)


phette23:
:+1:  yay! I tried this and it worked on my install. Only downside to running locally was missing out on the images so this is just great.

One comment: why not specify `"mkdirp": "~0.5.0"` in package.json instead of "*"? I realize mkdirp is unlikely to change its API but it's still best practice to give a specific version rather. There are already a couple "*" modules in async and jade that should probably be specified. My $0.02.

rkbodenner:
@phette23 Glad it worked for ya! Fixed that version spec, and will update `async` in another PR with some other related cleanup.

PR #123, opened 2015-02-23
Zaibatsu Loyalty

JoelCFC25:
Modeled after Feedback Filter (clicking automatically pays, manual trashing option). Have another way in my patch-5 branch that does it with a prompt like Sealed Vault.


mtgred:
The text that appears in the menu and in the log will be the same for both abilities.
Could please update the `:msg` to `"pay 1 [Credits] and prevent ..."` and `"trash it and prevent ..."` ?

PR #127, opened 2015-02-24
Fixed Crash Space (and other cards using recurring credits) menu item text

### LINKED ISSUE #125: Crash Space has a blank menu option
Looks like Crash Space has a blank menu option for using the credit on it for removing tags. Probably a general bug related to all token/credit-removal options, though.

Agnate:
This fixes #125 where Crash Space has a blank menu option for using the recurring credit on it. I've added the text to show the appropriate text to use the recurring credit. This fix will extend to all cards using recurring credits.

### DIFF
--- a/src/node/game/core.cljs
+++ b/src/node/game/core.cljs
@@ -297,7 +297,7 @@
                     (conj (:abilities cdef) "Use [Recurring Credits]")
                     (:abilities cdef))
         abilities (for [ab abilities]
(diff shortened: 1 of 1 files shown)


mtgred:
Thanks for reporting the issue and submitting the pull request. Your change will add a recurring credit label on cards without recurring credit if they don't have a proper label which can be an issue. I just pushed a fix to issue #125 so everything should be ok on the next server update.

PR #160, opened 2015-04-17
Add weasel dependency

mike529:
Building the client side js files fails without this dependency being added.


mtgred:
Hi, thanks for reporting this issue. Weasel is a dev dependency. The project.clj file has been updated with a :dev profile including Weasel as a dependency.

PR #184, opened 2015-05-06
Implement Security Testing; implement Mother Goddess dynamic subtyping

nealterrell:
This adds automated support for Security Testing. I have not worked with Clojure before, so comments on style are welcome.

When turn begins, S.T. prompts to choose a server. After the prompt, the card's state is updated with key :security-testing-target associated with a vector holding the target server. 

On successful run, :once :per-turn, S.T. checks its state to see if the :run's :server is the same as its :security-testing-target value. If so, we change the game state to override the run's :replace-access with a new function that gains 2 credits with no access. 

I have tested the following scenarios, beyond the simple run-on-S.T.-server:

1. Any event that triggers a run on a S.T. server (Indexing, Legwork, Account Siphon, etc.): card's instead-of-access is ignored, gain 2 credits, no option on whether to access (Siphon, Indexing).
2. Multiple S.T. on the same server: the first run on the server gains 2 credits with no access. Subsequent runs gain access.
3. Multiple S.T. on different servers: the first run on each server gains 2 credits with no access.
4. Bonus effects from other cards on successful run (Dirty Laundry, Desperado, John Masanori) still trigger.

I have not tested:

1. S.T. on HQ, Sneakdoor Beta. Expected: S.T. triggers after Sneakdoor, gain 2 credits no access. Actual: ???
2. S.T. on Archives, Sneakdoor Beta: Expected: Sneakdoor switches to H.Q., S.T. does not trigger. Actual: ???

(My instinct is that the Sneakdoor Beta interaction will depend on what order the successful-run events are called on the two cards.)

Known issues:

1. S.T. on Jackson Howard, J.H. removed from game before run is complete. No credits gained (EXPECTED). However, the game then "shifts" remove servers down to fill the hole left by Jackson. If you run the "new" server with the same number that Jackson had, S.T. will trigger. I don't know how to resolve, as I don't see a way of handling a "run is over because server disappeared" event. Same thing would happen with Self Destruct.

### DIFF
--- a/src/clj/game/cards.clj
+++ b/src/clj/game/cards.clj
@@ -2547,7 +2547,55 @@
                  :effect (req (corp-install state side target (:server run)))}]}
 
    "Mother Goddess"
-   {:abilities [{:msg "end the run" :effect (effect (end-run))}]}
+   {:abilities [{:msg "end the run" :effect (effect (end-run))}]
+    :events {
+      :rez {
+         :req (req (= (:type target) "ICE")) 
+         :effect (req (let [allice (->> (flatten (seq (:servers corp))) (map #(:ices %)) (filter #(not (nil? %))) flatten
+                              (filter #(and (:rezzed %) (not= (:cid card) (:cid %)))))
+                            allsubtypes (->> allice 
+                              (map #(vec (.split (:subtype %) " - "))) flatten)
+                            newtypes (clojure.string/join " - " (distinct (cons "Mythic" allsubtypes)))]
+            (resolve-ability state side {
+               :msg (msg "gain subtypes " newtypes)
+               :effect (req(let [c (assoc card :subtype newtypes)]
+                  (update! state side c)
+               ))
+            } card nil)
+         ))
+      }
+      :trash {
+         :req (req (= (:type target) "ICE")) 
+         :effect (req (let [allice (->> (flatten (seq (:servers corp))) (map #(:ices %)) (filter #(not (nil? %))) flatten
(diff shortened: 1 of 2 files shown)


nealterrell:
I have also added dynamic sub-typing for Mother Goddess. This generally is not needed except for interaction with NEXT Gold's "do 1 next damage for each rezzed NEXT ice" subroutine. 

I could have taken the easy way out and just added a manual check in NEXT Gold to also count a rezzed Mother Goddess, but I decided to do this the real way: any time an ICE is rezzed, trashed, or derezed, Mother Goddess will get the subtypes of all installed rezzed ICE and append them to its own. This automatically fixes the NEXT Gold interaction.

I had to add a :derez event to the core to allow Mother Goddess to rebuild its subtypes list when an ice is derezed.

mtgred:
I'll test your code tonight but it seems to be ok so good job. :+1: Those cards are not easy to implement. Also great description !

Regarding the code style, you can perhaps use LightTable, Emacs or IntelliJ + Cursive plugin to indent Clojure code properly. And instead of `(get (get-card state card) :security-testing-target)` it's more idiomatic to write `(:security-testing-target (get-card state card)` which is equivalent.

JoelCFC25:
Awesome contribution, Neal!

nealterrell:
Thanks @mtgred. Can you smell my OOP background? ;) 

I see some bloody Tabs snuck in there... I will be more careful with that in the future. I am looking forward to contributing more :).

PR #185, opened 2015-05-08
add Comet's +1 MU

JoelCFC25:
Comet was not adding 1 MU on install. Also fixed Paintbrush message.

### DIFF
--- a/src/clj/game/cards.clj
+++ b/src/clj/game/cards.clj
@@ -317,7 +317,8 @@
                    :effect (effect (draw :runner))}}}
 
    "Comet"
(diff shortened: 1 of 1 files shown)


mtgred:
:+1:

nealterrell:
Is it just me, or is this a breaking change?

"Comet"
-   {:events {:play-event
+   (:effect  

Isn't it supposed to be {:effect ?

JoelCFC25:
Yep, I apologize. I could've sworn that was a brace--it was the first time wrote this up, then I undid some other things and did it over. That's where the error came in.

PR #190, opened 2015-05-10
Implement Trick of Light.

nealterrell:
My implementation for Trick of Light. Based on code from Oversight AI (selecting an appropriate target) and Shipment from SanSan (choosing # of tokens, selecting destination). 

While implementing this card, I noticed a minor bug #191 in Shipment from Kaguya, Shipment from SanSan, and AstroScript Pilot Program: they can all place tokens on cards that can only be advanced while rezzed, like Tyrant or Woodcutter. (I doubt anyone has ever in the history of Netrunner used an Astro token to advance an unrezzed Woodcutter, but the code does allow it :P.) I addressed this issue in Trick of Light with the following req filter:

```(or (and (:advanceable %) (or (= (:advanceable %) "always") (:rezzed %)))
                                                                         (= (:type %) "Agenda"))```

requiring the destination to be `:advanceable :always` or `:advanceable` AND `:rezzed` or an agenda. If you like this implementation, I will add it to the other cards mentioned above.

### DIFF
--- a/src/clj/game/cards.clj
+++ b/src/clj/game/cards.clj
@@ -491,7 +491,7 @@
    "Dorm Computer"
    {:data {:counter 4}
     :abilities [{:counter-cost 1 :cost [:click 1]
-                 :prompt "Choose a server" :choices (req servers) 
+                 :prompt "Choose a server" :choices (req servers)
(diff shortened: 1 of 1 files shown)


mtgred:
:+1: 

Go ahead with the other cards :)

Issue #191, opened 2015-05-10
Placing advancement tokens on Tyrant, Woodcutter, and Salvage


nealterrell:
Cards that are marked advanceable-while-rezzed can be the target of Shipment from SanSan, Shipment from Kaguya, and AstroScript Pilot Program's token. I am filing this here just as a reminder to fix it later :).

PR #192, opened 2015-05-11
Implement Scavenge; fix placing tokens on :advanceable :while-rezzed

nealterrell:
My implementation for Scavenge, based on code for Test Run (select Heap or Grip to install from) and Modded (reduce cost of install by gaining credits first). Also fixes issue #191 by applying filter from Trick of Light to Shipment from Kaguya/SanSan and AstroScript Pilot Program.

Reasonably straightforward implementation. Feel free to tweak the messaging; right now, the players see:

> nealterrell plays Scavenge.
> nealterrell trashes Femme Fatale to install a card from Heap.
> nealterrell installs Femme Fatale

when using Scavenge to trash and reinstall Femme.

Also, the filter for choosing which card to install from Grip/Heap will only show programs that the Runner can afford to install with the Scavenge discount. Other cards like SMC don't add this filter, so feel free to remove it. If the filter is not there, the Runner can choose a program they cannot afford, but the install fails and nothing happens (as expected).

I have tested the expected interactions: Scavenge a Lady to refresh tokens; Scavenge a program played via Test Run to prevent it from going back on the Stack; Scavenge a Cache to refresh tokens.

### DIFF
--- a/src/clj/game/cards.clj
+++ b/src/clj/game/cards.clj
@@ -156,7 +156,7 @@
    "Blue Sun: Powering the Future"
    {:abilities [{:msg (msg "add " (:title target) " to HQ and gain " (:cost target) " [Credits]")
                  :choices {:req #(:rezzed %)}
-                 :effect (effect (gain :credit (:cost target)) (move (assoc target :rezzed false) :hand))}]}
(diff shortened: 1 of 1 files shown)


mtgred:
:+1:

PR #193, opened 2015-05-11
fix Comet message

JoelCFC25:
Comet was incorrectly using "install" in the message (it allows the play of events). Also fixed two small typos. 

I promise I'm not breaking Comet this time...


JoelCFC25:
Also added a fix for Manhunt's base trace.

mtgred:
:+1:

PR #196, opened 2015-05-17
Implement Hivemind

nealterrell:
This submission implements Hivemind and its interactions with existing virus programs.

### Core changes

1. Added a function get-virus-counters which returns the number of virus counters on target card, PLUS the number of counters on an installed Hivemind, if any. In effect, the function counts how many virus counters a card should "see".
2. Added support for a `:priority true` key for prompts in effects. If an effect has `:priority true` along with `:prompt`, the prompt shown will be placed at the beginning of the prompt queue instead of the end, forcing the player to resolve it immediately. Needed for Imp support.

### Hivemind implementation
1. Starts with 1 virus counter.
2. Its mere presence will cause some virus cards to "see" that they have extra counters for their effects (see below).
3. Has a single ability to transfer a virus token and immediately trigger an ability on another virus card.
  1. With `:priority`, a token is transferred from Hivemind to a target virus.
  2. If that target has only a single ability, its ability is immediately triggered. 
  3. If that target has multiple abilities, a prompt allows the user to select which to trigger. This functionality is currently unused, since all viruses have only a single ability.
  4. The target ability is run through resolve-ability, meaning any costs or requirements attached to the ability will be enforced. 

### Card updates
1. Medium: grants a token immediately upon successful run, before access. Grants an access-bonus equal to 1 less than its get-virus-count.
2. Nerve Agent: likewise.
3. Gorman Drip: gives 1 credit per get-virus-count.
4. Deep Thought: grants the reveal if get-virus-count is at least 3.

### Other viruses
Going through [this list](http://netrunnerdb.com/find/?q=s%3Avirus&sort=name&view=list&_locale=en)...

1. Cache: Hivemind can trigger.
2. Chakana: not implemented.
3. Clot: N/A.
4. Crypsis: did not fully test. Triggering with Hivemind will probably give it a token, but then trigger the Break 1 Subroutine ability, costing the runner 1cr. Credits can be adjusted manually by runner, and the transferred token can then be sacrificed to keep Crypsis alive.
5. Datasucker: untested, should consume counter and print message.
6. Gravedigger: untested, should consume and mill.
7. Imp: Hivemind can trigger.
8. Incubator: broken; if targeted by Hivemind, will transfer a token to Incubator and then do nothing. In reality, Hivemind should not be able to target Incubator. Will fix.
9. Ixodidae: N/A.
10. Lamprey: N/A.
11. Parasite: has to be manually tracked.
12. Pheromones: not implemented.

### DIFF
--- a/src/clj/game/cards.clj
+++ b/src/clj/game/cards.clj
@@ -704,8 +704,8 @@
    {:abilities [{:cost [:click 2] :effect (effect (gain :credit 3)) :msg "gain 3 [Credits]"}]}
 
    "Gorman Drip v1"
-   {:abilities [{:cost [:click 1] :effect (effect (gain :credit (:counter card)) (trash card))
-                 :msg (msg "gain " (:counter card) " [Credits]")}]
+   {:abilities [{:cost [:click 1] :effect (effect (gain :credit (get-virus-counters state side card)) (trash card))
+                 :msg (msg "gain " (get-virus-counters state side card) " [Credits]")}]
     :events {:corp-click-credit {:effect (effect (add-prop :runner card :counter 1))}
              :corp-click-draw {:effect (effect (add-prop :runner card :counter 1))}}}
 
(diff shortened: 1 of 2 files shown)


mtgred:
Hivemind is super tricky to implement so good job !

I found an issue with Imp. If Imp has already been used in a turn, using Imp again through Hivemind transfert 1 counter and remove the access prompt although nothing should happen. Add this edge case to your todo list. :)

Otherwise well done ! :+1:

nealterrell:
Yes, it was a very interesting card to work on :).

I will look into the Imp issue. I am learning more about how the general game flow works and believe I can get Hivemind to check the target's `:req` function before transferring the token.

Issue #197, opened 2015-05-18
Dev question: forcing synchronous prompts


nealterrell:
I am fooling around with damage increase/prevention hookups to allow interactions for The Cleaners, Plascrete Carapace, etc. My first attempt has been to add an event to the `(damage)` effect, BEFORE the damage is dealt, allowing event listeners to set a state variable to increase or decrease damage by a particular amount. This works well for The Cleaners: when the `:pre-damage` event fires, The Cleaners calls my new `(damage-bonus)` function to set a state variable increasing `:meat` damage by 1. The `(damage)` function uses this state variable to deal +1 meat damage from any source. Success! 

To implement Plascrete Carapace or any other pay-to-reduce card, I thought to listen for the `:pre-damage` event, and respond by showing the runner a prompt to select how much damage to absorb with the carapace. My first attempt showed the prompt to the Corp; whoops. The second attempt shows the prompt to the Runner, but the `(damage)` routine (triggered by the meat damage card, e.g., Scorched Earth) continues without waiting for the prompt to resolve. I am guessing this is because the prompt is run asynchronously; the `:pre-damage` event fires, Plascrete's listener shows the prompt and returns back to the damage function, which deals the 4 damage before the Runner can select the prevention amount.

@mtgred , am I missing any tech here that would help with this implementation? Is there a way to force a function which shows a prompt to NOT return until the prompt resolves? If not, do you have any other thoughts on how to handle damage prevention? My only other idea is to show a "Continue" button to the Corp each time damage is done, giving a clear window for the Runner to spend Plascrete Tokens, trigger Feedback Filter, etc.

JoelCFC25:
I don't have anything substantive other than to say this would be a *tremendous* improvement! Keep up the great work, Neal!!

mtgred:
There is currently no way to stop the execution of card abilities. A new interruption mechanism needs to be added.

What I had in mind for damage prevention and other type of prevention (eg. Zaibatsu Loyalty) is to keep a`:prevent` hashmap at the root the `game-state`. That hashmap would keep track of the number cards in play that can prevent certain type of event (eg: `:meat` `:net` `:expose`). When such an event occurs, the `:prevent` hashmap is checked to see of there are cards that can prevent it. In case there are, it prompts the player to use any prevent abilities he whishes before resolving the other ability.

[...]

PR #201, opened 2015-05-21
Implement Shoot the Moon

nealterrell:
Not the most pressing card to implement, but it was an easy job while waiting to hear about Plascrete Carapace in #197.

To support Shoot the Moon, I expanded the capabilities of `:choices` prompts so that the `:max` field can be a function instead of only a constant. The function will be evaluated in `(show-select)` and the result used as the constant for the maximum number of cards that can be selected. This allows Shoot the Moon to use `(:tag runner)` as the max number of ICE to be selected.

The only other detail of interest is that Shoot the Moon will automatically rez all ICE if the number of unrezzed ICE is <= the number of tags. Otherwise the Corp is given the prompt to choose the ICE to rez.

### DIFF
--- a/src/clj/game/cards.clj
+++ b/src/clj/game/cards.clj
@@ -1758,6 +1758,21 @@
    "Shock!"
    {:access {:msg "do 1 net damage" :effect (effect (damage :net 1))}}
 
+   "Shoot the Moon"
+   {:effect
+    (req (if (< (:tag runner)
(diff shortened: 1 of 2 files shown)


mtgred:
:+1: 

If you feel like, you can also update the trace mechanism with the optional function for the trace base strength. It would unlock Self Destruct and Searchlight

nealterrell:
Will do!

PR #206, opened 2015-05-22
Implement Self-destruct, Searchlight

nealterrell:
Implement dynamic trace strength for Self-destruct and Searchlight. `:trace :base` can now be a function instead of only a constant.

I also added a core function `card->server` to take a card and give back a server for the card's `:zone`. The server is returned as a map with `:content` and `:ices` keys. Self-destruct uses the code to get all cards "in or protecting this server". There may be a better way of implementing this, but I couldn't find any... the closest I saw was Singularity, but that uses a prompt to select the server, instead of the choice being implicit based on the card being used.

### DIFF
--- a/src/clj/game/cards.clj
+++ b/src/clj/game/cards.clj
@@ -1687,7 +1687,8 @@
                                                        :effect (effect (gain :credit 2))} st nil))})))}}}
 
    "Self-destruct"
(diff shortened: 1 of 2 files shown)


mtgred:
:+1:

PR #211, opened 2015-05-29
Implement damage prevention and boosting cards

nealterrell:
This adds a framework for preventing and boosting damage effects, and implements the most common related cards.

In summary, if a card that can prevent damage is installed when the runner suffers some sort of damage, a prompt will show to give the runner opportunity to prevent the damage. After spending tokens/credits/whatever, any unprevented damage is then applied.

### Core changes

0. Added state values (`:damage :prevent`) to store references to cards that can prevent damage. In a card's definition, the `:prevent` field can be used to signify that the card can potentially prevent some types of damage, specified by a vector, as in `:prevent [:net :brain]` or `:prevent [:meat]`.
1. Added functions `(damage-bonus)` and `(damage-prevent)` to core. Given a type and an amount, these functions will boost or prevent the next source of damage of that type, by updating state values `:damage :damage-bonus` and `:damage :damage-prevent` respectively.
2. Added an event `:pre-damage`, which fires in the `(damage)` function before the damage is applied. Cards can handle the event to automatically boost or prevent damage without user interaction (see below).
3. Added a function `(damage-count)` which calculates total damage to be applied after bonus and prevent effects.
4. Split `(damage)` into two functions. 
  1. `(damage)` itself fires the `:pre-damage` event, and then checks `:damage :prevent` for any cards that can prevent the damage. If there are cards, a prompt is shown to the runner to activate any cards he/she wishes, and then finish by clicking the "Done" button. After Done is clicked, or if there are no cards that can prevent, `(resolve-damage)` is called.
  2. `(resolve-damage)` applies the damage appropriately. By separating this into another function, we can tie its execution to the completion of the runner's damage prevention prompt.
5. Damage can be "unpreventable" by passing `:unpreventable` to `(damage)`.

### Card updates
Going through [this list](http://netrunnerdb.com/find/?q=x%3A%22prevent%22+x%3Adamage&sort=name&view=list&_locale=en)...

0. The Cleaners: boosts all sources of meat damage by 1. See Known Issues below.
1. Amped Up: damage is `:unpreventable`.
2. Chrome Parlor: not implemented. Not sure if it will work until `:pre-damage` also passes a card that is the source of the damage.
3. Crash Space: `:prevent [:meat]`
4. Deus Ex: `:prevent [:net]` (see Self-modifying Code below)
5. Feedback Filter: `:prevent [:net :brain]`
6. Flare: damage is `:unpreventable`
7. Leverage: not implemented, but should be doable.
8. Monolith: not implemented, should be doable.
9. Muresh Bodysuit: automatically (no user interaction) prevents only the first net damage.
10. Net Shield: can `:prevent [:net]` once per turn with user interaction. Technically can prevent ANY single net damage in the turn, not just the first overall, so this is a minor issue.
11. Plascrete Carapace: `:prevent [:meat]`
12. Sacrificial Clone: not implemented, should be doable.
13. Stim Dealer: damage is `:unpreventable`
14. Stimhack: damage is `:unpreventable`

### Miscellaneous

1. Self-modifying Code and Clone Chip have `:prevent [:net]` because the user can potentially use them to bring out Deus Ex. There are no programs to prevent meat or brain damage. If SMC or Clone Chip are installed, the runner will get the prompt to prevent any net damage (regardless if Deus Ex is in their deck or not).
2. I made the `:prompt`s for SMC and Clone Chip `:priority true`, so they will jump to the front of the prompt queue when used in the middle of a damage prevention prompt. Works great!

### Tests completed

1. Simple prevention: 
  1. Plascrete Carapace
  2. Feedback Filter (both net and brain damage)
  3. Crash Space (trash to prevent)
2. More complicated:
  1. Net Shield: works only once per turn.
  2. Muresh Bodysuit: works automatically on first net damage. Multiple copies still only prevent one damage (using `:once-key :muresh-bodysuit`)
  3. Deus Ex: can use Clone Chip or SMC to install Deus Ex and then use its ability to prevent all net damage. Very satisfying! :)

### Known issues

1. The Cleaners boosts the damage of Argus Security's meat damage effect, even though the cards are specifically worded to avoid this. (Argus says "suffer 2 meat damage", whereas Cleaners specifies "when you do meat damage".) To fix, need to add another parameter `:unboostable` to `(damage)` (maybe wrapping it into a map along with `:unpreventable`).

### DIFF
--- a/src/clj/game/cards.clj
+++ b/src/clj/game/cards.clj
@@ -294,6 +294,7 @@
    "Clone Chip"
    {:prevent [:net]
     :abilities [{:prompt "Choose a program to install" :msg (msg "install " (:title target))
+                 :priority true
                  :choices (req (filter #(and (has? % :type "Program")
                                              (<= (:cost %) (:credit runner))) (:discard runner)))
                  :effect (effect (trash card) (runner-install target))}]}
@@ -1381,7 +1382,7 @@
    "Plascrete Carapace"
    {:data [:counter 4]
     :prevent [:meat]
-    :abilities [{:counter-cost 1 :effect (effect (damage-prevent :meat 1))}]}
+    :abilities [{:counter-cost 1 :msg "prevent 1 meat damage" :effect (effect (damage-prevent :meat 1))}]}
 
    "Priority Requisition"
    {:choices {:req #(and (= (:type %) "ICE") (not (:rezzed %)))}
@@ -1747,6 +1748,7 @@
    "Self-modifying Code"
    {:prevent [:net]
     :abilities [{:prompt "Choose a program to install" :msg (msg "install " (:title target))
+                 :priority true
                  :choices (req (filter #(has? % :type "Program") (:deck runner)))
                  :cost [:credit 2]
                  :effect (effect (trash card) (runner-install target) (shuffle! :deck))}]}
--- a/src/clj/game/cards.clj
+++ b/src/clj/game/cards.clj
@@ -65,7 +65,7 @@
                  :msg "gain [Click][Click]"}]}
 
    "Amped Up"
-   {:effect (effect (gain :click 3) (damage :brain 1))}
+   {:effect (effect (gain :click 3) (damage :brain 1 :unpreventable))}
 
    "Andromeda: Dispossessed Ristie"
    {:effect (effect (gain :link 1) (draw 4)) :mulligan (effect (draw 4))}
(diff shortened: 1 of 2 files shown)


ZogStriP:
:heart_eyes: This might be the best PR I've ever seen :+1:

mtgred:
Indeed :)

nealterrell:
My pleasure. 

I'll be submitting the remaining logic in a little bit: Monolith, Sacrificial Clone, and Leverage (tricky, but I think I've got it); support for `:unboostable` damage (Argus meat damage, self-inflicted runner damage from Cybernetics); and some cleanup of the messages.

mtgred:
By the way, SMC doesn't prevent damage. When a damage ability is being resolved, there is no window to fetch and install Deus X or other prevention cards.

nealterrell:
Wow, TIL. I will push that into #212.

mtgred:
I played an opponent using Plascrete Carapace yesterday and notice a few things. The carapace is not trashed after its 4 counters are used. My opponent was still prompted for damage prevention although the carapace was empty. Two small things you can perhaps improve?

It was great to have Plascrete Carapace automated. Thanks again for your hard work.

nealterrell:
Submitted #236 which will trash Carapace (`desactivate` removes trashed cards from the prevent map).

PR #212, opened 2015-05-29
Finishing damage prevention

nealterrell:
A few fixes, and the last remaining damage prevention cards (I think?).

Turned the `:unpreventable` parameter of `(damage)` into a map that also supports `:unboostable` for damage that should NOT be boosted by corp effects (The Cleaners).  Applied `:unboostable` to Argus Security, Tri-Maf Contact, and cybernetics.

Implemented:

1. Monolith: trash a card with priority to prevent meat or brain damage.
2. Sacrificial Clone: prevent all damage.
3. Leverage: definitely the trickiest, but it works. Corp gets prompt to take 2 bad publicity or make runner immune to damage. If the latter is selected, we register temporary events to `(damage-prevent)` all damage of all types, plus an event to unregister our event on `:runner-turn-begins`. A small hack is necessary (putting empty `:runner-turn-begins` and `:pre-damage` events in Leverage's base definition) to get this to work.

### DIFF
--- a/src/clj/game/cards.clj
+++ b/src/clj/game/cards.clj
@@ -65,7 +65,7 @@
                  :msg "gain [Click][Click]"}]}
 
    "Amped Up"
-   {:effect (effect (gain :click 3) (damage :brain 1 :unpreventable))}
+   {:effect (effect (gain :click 3) (damage :brain 1 {:unpreventable true}))}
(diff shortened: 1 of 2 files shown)


mtgred:
Awesome :+1:

PR #217, opened 2015-06-02
Implement Oaktown Renovation.

nealterrell:
Implements Oaktown Renovation from Chrome City.

Adds a new tag `:install-rezzed true` which installs a corp card rezzed. I thought about doing this automatically for the "Public" sybtype, but I don't think we have confirmation on this being a consistent game mechanic, so I went with the tag instead. After advancing the card, the corp gains 2 or 3 credits depending on the number of counters now on the card.

I tested this by putting its code in The Cleaners, since the public card set doesn't include Chrome City yet.


mtgred:
:+1:

PR #218, opened 2015-06-02
Implement rez- and trash-cost increase/reduce effects

nealterrell:
I took the framework for damage bonus/prevention and applied it to Corp rez and trash costs.

### Core changes

1. Added functions `(trash-cost-bonus)` and `(rez-cost-bonus)` to increase the trash or rez cost of a corp card that is currently being accessed or rezzed, respectively. To decrease costs, a negative argument is used.
2. Added events `:pre-trash` and `:pre-rez`.
2. Modified `(rez)` and `(handle-access)` to use costs modified by the bonus functions. These functions fire the `:pre-X` events before calculating rez/trash costs, allowing cards to modify costs appropriately.

### Card updates
Rez costs:

1. Akitaro Watanabe: -2 rez cost for ice protecting his server.
2. Braintrust: -1 rez cost for all ice for each token on the scored agenda.
3. Xanadu: +1 rez cost for all ice.
4. Reina Roja: +1 rez cost for ice, `:once :per-turn`.
5. Brain-Taping Warehouse: -1 rez cost for `:subtype "Bioroid" ice for each `(:click runner)`.
6. Breaker Bay Grid: -5 rez cost for cards in the same server's `:content`.

Trash costs:

1. Encryption Protocol: +1 trash cost for installed cards.
2. Industrial Genomics: +1 trash cost for all cards in archives that are `(not (:seen))`.
3. Skulljack: -1 trash cost for all cards.
4. Oaktown Grid: +3 trash cost for all cards in the same server's `:content`.

### To-do

1. Rook: I don't know enough about hosting to get this guy working yet, but once it is hosted, it can listen to `:pre-rez`, check the target's server to see if it is the same as Rook's host's server, and if so, call `(rez-cost-bonus 2)`.
2. Running Interference: can use the framework from Leverage to register a `:pre-rez` event that only triggers until the end of the run. In `:pre-rez`, call `(rez-cost-bonus (:cost target))`.
3. IQ: I don't know if this one can be done, since I BELIEVE (?) installed cards won't receive events until they are rezzed

### DIFF
--- a/src/clj/game/cards.clj
+++ b/src/clj/game/cards.clj
@@ -192,6 +192,10 @@
    {:effect (effect (damage :brain 1) (gain :max-hand-size 3))
     :leave-play (effect (lose :max-hand-size 3))}
 
+   "Breaker Bay Grid"
+   {:events {:pre-rez {:req (req (= (:zone card) (:zone target)))
+                       :effect (effect (rez-cost-bonus 5))}}}
+
    "Breaking News"
    {:effect (effect (gain :runner :tag 2)) :msg "give the Runner 2 tags"
     :end-turn {:effect (effect (lose :runner :tag 2)) :msg "make the Runner lose 2 tags"}}
--- a/src/clj/game/cards.clj
+++ b/src/clj/game/cards.clj
@@ -54,6 +54,12 @@
    "Akamatsu Mem Chip"
    {:effect (effect (gain :memory 1)) :leave-play (effect (lose :memory 1))}
 
+
+   "Akitaro Watanabe"
+   {:events {:pre-rez {:req (req (and (= (:type target) "ICE")
+                                        (= (card->server state card) (card->server state target))))
+                       :effect (effect (rez-cost-bonus 2))}}}
+
    "Alix T4LB07"
    {:events {:corp-install {:effect (effect (add-prop card :counter 1))}}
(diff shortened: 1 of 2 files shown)


mtgred:
Good job. :+1: 

Something you can do next is to extend the mechanism to runner install. It would allow to automate Kate and Test Run. Some cards like Modded or Career Fair have to be adapted as they currently don't combine with other cost reduction ability.

mtgred:
I deployed the code this morning but soon after some players mentioned in the chat that the card access is broken. I'm currently at work and can't analyse the issue. So I rolled back the server to a previous commit. Could you have a look at it?

nealterrell:
Yeah, I saw that message, and can confirm that the rez- and trash-cost effects were not working on Live. I did a fresh clone of the most recent codebase (https://github.com/mtgred/netrunner.git), recompiled it, and tested on my home machine, and all the cards are working as intended. Is it possible the code wasn't actually published to Live?

mtgred:
I did some testing and this pull request indeed breaks the card access. Accessing a asset works fine but accessing an agenda doesn't steal it. It just hangs.

If you check out commit 10d3b1a, install an agenda and run it. Nothing happens when clicking on "Successful Run" button.

nealterrell:
You're right! I'll check into it right away.

PR #220, opened 2015-06-03
Implement Analog Dreamers, Constellation Protocol, Mark Yale

nealterrell:
Implementations for three new cards.

1. Analog Dreamers: I tested this by replacing Sneakdoor Beta's code (temporarily). While testing, I came across an interesting bug. If a `:replace-access` effect shows a prompt that is a map, thus triggering a `(show-select)` instead of a `(show-prompt)`, after the select resolves the run prompt is returned to the "Successful Run/Jack Out" options, and the runner can once again select Successful Run -> Use Card Ability to trigger the selection prompt again. I don't know the root cause of this, but it's something about how `(resolve-prompt)` vs `(resolve-select)` works with respect to ending runs. To hack around this, I added 

    `(swap! state update-in [:runner :prompt] rest)
(handle-end-run state side)`

    manually to the end of Analog Dreamers, which has the effect of fixing this issue.
2. Constellation Protocol: I tweaked selection prompts to use the `:prompt` string of an ability if it is present, instead of always automatically showing "Select a target for X". This is nice in Constellation Protocol, because otherwise you get two prompts in a row saying "Select a target for Constellation Protocol" and you aren't sure if your first click (select which ICE to move a token from) went through or not. Now the second prompt says "Move to where?" to be clear.

    I also had to tweak the code that removes prompts from the prompt queue when resolving them. Due to timing issues, any prompt that pushed a `:priority` prompt during its resolution would inadvertently pop the new prompt from the queue at the end of the original prompt's resolution. Now, the specific prompt that is being resolved will be filtered out of the queue, regardless of whether it's still first in the line. This is necessary for multiple copies of Constellation Protocol to work together: if the first prompt isn't priority, then you will start your turn with two "Use Constellation Protocol?" messages, then receive two "Select a target for Constellation Protocol" prompts, then two "Move to where?" prompts. With priority fixed, you fully resolve on CP before triggering the next.

3. Mark Yale: added an event `:agenda-counter-spent` any time an Agenda uses an ability with a `:counter-cost`. Otherwise straightforward. And yes, using Mark's ability to consume another Agenda's token does trigger his own 1-cr bonus for a total of 3 cr :).

### DIFF
--- a/src/clj/game/cards.clj
+++ b/src/clj/game/cards.clj
@@ -75,14 +75,17 @@
 
    "Analog Dreamers"
    {:abilities [{:cost [:click 1] :msg "make a run on R&D"
-                 :effect (effect (run :rd
-                                      {:replace-access
-                                       {:prompt "Choose a card to shuffle into R&D"
-                                        :choices {:req #(and (not (= (:type %) "ICE"))
-                                                             (not (:rezzed %))
-                                                             (not (:advance-counter %)))}
-                                        :effect (req (move state :corp target :deck) (shuffle! state :corp :deck))
-                                        :msg "shuffle a card into R&D"}}))}]}
+                 :effect (effect (run :rd {:replace-access
+                                           {:prompt "Choose a card to shuffle into R&D"
+                                            :choices {:req #(and (not (= (:type %) "ICE"))
+                                                                 (not (:rezzed %))
+                                                                 (not (:advance-counter %)))}
+                                            :effect (req (move state :corp target :deck)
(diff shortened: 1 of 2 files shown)


mtgred:
The end run code is currently buggy. It requires there is no prompt left before ending the run. The initial intent was to give Dirty Laundry money only after all the access/trash prompts are dismissed. So the code that ends runs is actually in resolve-prompt. It sometime breaks when there are select prompt such as Leela ability. The whole end run code is quite messy overall and should be reworked.

PR #222, opened 2015-06-04
Install cost bonuses; Chrome Parlor; card implementations

nealterrell:
A mishmash of implementations.

### Install cost bonuses

Extended the "bonus" framework to Runner installs via `:pre-install` and `(install-cost-bonus)`. Implemented these cards:

1. Kate "Mac" McCaffrey: bonus of -1 `:once :per-turn` for installing Hardware or Programs. Works in all expected cases: installing from SMC, Scavenge, Test Run (consumes the effect for no gain), Clone Chip, in conjunction with Modded, etc.
2. Modded: uses the tech from Leverage to register a temp event to handle the -3 install cost bonus. 
3. Career Fair: likewise for resources.
4. Scavenge: reworked to use install cost bonus.
5. Clone Chip: I couldn't figure out a way for the `req` filter to work nicely with Kate's ability. It currently only shows programs that the Runner can afford, so it won't show a program that could be afforded via Kate's ability. I removed the filter completely a la SMC, so the user can select a program they can't afford, and runner-install will handle the rest.

### Chrome Parlor

Extended `(damage)` and `:pre-damage` to pass the card dealing the damage as a parameter. `:pre-event` now receives a `target` of the damage type, and a `(second targets)` of the card. This allows Chrome Parlor to prevent damage from Cybernetics. Works perfectly.

### Running Interference
A fun card. Implemented with a temporary event that handles `:pre-rez` to increase all ICE rezzing on the run by the ICE's cost.

### DIFF
--- a/src/clj/game/cards.clj
+++ b/src/clj/game/cards.clj
@@ -1785,7 +1785,7 @@
                                       :run-ends
                                       {:effect (effect (unregister-events card))}}
                                      (assoc card :zone '(:discard))))
-    :events {:pre-rez nil}}
+    :events {:pre-rez nil :run-ends nil}}
 
    "Ryon Knight"
    {:abilities [{:msg "do 1 brain damage" :req (req (and this-server (zero? (:click runner))))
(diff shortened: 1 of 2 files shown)


mtgred:
Awesome :+1:

mtgred:
I did some testing with Kate and added a log message has been for her ability. With the bonus is not applied although the log message appears. It's probably related to the hosting.

Issue #129, opened 2015-02-28
Stacking Daily Business Show's


fabiodrucker:
When multiple DBS's are rezzed this happens:

1) You draw cards A, B, C.
2) It asks: Which of (A, B) you want in the bottom.
3) If you pick A, it asks (usually) which of (B, C) you want to put in the bottom. However, sometimes it asks you which of (A, C) you want to put in the bottom. If you pick A again, nothing goes to the bottom and you "draw" two cards total instead of one.

Since there's no manual way to put things on the bottom of R&D, recovering manually from this is hard.

mtgred:
Thanks for reporting the issue. It has been fixed.

Issue #223, opened 2015-06-04
Multiple Daily Business Show is Wrong


bdorfman:
It prompts you multiple times to put cards underneath, but it gives you the wrong options after the first which can lead to you having an extra card and no way to manually fix it. 

E,g, I drew A B and C. It asked whether i wanted to put card A or B on the bottom and I said A. Then it asked A or C (isntead of B or C) and I said A again. Now B and C are both in my hand.

nealterrell:
I submitted a fix for this in #224. Thanks for the report.

JoelCFC25:
When this gets closed out, the previous bug report for stacking DBS (#129) can get closed out as well.

PR #224, opened 2015-06-04
Fix for multiple Daily Business Shows.

nealterrell:
One more for today :). This is a fix for #223. Make the `:corp-draw` event only fire once regardless of DBS count, then drawn all the extra cards at once, and select the cards to return straight from HQ.


JoelCFC25:
This also fixes #129.

mtgred:
:+1:

PR #228, opened 2015-06-05
Fix for accessing agendas.

nealterrell:
Fix for the bug I introduced with accessing agendas in #218. Don't feed nil to + and max...


mtgred:
:+1:

Issue #231, opened 2015-06-08
Oaktown Renovation trashes itself


JoeyKL:
When Oaktown Renovation installs, it is immediately trashed.

nealterrell:
Thanks for the report, it's an issue with the code for rez cost discounts. I'll submit a fix soon.

Issue #230, opened 2015-06-07
Multiple Daily Business Show Interaction


Kijdhefgi:
When you have multiple Daily Business Shows installed, the unrezzed ones are counted as being rezzed as long as there is one rezzed.  I had two installed, one was unrezzed, and the next turn prompted me to draw three and discard two.
I don't really know much about Clojure so I doubt this will work, but I believe you'd just have to add a check to see if the cards you're filtering for are also rezzed.

    [dbs (->> (:corp @state) :servers seq flatten (mapcat :content)
    (filter #(= (:rezzed %) (filter #(= (:title %) "Daily Business Show"))) count)

nealterrell:
Thanks for the report. Right idea with the suggested fix, but the right syntax would be

`(filter #(and (:rezzed %) (= (:title %) "Daily Business Show")))`

Issue #214, opened 2015-05-30
Parasites disappear


percomis:
Played with Chaos Theory and every Parasite I used on 0 strength ICE (2 Yaguras) disappeared from the game. Not in heap or in stack.

mtgred:
Thanks for reporting the issue. The Parasite code automating the ICE trashing was quite buggy and caused all sorts of problem. It has been removed so now the ICE has be trashed manually but at least it shouldn't have ill effects anymore.

PR #232, opened 2015-06-08
Ice strength modifications; card fixes.

nealterrell:
This is an implementation of dynamic strengths for ice, including most cards that affect an ice's strength after it is rezzed. Although ultimately only Parasite really NEEDS this system, I believe it is a good step towards greater automation of ice interactions and breaking. I have done due diligence in testing this system, but I would recommend thorough testing by others before deploying it live, because the changes are fairly extensive.

Because I am a noob at GitHub, this PR also has a few miscellaneous fixes for #230 and #231.

### Core changes

Functions for handling dynamic ice strengths.

0. Ice cards now have a key `:current-strength` that gets updated at certain key moments or when requested, representing the ice's strength after modifying effects. `:strength` remains the ice's printed strength.
1. Ice card definitions can define key `:strength-bonus`, a function that returns an integer amount to modify the ice's `:strength` by. This is the starting point for dynamic ice strengths, and is used by any ice that has a simple "gain +X under this condition" bonuses. For example, Wraparound has `(req (if (some #(has? % :subtype "Fracter") (get-in runner [:rig :program])) 0 7))` as its `:strength-bonus` to grant 7 strength if there is no Fracter installed.
2. Similar to the systems for rez/install/trash cost bonuses, function `(ice-strength-bonus)` allows a third party to apply a bonus to a particular ice's strength. A paired function `(ice-strength)` accumulates all the bonuses and gives a particular ice's final `:current-strength`.
3. `(update-ice-strength)` can be called any time someone wants an ice's current strength to update. This function triggers event `:pre-ice-strength`, which a card definition can handle to set a `(ice-strength-bonus)` as applicable. `(update-ice-strength)` is called at several points in the game cycle:
    1. On ice rez
    2. On ice encounter (added a new event `:ice-encounter` for cards that trigger "on encounter")
    3. On ice install (for Curtain Wall)
    4. On adding a counter to an ice
4. Cards can call `(update-ice-strength)` if they have taken or responded to some action that modifies a target ice's strength (Datasucker, Parasite; see below).
5. Utility functions to update all ice in a server `(update-ice-in-server)` or all ice on the board `(update-all-ice)`. 
6. An event `:ice-strength-changed` which fires any time an ice's strength is recalculated. Needed for Parasite.

This may sound confusing, but these card implementation details may help...

### Card implementations

Self-boosting ice:

1. Location-based boosts (Gutenberg, Crick, Turing, Meru Mati, Curtain Wall): use `:strength-boost` to increase the ice's strength based on its server / position.
2. Counter-based boosts (Ice Wall, Fire Wall, Draco, Hadrian's Wall, Shadow): use `:strength-boost` based on `:counter` or `:advance-counter`.
3. Other conditions (Wraparound, IQ, NEXT Bronze): various techniques to count the required information for the boost.

Third-party boosts:

1. Permanent boosts (Superior Cyberwalls, Encrypted Portals, HB: Stronger Together, Lag Time, Experiential Data): handle `:pre-ice-strength` with appropriate `:req` to give bonuses to only certain types of ice. Example (Encypted Portals): 

  `:events {:pre-ice-strength {:req (req (has? target :subtype "Code Gate"))
                                :effect (effect (ice-strength-bonus 1))}`
2. Encounter-based boosts (Scrubbed, Ice Carver): handles the `:pre-ice-strength` triggered by the "on-encounter" core code; calls `(ice-strength-bonus)` when the target ice is the run's `current-ice`.
3. Boosts until end of turn (Corporate Troubleshooter, IT Department): after triggering the card's effect, "save" the target of the effect and call `(update-ice-strength)` to trigger a strength recalculation. Do fancy magic to apply a strength bonus.
4. Datasucker: if the runner is encountering an ice, keep track of how many tokens the runner has spent from the Datasucker, and call `(update-ice-strength)` each time a token is spent to update with the new (negative) strength bonus.
5. Parasite: on install, call `(update-ice-strength)`. Likewise on runner start of turn. Handle event `:ice-strength-changed` to check the ice for strength <= 0 to trash it. 
6. Patch: corp cards cannot currently be hosted; but once hosting is enabled, this card should work.

### UI

The current build shows an ice's current strength in orange, similar to a boosted icebreaker, if the current strength is not equal to the base strength. Picture:
image

This can be removed if it is distracting or unneeded.

### Tested scenarios

1. Parasite is the only card that REALLY needs this system (for now), and should work very well, including a fix for #214. Accounts for all boosting effects by requiring `:current-strength` to be <= 0, rather than base `:strength`. I have tested:
    1. Parasite on strength 0, immediate trash.
    2. Parasite on other strength, lowers current-strength by 1 each start of runner's turn.
    3. Parasite on ice, use Datasucker to drop strength to 0, immediate trash.
    4. Approach ice, Clone Chip a Parasite, use Datasucker to trash.
    5. Approach ice, corp uses IT Department to boost, Runner uses Datasucker to lower to a point that WOULD have trashed the ice... after passing ice, it resets to Parasited value.

### To-do

1. Bishop: still don't have a handle on hosting code, but should be simple to implement.
2. Quicksand: should be easy by handling the new event `:encounter-ice`.

Feedback welcome :).

### DIFF
--- a/src/clj/game/cards.clj
+++ b/src/clj/game/cards.clj
@@ -426,7 +426,22 @@
                                  state side
                                  {:choices {:req #(and (has? % :type "ICE") (:rezzed %))}
                                   :msg (msg "add " boost " strength to " (:title target))
-                                  :effect (effect (trash card))} card nil)))}]}
+                                  :effect (req
+                                            (update! state side (assoc card :troubleshooter-target target
+                                                                       :troubleshooter-amount boost))
+                                            (trash state side (get-card state card))
+                                            (update-ice-strength state side target))} card nil)))}]
+    :events {:pre-ice-strength nil :end-turn nil}
+    :trash-effect
+      {:effect (req (register-events
+                      state side
+                      (let [ct {:effect (req (unregister-events state side card)
+                                             (update! state side (dissoc card :troubleshooter-target))
+                                             (update-ice-strength state side (:troubleshooter-target card)))}]
+                        {:pre-ice-strength
+                          {:req (req (= (:cid target) (:cid (:troubleshooter-target card))))
+                           :effect (effect (ice-strength-bonus (:troubleshooter-amount card)))}
+                         :runner-turn-ends ct :corp-turn-ends ct}) card))}}
 
    "Corporate War"
    {:msg (msg (if (> (:credit corp) 6) "gain 7 [Credits]" "lose all credits"))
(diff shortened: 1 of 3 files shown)


mtgred:
Awesome ! :+1:

JoelCFC25:
Astonishingly good work...thanks a million!

PR #235, opened 2015-06-09
partial Power Shutdown and Shiro

JoelCFC25:
Using Power Shutdown right now is very arduous, so this is intended to at least ease the Corp trashing part. The Runner trash would still be manual.

Shiro just mimics Precognition and Kitsune. 

EDIT: Also discovered in a Reddit comment that Snowball was slightly bugged, so I fixed that.

### DIFF
--- a/src/clj/game/cards.clj
+++ b/src/clj/game/cards.clj
@@ -1501,7 +1501,8 @@
                :effect (effect (gain :bad-publicity 1) (gain :runner :tag 1) (forfeit card))}}
 
    "Power Shutdown"
-   {:prompt "Trash how many cards from R&D?" :choices (req (count (:deck corp)))
+   {:req (req (:made-run runner-reg))
+    :prompt "Trash how many cards from R&D?" :choices (req (count (:deck corp)))
     :msg (msg "make the Runner trash 1 installed program or hardware with an install cost of " target " or less")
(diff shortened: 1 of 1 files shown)


JoelCFC25:
I didn't notice at first a very minimal Power Shutdown was already in the code--sorry about that. Transferred over the run requirement to my new one and removed the old one.

mtgred:
`(req (count (:deck corp)))` won't work. The `resolve-ability` function expects an array of cards to be returned by the `req` function and not a number.

The handling of `:choices` in `resolve-ability` needs to be updated to be more generic before Power Shutdown can be implemented. I will try to do it in the next few days when I have time. I will park your pull request for now until it's done.

JoelCFC25:
Sorry about that. If Shiro is passable, I could pull it and the Snowball fix out and spin them off into a different pull request--what do you think?

JoelCFC25:
Thinking about this last night, I wonder if the Snowball fix attempt will even work--this particular program can receive conventional pumps *and* "until end of run" pumps. After passing a piece of ice I'm guessing it can't distinguish between the two kinds and either no pumps will be reset or all of them will.

JoelCFC25:
Closing this as there just isn't enough right with it and @mtgred said he was doing Power Shutdown himself.

PR #246, opened 2015-06-14
Leela fix; potential fix for freezing mid-run; improvements to ice strength code; Wyrm; Parasite iteration

### LINKED ISSUE #241: Leela ability trigger
Playing a game today with Leela, when I would access and score and agenda after the leela ability trigger it was not ending the run and allowing an additional access.  This was avoidable by clicking jack out instead of successful run.

### LINKED ISSUE #245: Trapped in a run
Sometimes, when the runner is trying to exit a run or when the run is completing, the runner instead gets "Continue" and "Jack Out" options instead, where neither option does anything. This seems to happen when the runner jacks out and the corp goes 'no further action' simultaneously, but also seems to happen with e.g. Argus Security, after triggering the Argus agenda steal penalty.

nealterrell:
A group of miscellaneous fixes.

### Current-ice fix

In macros.clj, the `effect`, `req`, and `msg` macros expose a binding `current-ice`. This binding was throwing when there either was no `:run` active or no `:run :position`. I don't know why this wasn't an issue before, but it's definitely an issue with the dynamic ice strengths, AND with Leela's ability on agenda steal. This may fix #245.

### Leela ability fix

Leela's ability was erroring when stealing an agenda because of the current-ice issue, so resolving that fixes her issue. I have also added an improvement that I learned doing Analog Dreamers, so that the "Continue/Jack Out" message will no longer be shown after triggering her ability. Fixes #241.

### Ice strength code improvements

Some minor improvements to ice strength code, including proper implementation of Curtain Wall.

### Parasite improvements

I have reworked parts of Parasite to account for:

1. Adding virus tokens to Parasite or Hivemind via Breeding Ground / Incubator / Surge / etc. (Added an event `:counter-added` for when non-agendas receive new counters.)
2. Grimoire's free token on install.

I believe this implementation is pretty solid overall.  I have tested this code on my server with @JoelCFC25; mediohxcore and kumohara also played a game on my server attempting to abuse Parasite with no issue reported. Interactions that we personally tested:

1. Auto-trash ice on install (Quandary at str 0, NEXT Silver when Grimoire is out)
2. Use Datasucker to lower ice to trashing distance on encounter.
3. Parasite an ice down to str 1; run with Ice Carver out; ice gets trashed on encounter.
4. Hivemind counters increasing Parasite strength.
5. Parasite installed; increase counters on Hivemind; trash ice when appropriate.

### Wyrm

Just in case there's someone out there dying to use Wyrm, I did implement its str-lowering subroutine.

### Misc

Pass the card that dealt the damage to the `:damage` event (which fires after the damage is done) to support @JoelCFC25's Defective Brainchips code (so Brainchips' extra damage doesn't trigger its own boosting ability).


mtgred:
:+1:

PR #253, opened 2015-06-16
Fix for Curtain Wall, Wraparound, Parasite

nealterrell:
Fix infinite recursion when Parasite trashes Curtain Wall or Wraparound. Don't listen to `:trash` events if you are the card being trashed. Same fix as for NEXT Bronze.

I am still a noob at GitHub so there are a few commits here that I had to revert because they weren't ready. Commit 93cb7c2a8ed761f2bff749a54fcd4fbc662d4f3a is the money commit.


mtgred:
:+1:

PR #257, opened 2015-06-16
remove click cost from Tallie Perrault

JoelCFC25:
Tallie's ability doesn't cost a click, it happens by trashing her at paid ability speed.


mtgred:
:+1:

PR #258, opened 2015-06-16
Grail ice with reveal and resolve

nealterrell:
Building off of @JoelCFC25's initial Grail code, here is an implementation of Grail ice that can:

1. Reveal up to 2 `:subtype "Grail"` ice from HQ; and then
2. Resolve one ability from a Grail ice in HQ at a time.

So on encounter, Corp clicks his rezzed Grail to reveal 2 Grail from hand. If the subs are not broken, Corp can choose which subs to fire in which order, by either trigger the rezzed ice's subroutine or by triggering the rezzed ice's ability to resolve another Grail's subroutine. It's a little click intensive, but gives the most control and accuracy over execution, and is important for making sure that Lancelot's Trash routine fully resolves before Merlin's net damage (in case Deus Ex is installed).

This code assumes that the ice's actual subroutine is first among its abilities, so the abilities should not be rearranged.

### DIFF
--- a/src/clj/game/cards.clj
+++ b/src/clj/game/cards.clj
@@ -2902,8 +2902,8 @@
                  :msg (msg "reveal "
                            (join ", " (map #(str (:title %) " ("
                                                  (:label (first (:abilities (card-def %)))) ")") targets)))}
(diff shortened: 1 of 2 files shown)


mtgred:
Awesome :+1:

PR #260, opened 2015-06-19
Agenda advancement cost effects: SanSan City Grid, et. al.

nealterrell:
Extension of the cost-bonus framework to agenda advancement requirements.

### Core changes

1. Added event `:pre-advancement-cost` that triggers when an agenda's advancement requirement is being calculated. Advancement requirements are calculated when:
    1. An agenda is installed.
    2. An agenda is advanced.
    2. Corp's turn begins.
    3. Requested by third parties.
2. Function `(update-advancement-cost)` updates an individual agenda's advancement requirement; utility `(update-all-advancement-costs)` does so for all installed agendas.
3. Cards can implement the key `:advancement-cost-bonus` as an `effect` to change their advancement requirement based on the state of the game. NAPD Contract uses this to give +1 advancement cost for each bad pub; Medical Breakthrough uses it to count the number of scored Medical Breakthroughs.
3. Function `(advancement-cost-bonus)` can be used by third parties to apply a bonus to the advancement requirement of an agenda that is currently being updated in `update-advancement-cost`, typically by handling the `:pre-advancement-cost` event. The Source uses this to always give a cost bonus of 1; Chakana uses it to give a cost bonus of 1 if there are at least 3 virus counters on it. 

### Card implementations

1. Chakana: bonus of 1 if there are at least 3 virus counters (including Hivemind).
2. Medical Breakthrough: bonus of -1 for each scored Medical Breakthrough.
3. NAPD Contract: self-inflicted +1 bonus for each bad publicity.
4. SanSan City Grid: triggers a cost update for any agenda already installed in the server, or installed later into the server. -1 bonus for any agenda in the same server.
5. The Source: +1 bonus for all agendas.
6. Traffic Jam: +1 bonus for each corp-scored agenda with the same title as the scorable agenda.

### Known issues

1. If a corp somehow gains a bad publicity AFTER installing NAPD Contract, AFTER the start of turn on the turn in which they attempt to score it, the bad publicity will not affect the contract's advancement requirement. Example: NAPD installed with 4 counters; Start of Turn; Draw; (SOMEHOW gain 1 bad publicity); score NAPD with only 4 counters.

### DIFF
--- a/src/clj/game/cards.clj
+++ b/src/clj/game/cards.clj
@@ -298,6 +298,15 @@
     :trash-effect {:req (req (:access @state))
                    :effect (effect (move :runner card :scored) (gain :runner :agenda-point 2))}}
 
+   "Chakana"
+   {:leave-play (effect (update-all-advancement-costs))
+    :events {:successful-run {:effect (effect (add-prop card :counter 1)) :req (req (= target :rd))}
+             :pre-advancement-cost {:req (req (>= (get-virus-counters state side card) 3))
+                               :effect (effect (advancement-cost-bonus 1))}
+             :counter-added
+             {:req (req (or (= (:title target) "Hivemind") (= (:cid target) (:cid card))))
+              :effect (effect (update-all-advancement-costs))}}}
+
    "Chaos Theory: Wünderkind"
    {:effect (effect (gain :memory 1))}
 
@@ -1313,6 +1322,12 @@
    {:events {:runner-turn-begins {:msg "trash the top 2 cards from Stack and draw 1 card"
                                   :effect (effect (mill 2) (draw))}}}
 
+   "Medical Breakthrough"
+   {:effect (effect (update-all-advancement-costs))
+    :stolen (effect (update-all-advancement-costs))
+    :advancement-cost-bonus (req (- (count (filter #(= (:title %) "Medical Breakthrough")
+                                                (concat (:scored corp) (:scored runner))))))}
+
    "Medical Research Fundraiser"
    {:effect (effect (gain :credit 8) (gain :runner :credit 3))}
 
@@ -1385,7 +1400,8 @@
     :effect (effect (corp-install (assoc target :advance-counter 3) "New remote"))}
 
    "NAPD Contract"
-   {:steal-cost [:credit 4]}
+   {:steal-cost [:credit 4]
+    :advancement-cost-bonus (req (:bad-publicity corp))}
 
    "NBN: Making News"
    {:recurring 2}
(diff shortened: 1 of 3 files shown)


mtgred:
Awesome ! :+1:

mtgred:
Perhaps you can extend this to additional cost for stealing? Eg: Red Herring, Strong Box, Predictive Algorithm, The Source, Utopia Fragment.

JoelCFC25:
Just brainstorming as I try to conceive of how to do this, `:steal-cost` is currently attached to a specific card definition. That handles Fetal AI and NAPD Contract, but needs to get broadened out to stack with: 

##### Global steal costs
The Source
Predictive Algorithm
Utopia Fragment *(2 credits per advancement on agenda being accessed)*

##### Server-specific steal costs
Strongbox
Red Herrings

What if `:steal-cost` became a sum of: 
`:steal-cost-card` (Fetal and NAPD change to this)
`:steal-cost-server`
`:steal-cost-all`

All three types could accept `credits` and `clicks`. Then on accessing an agenda, the Runner would get presented with sums of credit costs (and click costs if applicable). So a crazy unlikely case would be hitting a 3-advanced NAPD in a server with two Strongboxes and a Red Herrings with a Utopia Fragment scored and Predictive Algorithm in play (LOL): 

"Pay 17 credits and 2 clicks to steal NAPD Contract?"

nealterrell:
I abstracted all my "bonus" systems (install cost, rez cost, trash cost, advancement cost) as a little more general: bonuses can come from a card itself (usually through a function directly in the card definition), or from third parties (by handling events that are triggered at appropriate moments during calculations). If you can mimic the advancement cost system here, you should have a pretty flexible solution for stealing costs:

`:steal-cost` is analogous with `:advancement-cost-bonus` (defined by the card itself; can handle conditional increases that the card decides apply -- see NAPD Contact)
    `:steal-cost-server` and `:steal-cost-all` are analogous with handling the `:pre-advancement-cost` event and calling `(advancement-cost-bonus X)`. Third parties handle this event to add to the costs of other cards -- see Chakana, Traffic Jam, and SanSan City Grid. 

Then it's a matter of defining functions for the third parties to call to increase the bonus costs (see `(advancement-cost-bonus)`); a function to calculate the FINAL "cost" amount after bonuses are applied (`(advancement-cost)`; first asks the card itself for bonuses, then retrieves the bonuses registered by third parties; sums those values, doesn't let it go below 0); and then triggering events at appropriate times to update/use the calculations. Since steal costs are a little easier than advancement costs (which need to be known at all times, not just access time), you can probably work the event and calculations into the existing `handle-access` routine where it calculates steal costs. (I think that's where it happens...) That would be more analogous to the trash cost bonus system, which only has to calculate trash costs during access.

It's tricky, but I believe you can figure it out :).

PR #262, opened 2015-06-20
stop Caissa from revealing unrezzed ice

JoelCFC25:
Saw a report in chat tonight that Caissa are revealing the names of unrezzed ice when getting hosted. Added in a check used on cards like Psychographics, et al.


mtgred:
Awesome :+1:

PR #263, opened 2015-06-20
Targeted trashing and trash prevention

nealterrell:
Implements a Trash Resource button for Corp, refactors most "trash a ____" abilities for Corp, and adds trash prevention cards (Sacrificial Construct etc.).

### Core changes

1. Refactors the damage prevention code to unify all prevention effects under `@state`'s `:prevent` field. We now have `:prevent :damage` and `:prevent :trash`, and in the future can implement `:prevent :tag`. Feedback Filter now registers its prevention effect as `:prevent {:damage [:net :brain]}`, and Sacrifical Construct as `:prevent {:trash [:program :hardware]}`.
2. Adds code for trash prevention, giving the Runner time to activate any prevention cards when appropriate. Very similar to damage prevention code.
3. Adds a conditional "Trash Resource" button to the Corp's action area.

### Card implementations

Refactorings:

1. Refactors all Destroyer-type effects on ice to use a single shared ability defined at the top of `cards.clj`. I found several different implementations of "trash a program" abilities and decided to unify them all under one ability map. I am not 100% convinced that my submission is the BEST way of doing trash-a-program, but now we can change one set of code to affect all ice with this ability. For reference, the new ability is:

    ````clojure
 {:prompt "Choose a program to trash" :label "Trash a program"
                    :msg (msg "trash " (:title target))
                    :choices {:req #(= (:type %) "Program")}
    ````

    which is an improvement (IMHO) in that the Corp clicks on the ice to trash instead of selecting its name from a list, which has the problem of not distinguishing between multiple ice with the same name (relevant for Datasucker and Parasite in particular). I do see a potential issue outlined below.
2. Similarly refactors "trash a hardware" routines, which are far fewer.

New card implementations:

1. Aggressive Secretary: allow the Corp to select the programs to trash, instead of just printing a message.
2. Shattered Remains: likewise.
3. Fall Guy: `:prevent {:trash [:resource]}`. Fall Guy itself is trashed as `:unpreventable` just to make the code work... otherwise Fall Guy's self-trash triggers his own "Prevent trash?" routine which is just messy.
4. Sacrificial Construct: `:prevent {:trash [:program :hardware]}`.
5. LLDS Energy Regulator: `:prevent {:trash [:hardware]}`.

### Misc

1. Added a stack trace to exceptions caught by the server. Should help with debugging. If you want to send me a server log every few days, I can look into some of the error messages that crop up.
2. I really messed up my Git repository while trying to learn to `git rebase`. I am 95% sure I got everything restored for this submit, and have done my due diligence testing this latest submission, but I may have missed something.

### Known issues:

1. If a runner discards a card from hand, trash-prevention effects will trigger. Probably easy to fix but haven't looked into it. (Note to self: try `:unpreventable true` when calling `trash` in `move-card`.)
2. BIGGER ISSUE: when taking damage of any kind, if you have a prevention card installed, you will get a prevention prompt if the damage makes you trash a preventable card.
2. The refactored Trash a Program routine filters to cards with `:type "Program"`. This is IMO better than filtering to cards in the `:rig :program` zone, since Parasite and other hosted cards are not in this zone. However it does allow someone to trash a program on Personal Workshop, which to the best of my knowledge is not allowed...?

### DIFF
--- a/src/clj/game/cards.clj
+++ b/src/clj/game/cards.clj
@@ -1,5 +1,15 @@
 (in-ns 'game.core)
 
+(def trash-program {:prompt "Choose a program to trash" :label "Trash a program"
+                    :msg (msg "trash " (:title target))
+                    :choices {:req #(= (:type %) "Program")}
+                    :effect (effect (trash target))})
+
+(def trash-hardware {:prompt "Choose a piece of hardware to trash" :label "Trash a piece of hardware"
+                     :msg (msg "trash " (:title target))
+                     :choices {:req #(= (:type %) "Hardware")}
+                     :effect (effect (trash target))})
+
 (def cards
   {"Accelerated Beta Test"
    {:optional {:prompt "Look at the top 3 cards of R&D?"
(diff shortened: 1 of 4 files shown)


mtgred:
Awesome ! :+1:

Issue #264, opened 2015-06-21
Net Ready Eyes only shows up for first encounter


qmsm:
There was an old issue for Net Ready Eyes but didn't seem the same as this.

The increased strength for your target only shows up for the first encounter and then disappears rather than "for the remainder of the run".

JoelCFC25:
I thought this was easy, but was reminded that the code currently can't distinguish between pumps that last for the remainder of the run and those that only last for the duration of a single encounter if a program receives both kinds (this is also afflicts Snowball).

PR #268, opened 2015-06-22
fix Net-Ready Eyes strength boost

JoelCFC25:
The boost given to the targeted icebreaker should last for the rest of the run like Gordian Blade, Pipeline, et al.

Fix for #264.


JoelCFC25:
Not going to work, apparently .

PR #269, opened 2015-06-22
Fix for Rachel Beckman, prevention of trashes from taking damage

nealterrell:
Fix for Rachel Beckman's infinite recursion when trashed, by removing her watch effect before trashing her, so the trash effect does not trigger her watch effect in an infinite recursion loop.

Fix to disallow using trash-prevention effects when trashing cards from grip because of damage.


mtgred:
:+1:

Issue #270, opened 2015-06-22
Oaktown Renovation Triggers Hacktivist Meeting; It Shouldn't


scifantasy:
Installing Oaktown Renovation with Hacktivist Meeting in play triggers the "randomly trash one card from HQ" effect of Hacktivist Meeting, which it shouldn't.

It's easy to understand why: Oaktown appears to be implemented so that installing it triggers an immediate rez, and it's a non-ice card...

The problem is that Oaktown isn't _rezzed_, it's just installed faceup, so it shouldn't trigger Hacktivist.

nealterrell:
Fun with border conditions! Thanks for the report, I'll see what I can do.

mtgred:
The fix has been deployed this morning. It should be ok now.

PR #272, opened 2015-06-23
Fixes for Quicksand and Oaktown Renovation

nealterrell:
Fix for Quicksand to fit the new ice-strength system. Counter is gained during `:encounter-ice`, which is when the Corp clicks "No more action".

Fix for #270 by refactoring the "install rezzed" code. Changed `:install-rezzed true` to a more general `:install-state` key. Oaktown Renovation uses `:install-state :face-up`; if this feature is used on new cards in the future, we can easily expand to `:install-state :rezzed` or any other state. In `corp-install`, if a card is flagged `:install-state :face-up`, it is "pseudo-rezzed": we set `:rezzed` to true for display purposes and initialize the card, but notably do NOT trigger event `:rezzed`, so Hacktivist Meeting does not respond. If more cards make use of the distinction between rezzed and face-up, we can expand the UI code to visibly show "face-up" cards without setting the `:rezzed` key to true.

### DIFF
--- a/src/clj/game/cards.clj
+++ b/src/clj/game/cards.clj
@@ -3258,9 +3258,10 @@
    {:abilities [{:msg "end the run" :effect (effect (end-run))}]}
 
    "Quicksand"
-   {:abilities [{:msg "add 1 power counter"
-                 :effect (effect (add-prop card :counter 1) (add-prop card :strength 1))}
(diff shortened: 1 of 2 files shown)


mtgred:
Awesome :+1:

PR #275, opened 2015-06-23
Fix for IHW: passing variadic arguments to resolve-trash.

nealterrell:
Apparently passing a variadic argument sequence to another function taking a variadic argument sequence passes the original sequence as the first argument in the second function's sequence, so the second receives a sequence containing a sequence rather than a sequence containing the elements of the other sequence.

SEQUENCE.

This fixes I've Had Worse, which was receiving a `target` of `(:net)` instead of `:net` (likewise `:meat`).


mtgred:
:+1:

PR #284, opened 2015-06-25
Jinteki Biotech

nealterrell:
Implementation of Jinteki Biotech. Added an event `:pre-first-turn` that triggers after keeping your hand / using a mulligan. Biotech uses that event to require the corp to pick a flip identity, and announces that an identity (unrevealed) is chosen to give the Runner peace of mind :). 

On clicking the identity with 3 clicks, uses the selected ability and updates the card's state to make sure it can't be used again.


mtgred:
:+1:

PR #292, opened 2015-06-27
Fixes for Project Atlas and Geist.

nealterrell:
Remove debug message from Geist's ability. Fix Project Atlas receiving -1 counters when scored under SanSan City Grid.


mtgred:
:+1:

PR #306, opened 2015-07-01
Cards that modify their own rez cost while unrezzed: IQ, space ice

nealterrell:
Simple addition to allow a card to modify its own rez cost while it is unrezzed, as cards cannot participate in the event-based rez cost discount system while unrezzed. Implement IQ and the "space ice" Asteroid Belt, Nebula, Orion, Wormhole.


mtgred:
Awesome :+1:

Issue #299, opened 2015-06-30
Pheromones not implemented?


qmsm:
Thought I'd seen another log but couldn't find it.

Pheromones doesn't seem to do anything at the moment. It doesn't get a counter automatically and you can't add manually.

JoelCFC25:
It's not implemented at all. I could submit a patch that at least adds the virus counter on successful HQ runs, but right now the only way to support variable numbers of recurring credits (currently just this card and Net Police) is with a special function that I don't know how to write.

qmsm:
My naivety goes well beyond my play and into not checking the card status list and not realising that counters wouldn't fully fix it.

Having said that, I think having the virus counter would be a useful workaround for now if it's straightforward to implement.

mtgred:
Pheromones and Net Police has been implemented. The update will be deployed this afternoon.

JoelCFC25:
Implemented at c295d128ed84496e0e1a775c42545600e9a4e004.

PR #307, opened 2015-07-01
Retain counters on derezzed Corp assets

JoelCFC25:
Per the Official FAQ (page 4), a derezzed Corp card needs to *retain* all hosted credits, cards, and counters. 

This is relevant now with the introduction of Test Ground, which is useful for reloading an Adonis or Eve Campaign with credits prior to them running out and self-trashing (and if you're really on the ball you'll re-rez for free with Breaker Bay Grid). 

Currently we're only preserving `:advance-counter` on advanceable Weyland ice and advanceable assets when they derez. What this should do is stop dissociating `:counter` no matter what, then only proceed to dissociate them if it's a Runner card (Imps, Gravediggers, et al) or if `keep-counter true` has been passed via the `derez` function.

Adonis/Eve/Rex Campaigns and Private Contracts also get a minor modification so that if credits remain when derezzing, a later re-rez will **add** the placed counters instead of wiping out the remaining ones--i.e. an Eve with 2 left when derezzed by Test Ground will have 18 when later re-rezzed.

Also added a partial for Pheromones to address some of #299. It's not the whole thing, maybe later an `add-watch` can get slapped on and/or the CSS will allow showing both recurring credits and counters on a single card.

### DIFF
--- a/src/clj/game/cards.clj
+++ b/src/clj/game/cards.clj
@@ -42,7 +42,7 @@
                            :msg "gain [Click]" :effect (effect (gain :runner :click 1))}}}
 
    "Adonis Campaign"
-   {:data {:counter 12}
+   {:effect (effect (add-prop card :counter 12))
     :events {:corp-turn-begins {:msg "gain 3 [Credits]" :counter-cost 3
                                 :effect (req (gain state :corp :credit 3)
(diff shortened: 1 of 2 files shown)


mtgred:
:+1:

Issue #310, opened 2015-07-02
Strength Modification and Parasite bug


CrushU:
Strength reduction from Datasucker is carrying over if the ICE is killed by Parasite.

Recreation: 
  1. Run on a server with at least two ICE, and the outer is Parasited
  2. Use Datasucker during the run to reduce the outer ICE strength to 0.
  3. When you Continue to the next ICE, the strength of that ICE has also been modified by the Datasucker activations.

nealterrell:
Thanks for the report.

Note to self: this happens because the `:pass-ice` event is only triggered if an ice is found in the current run position. But since that ice was just trashed, no ice is found in the position, the event is not triggered, and Datasucker doesn't reset. Probably need to augment Datasucker to keep track of which ice it was used on, not just assume that the "current-ice" is being targeted.

nealterrell:
A fix went out for this today, let me know if it happens again!

JoelCFC25:
Tested this exact scenario just now, it seems to work as intended!

PR #312, opened 2015-07-02
Fix for Datasucker applying afte trashing ice with Parasite.

nealterrell:
Fix for #310, by triggering the `:ice-passed` event even when no ice can be found (because the ice was trashed by Parasite).


mtgred:
:+1:

nealterrell:
Well that was fast.

mtgred:
:)

I was about to commit something and wanted to merge your branch first.

Issue #301, opened 2015-06-30
Sundew grants credits even when running it first click


nealterrell:
Sundew gives the Corp 2 credits when you run it first click. I believe this is because `click-run` calls `pay` to pay the click cost BEFORE initiating the run. That triggers `:runner-spent-click` before the run is initiated; Sundew handles the `:runner-spent-click` event before there is anything in `(:run @state)`, so the attempt to check to see if the run is on Sundew's server fails because there is no run yet.

mtgred:
It used to work previously. I will check the issue this evening.

mtgred:
You are right. `pay` is called to check if the runner can pay for the run or the run ability before `run` is initiated. The condition actually never worked.

I don't see how this can be worked around with the current code. The `pay` function has to be split into a `can-pay?` function and a simpler `pay` function with no check. It would actually be beneficial for other card as well like Self Modifying Code or Clone Chip to check of the selected card can be installed.

PR #313, opened 2015-07-02
Better support for some "the first time _____" events

### LINKED ISSUE #281: Daily Business Show Rezzed Mid-Corp Turn Shouldn't Work That Turn
If the Corp rezzes Daily Business Show on its turn and then manually draws, DBS will trigger--even though it wasn't the first card of the turn. (DBS doesn't "know" that it wasn't.)

nealterrell:
Our use of `:once :per-turn` works well for most cards, but there are a few cards where this system doesn't work well. The crux of the issue is that the `:once` system only checks to see if the event in question has happened _since the card was installed/rezzed_. Thus, if the Runner spends a click to Draw, then installs Symmetrical Visage, then Draws again, Visage will grant a credit when it's not supposed to because it sees the second draw as the "first" time a draw occurred. This system works for cards that can reasonably expect to be installed/rezzed before the events they listen to could possibly occur (like all identities), or for cards with their own abilities that can only be used once... but not for cards that could be activated after their event has already happened once in the turn.

This is a small framework extension to support true "first time this turn" effects by saving every event that gets triggered in a vector in the state. Every time `trigger-event` is called, we push a pair into `@state :turn-events` consisting of the event's key and the targets/arguments passed to handlers of the event. From there, we just need a few utility functions to see if a specific event key has happened this turn or not.

I have applied this framework to some, but not all, of the cards that need it:

* Daily Business Show: to fix #281.
* Symmetrical Visage
* Adjusted Chronotype
* Autoscripter
* Enhanced Vision
* John Masanori
* Manhunt
* Synthetic Blood

These cards still need implementing. I passed on them because they were either difficult, weren't urgent, or are under bug review:

* Sundew (pending #301)
* Housekeeping (difficult. fortunately there is currently no way for Housekeeping to come into play during the Runner's turn, but that will change with a leaked NBN ID)
* Comet (difficult because of asynchronous prompt issues)
* NeoTokyo Grid
* Order of Sol (needs better event support)

### DIFF
--- a/src/clj/game/cards.clj
+++ b/src/clj/game/cards.clj
@@ -38,7 +38,9 @@
                           :effect (effect (gain :corp :bad-publicity 1))}}}
 
    "Adjusted Chronotype"
-   {:events {:runner-loss {:req (req (some #{:click} target)) :once :per-turn
+   {:events {:runner-loss {:req (req (and (some #{:click} target)
+                                          (empty? (filter #(= :click %)
+                                                          (mapcat first (turn-events state side :runner-loss))))))
                            :msg "gain [Click]" :effect (effect (gain :runner :click 1))}}}
 
    "Adonis Campaign"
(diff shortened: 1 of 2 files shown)


mtgred:
:+1:

PR #320, opened 2015-07-03
Fix for Autoscripter

nealterrell:
Thought I was cool and really understood what `mapcat` does, but failed. Should be `map`, plus the additional requirement of the new install being a program. Whoops.


nealterrell:
Closing this until I figure out why it can't be merged automatically.

PR #321, opened 2015-07-03
Autoscripter fix

nealterrell:
Here we go -- fix for Autoscripter. Will only trigger now on PROGRAM install, and only if it's the first program this turn.


mtgred:
:+1:

PR #349, opened 2015-07-09
miscellaneous fixes

JoelCFC25:
Removes the new lime-green recurring credits from cards that get trashed or uninstalled by adding `rec-counter` to the list of things that get dissociated.

Fixes: 

* Typo in Dinosaurus hosting message
* Add click cost to the "Install and host" option of Off-Campus Apartment
* Add HQ run requirement to Leverage
* Add partial stub for Bug

Cosmetic changes: 
* Move Zona Sul Shipping out of ice section and Deep Thought up into the section of working cards


mtgred:
:+1:

Issue #308, opened 2015-07-01
Gang sign does not stack with HQ Interface


darlingsensei:
The corp scoring yields only one access when Gang Sign and HQ interface are present. Gang sign also sometimes results in a "Jack Out or Continue" prompt on the corp's turn.

JoelCFC25:
My thought on this today...it could be as simple as how the access is being done. Currently Gang Sign does: `:effect (req (let [c (first (shuffle (:hand corp)))`

But we could simply do it the same way the `(defn access :hq` works in `core.clj`. The default for `:hq-access` is 1, and simply gets incremented when you install an HQ Interface. 

So something in Gang Sign along the lines of: 
`:effect (req (let [c (take :hq-access (shuffle (:hand corp)))]`

...ought to give as many accesses as desired from HQ while ignoring any upgrades, right? However, we're still left with the problem of ordering simultaneous effects if you have multiple Gang Signs and/or multiple HQ Interfaces.

nealterrell:
That feels like it should work, even for multiple interfaces, if you turn the `let` into a `doseq`. But you're right, it won't solve multiple Gang Signs... You'll get the second access procedure while the runner is resolving prompts to trash, so the second access will use the same card options as the first.

JoelCFC25:
Would there be a way to put logic into Gang Sign to first count how many Gang Signs are installed, then "for each" do <insert access code here>?

PR #351, opened 2015-07-10
implement Exploratory Romp and Raymond Flint

JoelCFC25:
@mtgred @nealterrell 

This is the first thing I successfully got working with repeated tweaking/reloading using the REPL in the test environment (yay!). I can verify this works except for *one* part--even after choosing to use the Run ability instead of accessing and then removing advancements from a card, you get the "Successful run / Jack out" prompt like we used to see after Leela's ability sometimes. 

I tried adding `(handle-run-end state side)` in a couple different places but couldn't get it to take--it would then stop removing advancements and hang. In the REPL it complained about the wrong number of arguments being passed. Was that because I wasn't preceding it with `(swap! state update-in [:runner :prompt] rest)`? 

So I'm wondering if adding these two lines immediately after the `(add-prop :corp target :advance-counter (- c))` will tidy it up: 

`(swap! state update-in [:runner :prompt] rest)`
`(handle-run-end state side)`

### DIFF
--- a/src/clj/game/cards.clj
+++ b/src/clj/game/cards.clj
@@ -890,17 +890,19 @@
    "Exploratory Romp"
    {:prompt "Choose a server" :choices (req servers)
     :effect (effect (run target
-	                   {:replace-access
-		                  {:prompt "Remove how many advancements from a card in or protecting this server?"
-		                   :choices ["0", "1", "2", "3"]
-		                   :effect (req (let [c (Integer/parseInt target)]
-				                          (resolve-ability state side
-                                           {:choices {:req #(and (contains? % :advance-counter)
-							                                    (or (= (last (:zone %)) :ices)
-                                                      (= (last (:zone %)) :content)))}
-				                            :msg (msg "remove " c " advancements from " (if (:seen target) (:title target) "a card"))
-				                            :effect (effect (add-prop :corp target :advance-counter (- c)))}
-				                           card nil)))}}) card)}
+	              {:replace-access
+		       {:prompt "Remove how many advancements from a card in or protecting this server?"
+		        :choices ["0", "1", "2", "3"]
+		        :effect (req (let [c (Integer/parseInt target)]
+				       (resolve-ability state side
+                                        {:choices {:req #(and (contains? % :advance-counter)
+							      (or (= (last (:zone %)) :ices)
+                                                                  (= (last (:zone %)) :content)))}
+				                   :msg (msg "remove " c " advancements from " (if (:seen target) (:title target) "a card"))
+				                   :effect (req (add-prop state :corp target :advance-counter (- c))
+				                                (swap! state update-in [:runner :prompt] rest)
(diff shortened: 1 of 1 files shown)


nealterrell:
If you use `(effect` then you do not put `state side` in function calls, as they are inserted automatically. This also means you can only call functions that need state and side as their first arguments. You can use `req` instead of `effect` for flexibility but you then have to pass state and side manually. 

See Analog Dreamers for the fix for the end run code. It does require both statements, one to update the state and one to end the run.

JoelCFC25:
I'm still stumped on finishing up Exploratory Romp--when I add in the lines stolen from Analog Dreamers, I get: 

`java.lang.IllegalArgumentException: Don't know how to create ISeq from: clojure.lang.Symbol`

...pointing at the first paren in the 3rd line:  `:effect (effect (run target`

Moving on for now, I've added Raymond Flint. This one worked on the first try! The only limitation it has is that the HQ access granted when BP is gained is limited to 1 card even if HQ Interface is installed (the same problem Gang Sign currently has).

JoelCFC25:
I had an idea in #308 for how to let Raymond's HQ access stack with an HQ Interface.

nealterrell:
I don't definitively know the cause of your new error, Joel, but sometimes switching all `(effect` to `(req` (inserting `state side` when needed) fixes weird stuff like that. I can't explain it.

nealterrell:
Any updates on your problem here, @JoelCFC25 ?

JoelCFC25:
Nope, I haven't been able to revisit Exploratory Romp yet. 

As for Gang Sign/Raymond Flint taking account of an HQ Interface, I did try the `(take :hq-access (:shuffle (:hand corp)))` idea on Raymond and got some sort of "illegal cast" error. I needed to call it a night so I left it there and didn't investigate further.

nealterrell:
It's probably because `:hq-access` is a keyword, and `take` wants a number as the first argument. You need the `:hq-access` of the Runner, which would be in `(:hq-access runner)` (if in `effect` or `req`; otherwise `(get-in @state [:runner :hq-access])`).

mtgred:
So what's the status? Does it work properly?

JoelCFC25:
Raymond Flint works--maybe I should split him off into a separate pull to get him added? 

So far I have only gotten Exploratory Romp to work in a way that still gives the Runner the extra Successful run / Jack Out prompt after doing the run ability. My attempts to dispose of the prompt and add in the `handle-run-end` haven't worked.

JoelCFC25:
I'm going to split out Raymond Flint into a new pull and put Exploratory Romp on the shelf for now.

mtgred:
Yeah it might better to put Raymond in another pull request.

There is also no way to enforce that the target in the attacked server because the `:choices {:req #()}` function only knows about the clicked card and doesn't have access the game state. Ideally it should be changed to have a `(req)` instead with the clicked card as `target`.

PR #359, opened 2015-07-13
fix Frame Job, Monolith tweak, add Helium-3 Deposit

### LINKED ISSUE #277: Frame Job is unplayable
Frame Job cannot be played from hand at all (no clicks or credits deducted, card is immobile and doesn't respond at all). I manually boosted my clicks and credits prior to testing it to be sure I met all the requirements. 

The current code is: 

`{:additional-cost [:forfeit] :effect (effect (gain :corp :bad-publicity 1))}`

I'm not sure if this is where it's bogging down, but the forfeit is not actually an additional cost, it's the outcome of the event. If `:additional-cost` got dropped and `:forfeit` was moved into the `:effect`, would that make it work? Basically it would be altered to look more like Data Dealer.

JoelCFC25:
Fixes #277. Removes the forfeit from the cost and makes it the effect instead. Tested and verified working.

### DIFF
--- a/src/clj/game/cards.clj
+++ b/src/clj/game/cards.clj
@@ -1122,6 +1122,15 @@
    "Hedge Fund"
    {:effect (effect (gain :credit 9))}
 
+   "Helium-3 Deposit"
+   {:choices ["0", "1", "2"] :prompt "How many power counters?"
+    :effect (req (let [c (Integer/parseInt target)]
+                   (resolve-ability
+                     state side
+                     {:choices {:req #(contains? % :counter)}
+                      :msg (msg "add " c " power counters on " (:title target))
+                      :effect (effect (add-prop target :counter c))} card nil)))}
+
    "Hemorrhage"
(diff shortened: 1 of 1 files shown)


JoelCFC25:
Implemented Helium-3 Deposit. Since the system doesn't (yet) distinguish between agenda counters, virus counters, or power counters, technically this can be used incorrectly, so it does require players to choose a valid target. Tested and verified that it adds power counters to Corp or Runner cards (Atman being the test case as the only Runner card you'd likely ever target).

JoelCFC25:
I'm on a roll tonight--with the insight granted by @nealterrell about the differences between `:req` and `:effect` I finally pulled together Monolith's discounted program installs after getting stuck for several days.

DominicKexel:
@JoelCFC25 So, what is the difference between `:req` and `:effect`? :-)

nealterrell:
These look good, Joel. Nice job :).

mtgred:
:+1:

Issue #358, opened 2015-07-11
Architect doesn't print to the console (possible to cheat)


eah13:
Architect, when rezzed, does not print that the corp is using the look at 5 cards sub.  This allows the corp to cheat by looking at the top 5 cards then choosing no install.  Seems like it should print to the chat that the corp is looking at the top 5 cards and also that the corp chooses not to install

nealterrell:
Architect does print a message, but only if the Corp does not click "No install". So you're right, a Corp can fire the subroutine to peek at the top of R&D without alerting the Runner. 

My repo is in a difficult state right now, so I can't fix this quite yet, but if anyone else wants to take a crack at it... maybe try removing the `:req` and augmenting the effect to make sure target is not `"No install"` before calling `corp-install`. The message _should_ then print regardless, and a second message to indicate the installation will fire from `corp-install`.

DominicKexel:
Maybe I'll take a look if I find some time this weekend

PR #366, opened 2015-07-15
Added option to display a msg when an ability gets activated

DominicKexel:
Fixes mtgred/netrunner#358.

Also added it to Executive Boot Camp; it's probably usefull to a lot of
other cards where the standard :msg is not sufficient.

It enables your opponent to see what you're doing and helps prevent
cheating.

Let me know what you think.


DominicKexel:
Related to mtgred/netrunner#301 and mtgred/netrunner#367 and contains the fix for mtgred/netrunner
#358 which is already in the other pull request, because I suck at using git.

mtgred:
:+1:

PR #371, opened 2015-07-15
Fix get-virus-counters to see Hiveminds that are hosted.

nealterrell:
Simple fix to find Hiveminds that are hosted and count their tokens. Fixes the un-reported issues with Medium etc. not getting Hivemind bonuses when Hivemind is on Progenitor.


mtgred:
:+1:

PR #372, opened 2015-07-15
Fix Noise's ability interaction with 0-strength Parasite installs

### LINKED ISSUE #364: Parasite onto 0-str ice doesn't trigger Noise's ability
When playing Noise, installing a Parasite onto a 0 strength ice correctly kills the ice immediately, but doesn't trigger Noise's ability to mill the Corp due to the virus install. 

For all other ice strengths the mill is occurring.

nealterrell:
Fix #364, by manually triggering the `:runner-install` event if we note that Parasite's trash condition is triggered during the normal install process. Since Parasite ends up trashed as part of the install process, the `trigger-event :runner-install` in `(runner-install)` does not fire normally.

### DIFF
--- a/src/clj/game/cards.clj
+++ b/src/clj/game/cards.clj
@@ -1831,7 +1831,11 @@
 
    "Parasite"
    {:hosting {:req #(and (= (:type %) "ICE") (:rezzed %))}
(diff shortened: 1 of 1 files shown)


mtgred:
:+1:

Issue #361, opened 2015-07-13
The Source doesn't work


JoelCFC25:
Installing The Source leaves the Corp in a state where no agendas can be advanced at all. It also isn't trashing itself when the Runner steals an agenda. 

It got modified by #260, but it looks like there is an `{:effect` missing in a couple places in the `:events` section. I'll try to look at it tonight.

nealterrell:
Whoops. Thanks for cleaning this up.

JoelCFC25:
Fixed at #374.

PR #374, opened 2015-07-16
fix Gang Sign to work with HQ Interface

### LINKED ISSUE #308: Gang sign does not stack with HQ Interface

JoelCFC25:
Partially fixes #308. 

Still left with the issue of how to handle multiple Gang Signs, but at least for now this will grant an additional access for every HQ Interface installed.

### DIFF
--- a/src/clj/game/cards.clj
+++ b/src/clj/game/cards.clj
@@ -992,7 +992,7 @@
 
    "Gang Sign"
    {:events {:agenda-scored {:msg "access 1 card from HQ"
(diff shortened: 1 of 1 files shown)


JoelCFC25:
Fixed #361 so The Source works again.

mtgred:
:+1:

Issue #370, opened 2015-07-15
Grifter freezes turn


alcohollywood:
In a rig with Grifter installed, the End Turn button does not activate until Grifter is trashed, even when the runner makes a successful run. The 1 credit the runner gets for a successful run is not supplied as well.

PR #375, opened 2015-07-16
Fix grifter fixes mtgred/netrunner#370

DominicKexel:
The gain function does not take a vector or similiar, since it uses & args.
Maybe the gain function should check for an even amount of args.

The function in :effect is called with 4 arguments, so the anynomous function has to use %4 instead of nil.


DominicKexel:
It seems this time I was not too stupid to create a new branch

mtgred:
:+1:

Issue #141, opened 2015-03-24
Networking 


HongkongKoma:
When playing "Networking" (Humanity's Shadow 84), I paid the 1 cred to return it to my hand but the card instead went to heap. Manually dragging it into HQ did not work.

mtgred:
I just tested Networking and it seems to work fine. Can you describe the situation when it didn't work for you? It will help me reproduce the issue and analyse what's causing it.

HongkongKoma:
Hi mtgred, thanks for your reply. 
When I play Networking, the tag gets removed and a box appears asking me if I want to pay the 1 cred to get the card back in hand. Click: Yes. The 1 cred gets paid, but Networking does not go to hand. The card stays in my Heap. And I can not move it manually to HQ either. 

Yesterday, I got a Sure Gamble back in hand instead. 

I hope this helps  you to find the error.

JoelCFC25:
I replicated this bug. Had only a Sure Gamble in the Heap while carrying multiple tags, played the Networking--1 credit was lost, but Networking remained in the Heap and Sure Gamble was returned to my Grip.

HongkongKoma:
Thanks for fixing!

PR #376, opened 2015-07-16
Fixed networking. Fixes mtgred/netrunner#141

DominicKexel:
If a card is played and is added to the :discard zone, it is added at the
end of the collection (since the front parameter is not set).

So, to take the "top" card of archive, you have to take the last card of
the :discard collection.


mtgred:
:+1:

PR #378, opened 2015-07-16
implement Raymond Flint

JoelCFC25:
Tested and verified as working--and it will grant extra accesses for installed HQ Interfaces.


JoelCFC25:
Added the trace to Flare...someone in chat had reported it was missing.

mtgred:
:+1:

PR #384, opened 2015-07-17
link updated site guide on About page

JoelCFC25:
Mythbuilder's guide is getting increasingly out of date due to all the changes in the last couple months. @darlingsensei wrote up a nice new one right here in the wiki that can more easily be kept current.


mtgred:
:+1:

PR #385, opened 2015-07-17
Fix Parasite infinite recursion with Grimoire out.

nealterrell:
Introduced by #372, installing Parasite onto a strength 0 ice with Grimoire installed causes an infinite recursion loop, whereby Parasite wants to trigger its Noise ability fix, which triggers `:runner-install`, which Grimoire handles to add a counter to the virus, which triggers Parasite to re-check the ice strength, which triggers the Noise ability fix, which triggers `:runner-install` etc.

Fix: unregister all events from Parasite the moment it notices that its host strength is 0 or less. It will then not listen to any more events causing these types of situations. Incidentally, this would have fixed the crashes from NEXT Bronze and Wraparound a long time ago.


mtgred:
:+1:

PR #391, opened 2015-07-18
Fix for Autoscripter

DominicKexel:
Hosted cards no longer stay in thier original zone, but gets moved to a zone called `:onhost`.

When a card gets moved, the previous zone is now saved in `:previous-zone`

### DIFF
--- a/src/clj/game/cards.clj
+++ b/src/clj/game/cards.clj
@@ -179,8 +179,12 @@
    "Autoscripter"
    {:events {:runner-install {:req (req (and (has? target :type "Program")
                                              (= (:active-player @state) :runner)
-                                             (empty? (filter #(has? % :type "Program")
-                                                             (map first (turn-events state side :runner-install))))))
+                                             ;; only trigger when played a programm from grip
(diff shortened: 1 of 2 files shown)


nealterrell:
Nice job with Exile and Autoscripter. I don't know the hosting mechanism well enough to predict any complications from using `:zone :onhost`, so @mtgred will have to chime in on that. Have you tested this change with cards that are currently able to target hosted cards when applicable? (Foxfire and The Personal Touch would be two good prototypical cards to test.)

DominicKexel:
Thanks.

The zone of a card does not matter in this case. Important is if a card is installed or not, regardless if it is hosted or not. 

Example: 

- when you play a Parasite on a piece of ICE, it's installed and hosted. 
- if you use PW's ability to host a card, that card is hosted but *not* installed.

The cards you mentioned correctly check the  `:installed` property of thier target, so there's no problem :-)

DominicKexel:
So, the FAQ says that Foxfire can only target installed cards, so nevermind about what I said about it. Thanks for the info, Neal

mtgred:
:+1:

Issue #393, opened 2015-07-18
Servers merging


nealterrell:
This is related to #156, but with reproducible steps and a theory on why.

1. Install an asset/agenda into Server 0 (call this Asset0).
2. Install an asset/agenda into Server 1 (call this Asset1).
3. Install an ice into Server 1 (optional, helps visualize the bug).
4. Install an asset/agenda into Server 0 (call this Asset2).

Expected: Asset2 replaces Asset0 in Server 0. Server 1 has Asset1 and ice.

Actual: Server 1 is no more. Asset1, ice, and Asset2 are now all in Server 0.

Why: I believe this happens because of the following logic flow:

* Corp chooses to install Asset2 in Server 0.
* `corp-install` checks to see if an asset/agenda is already in server 0, sees that there is, calls `(trash)` to get rid of it.
* `trash` calls `move`. Asset0 gets moved to Archives. `move` checks to see if Asset0 was in a remote, and if that remote has no content and no ice. Because this is true, the remote is removed from `[:corp :servers :remote]`. 
* Because remotes are numbered implicitly by their position in `[:corp :servers :remote]`, Server 1 is now Server 0 by virtue of being first in the list.
* `corp-install` resumes its command to install Asset2 in Server 0, unaware that the old Server 0 is gone and Server 1 is now Server 0. Asset2 is installed into Server 0 `:content`, blissfully unaware that something has gone horribly wrong.

DominicKexel:
Interesting. If no one is already working on this, I'll try to fix it this evening.

nealterrell:
Go for it. The easiest solution might be to just wait to trash the old asset until after the new one `move`d to the server. That will technically trigger events in the wrong order though... it will trigger `:corp-install` before `:trash` when according to game rules it should go the other way. Not that I know of any interactions where that would matter... hmm...

DominicKexel:
I had the same idea, and the same worry (don't want to move the bug to the future when a card comes out where it actually does matter).

Ideally Servers would not be numbered implicitly IMHO, but I guess that would requiere a lot of code changes (but I'll have to take a look first).

I'll have to look at the source first to decide what would be the best approach.

Issue #156, opened 2015-04-14
Installed a card over one of my servers, second server merged into it?


Agnate:
I had a rezzed Director Haas installed in Server 1 and an unrezzed Adonis Campaign in Server 2. I installed a Jackson Howard over top, causing Director Haas to be trashed (as expected). Server 2 then merged into Server 1, such that there was a Jackson Howard AND an Adonis Campaign in my Server 1. It said "Server 1 (2)" meaning it even registered that there were 2 cards there.

Server 0 contained an unrezzed Eli 1.0 ICE and a rezzed Haas Arcology AI with 2 counters on it.

mtgred:
Thanks for reporting the issue. It will fixed soon.

Agnate:
Just replying here to say that this happened to me again today.

JoelCFC25:
Fixed at #394.

PR #394, opened 2015-07-18
When installing a card in a remote, don't remove that remote

DominicKexel:
fixes mtgred/netrunner#393 and mtgred/netrunner#156

When you installed an asset/agenda in a remote n that already had an
asset/agenda, that already installed card got trashed.

Then that server was removed from the game, and the remove server n+1
became server n, since the remotes are implicitly numbered.

Then the new asset/agenda was installed in server n, which was the wrong
server.

### DIFF
--- a/src/clj/game/core.clj
+++ b/src/clj/game/core.clj
@@ -115,7 +115,7 @@
 
 (defn move
   ([state side card to] (move state side card to nil))
-  ([state side {:keys [zone cid host installed] :as card} to front]
+  ([state side {:keys [zone cid host installed] :as card} to {:keys [front keep-server-alive] :as options}]
    (let [zone (if host (map to-keyword (:zone host)) zone)]
      (when (and card (or host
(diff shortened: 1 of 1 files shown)


mtgred:
:+1:

mtgred:
The server has been updated with the latest commit. There seems to be an issue with remote server removal now. Server 0 contains a rezzed Eve Campaign with no ICE, Server 1 contains an agenda with no ICE. I run server 0 and trash Eve Campaign. Server 0 is not removed. Agenda in Server 1 is scored. Server 1 is removed but Server 0 is still there.

DominicKexel:
Mmmh.... Didn't noticed while testing. I'll take a look.

DominicKexel:
Sorry for the inconvenience. Here's the fix for the fix: #406.

Trashing should only prevent the destruction of the server if the `:keep-server-alive` flag is set.

Issue #396, opened 2015-07-18
OH MY GOD FOXFIRE IS BROKEN!!!


DominicKexel:
Yeah, I couldn't believe it myself, but Foxfire does not work correctly.

Foxfire says:

>  If successful, trash 1 virtual resource or 1 link. 

There's no limitation that the card has to be installed, compared to, say Power Shutdown:

>   ...The Runner trashes 1 **installed** program...

So if the Runner hosts a The Helpful AI on The Supplier, you can't trash the The Helpful AI with Foxfire!!!

#### UNBEARABLE!!!

nealterrell:
So I actually disagree with this conclusion. The [newest FAQ](https://images-cdn.fantasyflightgames.com/filer_public/f5/2e/f52e015b-dd10-4290-9fba-9b9d0303e0cc/adn_faq_21.pdf) has a new entry that reads 

##### "Installed Cards

Unless otherwise noted, all card abilities that do not specify the state of a card can only affect installed cards.

_Example: The Corporation resolves a subroutine that says “|The Runner trashes 1 program.” The Runner must trash one of his or her installed programs_"

It seems this should apply to Foxfire as well.

DominicKexel:
Woah. Thanks for the information. Didn't take a look at the new FAQ yet. Right, then Foxfire is fine  as it is.

nealterrell:
Yeah, this clears up a lot of poorly-worded cards in a way that makes the most sense, including Net-Ready Eyes.

DominicKexel:
True. So no more trashing programs from Personal Workshop.

Now that Foxfire is nerved I won't play it anymore!

nealterrell:
Literally ruins SO MANY Corp builds.

PR #397, opened 2015-07-18
Foxfire can trash hosted but not installed cards

DominicKexel:
since there's not limitation that the target card has to be installed

fixes mtgred/netrunner#396


DominicKexel:
The new FAQ changed how Foxfire works, so no need for this change.

PR #400, opened 2015-07-19
Temporary Run Credits

laganojunior:
I added tracking of temporary credits that are given at the beginning of the run and are returned at the end of the run. The amount of credits available on just the run (and will be returned at the end) is displayed on the credit count if applicable. I've implemented stimhack and bad publicity credits here.

### DIFF
--- a/src/clj/game/cards.clj
+++ b/src/clj/game/cards.clj
@@ -2606,6 +2606,7 @@
    "Stimhack"
    {:prompt "Choose a server" :choices (req servers)
     :effect (effect (gain :credit 9)
+                    (gain :run-credit 9)
                     (run target {:end-run
                                  {:msg " take 1 brain damage"
                                   :effect (effect (damage :brain 1 {:unpreventable true :card card}))}}
--- a/src/clj/game/core.clj
+++ b/src/clj/game/core.clj
@@ -738,6 +738,10 @@
       "New remote" [:servers :remote (count (get-in @state [:corp :servers :remote]))]
       [:servers :remote (-> (split server #" ") last Integer/parseInt)])))
 
+(defn add-bad-publicity-credit [state]
+  (swap! state update-in [:runner :run-credit] + (get-in @state [:corp :bad-publicity]))
+  (swap! state update-in [:runner :credit] + (get-in @state [:corp :bad-publicity])))
+
 (defn run
   ([state side server] (run state side server nil nil))
   ([state side server run-effect card]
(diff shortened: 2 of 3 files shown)


nealterrell:
This is excellent :+1:. I've done some testing on my dev machine and things are working beautifully.

1. Making a run with 1 real credit and 1 BP, my UI shows "2 credits (1 for run)". I can pop SMC to go to "0 credits", install a 0-cost program, success.
2. Using Stimhack with 1 real credit and 1 BP, my UI shows "11 credits (10 for run)". Pop SMC to install Torch, 0 credits, success.
3. Using Stimhack with 0 real credits, "9 credits (9 for run)". End run, back to 0 credits. Success.
4. Using Stimhack with 0 real credits, blah blah. Use 4 credits to pay for R&D Interface on Personal Workshop. Successful run, back to 0 credits. Success.

@mtgred may want to comment on the UI, as I believe he had a plan for how to show the run credits (in orange? I forget), but otherwise this is a great first contribution :).

DominicKexel:
Awesome :+1

nealterrell:
@DominicKexel, perhaps you could help @laganojunior update `can-pay?` with `:run-credit` taken into account?

laganojunior:
@nealterrell No, I just picked up the source code and worked on it because I was running a deck with lots of BP and got annoyed that I had to keep track of BP credits for the runner. I figured that someone else was probably working on this, and perhaps they have a better design.

The UI does need work. Particularly, the run credit count isn't manually adjustable or back-trackable. E.g. if you want to redo a stimhack run and you manually adjust credits, or you manually adjust credits to pay through a pop-up window or Gagarin ability. Maybe something like "Added X run credits" and "Returned X run credits" in the chat log would help at least with the back-tracking part.

@nealterrell 
I believe that `can-pay?` already "just works" since I just add all of the run credits to the general credit pool during the run.

nealterrell:
Right on. I see what you mean about using the manual credit adjust -- the up/down arrows use `change` instead of `pay`. Try playing around with the UI -- maybe you can get a completely separate line for Run Credits with its own adjustment controls that only shows during a run when there are run credits to spend.

justinliew:
Looks great. I'll sync up and use this as a basis for the recurring credits.

mtgred:
Awesome :+1:

mtgred:
Using Tollbooth during a run doesn't take away run credit. I think it is because Tollbooth uses `lose` instead of `pay`.

@laganojunior Can you update the run code to also include `lose`?

laganojunior:
@mtgred I imagine it won't be hard to add more ways to use run credits. Can you take a look at https://github.com/mtgred/netrunner/pull/417 and tell me if that's a better way to represent run credits? It'll change the way I would have to implement this fix for tollbooth, and I don't want to have to implement the change twice.

PR #401, opened 2015-07-19
Dockerize

gaverhae:
With these changes, the application runs completely inside of Docker containers. This has two advantages:

* All dependencies are made explicit in the `docker-compose.yml`.
* For people who have `docker-compose` installed, running the project is simply `docker-compose up {dev|prod}`.

For people who do not want to use `docker`, the `docker-compose.yml` file is still valuable, up-to-date documentation that describes how to start the complete system.

`docker-compose up prod` will start the Clojure server in "production" settings, which means it will compile the uberjar and run the server from there.
`docker-compose up dev` will start a Clojure REPL server on `localhost:7001` for development. The server itself is not started and must be run manually by evaluating `(future-call dev)` in the `game.main` namespace as before.

To connect to the REPL server, one can run `lein repl :connect localhost:7001` from the host machine. A similar command should allow any Clojure development tool to connect to the running REPL server.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,13 +2,10 @@ mongo:
     image: mongo:3.1.5
     volumes:
         - /tmp/netrunner/mongo-data:/data/db
-node:
+nodeBase:
     image: node:0.12.7
     volumes:
         - .:/workdir
-    links:
-        - mongo:mongo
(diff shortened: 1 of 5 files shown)


mtgred:
:+1:

Issue #259, opened 2015-06-17
Feature Request - Icebreaker break ICE button


MasterAir:
Great work. This is really looking good.

One thing I thought might be useful and not too onerous to code in is a 'break all subroutines' button for Icebreakers with multiple paid abilities. If you're breaking curtain wall with corroder (something that shouldn't be done too often but is occasionally advisable) it takes 22 mouse clicks. A button that matches strength and breaks all subroutines of the Ice being encountered (and tells you how much it costs) would be really good.

mtgred:
I like the idea and will try to code it this week.

nealterrell:
This would be a direct nerf to Curtain Wall -- inducing carpal tunnel is one of its most important uses :+1:

mtgred:
Neal, you want to do it?

I just realised i don't know how the ICE strength code works while working on Patch.

BTW, the commit with the hosting support for Operations has just been pushed. Patch doesn't seems to boost strength though. Can you have a look at it?

nealterrell:
I can look at Patch.

This issue could be a tricky implementation, though. I don't see there being enough information in the current icebreaker code to figure this out at runtime... but maybe I'm overlooking some things.

```clojure
{:abilities [
  {:cost [:credit 1] :msg "break 1 code gate subroutine"}
  {:cost [:credit 1] :msg "add 1 strength" :effect (effect (pump card 1))}]}
```

how would you examine that structure to and determine the cost to break a strength 5 code gate? OK, the first ability is consistently an ability to break a subroutine, and it has a cost... but there's no non-text data about how many subroutines are broken. (Can't just parse it, either... see Switchblade). The pump effect has a number to pump by, but it's not true that the second ability of all icebreakers is a pump effect (Deus Ex, fixed-strength). I might be able to figure out if the :effect of an ability is calling the `pump` function, but I don't know if that is a good long-term solution.

This feels like a lot of work, needing a solid design to be drafted, but maybe I'm not thinking Clojure-ly.

nealterrell:
Also, ideally this would display the cost of the automation in the click menu of the icebreaker, but to the best of my knowledge, labels on abilities are not dynamic -- they are set once and never updated. So I don't know how I could make the dynamic cost display work.

MasterAir:
Sorry, I had no idea how the ice was implemented when i suggested this. An alternative improvement to the UX that has a similar function would be to implement an input field where you could ask for a certain effect to be applied an integer number of times.

[...]

PR #405, opened 2015-07-20
Auto-pumping icebreakers; ability costs in UI; cards.clj split

nealterrell:
Implements a "Match strength with [ice]" fake ability for pumpable icebreakers that automates the "pump X times" tedium that the current one-pump-at-a-time UI requires (as requested in #259). Adds the cost of an ability to the card-click UI. Splits cards.clj into multiple files by card type to address a looming Clojure issue regarding our `cards` map.

### Icebreaker auto-pumping

When encountering an ice, any icebreaker that can break the ice's subtype will get a "fake" (not present in the card's ability definitions) ability to match strength with the encountered ice. That ability will (if money is available) repeatedly resolve an ability from the icebreaker's `:abilities` vector that is flagged with the `:pump` key. The UI for the button shows how much it will cost to match strength. 

curtain1

There's a lot of trickery to make this happen. For interested devs, see the definition of `auto-icebreaker` and `breaker-auto-pump` in cards-icebreakers.clj. In response to many game events (approach-ice, start-run, ice-strength-changed, ...), we check to see if we are encountering an ice and if the ice has a subtype that we (the icebreaker) can break. We calculate how many pumps it would take to match or exceed the ice's strength, and how much that would cost, then display that as a fake ability. In gameboard.cljs, we hook up a fake ability button that does not link to `game.core.play-ability`, but instead to `game.core.auto-pump` which applies the repeated pumping logic. 

Future icebreaker implementations should follow the `auto-icebreaker` pattern if they have abilities to increase their strength.

### Ability costs in UI

This only matters to cards with multiple abilities, but I extended the auto-pump cost display to all abilities with a `:cost`. Clicks and credits are both shown in the button for selecting the ability. 
contract-killer

Trashing a card is not part of its `:cost` vector, so trash icons are not shown. They can be faked in some cases, by adding "[Trash]: " to the `:label` of the ability.

### Cards.clj split

I had to split the `cards` map into multiple maps due to an issue with the JVM not wanting to compile large maps. While I was at it, I decided to split cards.clj into 11 files by card type, and then merge those files at runtime to keep our existing single-map structure for all cards. No changes are necessary to the rest of the code, __except__ that future cards need to be added to the correct file. I have merged all card changes up to commit 07b757a2b998a28151fc9142ede167aaf4465bf5.


mtgred:
Awesome! :+1:

DominicKexel:
Agree. Awesome work! :+1:

PR #406, opened 2015-07-21
Fix for the server trash fix

DominicKexel:
the fact if a server should be keeped alive must be passed as a parameter

to resolve-trash from corp install.


mtgred:
:+1:

PR #410, opened 2015-07-21
Fix for hosted icebreakers.

### LINKED ISSUE #408: London library + activate hosted icebreaker problem
After the latest patch i can no longer activate an icebreaker hosted on london library (tested with overmind and chameleon). I click the card but no menu is shown.

nealterrell:
UI only builds a card-view for cards that are in runner's `:rig`, which #391 changed. Adds `:onhost` to the list of zones for which card-views should be built. Fixes #408.

### DIFF
--- a/src/cljs/netrunner/gameboard.cljs
+++ b/src/cljs/netrunner/gameboard.cljs
@@ -90,7 +90,7 @@
                      (when (:installed card)
                        (handle-abilities card owner))
                      (send-command "play" {:card card}))
-            ("rig" "current") (handle-abilities card owner)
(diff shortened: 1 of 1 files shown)


DominicKexel:
Beat me to it. I was about to push a similar fix :-)

Before the change the `:zone` was still `[:hand]`, so the 

```
 "hand" (if (:host card)
                     (when (:installed card)
```
`case`/`if` fired.

Apparently I didn't test hosted ICE breakers; I'm very sorry for the inconvenience. Mea culpa.

nealterrell:
I've done worse :).

mtgred:
:+1:

PR #411, opened 2015-07-21
Fix for Progenitor granting tokens on purge.

### LINKED ISSUE #388: Medium on progenitor gains virus token after purge
If you have medium on progenitor with 0 virus tokens and the corp purges then medium gains one virus token.

nealterrell:
New event :pre-purge that happens before tokens are purged, so Progenitor can ensure the hosted card had at least one token before the purge happened. Fixes #388.


mtgred:
:+1:

PR #413, opened 2015-07-21
Fix for Kate discount when installing Parasite.

nealterrell:
Kate will now receive her discount when installing Parasite, or more generally when using any "install onto host" ability. Only call `:pre-install-cost` once we have determined where the card will be hosted.


mtgred:
:+1:

Issue #379, opened 2015-07-16
Edward Kim Ability Activating on Archive Run


timeh300:
Hi, Edward's ability should be disabled but not activate to re-trash an operation in archives on a run. This is completely irrelevant in 99.9% of cases, but can have an impact if Hostile Infrastructure is in play.

DominicKexel:
Shouldn't be that hard to fix. Either some lambda mumbo jumbo in the card itself, or better change the trash function to ignore cards that are already in :discard.

gammanet:
if i recall the ruling, his ability actually does trigger when you run archives. 

28 Edward Kim
•	 If the first operation Edward Kim accesses is in Archives,
then he cannot trash another operation with his ability that
turn.https://images-cdn.fantasyflightgames.com/filer_public/f5/2e/f52e015b-dd10-4290-9fba-9b9d0303e0cc/adn_faq_21.pdf

realize this is annoying with hostile infrastruture, perhaps if it was a prompt on the first one you access?

nealterrell:
What needs to happen here is for Kim's ability to not call `(trash target)` if target is in Archives. Everything else is working fine.

DominicKexel:
Since the rulebook says

> Trash: The act of moving a card to its owner's trash pile.

and accessing archives:

> Archives: The Runner accesses all cards in Archives and any upgrades in its root. The Runner turns all cards faceup when accessing them, and does not need to keep them in order. The Runner steals all agendas in Archives and **cannot trash cards that are already in Archives**. After accessing Archives, all cards in Archives return to Archives faceup.

I think this should be handled by the `trash` function; this is nothing special to Kim's ability, but a general rule that cards in archive can never be trashed. The game engine should not allow that.

PR #420, opened 2015-07-22
don't trash  cards already in :discard

DominicKexel:
fixes mtgred/netrunner#379

Show different message if a card can't be trashed by Edwards ability
because it is already trashed

### DIFF
--- a/src/clj/game/cards-identities.clj
+++ b/src/clj/game/cards-identities.clj
@@ -39,8 +39,10 @@
 
    "Edward Kim: Humanitys Hammer"
    {:effect (effect (gain :link 1))
-    :events {:access {:req (req (= (:type target) "Operation")) :once :per-turn
-                      :msg (msg "trash " (:title target)) :effect (effect (trash target))}}}
+    :events {:access {:once :per-turn
+                      :req (req (= (:type target) "Operation"))
+                      :effect (effect (trash target))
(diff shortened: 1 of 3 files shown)


DominicKexel:
So, I tuned up my git skills to create a nice, single-commit pull request that is up-to-date with the master branch.

nealterrell:
@mtgred : the purpose of this fix is to stop Edward Kim from receiving Hostile Infrastructure damage when accessing an operation in Archives. Because his ability auto-`trash`es any operation he accesses, he was triggering Hostile even if the operation was in archives. Dom took the approach of preventing `trash` from doing anything if the card is already in `:discard`, so any future effect that "trashes" things in Archives won't trigger `:trash` events.

mtgred:
What happens if Edward accesses an Operation in Archives and then another elsewhere? Does the 2nd operation gets trashed?

nealterrell:
According to [this link](https://www.reddit.com/r/Netrunner/comments/2uh36z/rules_question_does_edward_kims_ability_fire_on/), if his first Operation is accessed in Archives, he loses his ability for the turn. We currently do this.

DominicKexel:
The Operation in archives is still accessed, but the rules states that cards in archives are never trashed. So accessing a second Operation does not trigger Edward again.

Edward will now tell you that the accessed Operation can't be trashed because it's already in archives.

mtgred:
:+1:

Issue #389, opened 2015-07-18
Kate's ability fires even if you can't install the card...


DominicKexel:
Example:

Have 1 Credit, click on a programm in your hand that cost 3 credits to install it.

Kate's ability fires, but you didn't install the card (since you didn't have enough credits).

When you then try to play another hardware/programm, e.g. Clone Chip which costs 1, you don't get the discount.

JoelCFC25:
Could we just steal from Street Peddler and add a condition to the `req` that the cost of the targeted item be no larger than credits + 1?   I.e., `(<= (:cost %) (inc (:credit runner)))`

mtgred:
There would be an issue if another cost modifier like Modded or Street Peddler is used in combination with Kate.

Ideally the `can-pay?` function should take the `install-cost-bonus` into account. That way the `:req` in Kate, Street Peddler, Clone Chip or Self-Modyfing could check the discounted cost properly.

nealterrell:
I know that system best, I'll take a look. `:pre-install-cost` will need an argument to indicate if it's being called as part of the actual install or if it's "just to check if I can afford this."

PR #422, opened 2015-07-22
Make Kate's ability apply only if the install succeeded.

nealterrell:
Fix for #389 so that Kate's ability is not disabled when attempting to install something you can't actually afford. Move the "once per turn" logic to runner-install so her ability only gets disabled once an install succeeds. Works with installs off SMC, Clone Chip, and Modded.


mtgred:
:+1:

justinliew:
Cool, that clears it up. Thanks!

Issue #429, opened 2015-07-23
Card name hover/highlight issues


nealterrell:
Minor issues from #424 today:

1. Usernames are being linked to cards. "TheCodeTroll" highlights and links Troll.
2. Gimoire links as Grim.
3. e3 Feedback Implants -> Imp.

![highlights](https://cloud.githubusercontent.com/assets/10083341/8998141/8bf21f70-36de-11e5-85d1-46cf714f3eb8.png)

nealterrell:
Sometimes cards have strange output:
![image](https://cloud.githubusercontent.com/assets/10083341/8859649/59da84ec-3133-11e5-8bcb-8aad6cccbb68.png)

Saw this with Plascete Carapace and Grimoire as well.

mtgred:
I had the same issue with Drive By played on Adonis Campaign.

Upgrades in remote servers are also positioned weirdly now (next to each other instead of overlapping).

<img width="1552" alt="screen shot 2015-07-23 at 20 57 17" src="https://cloud.githubusercontent.com/assets/284114/8859715/603ef5f6-317f-11e5-9160-edd142dbc2a7.png">

mtgred:
I was just thinking Kati Jones, Plascrete Carapace, Adonis Campaign... it might be due to alternate art cards.

Grimoire has not alt art though.

DominicKexel:
OK, the cards needs to be sorted before changing the message. I'll provide a patch later this evening.

DominicKexel:
I proved a simply fix by using a regular expression with word boundaries. Don't know why I didn't do it from start.

"TheCodeTroll" should also no longer link Troll.

A user name that matches exactly a card title is still linked to the card. The question is: Is this worth "fixing" or should be keep this behaviour as easter egg?

DominicKexel:
BTW, is jinteki.net up ATM? I always get a 522 error from cloudfare...

DominicKexel:
Are you both running OS X? Just asking because I can't reproduce the issue where the card ID is shown in the log window. Makes me wonder if it's some kind of browser specific javascript issue.

nealterrell:
Chrome on Windows.

mtgred:
All the alt art cards consistently have the issue.

Dominic you have access to alt art cards on the live server. I'll mail you a json file to import then into your local db.

DominicKexel:
@mtgred my email is my github username @gmail.com 
What's special about the alt arts? Is there just another image code used?

[...]

PR #430, opened 2015-07-23
use regex patter with word boundaries...

DominicKexel:
..to match only the exact name.

Relates to case mtgred/netrunner#429

### DIFF
--- a/src/cljs/netrunner/gameboard.cljs
+++ b/src/cljs/netrunner/gameboard.cljs
@@ -158,7 +158,7 @@
   [:span item])))
 
 (defn add-image-codes [text]
-  (reduce #(.replace %1 (:title %2) (str "[" (:title %2) "~"(:code %2) "]")) text (:cards @app-state)))
+  (reduce #(.replace %1 (js/RegExp. (str "\\b" (:title %2) "\\b") "g") (str "[" (:title %2) "~"(:code %2) "]")) text (:cards @app-state)))
 
 (defn get-message-parts [text]
   (let [with-image-codes (add-image-codes (if (nil? text) "" text))]


mtgred:
:+1:

hydrozen:
@mtgred @queueseven  I’m trying to figure out a solution, but this seems to have broken links in the message pane for cards such as Snare!, Shock! and Shi.Kyù, I’m guessing because `!` and `.` are related to word boundaries.

queueseven:
@hydrozen yeah, there's already a fix in another currently open pull request

hydrozen:
@queueseven which PR?

JoelCFC25:
#490

hydrozen:
@queueseven ah its been merged already. Not sure I understand the new regexp ;) Cool that it is fixed though nice work.

queueseven:
@hydrozen it's a kind of hack because the javascript regex engine has neither balancing groups nor look-behind expressions.

hydrozen:
@queueseven mind explaining what it’s matching exactly, in words? I’m just curious.

It looks like this: `(^|[^\[\S])Snare!(?![~\w])` right? I’m not sure I understand both sides of the regexp.

queueseven:
The first part `(^|[^\[\S])` matches the start of the string or not [ or not non-white space; maybe I could just have used `(^|\s)`, but my brain as probably already melted at that time I wrote that.

The `(?![~\w])` at the end only matches if the string does not end with a word character or `~`.

The function scans the text for card names, like *Gang Sign*, and then replaces it with `[Gang Sign~01234]`, where `01234` is basically the name of the image file. So the `(?![~\w])` part ensures that after creating `[Pop-up Window~98787`], the `Window`-part does not get replaced with `[Window~56473]`, since there's also the card `Window`.

One idea was to simply change the server side code to send the `[card title~number]` strings itself, but I didn't want to change every line in `core.clj` and `cards.clj` where the title of a card is used in a message. Also, I think it's quite cool that you can type *"Is Gang Sign broken?"* in the chat and the card is linked automatically.

PR #438, opened 2015-07-24
implement Shiro and Hellion Alpha Test

JoelCFC25:
Shiro is tested and works with R&D Interface(s), stealing the same method as used with Raymond Flint/Gang Sign and HQ Interface. 

Added a line to `runner-install` to note in the Runner's `:register` when a Resource has been installed--this provides the support for Hellion Alpha Test. 

Made a very small tweak to the Invasion of Privacy message to make clear how many cards the Corp is eligible to trash.


mtgred:
:+1:

Issue #426, opened 2015-07-23
Faerie installed mid-run has bonus strength


MasterAir:
When playing with Geist I installed Faerie during a run using a clone chip. The Faerie was installed and immediately pumped to match the strength of the ICE currently being encountered (apparently at no cost).

DominicKexel:
Remeber what ICE it was?

nealterrell:
Did you get a chat log message about Faerie being boosted?

nealterrell:
Was the Faerie used previously to break the same ice? I am now seeing that a trashed icebreaker retains its "current strength", so the next time you install it will have the strength it was boosted to previously. This is really only visible to Sharpshooter and Faerie, who boost and then trash themselves before their strength would have been reset by the normal run flow.

MasterAir:
I'm unsure. If I see it again, I'll be more careful to note the specifics and report back.

The strength boost was visible only from the bottom right of Faerie, if I recall correctly. I'm not sure if I had used the Faerie on the same run. I do recall it happening multiple times in one game though.

nealterrell:
Thanks. I have confirmed what is likely the cause, that trashing an icebreaker while it is pumped does not clear the pump effect, so next time you install it from heap it will have the previously-pumped strength. Let me know if this is not the pattern you're seeing :).

PR #439, opened 2015-07-24
Dissoc pump effects when trashing icebreakers.

nealterrell:
Fix for #426. Dissoc `:pump` when deactivating cards so they do not keep their pumped strength when reinstalled from heap.


mtgred:
:+1:

Issue #441, opened 2015-07-24
Predictive algorithm doesn't prevent stealing


Nipazz:
Title. I was at 5 creds and was able to steal NAPD with predictive algorithm active.

JoelCFC25:
Other than the per-card costs on Fetal AI and NAPD Contract, steal costs in general are a yet-to-be-implemented feature:   https://github.com/mtgred/netrunner/wiki/Jinteki.net-Guide#gameplay-limitations

It's on the radar but there is no timeline. See some discussion in #260.

nealterrell:
This is now implemented!

mtgred:
It's deployed now.

PR #446, opened 2015-07-25
Steal cost bonuses: Red Herrings, etc.

nealterrell:
Implementation of the framework discussed in #260, #441, etc. Extends the current support for NAPD Contract and Fetal AI to third-party cards that can also impose costs on stealing agendas.

### Core changes

1. Added event `:pre-steal-cost` for third parties to handle to increase agenda steal costs. Triggered during `handle-access` once we know the card is an agenda.
2. Added function `(steal-cost-bonus)` taking a vector of costs to apply to the current steal attempt. (Example: Red Herrings applies `(steal-cost-bonus [:credit 5])` to agendas in the same server.)
2. Tweaked the `:steal-cost` tag of card definitions to rename as `:steal-cost-bonus` (matching other "bonus" frameworks) and be a `req`. Updated definitions of NAPD Contract and Fetal AI to match. (This change is future-proofing, as the `req` allows the card to define a steal cost in terms of the game state. This would allow easy implementation of a theoretical "Pay X credits to steal this card if it is not installed"-type agendas.)
3. In `handle-access`, merge the steal costs from `(steal-cost-bonus)` and from a card's `:steal-cost-bonus` function into one cost vector, and apply the existing `optional` prompt to resolve the additional costs.

### Card implementations

1. The Source: `(steal-cost-bonus [:credit 3])`.
2. Red Herrings: see above.
3. Strongbox: `(steal-cost-bonus [:click 1])` for agendas in the same server.
4. Predictive Algorithm: 2 credits for all agendas.
5. Utopia Fragment: 2 credits for each `:advance-counter` on the target agenda.

### All together now
steal-costs

NAPD with 3 tokens, Red Herrings, Strongbox, Predictive Algorithm, The Source, and scored Utopia Fragment

4cr (NAPD) + 5cr (Herrings) + 1click (Strongbox) + 2cr (Predictive) + 3cr (The Source) + 6cr (Utopia) = 20cr, 1click :dancers:

### DIFF
--- a/src/clj/game/cards-agendas.clj
+++ b/src/clj/game/cards-agendas.clj
@@ -90,7 +90,7 @@
    "Fetal AI"
    {:access {:req (req (not= (first (:zone card)) :discard)) :msg "do 2 net damage"
              :effect (effect (damage :net 2 {:card card}))}
-    :steal-cost [:credit 2]}
+    :steal-cost-bonus (req [:credit 2])}
 
    "Firmware Updates"
    {:data [:counter 3]
@@ -185,7 +185,7 @@
 
 
    "NAPD Contract"
-   {:steal-cost [:credit 4]
+   {:steal-cost-bonus (req [:credit 4])
     :advancement-cost-bonus (req (:bad-publicity corp))}
 
    "Nisei MK II"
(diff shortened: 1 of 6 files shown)


DominicKexel:
Nice!

JoelCFC25:
Superb! I wouldn't have gotten this in a million years.

mtgred:
Awesome! :+1:

nealterrell:
Thanks. This one was more interesting than the other bonus systems because of Strongbox, meaning the bonuses couldn't just be a number but instead had to be a pair like `[:click 1]`. `merge-costs` is an MVP function.

sperry023:
I just had an awkward Red Herrings interaction against a player who appeared to be new, so I'm not sure exactly what happened:
He Stimhacked from 0 to 9.
Spent 4 on ICE.
I rezzed Red Herrings.
He accessed, stole agenda, trashed Red Herrings for 2 (Encryption Protocol).
He went to -2.

This makes it look like it either applied Red Herrings as a penalty after the fact (the agenda was in the score area before the Red Herrings trash) instead of as a cost.

nealterrell:
Red Herrings has not been deployed to the server; your opponent did not get a prompt to pay more in order to steal the agenda. This sounds like a bug with Stimhack instead.

PR #450, opened 2015-07-27
Make lose and pay use a common function to apply costs

laganojunior:
Fixes the tollbooth credit loss to use run credits if possible. Will probably fix other ways to lose credits rather than paying credits, such as datapike.

See #400


mtgred:
:+1:

PR #452, opened 2015-07-28
Improved card list scrolling performance

Leffelin:
A lot of elements are being repainted when scrolling the card list.
This is taking a lot of time and the frame rate is restricted to around 20fps.
By moving the cards to a seperate layer, no more painting is needed when scrolling.
We are left with compositing only, and this is hardware accelerated on the gpu.
This boosts the performance to around 60fps.

### DIFF
--- a/src/css/base.styl
+++ b/src/css/base.styl
@@ -595,6 +595,7 @@ nav ul
   overflow: auto
   -webkit-overflow-scrolling: touch
   padding-top: 40px
+  transform: translateZ(0)
   flex(1)
 
 


mtgred:
Nice :+1:

DominicKexel:
There's still another thing that could be improved in the log window. Whenever a new message is created, each span element is recreated in the IRenderState function of the log-pane view. I don't know how you can improve this because this is just the way Om works. I thought about checking the number of log messages and only adding New ones to the DOM using e.g. infinity.js's ListView, but also don't know if it's worth the effort and if this even works with Om.....

PR #459, opened 2015-07-29
Cardpreview

DominicKexel:
This enables the card preview for prompts when cards are used. It doesn't create a span with fake-link. The entire button enables the preview, not just the text.

Also, I changed the card preview to use the :id attribute instead of the :title attribute. When using the :title attribute, it enables a tooltip window that shows the card id, which is not very nice. Sadly, you can't use custom attributes with Om/React, so the options are limited to :title or :id.


mtgred:
:+1:

Issue #213, opened 2015-05-30
hotkey for chat input selection


gparmer:
It would be quite nice if we could hit a hotkey (e.g. <TAB>), and auto-select the chat input box.  This is useful because:

1. the system works best with high resolutions, which makes mouse travel to the lower-right tedious.
2. since the chat input box is at the bottom right, it conflicts with "mouse-over" OS bars that are often at the bottom of the screen (i.e. where all running applications are listed).  Thus, attempting to select the chat input often results in that animation that (on Windows, at least) cover up the input box.
3. it would make it easier to chat, thus building community.

Great work, I hope this provides useful input.  Sorry I don't have the time or JS knowledge to help out with the implementation.

PR #460, opened 2015-07-29
enable chatbox with [hotkey] + [l]

DominicKexel:
Fixes mtgred/netrunner#213

You can now focus the chat inputbox with the access key `l`.

On Chrome, you can use [ALT] + [l]
On Safari, it's [ALT] + [l] on Windows and [ALT] + [CTRL] + [l] on Mac

The exact key combination is dependant on your [browser](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/accesskey).


mtgred:
:+1:

Issue #455, opened 2015-07-28
Leela is able to push runner cards into HQ


justinliew:
I was at 4 points and accessed Superior Cyberwalls from R&D. I had Turntable out and tried to click it to see if it was implemented and if there was a UI for it (it isn't, which I realized later) as I would have won switching it for Execute Bootcamp which my opponent had scored. This accidentally put Turntable into the Corp's HQ.

nealterrell:
Nice. Looks like Leela's ability only filters for `not :rezzed`, which of course applies to Runner's cards.

justinliew:
So would it be as simple as adding a condition of `= side :corp` as well?

mtgred:
@justinliew Yes that would do it.

JoelCFC25:
Fixed at #463.

PR #463, opened 2015-07-29
Fix Leela's ability to only allow her to push corp cards back to HQ

justinliew:
Previously the choices were not checking that it was a Corp card.


justinliew:
Addresses https://github.com/mtgred/netrunner/issues/455
I think I have the github/pull/branch stuff worked out now.

mtgred:
:+1:

Issue #462, opened 2015-07-29
Lockdown does not stop I've Had Worse from firing


thrdechelonelite:
Opponent ran into a Lockdown and did not break the subroutine. Then they ran a Fetal AI in the remote, which hit an IHW, causing the runner to draw 3 cards from it.

JoelCFC25:
Unfortunately Lockdown is one of several cards with far-reaching prevention effects that doesn't work at this time--others include Blacklist, Crisium Grid, etc.  

You can read about other current limitations in the [help guide](https://github.com/mtgred/netrunner/wiki/Jinteki.net-Guide).

PR #464, opened 2015-07-30
Implemented Lockdown

DominicKexel:
fixes mtgred/netrunner#462

Added a `:cannot-draw` flag that works mostly like the `:cannot-run` flag

### DIFF
--- a/src/clj/game/cards-ice.clj
+++ b/src/clj/game/cards-ice.clj
@@ -306,6 +306,10 @@
    {:abilities [{:msg "end the run" :effect (effect (end-run))}
                 {:msg "make the Runner gain 5 [Credits]" :effect (effect (gain :runner :credit 5))}]}
 
+   "Lockdown"
+   {:abilities [{:label "The Runner cannot draw cards for the remainder of this turn"
+                 :msg "prevent the Runner from drawing cards" :effect (effect (prevent-draw))}]}
+
    "Lotus Field"
    {:abilities [{:msg "end the run" :effect (effect (end-run))}]}
 
--- a/src/clj/game/core.clj
+++ b/src/clj/game/core.clj
@@ -192,10 +192,12 @@
 (defn draw
   ([state side] (draw state side 1))
   ([state side n]
-     (let [drawn (zone :hand (take n (get-in @state [side :deck])))]
-       (swap! state update-in [side :hand] #(concat % drawn)))
-     (swap! state update-in [side :deck] (partial drop n))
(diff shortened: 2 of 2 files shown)


mtgred:
:+1:

JoelCFC25:
Very cool--will this be easily modified to limit drawing to a specified number of cards in a turn? Because Genetics Pavilion is coming in the very next datapack (Runner cannot draw more than 2 cards in a turn).

DominicKexel:
@JoelCFC25 Sadly, I think not. `:cannot-draw` and `:cannot-run` use `(-> @State :runner :register)` which gets cleared when the turn starts. But it shouldn't be to hard to add another flag like `(-> @State :runner :cards-left)` and let Genetics Pavilion set it to 2, and `dec` it whenever a card is drawn. When it's 0, forbid drawing a card.

mtgred:
It's actually possible to modify function at runtime in Clojure. These libraries facilitate the process: https://github.com/technomancy/robert-hooke, https://github.com/MichaelDrogalis/dire

Implementing such kind of abilities with these libraries would be a piece of cake and wouldn't require to statically add new event triggers in every function that can modified by cards.

I initially wanted to use robert-hooke when I started the project. But the game engine was initially in Clojurescript which doesn't support the meta programming functionalities required by robert-hooke. Now that the game engine has been ported to Clojure it might be time to convert the event system with either robert-hooke or dire. It's quite a big overhaul and would require a ton of time and testing. I don't have enough time at the moment to commit to such a big refactoring. But if someone is up for it, it would simplify many many things. Even things like cost modification (eg. Kate, Industrial Genomic), damage modification (eg. The Cleaners), prevention (eg. Black List, Crisium Grid) could be done easily with these library.

DominicKexel:
Cool, I take a look at those libraries (But I won't promise anything :-)

DominicKexel:
OK, so I played a little with robert-hooke. 

Some stuff would have to be changed. 

For example, when an :effect is fired, the card causing the effect has to be passed to the function called, so Crisium Grid could create a hook function that has access to the Crisium Grid card, so the hook function has a way to check if the server that the run is made against is the server where Crisium Grid is installed. Or the hook function has to manually check if there's a server with a rezzed Crisium Grid.

Events are already fired with the `trigger-event` function, so Crisium Grid could wrap that and simply don't call it if the event is :successful-run and the arg is Crisium Grid's server.

But checking for stuff in the `:register` has also be encapsulated into a function, so Crisium Grid can also wrap that (or prevent putting a `:successful-run` into there).

Blacklist could wrap `move` and `shuffle-into-deck` to prevent cards from being moved from the heap.

Issue #428, opened 2015-07-23
Reina's ability is lost when Corp attempts to rez ice it can't afford


nealterrell:
Similar to #389, if Corp tries to rez an ice it can't afford, Reina's ability will be consumed. Easiest way to reproduce is to attempt to rez an ice you have the exact credits for; the first attempt will not go through because `pay` will fail, but the second attempt will succeed because Reina's ability won't affect the second attempt.

Could use a similar treatment as in #422. It's not a pressing issue so I will leave it here for someone else to use as a learning exercise if they are interested, otherwise I will fix it in a week.

justinliew:
Hmm, maybe I'll take this on first instead of the daunting recurring credits.

PR #467, opened 2015-07-30
Reina's ability is only consumed if the ICE was rezzed successfully.

justinliew:
Addresses https://github.com/mtgred/netrunner/issues/428


mtgred:
:+1:

PR #471, opened 2015-07-31
Implement The Helpful AI strength boost. Fix Corporate Troubleshooter.

nealterrell:
Not much to say. Use a temporary event handler to handle `:pre-breaker-strength` to give the targeted breaker +2 strength until end of turn. Also fix Corporate Troubleshooter which had a minor bug in it. (There is no `:end-turn` event.)


mtgred:
:+1:

PR #472, opened 2015-07-31
Code Siphon discount for install. 

nealterrell:
Apply effects in same order as stated on the card, just in case it ever matters. (Search and install; take a tag; shuffle stack.)


mtgred:
:+1:

PR #474, opened 2015-07-31
arrow up in chat box to get the last message

queueseven:
After sending a message, you can now press the arrow up key to copy the
last message into the text box again.

### DIFF
--- a/src/cljs/netrunner/chat.cljs
+++ b/src/cljs/netrunner/chat.cljs
@@ -7,6 +7,8 @@
             [netrunner.ajax :refer [GET]]
             [netrunner.main :as main]))
 
+(defonce last_msg (atom ""))
+            
 (def app-state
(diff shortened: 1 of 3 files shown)


mtgred:
In general I try to minimise the amount of code and features added. More code means more risk for bugs, performance hits and more stuffs to understand and maintain. So I tend to add features very parsimoniously and only if it's a big improvement.

In this case i'm not sure what would be a typical use case for this feature? Are there cases where a user want to retype something he just typed?

queueseven:
I really miss this feature sometimes, but as you said yourself it's just a very minor feature that most people probably won't miss. 

Actually I implemented this for myself so I could better test stuff regarding the client's log window, so the code was already there anyway. I understand you fully if you want to keep the code base small, no worries.

Issue #355, opened 2015-07-10
Suggestion: A player ranking system. Feature request


lpoulter:
It would be great if players had a ranking either elo or gilcko. This could be used to find even skilled opponents and to track your own performance.

eah13:
In addition to skill, a rating of sportsmanship or etc would be nice and along the lines of #354

nealterrell:
As someone with personal experience implementing a ranked ladder system for an otherwise casual online play community... I recommend passing on this. Once you attach a number to wins/losses, people get very aggressive and antagonistic over matches, even more than what we're already seeing. And as long as we have many unimplemented/bugged cards, we'll get no end of complaints about losing e-Penis Points due to site errors, demanding that points be adjusted manually, etc. 

I think a better solution has been proposed: adding self-chosen labels to games, along the lines of "Beginner", "Intermediate", "Advanced", "Tournament", etc.

DominicKexel:
I'll agree with @nealterrell here. Let's keep it casual by default, so we maintain a friendly environment. 

I could think of keeping some anonymous statistics like number of plays/wins for each fraction/deck like OCTGN does...

thehunter518:
Could implement a selection of  buttons for game creation that accomplish the same thing as typed labels, but with a standardized meaning (like the beginner/intermediate/advanced suggestion). This could be implemented in a way that was more visually obvious in the games lobby (e.g. all "beginner" games are highlighted green, or have a special icon) and could theoretically be used to filter desired game types.

Downsides? (other than a lot of extra programming work) 
Without an actual matchmaking system, people could just lie about their experience level in order to lure in less experienced players. While you can technically still do this now, the suggested system would enable people to do it more readily.

athayn:
i have to third @nealterrell. skill and sportsmanship ratings will make the jinteki atmosphere more competitive. the anonymity of online play has already brought out [hostility and harassment](https://github.com/mtgred/netrunner/issues/356). do we want it to feel like league night, a tournament, or casual play?

DominicKexel:
At least I want a way to report rage quitters. Maybe add a :-( smiley to the user name if someone joins your game that was reported several times as rage quitter or douche bag.

zaroth:
A tiny bit in this direction was done in the room split in #1215 (it's worth to read mtgred's comment there, too). However, we are still quite far from a full-blown ranking system, soon I'll try to write up a proposed roadmap for that.

Issue #345, opened 2015-07-07
Suggestion: Game type prompt


qmsm:
When starting a new game it would be good to have a drop down or some sort of prompt asking for what type of game you're looking for; tournament practice, social, new player etc. I know you can manually type something in the game name but most people don't. A quick mandatory field would force it.

JoelCFC25:
Related to #179. I think it's a good idea--not sure about it being mandatory. In my experience using game names to explicitly mark things as for testing or "not a real game" people don't read them anyway, so I have pretty low expectations about people respecting tagged/classified open games either.

DominicKexel:
I noticed that on jinteki.net nobody cares about the game title (it's always FooBar's game).

On OCTGN its common to use a pattern like `(league) LF  [side/playername] (level)`, for example:

- LF ThatOneUser
- LF either - beginner
- LF runner (advanced)
- SHL4 LF corp

A dropdown would be cool, but it should not be mandatory (or the default could be something generic like `regular`).

qmsm:
Haha, sounds like the argument I have with the President of my rugby club:

"People don't do as they're told!"
"We should put signs up to remind them."
"People don't read signs"
Repeat.

Quarg:
Perhaps a drop-down box is the correct solution, it would allow for having a few different "tags" for games, such as "private", "rematch", "tournament" etc, which can be shown in bold and brightly coloured so they can be distinguished at a glance; thus not requireing the users to actually read what is written.

JoelCFC25:
Would be handled by #755. I really want this!

Issue #179, opened 2015-05-03
New Feature: Play level


CptHilts:
I am not very good. I would enjoy games against other casual players. It would be nice to set a level of some sort that is public info for other player. This could include: Newbie, casual, prepping for tourney, jank list, hardcore

This could be toggled at any time.

JoelCFC25:
For the time being, use the title of the room you create to communicate the type of game you're seeking.

PR #481, opened 2015-07-31
User-selected indicators for skill level and seriousness when creating games

nealterrell:
Add options for "Skill Level" and "Game Type" ("seriousness") when creating a game. This information is displayed in the game list with colored icons that can be hovered over.  Addresses #179, #345, and #355, giving a simple visual cue as to what type of game a player is looking for. Attempts to be nonintrusive by setting default options to "Intermediate/Serious", which is probably what most site users look for.

skill-create

Skill options:
* Beginner (green)
* Intermediate (yellow) DEFAULT
* Expert (red)

Game Type options:
* Private (black)
* Friendly (green)
* Serious (yellow) DEFAULT
* Tournament (red)

Game lobby list:
skill-list

Hovering the icons shows a descriptive popup message, along the lines of "Looking for Expert opponent" and "This is a Friendly match".

I used built-in CSS color names "red", "green", and "yellow", which obviously can be tweaked.

### DIFF
--- a/server.coffee
+++ b/server.coffee
@@ -88,7 +88,7 @@ lobby = io.of('/lobby').on 'connection', (socket) ->
   socket.on 'netrunner', (msg) ->
     switch msg.action
       when "create"
-        game = {date: new Date(), gameid: ++gameid, title: msg.title,\
+        game = {date: new Date(), gameid: ++gameid, title: msg.title, skill: msg.skill, type:msg.type,\
                 players: [{user: socket.request.user, id: socket.id, side: "Corp"}]}
         games[gameid] = game
         socket.join(gameid)
--- a/src/cljs/netrunner/gamelobby.cljs
+++ b/src/cljs/netrunner/gamelobby.cljs
@@ -44,6 +44,8 @@
   (authenticated
    (fn [user]
      (om/set-state! owner :title (str (:username user) "'s game"))
+     (om/set-state! owner :skill "Intermediate")
+     (om/set-state! owner :game-type "Serious")
      (om/set-state! owner :editing true)
      (-> ".game-title" js/$ .select))))
 
@@ -54,7 +56,8 @@
        (om/set-state! owner :flash-message "Please fill a game title.")
        (do (om/set-state! owner :editing false)
            (swap! app-state assoc :messages [])
-           (send {:action "create" :title (om/get-state owner :title)}))))))
+           (send {:action "create" :title (om/get-state owner :title) :skill (om/get-state owner :skill)
+                  :type (om/get-state owner :game-type)}))))))
 
 (defn join-game [gameid owner]
   (authenticated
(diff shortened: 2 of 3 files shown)


DominicKexel:
Very cool

thehunter518:
I love this! Even experienced players like to play more casually sometimes. And I know I personally feel guilty when I'm trying to hard-test a tournament deck and my opponent was just trying to have some fun with his home-made janky Gagarin or Ken deck. Curious to see how well people will self-assess their skill/seriousness levels.

ZogStriP:
:+1:

mtgred:
I personally don't like forcing the player to choose a skill level or game type from a set list. What should he choose if it doesn't fit in the allowed option?

In general, I also try to keep the interface as simple and as intuitive as possible. In this case I'm afraid new users might be wondering what's the meaning E and F or the difference between "Serious" and "Tournament". Having to set these 2 options also increases the number of clicks required to create a game.

So I'm currently in favour of letting the creator write remarks about the game in the title. But this feature seems to generate a lot of enthusiasm so i will put it on hold for now and ask users about their opinion on it.

nealterrell:
I understand your points. I have seen some friction in chat lately between players with different ideas on how serious / fast they should be playing, and others who openly stress about bring new and feel intimidated about matching up with strangers. I thought these issues could be avoided pretty easily by communicating player intents when creating games. There is no objective standard for what a Beginner or Intermediate is, and I don't think there needs to be; the game types too aren't perfectly defined, but I think that's something that will organically come to a consensus in the community. And even if it never takes off, the default selections of Intermediate / Serious reflect what I perceive to be the norm for our users and wouldn't be any more clicks in most cases. 

But it's not a big deal if you think it's outside the scope of the client :).

queueseven:
I think the problem is the following:

If you create a game "Please don't join - testing stuff" or "looking for foobarsomeone", people still join your game, because nobody reads the game title: The default is *foo's game*, so it's used in 98% of all games. You see games you can't join, and read even more *foo's game* titles; so you become blind for the game titles.


Compare that to OCTGN, where you only see games that not already running, and you have to type a game title yourself; and even new players get after 2 minutes that you usually type *LF Corp* or *LF Runner (beginner welcome)* etc.

I think we could create a default game title by dynamically. When you use the default setting (Intermediate/Serious), the game title generated is *Serious game* or something like that. When you select e.g. Beginner/Friendly the generated title is *Friendly game for beginners*, Expert/Serious generates *Serious game for experts* etc. That would make it more clear what you're looking for, and you don't need the icons. No need to put the user name in the game title, it's redundant.

So neal's game example from above would look like this:

bild 2

I think that would make the whole thing clearer, and you really don't have to make more clicks if you just use the default options.

nealterrell:
I like that idea, Dom. A second hybrid could be what you suggest, but still allow custom titles. Use the game type selectors in the client only to generate a name; let game creator change to something more specific if desired.

queueseven:
@nealterrell Yes, that was my idea. The generated title is just a suggestion; you should still be able to change it e.g. when you look for a specific player or want to do a league game.

nealterrell:
@mtgred: I pushed the changes suggested by @queueseven. Removed the colored icons on the game list; use the game type and skill level lists to auto-create a game title instead of the default "nealpro's game".

The format for auto-generated game names is:

* "Private game" if the game type is "Private"
* "Friendly/Serious/Tournament game" if the skill level is "Intermediate"
* "Friendly/Serious/Tournament game for beginners/experts" otherwise

The default selection is Friendly/Intermediate, with a game title of "Friendly game". This can easily change. You can ignore the two lists completely and type out a custom name as before.

The implication is that the average user is an intermediate player looking for a friendly game; exceptions to this will cause longer game titles and draw attention in that way. 

We do not necessarily need the "Tournament" type, or perhaps it can be renamed -- my idea was that there is indeed a higher tier than just "serious" play, and by calling your game Tournament you might be more likely to get opponents playing Tier 1 decks (which chat users often request).

queueseven:
@nealterrell I think the "Tournament" type can stay, because there are already tournaments run on jinteki.net

nealterrell:
See #755.

Issue #470, opened 2015-07-30
Choosing order of access


nealterrell:
I am working on cleaning up the access routines, mostly in preparation for Haarpsichord Studios. Right now our access code works like this:

1. On successful run, gather all cards to access. For remote servers, this is all cards in the server's `:content`. For HQ, this is `:hq-access` random cards from hand and all cards in `:hq :content`. For R&D, this is the top `:rd-access` cards from R&D, plus cards in `:rd :content`. For Archives, this is all cards in `:corp :discard` and all cards in `:archives :content`. Note that cards in the server are gathered last for central servers.
2. Pass the cards to `handle-access`.
3. `handle-access` iterates through the group of cards and decides what to do with each. For non-trashable cards, a "Done" prompt is displayed with the card name. For assets/upgrades, an "optional" prompt is displayed to trash. For agendas, steal costs are calculated, and the agenda is sent to `steal`. 

We have two big issues with this approach:

1. Runner is supposed to choose order of access, even so far as being able to pick 1 card from R&D, then access an upgrade in root, then another from R&D on a multi-access run. This is especially important for remotes, where you need to be able to trash (for example) Oaktown Grid before trashing the asset whose trash cost the Grid is increasing. Also incredibly important for Archives (again, see Haarpsichord.)
2. All the access logic is run immediately without waiting for one card's access to resolve. Going back to Oaktown Grid, even if you get lucky and get to access Oaktown Grid first to trash it, its effect has already been applied to the trash cost of other cards in the server, because all the trash prompts are created at once without waiting for the resolution of accessing Oaktown Grid. Another problem scenario: Runner users Legwork; 3 cards are accessed, the first is NAPD Contract, the next is a 3 point agenda. The second agenda automatically shows up in Scored area while the NAPD prompt is active, but the Runner is supposed to completely resolve NAPD's access before gaining knowledge about the next card. Fearing Punitive, perhaps the Runner denies NAPD with his illegal knowledge.

There are other smaller issues, but let's cut to the chase.

I have a working prototype for choosing access in remote servers. It looks like this:

```clojure
(defn access-helper-remote [cards]
  {:prompt "Select a card to access. You must access all cards in this server."
   :choices {:req #(some (fn [c] (= (:cid %) (:cid c))) cards)}
   :effect (req (handle-access state side [target])
                (when (< 1 (count cards))
                  (resolve-ability state side (access-helper-remote (filter #(not= (:cid %) (:cid target)) cards))
                                   card nil)))})

[...]

PR #485, opened 2015-08-03
Choose order of accessing cards

nealterrell:
My system for letting the runner choose order of access when it is appropriate and unobtrusive. See #470 for pictures and a brief description. In general, we use both click-on-card and click-on-prompt to allow the runner to choose in what order they will access cards in a server. Intelligently (mostly) auto-selects cards when the context is clear. For example, if HQ has two rezzed upgrades and one unrezzed, and the runner selects to access "Upgrade in HQ" (as opposed to the names of the other two upgrades), they will not have to further specify; but if there are two unrezzed upgrades, they will then have to click on the one they want.

Addresses a number of issues with the simpler access system:

1. Oaktown Grid and Snare! in a server. Oaktown Grid rezzed prior to access. Live behavior: runner accesses in "random" order (based on install order), possibly having to pay 3cr to trash Snare and 4cr to trash Oaktown. Fixed behavior: runner can choose to access Oaktown and trash for 4cr first, then access Snare and trash for 0cr.
2. Accessing multiple unrezzed upgrades in a server. Live behavior: all cards accessed are printed to chat log, and then accessed in "random" order; runner can peek at what will be accessed later and decide to not trash the first thing they access to save money for the next. Fixed behavior: no reveal of what will be accessed until it is clicked.
3. Likewise, accessing multiple cards from HQ. Live behavior: runner sees he accesses NAPD and The Future Perfect; having only 4 cr, he chooses not to steal NAPD so he can play Psi game. Fixed behavior: must resolve the first access with no idea what's coming.
4. For Haarpsichord Studios, choosing order of access from Archives will be very important. Live behavior will theoretically give you the first agenda to enter Archives.

Known issues:

1. Ironically, Red Herrings and Strongbox like the old system and will no longer enforce their "including during the run in which Red Herrings is trashed" effect. Easy fix.
2. This prevents Edward Kim from accessing operations in Archives, which is supposed to disable his ability for the rest of the turn. Medium fix.

### DIFF
--- a/src/clj/game/cards-resources.clj
+++ b/src/clj/game/cards-resources.clj
@@ -124,9 +124,8 @@
 
    "Gang Sign"
    {:events {:agenda-scored {:msg "access 1 card from HQ"
-                             :effect (req (doseq [c (take (get-in @state [:runner :hq-access]) (shuffle (:hand corp)))]
-                                            (system-msg state :runner (str "accesses " (:title c)))
-                                            (handle-access state :runner [c])))}}}
+                             :effect (req (let [c (take (get-in @state [:runner :hq-access]) (shuffle (:hand corp)))]
+                                            (choose-access c '(:hq))))}}}
 
    "Ghost Runner"
    {:data {:counter 3}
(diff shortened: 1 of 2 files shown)


mtgred:
Awesome! :+1:

DominicKexel:
Great work as always. :+1:

PR #487, opened 2015-08-03
Implement Cerebral Static

nealterrell:
A simple implementation of Cerebral Static. Not 100% perfect, but good enough for now.

Unregisters the runner identity's events, so it won't receive notifications about board state changes. The overwhelming majority of runner identities simply do things in response to game events, so unregistering these events makes the identity effectively blank. Example: Kate will not receive `:pre-install-cost` and won't apply her discount; Noise won't receive `:runner-install` and can't mill when viruses are installed; Gabe won't receive `:successful-run` for his credits, etc.

A few identities don't handle events and are not correctly handled. All these can be manually accounted for with minimal effort:

1. Whizzard - keeps his recurring credits. Just don't click them!
2. Chaos Theory - keeps her +1 MU. Adjust MU in interface.
3. Quetzal - can still click her identity to "break 1 barrier subroutine"... but all it really does is print to the log.

I had to hook up a few things in core to make this work, but it's simple: currents can now receive `:leave-play` when they are trashed, and currents won't have their events hooked twice (an old bug that isn't noticeable, because most currents are "the first time" events.)

### DIFF
--- a/src/clj/game/cards-operations.clj
+++ b/src/clj/game/cards-operations.clj
@@ -46,6 +46,12 @@
                                      (gain state side :tag 1)
                                      (damage state side :brain 1 {:card card})))}}}
 
+   "Cerebral Static"
+   {:msg "disable the Runner's identity"
+    :effect (req (unregister-events state side (:identity runner)))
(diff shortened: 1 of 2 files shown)


mtgred:
:+1:

PR #492, opened 2015-08-03
Don't reveal R&D accesses. Fix Raymond Flint and Gang Sign.

### LINKED ISSUE #488: Corp can view Cards from R&D Accesses
After the latest server restart I noticed that the chat bar will tell the corp what cards were accessed by the runner

nealterrell:
My bad :). Fixes #488.


mtgred:
:+1:

PR #493, opened 2015-08-03
Haarpsichord Studios and prevent-steal framework.

nealterrell:
The real reason I was working on choosing access order. Uses a new register `[:runner :register :cannot-steal]` to prevent stealing agendas when the number of points stolen this turn is non-zero. Doesn't stop the `:access` events, so Explode-a-palooza, TGTBT, etc. still trigger their on-access effects :D.


nealterrell:
I realize there is a flaw here with Domestic Sleepers. I'll look into it.

mtgred:
:+1:

PR #498, opened 2015-08-05
Implement Escher.

nealterrell:
Implements Escher by letting the runner click two ice to swap, as many times as he/she wants to. Works pretty well! Thanks to mtgred and Joel for diverting me from the recursion-hell I was heading for with my first thought.

The `:prompt` here can be lifted for Tenma Line as long as you remove the recursive `resolve-ability (es)` at the end. Sunset can adapt the recursive helper to specify a server after asking the corp.

### DIFF
--- a/src/clj/game/cards-events.clj
+++ b/src/clj/game/cards-events.clj
@@ -85,6 +85,27 @@
    {:req (req (some #{:hq} (:successful-run runner-reg))) :msg (msg "derez " (:title target))
     :choices {:req #(and (has? % :type "ICE") (:rezzed %))} :effect (effect (derez target))}
 
+   "Escher"
+   (let [ice-index (fn [state i] (first (keep-indexed #(when (= (:cid %2) (:cid i)) %1)
+                                                      (get-in @state (cons :corp (:zone i))))))
+         eshelp (fn es [] {:prompt "Select two pieces of ice to swap positions"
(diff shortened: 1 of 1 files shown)


queueseven:
Cool. Very nice and clean.

mtgred:
Nice :+1: 

I had a different implementation in mind with a generic card swapping mechanism similar to the targeting mechanism. It would allow to specify conditions on the source card and target card, how many swaps can be done and a prompt. The swap would be done by drag & droping a card onto another. That would to allow implement cards Escher, Tenma Line, Sunset and Allele Repression in a simpler way.

The :replace-access for Escher would be something like:
```{:swap {:prompt "Swap ICEs" :req (req (and (:installed target) (= (:type target) "ICE") (:installed (second targets)) (= (:type (second targets)) "ICE")}}```

Tenma Line
```{:swap {:max 1 :prompt "Swap 2 ICEs"  :req (req (and (:installed target) (= (:type target) "ICE") (:installed (second targets)) (= (:type (second targets)) "ICE")}}```

Allele Repression
```{:swap {:max (req (:advance-counter card)) :prompt "Swap "
                :req (req (and (= (:zone target) [:hand]) (= (:zone (second targets)) [:discard])))}}```

If you feel like doing it, feel free go ahead. :) Else I will have more time in September to do it.

PR #500, opened 2015-08-06
Fix for Architect installing top card of R&D in NEH.

### LINKED ISSUE #278: Near-Earth Hub Installing Top Card with Architect Prematurely Draws
If, as Near-Earth Hub, you fire Architect's subroutine to install from R&D and install the top card creating a new server, the card ends up in HQ instead of on the table.

As far as I can tell, this happens because the "draw" condition technically fires before the "install" condition, so the top card--the one intended to be installed--ends up in HQ, and therefore isn't there to be installed.

(In the best tradition of not teaching granny to suck eggs, I'll stop here...)

nealterrell:
Fixes #278 by moving the card selected by Architect to `:play-area`, so that when `corp-install` triggers the `:server-created` event, the card is no longer on top of R&D and drawn by NEH's ability. The selected card stays in play area until the corp selects a server for it.

### DIFF
--- a/src/clj/game/cards-ice.clj
+++ b/src/clj/game/cards-ice.clj
@@ -15,7 +15,7 @@
                  :activatemsg "uses Architect to look at the top 5 cards of R&D"
                  :req (req (not (string? target))) :not-distinct true
                  :choices (req (conj (take 5 (:deck corp)) "No install"))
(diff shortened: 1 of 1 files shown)


mtgred:
:+1:

PR #501, opened 2015-08-07
New trash-installed function, add to ice that use it

JoelCFC25:
This adds a more generalized version of the `trash-program` and `trash-hardware` functions in `cards.clj`. Instead of relying on the Runner to trash something manually, we handle it with a prompt. Markus 1.0 and Information Overload now use it. 

Burke Bugs is fixed up to add the trace and have the Runner choose the program to trash--it was erroneously having the Corp choose.


mtgred:
:+1:

PR #502, opened 2015-08-07
implement Project Ares, fix Exile, partial Immolation Script

JoelCFC25:
Exile was wrongly generating the card draw only during the Runner's turn--it should happen every time regardless of whose turn it is. Added in a stub for Immolation Script to at least have it initiate an Archives run. 

Project Ares checks to see if it's over-advanced and that the Runner's rig isn't empty, whereupon it passes a multi-select prompt to the Runner to trash the requisite number of cards and then gives the Corp a bad publicity.


queueseven:
Hey! My extra newline!

mtgred:
:+1:

PR #505, opened 2015-08-07
Added a prompt for Paintbrush

queueseven:
This adds a prompt to Paintbrush so the Runner can select sentry, code gate or barrier.


mtgred:
:+1:

PR #518, opened 2015-08-09
implement Nasir Meidan, add rest of Muertos Gang Member

JoelCFC25:
This gives Nasir a "click ID to use him" ability a la Blue Sun. Nasir likes to spend down his credits with all kinds of shenanigans in step 2.3 of the timing chart, so the Runner needs to pop Clone Chips, SMCs, and do whatever else before clicking him. I've confirmed that this will add back credits equal to modified rez cost--so e.g. hitting a twice-advanced Weyland space ice will leave Nasir sitting at 3 credits.  

Muertos Gang Member will now prompt the Corp on install to derez something, and give a free rez prompt when it leaves play.

### DIFF
--- a/src/clj/game/cards-identities.clj
+++ b/src/clj/game/cards-identities.clj
@@ -157,7 +157,11 @@
                                   :effect (effect (mill 2) (draw))}}}
 
    "Nasir Meidan: Cyber Explorer"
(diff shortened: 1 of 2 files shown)


JoelCFC25:
Whoops, back to my original plan with Nasir--this won't work with Rook/Xanadu.

PR #520, opened 2015-08-09
add missing Muertos Gang Member effects, partially implement Off the Grid

JoelCFC25:
Completes Muertos Gang Member with the missing Corp derez and discounted rez effects when MGM is installed and leaves play. 

Added a partial Off the Grid--it trashes itself whenever a run on HQ is successful.


mtgred:
:+1:

PR #524, opened 2015-08-10
add #uk channel as per request

queueseven:
also, use a array-map so the order of channels is fix


queueseven:
@mtgred Feel free to just close this if you thing we don't need more channels

Also, I took the liberty to sort the channels alphabetically, with general always on top

queueseven:
There's not much action in the other channels, so I think we don't need another one.

PR #527, opened 2015-08-11
Removed gender-specific language

AndrewHynes:
I aimed to change all instances of "his" with "his or her" like the program tends to have, but went one step further and changed both "his" and "his or her" to "their". Hopefully this makes the platform more
friendly to use for people who aren't male!

Note that this is for all instances of "his" or "his or her" unless my grep-fu has failed me.


mtgred:
:+1:

PR #529, opened 2015-08-11
add :previous-zone when hosting cards...

queueseven:
...so it behaves like the move function.

This enables Autoscripter to work correctly when the card is not directly
installed but due to a host'n'install ability like Scheherazade.


mtgred:
:+1:

Issue #511, opened 2015-08-08
Parasite on ice + Patch + Lag Time


hydrozen:
The ice in question was strength 4 and so had a total strength of 7 with Patch and Lag Time. Parasite trashed the ice when it had 5 tokens on it I believe so I’m guessing it ignored Patch. Possibly a bug.

queueseven:
I played Data Raven, Lag Time, then Patch.

![bild 4](https://cloud.githubusercontent.com/assets/3126597/9149557/376356ee-3db0-11e5-9a53-407772a4dd95.png)

Now Data Raven has strength 5 as you see in the screenshot, but it should be 7.

hydrozen:
I tried only playing patch, and it doesn’t display 6 as the STR on the card so maybe it’s only a problem with Patch? or is it only currents that add that orange token with the correct strength?

queueseven:
That orange token should be visible whenever the strength of an ICE is different from its base strength.

nealterrell:
Patch wasn't working the last time I played it.

swankidelic:
I just had a game with a similar issue. We were using Patch, Lag Time, and Parasite, so there were a lot of modifiers.

First noticed the issue when I scored an agenda and trashed Lag Time. The ice with Parasite on it updated correctly, but the rest of the ice had not.

http://i.imgur.com/Qe5jXQw.png

We continued the game keeping this in mind, and when the runner made another run and encountered a problem ice, the strength of the ice updated to the correct value for the current game state.

It appears that trashing Lag Time doesn't trigger the "update ice strength" function that's triggered when Parasite adds a virus token, or when the ice is encountered.

We hadn't trashed Patch while it was installed (or, the "condition counter" was installed) so I don't know if this problem affects Patch as well.

nealterrell:
Fix submitted in #532.

hydrozen:
Coolio, thanks buddy.

swankidelic:
Super thank you, Neal! Love all the work you guys are doing =)

queueseven:
#532 was merged so I think this issue can be closed, or do you wait until it is actually deployed on the server?

nealterrell:
Usually wait until deployed.

JoelCFC25:
This one has been fixed and deployed.

PR #532, opened 2015-08-11
Patch for Patch and Lag Time

nealterrell:
Patch was broken by some of the cleanup code for hosted card events. It is now treated as "installed" and thus can receive events like `:pre-ice-strength`. 

Lag Time will now update all ice strength when it leaves play. 

Addresses #511.


mtgred:
:+1:

PR #534, opened 2015-08-12
re-enable Geist

JoelCFC25:
Armand "Geist" Walker has stopped working altogether--then I suddenly noticed that his subtitle is missing, so the card definition isn't matching the correct card title as pulled from NetrunnerDB!


mtgred:
:+1:

queueseven:
Has this changed on NetrunnerDB's side?

PR #535, opened 2015-08-13
Blue Sun accounts for modifications to rez cost

### LINKED ISSUE #517: Blue Sun power does not take modified rez costs into account
If a piece of Space ICE has been advanced (Orion, Wormhole, etc), once retrieved from board it should only give the reduced rez cost back.
If a piece of ICE has been made more expensive by a PERMANENT effect (Rook, Xanadu) it should give the increased money cost back.

nealterrell:
This turned out to be harder than anticipated because of the red queen Reina Roja. Reina's ability notably does NOT grant Blue Sun an extra credit, whereas every other source of rez cost modification does. This is justified on ANCUR by Lukas due to Reina's ability only being active while the rez action is taking place, while the others are constant effects. 

So we separate `:pre-rez` into a second event, `:pre-rez-cost`, which we can trigger to only alert cards that have constant effects on rez cost. Reina remains as the only card handling `:pre-rez` for the moment, as she should only be affecting rez cost as the card is being rezzed.

Fixes #517.

### DIFF
--- a/src/clj/game/cards-agendas.clj
+++ b/src/clj/game/cards-agendas.clj
@@ -42,9 +42,8 @@
 
    "Braintrust"
    {:effect (effect (set-prop card :counter (quot (- (:advance-counter card) 3) 2)))
-    :events {:pre-rez
-             {:req (req (= (:type target) "ICE"))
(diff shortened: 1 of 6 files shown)


mtgred:
:+1:

PR #540, opened 2015-08-13
Fix for Imp in multi-card servers

### LINKED ISSUE #533: Imp cannot trash while accessing remote servers
This may be tied to the new system of accessing cards, but in remote servers when I access a card and click Imp, nothing happens. It does work when accessing cards in HQ (I assume R&D too, but didn't test it).

nealterrell:
Another interesting issue.

The UI will ignore card clicks (for the purpose of activating an ability) if there is a Selection prompt active, because the user is supposed to be clicking cards to resolve some ability on. However the UI wasn't doing a bright job of this -- it was looking to see if there were ANY Selection prompts queued up, not if the currently active prompt was a Selection. This has never been an issue until the multi-access revamp, which puts Selection prompts in the prompt queue behind the Handle Access prompt when accessing from a multi-card server. The UI was therefore trying to Select Imp as the target card to access in the server, even though the "select a card to access" prompt wasn't the active prompt.

The fix involves communicating to the UI what a prompt's "type" is. Only Selection prompts use this for now. The UI then will then only ignore card clicks if the active prompt is a "select" prompt, otherwise it will let the click go through.

Fixes #533.

### DIFF
--- a/src/clj/game/core.clj
+++ b/src/clj/game/core.clj
@@ -204,13 +204,13 @@
 
 (defn show-prompt
   ([state side card msg choices f] (show-prompt state side card msg choices f nil))
-  ([state side card msg choices f priority]
+  ([state side card msg choices f {:keys [priority prompt-type] :as args}]
    (let [prompt (if (string? msg) msg (msg state side card nil))]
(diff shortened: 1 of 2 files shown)


mtgred:
:+1:

PR #544, opened 2015-08-14
Implement Casting Call

nealterrell:
Implements Casting Call from Old Hollywood.

Had to do some surgery on the core routines for this to work.

1. `corp-install`: instead of a parameter `:rezzed`, take the more general `:install-state` which can be either `:rezzed` or `:face-up`. Migrated License Acquisition and Accelerated Beta Test to this system.
2. `card-init`: take a parameter `resolve`, which is true by default, and only resolves the card's ability if `resolve` is true. This allows us to not resolve cards that are being installed `:face-up`, yet still take advantage of the rest of `card-init` (which is useful otherwise). Without this, using Casting Call on an agenda with a "when scored" ability would trigger that ability on install. Super OP.
3. `move`: when a server is collapsed, all the cards in subsequent servers have their zones updated with their new server numbers. This was not applying to cards HOSTED in those servers. Likewise, hosted cards were not having their events re-registered (necessary because the zones have now changed). This has been fixed.
4. I moved the trigger for the `:access` event to before the access happens rather than after... otherwise accessing the agenda hosting Casting Call will cause a steal, trashing Casting Call, preventing it from handling events.

I have tested this with a variety of agendas, also in Haarpsichord and Argus.

### DIFF
--- a/src/clj/game/cards-agendas.clj
+++ b/src/clj/game/cards-agendas.clj
@@ -7,7 +7,7 @@
                            (str "install " c " ICE and trash " (- 3 c) " cards")))
                :effect (req (doseq [c (take 3 (:deck corp))]
                               (if (= (:type c) "ICE")
-                                (corp-install state side c nil {:no-install-cost true :rezzed true})
+                                (corp-install state side c nil {:no-install-cost true :install-state :rezzed})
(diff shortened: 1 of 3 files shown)


mtgred:
:+1:

PR #548, opened 2015-08-17
Fixed path change on netrunnerdb for image files.

Makai:
NetrunnerDB seems to have changed the path where it stores its images. Changed fetch.coffee to reflect that so actual images are being properly retrieved again.


JoelCFC25:
Nice--I discovered this as well last week when I had to reclone the repo but hadn't submitted it yet.

mtgred:
:+1:

PR #553, opened 2015-08-18
Implement Nasir; fix obscure hosting issue

nealterrell:
This is @JoelCFC25's code for Nasir. While helping him debug its interaction with Rook I came across an obscure issue with the hosting code. That fix is bundled with this submission.

This is a partial implementation of Nasir: click the ID when in a run and the corp has just rezzed an ice to lose all your credits and gain back credits for the modified rez cost of the ice. There are too many edge cases to make this completely automated, and I agree with Joel that this is the best implementation for our system.

The hosting issue is thus: when rezzing a card, the zones of any hosted cards were retaining the ClojureScript version of the `:zone` field, which uses strings instead of keywords. Thus the `:zone` of the `:host` of a card (say, Rook) is a vector of strings, whereas the `:zone` of the host card itself is a seq of keywords. Complicated, but fixed with the same code we apply to fix the `:zone` of the host card itself.

### DIFF
--- a/src/clj/game/cards-identities.clj
+++ b/src/clj/game/cards-identities.clj
@@ -161,7 +161,17 @@
                                   :effect (effect (mill 2) (draw))}}}
 
    "Nasir Meidan: Cyber Explorer"
-   {:effect (effect (gain :link 1))}
+   {:effect (effect (gain :link 1))
+    :abilities [{:req (req (and (:run @state)
+                                (:rezzed (get-card state current-ice))))
+                 :effect (req (let [current-ice (get-card state current-ice)]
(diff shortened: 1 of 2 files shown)


mtgred:
Clojurescript has some quirks. One is that keywords are converted to strings on the client side. So when card data are sent back to the server (eg. when selecting a card as ability targets) the elements in `:zone` has to be converted back to keywords. The conversion is swallow. With hosting implemented, the conversion had to be extended to the host and hosted cards, potentially recursively (eg. Leprechaun host Leprechaun). I didn't have time to test the conversion extensively so there might still be some `:zone` with strings. But you figured it out so good job. :+1:

PR #554, opened 2015-08-19
implement Power to the People, Bookmark

JoelCFC25:
Adding a couple more cards from The Universe of Tomorrow. Both are tested and seem to be working--somewhat surprisingly I got Bookmark working on the first try, and only stumbled briefly on Power to the People (getting the `:once :per-turn` in the right place).


mtgred:
:+1:

PR #556, opened 2015-08-19
Checkpoint fix; update older card routines

### LINKED ISSUE #373: Checkpoint. dmg doenst fire and trace causes long lag spike
So i recently played around with checkpoint ice.

-If u let the trace fire, the corp can set the trace,  the runner gets the trace window.  but setting it and clicking on ok  creates a 2 - 3 min lag burst before actually doing something

-if the trace fires, and the run is succesfull.  theres no way to apply he dmg ?

this was discovered when playing with online vs someome.  Then i tried it playing a agame against myself using 2 browsers  on the same pc.  Same thing happens.
Normal traces (eg. sea source)  where no problem.

nealterrell:
Checkpoint was using an older style of code for its trace effect, via an anonymous function instead of `req` or `effect` macros. Simple fix. I also fixed three other cards with the same code style: Cell Portal, Oracle May, and Grifter.

Fixes #373.

### DIFF
--- a/src/clj/game/cards-ice.clj
+++ b/src/clj/game/cards-ice.clj
@@ -70,7 +70,7 @@
 
    "Cell Portal"
    {:abilities [{:msg "make the Runner approach the outermost ICE"
-                 :effect #(do (swap! %1 assoc-in [:run :position] 0) (derez %1 %2 %3))}]}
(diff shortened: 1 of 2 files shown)


mtgred:
:+1:

PR #558, opened 2015-08-19
Network improvements, fix for Personal Workshop and Off-Campus Apartment

### LINKED ISSUE #354: Off-Campus Apartment Creates Tremendous Lag
Myself (and others) have experienced an extremely high amount of lag when hosting multiple resources on Off-Campus Apartment. I found it happening after the number of hosted connections reached 4+, and the lag manifested itself in the form of server-response delay in clicking buttons and typing into chat (delay between 5 and 20 seconds). 

Interestingly, it ONLY seemed to effect me (the runner): the corp reported experiencing no delays, and their turn seemed to proceed smoothly. Lag returns as soon as next runner's turn begins.

This may be a more general issue with card hosting? I have not experienced this issue with Personal Workshop or The Supplier, but it is admittedly unusual for either of those cards to have more than 3 cards hosted on them.

nealterrell:
As discussed on Slack. Fixes #354.

Remove unnecessary card data from game state: `:setname :text :_id :influence :number :uniqueness :influencelimit :faction :factioncost`.

Remove the `:hosted` array when associating a card like Personal Workshop as the `:host` of another card that it is hosting. Dramatic reduction in state size when many cards are hosted.

I'd estimate a 30-40% reduction in state size for most games, and significantly greater reduction for decks running PW/Off-Campus Apartment.

### DIFF
--- a/src/clj/game/core.clj
+++ b/src/clj/game/core.clj
@@ -578,7 +578,9 @@
 
 (defn create-deck [deck]
   (shuffle (mapcat #(map (fn [card]
-                           (let [c (assoc card :cid (make-cid))]
+                           (let [c (assoc card :cid (make-cid))
+                                 c (dissoc c :setname :text :_id :influence :number :uniqueness :influencelimit
+                                           :faction :factioncost)]
                              (if-let [init (:init (card-def c))] (merge c init) c)))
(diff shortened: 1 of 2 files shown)


mtgred:
Awesome :+1:

PR #559, opened 2015-08-19
partial implementation of Zed 1.0

JoelCFC25:
This is up to the players to police for themselves, obviously. But in the *extremely* unlikely event that no other rezzed cards can deal a brain damage, someone is using Zed, **and** Zed's subs fire...1 brain can now be dealt. ;)


mtgred:
:+1:

Issue #383, opened 2015-07-17
Personal Workshop does not seem to work with 0 cost programs


queueseven:
If you host e.g. a Paricia on Personal Workshop, it should get installed immediatly in your rig, but it just stays on Personal Workshop forever

JoelCFC25:
Fixed at #560!

PR #560, opened 2015-08-19
Zero cost program or hardware hosted on PW installs immediately now

Makai:
Fix for issue #383. Hosting a zero cost program or hardware on Personal Workshop will now immediately install it instead of hosting it forever.


mtgred:
:+1:

PR #563, opened 2015-08-20
Implement Apex

queueseven:
Added a new zone [:runner :rig :facedown] for facedown cards.
When adding a card to that zone, it will be deactivated.
Facedown runner cards can't use abilities.
Installing a runner card facedown never has any costs.
Installing a runner card facedown does ignores unique cards.

The 'You cannot install non-virtual resources.'-part is not enforced.


mtgred:
Nice :+1:

PR #566, opened 2015-08-20
Made auto-pump functionality more generic.

justinliew:
Making the auto-pump addition of dynamic abilities more generic so we can re-use them for other dynamic abilities.
Will be using this next to implement proper recurring credit abilities based on type.


justinliew:
This isn't super high priority but will be useful when my recurring credit changes eventually get  done. I can always resubmit this with those changes if that makes more sense.

PR #569, opened 2015-08-21
Add several Data & Destiny cards

JoelCFC25:
All are fully working except Archangel and Improved Tracers. For Archangel, the Corp is prompted to choose whether to pay 3 credits when Archangel is accessed, but actually executing the subroutine would require it to be manually installed and rezzed--unless we want to try passing a prompt over to the Runner asking if they plan to break it, then doing the trace if they don't/can't. 

Improved Tracers only adds strength to Tracer ice currently--increasing base trace strength can be done by taking an extra credit and boosting by 1 in addition to what would be spent.

### DIFF
--- a/src/clj/game/cards-agendas.clj
+++ b/src/clj/game/cards-agendas.clj
@@ -188,6 +188,10 @@
     :abilities [{:counter-cost 1 :msg "do 1 net damage" :req (req (:run @state)) :once :per-run
                  :effect (effect (damage :net 1 {:card card}))}]}
 
+   "Improved Tracers"
+   {:events {:pre-ice-strength {:req (req (has? target :subtype "Tracer"))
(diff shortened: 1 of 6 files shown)


mtgred:
:+1:

mtgred:
Few remarks:
* You can use `:msg` instesd of `(system-msg)` in Reality Threedee, Archangel, Special Offer and Technical Writer.
* For Windfall, I think the `top card` will be the top card of the deck before the shuffling. `(trash state side top card)` is present in both cases of the `if` so it's better to just put it outside the `if`

JoelCFC25:
OK thanks, I'll double back on those and check them out. I did Technical Writer the way it is on purpose so the message wouldn't look goofy: "joel uses Technical Writer to place 1 credit on Technical Writer"

Issue #568, opened 2015-08-21
Tinkering errors


nealterrell:
Tinkering grants sentry, code gate, __and__ barrier to a piece of ice, not just one of those; we show a prompt to choose one subtype. It is also printing the name of the targeted ice to the log even if that ice is not rezzed.

queueseven:
@nealterrell Tinkering granting sentry, code gate, and barrier is correct. The card reads: "Choose a piece of ice. That ice gains sentry, code gate, and barrier until the end of the turn."

queueseven:
I hang my head in shame...

PR #570, opened 2015-08-21
Fix Tinkering, fixes case mtgred/netrunner#568

queueseven:


mtgred:
The `resolve-ability` is not necessary anymore so it can be simplified to e3e0d3f0

queueseven:
You're right, stupid me :-)

PR #574, opened 2015-08-21
Fix for Fetal AI

nealterrell:
When declining additional costs on an agenda, we still need to fire the `:access` events. Also applies to Red Herrings and Strongbox.

I upgrade `:no-msg` in optional prompts to be a full-fledged ability called `:no-effect`, so that the agenda access code could run functions when the optional prompt is declined instead of just printing a message. I updated Woman in the Red Dress (the only use of `:no-msg`!) to reflect this change.


mtgred:
:+1:

Issue #416, opened 2015-07-22
SMC trashes itself if selected program can't be installed


laganojunior:
If you have at least 2 credits and trigger Self-modifying code (SMC), it will show all programs left in the deck, even if you don't have the credits to immediately install it. If you do this, then SMC trashes itself and you lose 2 credits (as if you trigger the trash cost).

There should generally be a way to gracefully back out of the SMC action, like a "None" option in the SMC program select. Perhaps don't even show programs that can't be installed as a choice

DominicKexel:
If you activate SMC, you have to play 2c and trash it. 

If you can't find one that you can install, you already paid 2c and trashed SMC. It was the cost to activate the ability. There's no way to say: "Oh, didn't find one, I take my credits and SMC back!".

Showing only the programs you can install would be an improvement, though.

laganojunior:
I don't think you can even legally pop SMC without using its effect, similar to the street peddler ruling if it has only events (https://twitter.com/RukasuFox/status/618047088308432896). This would probably only be relevant for Geist desperately trying to draw cards though.

DominicKexel:
But you probably don't know in advance if there's a program you can install or not.

If you pop SMC or play Special Order or Hostage or any card that let you search your deck, there's always the posibility of failure to find a card. 

The FAQ says:

> If a player is searching for a card, he must find the card, if able. If a player is unable to fulfill the condition of the search, then nothing happens, but the deck is always reshuffled.

laganojunior:
Ah, you're right about the popping just to shuffle. I guess then I'm just asking for only showing programs that the system will allow installing.

steveklabnik:
This messes up a _lot_ of my games. The searched-for card seems to straight-up disappear when this happens, on occasion.

(I'm not usually the one playing SMC, my opponents are)

josemgarcia:
One thing related to this issue is when you have some source of recurring credits that you can use to install programs (e.g. Sahasrara).

For instance, if you trash SMC, and then get recurring credits from Sahasrara while the SMC prompt is listing the programs that can be installed, you can click on a program that you should be able to install, but it is not installed in the end. However, the log even shows the message that the program was installed!

JoelCFC25:
[...]

PR #575, opened 2015-08-21
When triggering SMC show only programs the runner can install

Makai:
Implementing feature #416. Not sure if this is entirely desirable as it could potentially confuse the user as to why certain programs do not appear as options. Then again, it might be better than having them select a program they do not have the credits to install and have SMC trash and nothing happen.

### DIFF
--- a/src/clj/game/cards-programs.clj
+++ b/src/clj/game/cards-programs.clj
@@ -380,7 +380,9 @@
    "Self-modifying Code"
    {:abilities [{:prompt "Choose a program to install" :msg (msg "install " (:title target))
                  :priority true
-                 :choices (req (filter #(has? % :type "Program") (:deck runner)))
+                 :choices (req (filter #(and (has? % :type "Program") 
+                                             (<= (:cost %) (- (:credit runner) 2)))
+                                       (:deck runner)))
                  :cost [:credit 2]
                  :effect (effect (trash card {:cause :ability-cost}) (runner-install target) (shuffle! :deck))}]}
 


queueseven:
I think simpy comparing :cost and :credits is not enough. If you're playing e.g. Kate and have 6 credits, you should be able to fetch your Magnum Opus (2c for SMC, 5c for MO, -1c for Kate's install bonus).

I don't know exactly how the instal bonus code works, maybe you can find it out yourself or page @nealterrell :-)

Makai:
Good point. I assumed ```install-cost``` would take the install bonuses into consideration but that doesn't seem to be the case. I'll look into it.

Makai:
Incidentally, there seems to be an issue with cards such as Scheherazad suffering from the same issue.

queueseven:
Maybe you want also to look into the MU issue? We (I mean Joel) had the idea that MU should be allowed to get negative. So you could fire SMC even when you have no MU left, and the program gets installed, and then you have the option to trash programs until your MU is >= 0 again,

Makai:
Changing from ```(:cost %)``` to ```(install-cost state side %)``` doesn't seem to help as Kate's discount is calculated at pre-install which is after we filter for cards. 
I'll open up a defect for the cards where this is already happening (Scheherazad, Street Peddler) and try to look for a fix.

Allowing MU to go negative sounds like a good idea, but we might want some interface alert of some kind to go with that to warn the players not to continue play until MU is brought back to a legal state.

queueseven:
Yeah, maybe adding some warning sign icon next to the MU?

PR #577, opened 2015-08-21
Implement Tenma Line.

nealterrell:
As coded on https://www.livecoding.tv/video/jintekinet-intro-tenma-line-12/ :).

The double `doseq` is logic that will need to be applied back to Escher and probably in other cases. I'll get on that.


mtgred:
:+1:

PR #584, opened 2015-08-23
implement Victoria Jenkins, fix trash effect of other executives

JoelCFC25:
Adds Victoria Jenkins. Chairman Hiro, Director Haas, and The Board had too many arguments being passed to `gain-agenda-point`--they shouldn't bug games when trashed anymore. 

Tacked on Resistor and a small tweak to Improved Tracers.


mtgred:
:+1:

PR #585, opened 2015-08-23
finish Docklands Crackdown, Executive Boot Camp

JoelCFC25:
Making Docklands Crackdown and Executive Boot Camp fully functional. Code from Kate was almost a direct drop-in to Docklands, and EBC got slightly modified bits from Eliza's Toybox. The discounted rez option should be used prior to hitting Start Turn, just like with Blue Sun's ability and several other "When your turn begins" abilities.


mtgred:
:+1:

PR #586, opened 2015-08-23
fix Order of Sol turn start issue

### LINKED ISSUE #344: Order of Sol not firing when you start turn with 0 creds
When I've got Order Of Sol installed, it works fine if you have at least 1 credit during your turn and you go down to 0-- it pumps you back to 1. However, if you start a turn with 0 creds, it doesn't automatically bump you to 1. And if you do it manually and then (inevitably) spend it, then OOS will fire, and you'll have to decrease the cred. Sort of a hassle.

JoelCFC25:
Fixes #344. Add a `runner-turn-begins` to get Order of Sol to add a credit when starting the Runner turn on 0 credits, then update `:per-turn` to make sure it doesn't fire again if you drop to 0 during that same turn. 

EDIT: Forgot it needed a `corp-turn-begins` for the same reasons.


mtgred:
:+1:

PR #589, opened 2015-08-23
Sunny Labeau starting link is 2. Multithreader has two recurring credits.

dtelad11:


JoelCFC25:
Change Sunny's last name to `Lebeau` and you'll be all set. :+1:

mtgred:
Thanks for the contribution El-Ad !

dtelad11:
Thanks for fixing the typo.

PR #591, opened 2015-08-23
implement some Sunny Lebeau cards

JoelCFC25:
Implements Sunny's cloud breakers (except for the automatic MU savings, which isn't supported yet). Globalsec Security Clearance steals from Joshua B. and Motivation to show only the Runner a prompt with the title of the top card of R&D.


mtgred:
Awesome :+1:

PR #592, opened 2015-08-23
Expand install-cost-bonus framework; implement Utopia, Eden, Hades Shard installs

nealterrell:
As shown here: https://www.livecoding.tv/video/jintekinet-events-utopia-shard-12/

Expand the `install-cost-bonus` framework with two additions:

1. A card can modify its own install cost by implementing `:install-cost-bonus` in the card-def, similar to corp cards that modify their own rez cost.
2. All modifications to install cost (whether through `:install-cost-bonus` or the `(install-cost-bonus)` function) are now of the form `[type amount]`. Rather than assuming (and restricting) bonuses to be regarding the credit cost, we now let the modifications be to any cost that can be paid through `pay`. This turns Modded into `(install-cost-bonus [:credit -3])`, for example, and allows full implementation of installing the Shards by clicking on them during a run.

I implemented the Shards by incorporating a `:install-cost-bonus` of `[:credit -7 :click -1]` when you install them during a run on the appropriate server after passing all ice (at `:position 0`). The new framework allows us to modify not only the credit cost of the Shard, but also cancel out the `:click 1` that comes from trying to install the card from hand. 

Any cost that can be paid through `deduce` can be modified by this bonus framework. That includes `:memory`... which would allow the Cloud breakers to cancel their memory cost on install if the runner has >= 2 link. I'll leave that to someone else.

### DIFF
--- a/src/clj/game/cards-assets.clj
+++ b/src/clj/game/cards-assets.clj
@@ -138,7 +138,7 @@
    {:abilities [{:cost [:click 2] :msg "add 1 power counter" :effect (effect (add-prop card :counter 1))}]
     :events {:pre-install {:req (req (and (not (zero? (:counter card)))
                                           (not (get-in @state [:per-turn (:cid card)]))))
-                           :effect (effect (install-cost-bonus (:counter card)))}
+                           :effect (effect (install-cost-bonus [:credit (:counter card)]))}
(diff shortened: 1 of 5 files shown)


mtgred:
Awesome! :+1: 

Great video too!

PR #594, opened 2015-08-23
Show regular card images in card preview

queueseven:
not alt arts...


queueseven:
So, what's up with this PR, @mtgred? Do we keep alt arts in the card preview? Then I would close this PR. Of course, another way would be to check if the user has the show-alt-art-flag, and then decide if we show the alt art or regular art. What do you think?

mtgred:
I misunderstood what this PR. Showing non alt is fine.

Sorry for the delay.

queueseven:
np :-)

Issue #536, opened 2015-08-13
Rolodex tells the corp which cards the runner looks at...


queueseven:
...and in which order they are returned to the stack.

JHonaker:
So I've narrowed it down. This is not a problem with the implementation of Rolodex. It's from the implementation of move-card in core.clj. When the runner moves a card to the Stack, it doesn't hide the name of the card, but when the runner moves it to R&D it does.

mtgred:
This should be fixed now. Thanks @JHonaker !

PR #595, opened 2015-08-23
move-card no longer shows name from :play-area

JHonaker:
The problem with Rolodex displaying the moved cards to the Corp wasn't with the card itself. It was the fact that runner cards moved onto the Stack were not being hid by move-card. This is why Indexing didn't have this problem. It moved Corp cards to R&D.

So, this prevents cards in the :play-area region from being displayed when moved. This should prevent future problems with cards moving from the :play-area from being displayed too.

Fixes Issue #536

### DIFF
--- a/src/clj/game/core.clj
+++ b/src/clj/game/core.clj
@@ -1377,10 +1377,12 @@
 
 (defn move-card [state side {:keys [card server]}]
   (let [c (update-in card [:zone] #(map to-keyword %))
-        label (if (or (and (= (:side c) "Runner") (not (:facedown c)))
-                      (:rezzed c) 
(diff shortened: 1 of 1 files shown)


nealterrell:
Nice work! Our development team uses Slack for communication and Trello for task management. Would you like invitations to each by @mtgred and @queueseven?

JHonaker:
Yes! I was actually going to ask about that next.

mtgred:
:+1: 

If you provide me your email, I'll send you an invite to our Slack.

JHonaker:
I sent it to your gmail account. Thanks a lot!

Issue #596, opened 2015-08-24
Hayley Install


ihmcallister:
After the first install from Hayley on a turn if you choose no to installing another card with Hayley's ability the interface keeps asking if you want to install another from grip on all subsequent installs. Hayley's ability is a 1 time thing each turn, if you don't use it you have missed your opportunity and you shouldn't be able to trigger the ability on subsequent installs.

JHonaker:
I think this has something to do with the optional flag. The same bug is affecting Laramy Fisk, Comet, and (unverified) the rest of the cards that say "The first time X, you may Y."

JoelCFC25:
Deployed!

Issue #63, opened 2015-01-17
The Foundry's ability triggers more than once per turn


H4pp7B1rd:
I rezzed two pieces of ICE on the runners turn. It asked me both times if I want to add a copy of that ICE to HQ.

mtgred:
Thanks for reporting the issues. The Foundry is now fixed.

JHonaker:
The Foundry was actually not fixed, but I've fixed it in an incoming pull request.

PR #597, opened 2015-08-24
Fixing "First time X you may Y" abilities

JHonaker:
I've just finished updating the cards with rules text that says, "the first time X happens, you may Y." I believe I've fixed them all. The only one that remains unimplemented is Tori Hanzo because it is a good bit more difficult.

The problem was with the `:once :per-turn` tags not working properly with the `:optional` abilities. I've added `:req` calls that make use of the very handy `(turn-events)` function to make sure the game state is proper for the call.

This pull request addresses issues with:
Laramy Fisk
Hayley
Comet
The Foundry

It also fixes issues #596 and #63.


mtgred:
:+1:

PR #600, opened 2015-08-25
Implement cloud icebreaker

queueseven:
The first step was to enable MU to be negative. Image you have a bunch of cloud breaker and other programs installed, and then you somehow lose your 2nd link. Suddenly you have to trash a bunch of programs, because your MU is "negative" until you trashed enough programs.

This also fixes the annoyance when you e.g. clone chip or SMC a program with not enough memory left: the clone chip or SMC would get trashed, but the program would not get installed.

Implementing the cloud breaker was straight forward thanks to @nealterrell's `:install-cost-bonus` and @mtgred's tip with simply using `add-watch`. First I wanted to implement them directly in `core.clj`, but I couldn't get it to work because the `:install-cost-bonus` has to be part of the card definition before the card is in play. I created a function `cloud-icebreaker` that simply wraps a card definition (like `auto-icebreaker`), so there's no code duplication.

I added a little visual clue to show when the runner's MU is negative, and, while at it, I also show it when the runner is tagged, since that's also an important information and unfortunately easy to miss:

bild 27

### DIFF
--- a/src/clj/game/cards-icebreakers.clj
+++ b/src/clj/game/cards-icebreakers.clj
@@ -30,6 +30,20 @@
                               :run-ends breaker-auto-pump :ice-strength-changed breaker-auto-pump
                               :breaker-strength-changed breaker-auto-pump :approach-ice breaker-auto-pump })))
 
+(defn cloud-icebreaker [cdef]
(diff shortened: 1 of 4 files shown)


mtgred:
:+1:

PR #603, opened 2015-08-25
Implement Howler.

nealterrell:
Silly card, fun card. I avoided using `move` since it doesn't let you specify position (only front or back) and this is a very specific interaction that I didn't think needed to be generalized. Manually places the Bioroid ice immediately after Howler, makes the next ice encountered be the installed ice, and trashes Howler / derezzes the Bioroid at the end of the run.

Video: https://www.livecoding.tv/video/jintekinet-resolving-abilities-howler/


mtgred:
Awesome :+1:

PR #604, opened 2015-08-26
improvement to Bad Times, Improved Tracers

JoelCFC25:
Now that MU can go negative with the new cloud breaker subsystem, Bad Times can reduce the Runner's MU by 2 until the end of the Corp turn and then restore it. If MU dips below 0, in addition to the new visual indicator, a message is put into the log reminder the Runner to trash programs.


mtgred:
Instead of using registering on `:corp-turn-ends` and then unregister it, you can use the `:end-turn` key. See Breaking News as an example.

JoelCFC25:
Sounds good, I'll tidy it up tonight.

JoelCFC25:
Closing this and repackaging it with the completion of Improved Tracers.

PR #605, opened 2015-08-26
implement Exploratory Romp

JoelCFC25:
Another chapter in my ongoing "crap cards no one uses" saga...

It took me way longer than it should have to figure out how to enforce legal targeting (card to remove advancements from must belong to the server being run).


mtgred:
:+1:

PR #607, opened 2015-08-26
Adam and his directives

queueseven:
Adam: you have to have the 3 directives in your deck. They will be installed automatically when the game begins.

Neutralize All Threats: If you can't pay the trash cost, the regular prompt is shown, so you can use e.g. Ghost Runner or recurring credits. If you can pay, the card is automatically trashed.

Safety First: Nothing special, fully automated

Always Be Running: The 'Your first Click...' condition is not enforced; and I think it's not a big deal. There was a discussion on Slack on adding a `:runner :restrictions` map to the state. I think we could fully implement *Always Be Running* once we have *Replicating Perfection* fully implemented.

### DIFF
--- a/src/clj/game/cards-identities.clj
+++ b/src/clj/game/cards-identities.clj
@@ -1,7 +1,16 @@
 (in-ns 'game.core)
 
 (def cards-identities
-  {"Andromeda: Dispossessed Ristie"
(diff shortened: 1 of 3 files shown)


mtgred:
:+1:

PR #612, opened 2015-08-26
Fixed Hellion Alpha Test and Sherlock 1.0

queueseven:
The move function was changed to take a map as options, so we need to pass
{:front true} instead of simply true.


mtgred:
:+1:

PR #614, opened 2015-08-26
Installing programs at no cost should still charge their MU.

### LINKED ISSUE #613: London Library + Overmind not counting MUs correctly
When Hosting Overmind on London Library, Overmind is getting counters equal to the MUs before hosting Overmind.  It should be getting counters equal to the MUs after Overmind is installed.  As a result, when bringing Overmind back to one's grip, MU amount goes up one above total MU.

nealterrell:
Explicitly pass all known costs to `install-cost` so it only has to worry about modifications to those costs. Pass `:memoryunits` as a cost when installing a program at "no cost" (which really means "no credit cost).

Fixes #613.

### DIFF
--- a/src/clj/game/core.clj
+++ b/src/clj/game/core.clj
@@ -530,9 +530,9 @@
 (defn install-cost-bonus [state side n]
   (swap! state update-in [:bonus :install-cost] #(merge-costs (concat % n))))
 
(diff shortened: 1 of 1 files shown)


mtgred:
:+1:

PR #616, opened 2015-08-27
fix for tag check in Reality Threedee

mharris717:
Before this change, Reality Threedee was giving $2 every time. After this change it is working correctly. 

I am still very new to Clojure, and I only know enough to make sense of the code at a basic level.  You'll need to look at the change and see if it makes sense. I can only speak to the results of my testing. 

Thank you so much for jinteki.net, it's amazing.

### DIFF
--- a/src/clj/game/cards-assets.clj
+++ b/src/clj/game/cards-assets.clj
@@ -378,8 +378,8 @@
    "Reality Threedee"
    {:effect (effect (gain :bad-publicity 1) (system-msg "takes 1 bad publicity"))
     :events {:corp-turn-begins
-             {:effect (req (gain state side :credit (if '(tagged) 2 1)))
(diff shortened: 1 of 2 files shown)


mharris717:
Just saw the failing circleci build. I'll check it out tomorrow if I have time. 

I suspect the fix would take 10 seconds for you, but I'm enjoying learning the codebase.

mharris717:
Do you have any tips for recreating the lein test on circleci on a dev machine? Doing lein test locally gives 0 tests run.

mharris717:
Ah, needed to add tagged to the let in the msg macro. 

Again, I have no idea if this is a start change to make. I won't be offended if you reject it. 

Thanks again for all your hard work.

mtgred:
You commit looks good. :+1:

mtgred:
There is currently no test running. But Circle CI tries to compile the Clojure code. To simulate that, you can just do `lein uberjar`.

PR #622, opened 2015-08-27
improve Bad Times, fully implement Improved Tracers

JoelCFC25:
Bad Times does the temporary MU loss to the Runner and uses a message to remind them to trash programs if they dip below 0 MU. 

Improved Tracers is now 100% complete thanks to the `init-trace-bonus` feature done by @queueseven!


mtgred:
:+1:

PR #625, opened 2015-08-27
implement News Hound

JoelCFC25:
Pretty straightforward--the ETR only fires if a Current is found in either `[:runner :current]` or `[:corp :current]`.


mtgred:
:+1:

PR #627, opened 2015-08-27
Fix trash and steal cost not being correctly applied.

JHonaker:
I missed these flags in the agenda scoring/stealing process because of the lack of an `:optional` key. Trashing and steal costs should work now.


mtgred:
:+1:

PR #629, opened 2015-08-27
Don't show trash prevention prompts when inappropriate

nealterrell:
Don't show trash prevention prompts for:

1. Discarding from hand. 
2. Self-trashing a card from a lack of counters.


mtgred:
Awesome :+1:

Issue #617, opened 2015-08-27
Alpha cannot break subroutines on outermost ice


Merlin8000:

nealterrell:
Thanks for the report. Looks like it's backwards with Omega, and off by one at that.

nealterrell:
Fix submitted in #630.

PR #630, opened 2015-08-28
Fix Alpha and Omega.

nealterrell:
Fix for #617.

Alpha and Omega were backwards, and both off by 1. Position 1 is innermost, not 0; when there are N ices, position N is outermost, not N-1.


mtgred:
:+1:

PR #631, opened 2015-08-28
implement Employee Strike and Sunset

JoelCFC25:
Employee Strike is just the mirror image of Cerebral Static, not sure why I missed doing this one earlier. 

Sunset is mostly blatantly stolen from @nealterrell's Escher code with the twist of supplying the chosen server to the helper function.


mtgred:
:+1:

Issue #634, opened 2015-08-28
Woman in the Red Dress


CptHilts:
With what I think is the most recent update. At the start of Runner's turn the choice box is given to the runner not the corp.

JHonaker:
The corp still draws the card though, correct?

CptHilts:
The corp must manually click up to draw a card

JHonaker:
I'll take a look at it.

CptHilts:
cool thanks!

JHonaker:
Fixed it with the latest pull request.

PR #635, opened 2015-08-28
WitRD now correctly addresses the Corporation.

JHonaker:
The :player :corp key was in the wrong position.

Fixes issue #634


nealterrell:
This also fixes the bug with Account Siphon not granting the runner credits.

mtgred:
:+1:

PR #639, opened 2015-08-29
Fix messages and costs for optional abilities

nealterrell:
Fixes a few issues related to the recent `:optional` refactoring. `:msg` and `:cost` belong inside the `:yes-ability` of the optionals, rather than outside. Fixes issues with traps not charging the corp money, and with messages not printing when the optional is used.


mtgred:
:+1:

PR #644, opened 2015-08-30
restore Frame Job

JoelCFC25:
Frame Job was working back at #359, but it got reverted somehow between then and now.


mtgred:
:+1:

PR #645, opened 2015-08-30
Give :cid to ID cards

### LINKED ISSUE #640: Kate / Spark Agency Block Each Other
If the runner has received a Kate discount in a given turn, they don't then lose credits if Spark Agency rezzes an Advertisement. If Spark rezzes the ad first, Kate's discount then fails to trigger.

nealterrell:
Fixes #640. IDs weren't being given `:cid` tags, so `:once` was using `nil` in the `:per-turn` map. That worked fine as long as there was only one ID with a `:once` key, but fails if both sides have a "once per turn" ability on the ID.


mtgred:
:+1:

PR #646, opened 2015-08-31
hotfix for register redirect if there is no demo user

domtancredi:
Basically, the issue is that there's never a redirect when you register a user (developing locally here).

The issue is because a Response (code: 200) is never sent back. This is because when a new user registers, the system checks for new decks under "__demo__" user and then copies // inserts those into the newly registered user's "decks". However, there is no "__demo__" user in a local database (hint: feature to pre-seed this data, along with their decks).

I've just added some minor business logic to account for this and allow a redirect.

### DIFF
--- a/server.coffee
+++ b/server.coffee
@@ -249,9 +249,12 @@ app.post '/register', (req, res) ->
             for deck in demoDecks
               delete deck._id
               deck.username = req.body.username
-            db.collection('decks').insert demoDecks, (err, newDecks) ->
-              throw err if err
-              res.json(200, {user: req.user, decks: newDecks})
+            if demoDecks.length > 0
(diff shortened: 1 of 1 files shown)


queueseven:
You're right, that's annoying. But it's usually not a big deal since you can click *login* right after *sign up* (or whatever the button text is). Maybe we could also update the demo user to have some more recent decks (all worlds winner deck for example)?

domtancredi:
Sure. Just to be clear, the reason the issue occurs is because we haven't pre-seeded any data in the database to have the demo user. If we set that up (should be straightforward), this behavior goes away. 

And yes, more decks for the demo user would be pretty neat! I'm all for that. If we add a "Seed Database" Issue in GH, we can include that as a requirement.

Issue #497, opened 2015-08-04
[Feature request] Reset your password


davide-ceretti:
It would be nice to have the ability to reset the password. I understand that there isn't any sensitive data on the website but still it would be helpful.

nealterrell:
Implemented in #647!

PR #647, opened 2015-08-31
Password Reset

domtancredi:
Password reset for Issue #497 

Recommendations for a v2 are to:

* style the reset form
* create a template system for sending emails (instead of coding within business logic of server.coffee)
* more validation for reset


mtgred:
:+1:

PR #654, opened 2015-09-01
fully implement Kraken and Hemorrhage

JoelCFC25:
This just automates Kraken all the way, giving the Corp a prompt to trash a piece ice from the Runner's chosen server (and ensuring that the trash happens from that exact server). 

Same goes for Hemorrhage (just eliminates having to drag the card to Archives).


mtgred:
:+1:

PR #658, opened 2015-09-02
Completed Tori Hanzo!

JHonaker:
This implements Tori Hanzo fully, and also fixes the mistaken print from The All-Seeing I that I left in.

### DIFF
--- a/src/clj/game/cards-operations.clj
+++ b/src/clj/game/cards-operations.clj
@@ -392,19 +392,18 @@
                                              (trash state side resource)))
                               :msg (msg "trash all resources")}]
        {:req (req tagged)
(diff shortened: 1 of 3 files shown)


nealterrell:
So what happens if the runner has Feedback Filter or Deus X installed? I am concerned that the Tori code will run while the runner is deciding about blocking the net damage.

JHonaker:
Ah yes. I forgot about that. Hold off on merging this. I thought of this as I was drifting off last night, but it slipped my mind when I was cleaning it up this afternoon. We should discuss how to deal with this on Slack.

JHonaker:
Working on this a bit more.

PR #661, opened 2015-09-03
Fix "add as an agenda" cards to be forfeitable

### LINKED ISSUE #655: after Franchise City is used the runner cannot steal the agenda
One or more rezzed Franchise Cities in remote servers. The runner runs on remote and accesses an agenda. Franchise City then gets added to the score area. That leads the servers to change their position. The run arrow then points on the next outer server or to nothing when there is none. Now the game ist stuck. Nothing happens when the runner clicks on "steal [agenda]". They also cannot jack out.

JoelCFC25:
This fixes #655, so Franchise City will no longer break games. 

The new `as-agenda` function is applied to Franchise City, Notoriety, Fan Site, News Team, and Shi.Kyu to make them forfeitable (e.g. to Data Dealer, or to rez an Archer in the case of Franchise City). Messaging is also beefed up to be clearer in the log what has happened. 

This also fixes a bug that was reported in chat--News Team was not firing from R&D as it should have been.

### DIFF
--- a/src/clj/game/cards-assets.clj
+++ b/src/clj/game/cards-assets.clj
@@ -484,13 +484,13 @@
              :prompt "How many [Credits] for Shi.Kyū?" :choices :credit
              :msg (msg "attempt to do " target " net damage")
              :effect (effect (resolve-ability
-                               {:player :runner :msg (msg target)
-                                :prompt (str "Take " target " net damage or take Shi.Kyū as -1 agenda points?")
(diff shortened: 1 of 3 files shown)


mtgred:
:+1:

JoelCFC25:
Dang, I didn't add a message in time--there were a couple things I needed to clean up in the messaging for Shi.Kyu and News Team.

PR #663, opened 2015-09-03
fix take 2 tags option for News Team, improve Shi.Kyu message

JoelCFC25:
Sorry, I goofed in #661 and broke the option on News Team to take 2 tags (missed passing in `state`). This also adds messages to the log for both options the Runner can choose when hitting News Team or Shi.Kyu.


mtgred:
:+1:

PR #666, opened 2015-09-04
Add recursion sanity check for nested hosting

nealterrell:
If you have a host and that host can be found with `get-card`, update that card recursively.

If you do not have a host, use `update!`. Since you don't have a host, `update!` won't recurse back to `update-hosted!`.


mtgred:
:+1:

nealterrell:
#666... Recursion is the devil...?

PR #667, opened 2015-09-04
remove prompt when triggering Geist ability as draw is not optional

sodwyer:
The original Geist implementation offered the runner the choice of drawing a card or not when triggering his ability. The wording on the card implies that the draw is not optional, so I removed the prompt.

Does this seem ok to you?


queueseven:
Seems right. I think when the card was spoiled, it said '*may* draw 1 card' but was changed when the card was actually released. The card text on acoo.net e.g. is still wrong about that.

mtgred:
:+1:

Issue #670, opened 2015-09-05
Crisium Grid blocking R&D accesses when Medium is installed


JDCaron:
Have not tried to replicate on other servers, but the temporary solution was to bounce Medium to the grip upon a successful run.

nealterrell:
Could you be more specific about this? What do you mean by "block"? I just tried this with a simple setup: Medium for runner, Crisium on R&D; run R&D, rez Crisium before successful run. On success, Medium gains no tokens, runner accesses 1 card from hand, 1 upgrade in R&D.

nealterrell:
OK, I have a fix. Was this specifically with a Medium that had no counters on it during the run in which Crisium was used?

JDCaron:
Sorry for the delay, yes it was. Medium(s) were empty. Thanks!

Issue #407, opened 2015-07-21
Two or more Chakanas disable advancing agendas


percomis:
At one point my opponent had 3 Chakanas out and I was not able to advance my agenda at all. Later he had 2, same happened. When he only had 1 I was able to advance. When I was not able to advance the log said I advanced a card (didn't specify in which server) but the agenda never got an advancement token, even after multiple times.

DominicKexel:
I wanted to try this and the following happened:

New game. 

Corp installs agenda, take credit, take credit.

Runner installs Chakana, take credit, take credit, take credit.

Corp starts turn, exception:

> game.main=> Error  do start-turn nil #<NullPointerException java.lang.NullPointerException>
Stack trace: [clojure.lang.Numbers.ops(Numbers.java:961), clojure.lang.Numbers.gte(Numbers.java:231), clojure.lang.Numbers.gte(Numbers.java:3811), game.core$fn__7716.invoke(cards-programs.clj:46), game.core$trigger_event.doInvoke(core.clj:409), clojure.lang.RestFn.invoke(RestFn.java:467), game.core$update_advancement_cost.invoke(core.clj:1110), game.core$update_all_advancement_costs.invoke(core.clj:1106), game.core$start_turn.invoke(core.clj:917), game.main$run$fn__10149.invoke(main.clj:61), game.main$run.invoke(main.clj:57), game.main$dev.invoke(main.clj:83), clojure.core$binding_conveyor_fn$fn__4145.invoke(core.clj:1910), clojure.lang.AFn.call(AFn.java:18), java.util.concurrent.FutureTask.run(Unknown Source), java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source), java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source), java.lang.Thread.run(Unknown Source)]

DominicKexel:
The problem is that the card that gets passed to Chakana's `{:req (req...` via the event system is not the actual card that is in play. It has no `:counter` property, hence  `get-virus-counters` returns `nil`.

It could be improved to 
```
--- a/src/clj/game/core.clj
+++ b/src/clj/game/core.clj
@@ -941,7 +941,7 @@

 (defn get-virus-counters [state side card]
    (let [hiveminds (filter #(= (:title %) "Hivemind") (all-installed state :runner))]
-        (reduce + (map :counter (cons card hiveminds)))))
[...]

PR #671, opened 2015-09-05
Make get-virus-counter handle nil :counter without crashing

nealterrell:
`get-virus-counters` was barfing when a virus never had any counters installed on it, giving a `:counter` of `nil`. This caused #407 and most likely #670.


mtgred:
:+1:

Issue #515, opened 2015-08-08
Laramy Fisk ability timing


JDCaron:
Hey there, just wanted to site a problem with the new Fisk ID and the timing of his ability.  It should be successful run option, force draw option and then access.  Seems like it is successful run, access, force draw at the moment.  

Thanks for the hard work getting things together, the site is extremely helpful!  Keep up the great work.

-Jack

aiannacc:
This is a bug.  It's particularly ugly when the run is on R&D:

Expected behavior: Corp draws the top card of R&D; runner accesses the second card.

Actual behavior: Corp draws the top card AND runner accesses that same card.  If it's an asset, the runner is given the option to trash it, but it ends up in the corp's hand regardless.

qmsm:
Just had the same issue which crashed the game. Corp drew Hostile Takeover while I was given the option to steal it, then BOOM.

nealterrell:
It's going to be a tricky fix for obscure engine reasons.

PR #672, opened 2015-09-05
Laramy Fisk ability timing bug fix 

chua-mbt:
I'm a new contributor to this project. I'm also new to netrunner in general. I've only played a small number of games, so I'm not very well versed on the rules yet. I'd like to contribute because I'm interested in the game and in clojurescript.

This is a fix for #515. I delayed the event trigger for successful runs in the successful-run function. The interaction for Fisk seems to work correctly now (access first, then prompt for draw afterwards). I'm not 100% sure I haven't caused side effects in other cards that have successful run event triggers, since I don't know the game very well yet.  

I'm also not sure if you have particular rules for pull requests. Please provide further instruction if this is unsatisfactory.

### DIFF
--- a/src/clj/game/core.clj
+++ b/src/clj/game/core.clj
@@ -297,12 +297,12 @@
 (defn init-trace [state side card {:keys [base] :as ability} boost]
   (trigger-event state side :pre-init-trace card)
   (let [bonus (or (get-in @state [:bonus :trace]) 0)
-        base (if (fn? base) (base state side card nil) base) 
+        base (if (fn? base) (base state side card nil) base)
         total (+ base boost bonus)]
-    (system-msg state :corp (str "uses " (:title card) 
-                                 " to initiate a trace with strength " total 
+    (system-msg state :corp (str "uses " (:title card)
+                                 " to initiate a trace with strength " total
                                  " (" base
-                                 (when (> bonus 0) (str " + " bonus " bonus")) 
+                                 (when (> bonus 0) (str " + " bonus " bonus"))
                                  " + " boost " [Credits])"))
     (swap! state update-in [:bonus] dissoc :trace)
     (show-prompt state :runner card (str "Boost link strength?") :credit #(resolve-trace state side %))
@@ -527,8 +527,8 @@
   (swap! state update-in [:bonus :cost] (fnil #(+ % n) 0)))
 
 (defn init-trace-bonus [state side n]
-  (swap! state update-in [:bonus :trace] (fnil #(+ % n) 0)))  
-  
+  (swap! state update-in [:bonus :trace] (fnil #(+ % n) 0)))
+
 (defn rez-cost [state side {:keys [cost] :as card}]
   (if (nil? cost)
     nil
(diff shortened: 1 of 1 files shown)


nealterrell:
Hello, thanks for the submission :).

I think you have Fisk's ability wrong here. The correct order is actually to decide whether to force the draw, THEN do the access. A run is declared successful before access occurs (step 4.4 in [this diagram](http://ancur.wikia.com/wiki/Timing_Structure_of_a_Run)). If Fisk runs R&D, he can force the top card to be drawn, thereby accessing the second. If he runs HQ, the card on top of R&D goes into the corp's hand before accessing.

The current problem with Fisk's code isn't the order of the events, but the fact that his ability is optional, which means we have to put up a prompt for it. Due to simplicity concerns, our prompts are asynchronous, meaning that showing a prompt does not delay the execution of other code in the core engine. When `:successful-run` is triggered in the run code, Fisk shows his "yes/no" prompt, but the rest of the run code then resumes before the prompt is resolved... which means the Access code is then executed, before the runner has forced the corp to draw. We would be fine if his ability was automatic like Gabe's, or didn't affect the access itself like Silhouette's... alas, we are in this predicament currently.

If you are interested in learning more about the engine and continuing to attack this issue, we can invite you to two services that we use for developer communication. @mtgred can send you an invitation to our Slack message board if you send him your email address. We have a number of people active on the board who can help guide you through the engine and the game itself :).

What follows was my first response before looking closely at Fisk's timing in general. It is my attempt to be thorough in predicting what kinds of side-effects we might have from moving an event trigger. I'll leave it here for other developers to see. I am not trying to shame you, I just want to demonstrate potential effects from something as seemingly small as one event change. :)

------

Delaying the event trigger doesn't seem like the right solution to this problem. From a game rules perspective, we are now declaring runs to be successful _after_ cards are accessed, which is incorrect. It's possible that our engine is flexible enough that other interactions won't break from this, but I don't think that diverging from the strict rules of the game in order to make 1 card easier is the right call. 

Some potential issues with this delay, off the top of my head: 

* Access will occur before Desperado grants +1 cr. With our "choose access order" code this might not actually matter, I believe the runner will still have the credit by the time they choose to trash something, but it demonstrates one issue.
* Legwork and Maker's Eye both use `:successful-run` to set their bonus access amounts.
* Lots of effects need to happen on successful run, before accessing:
  * Silhouette's shows a prompt to expose a card
  * Manhunt's trace needs to fire
  * Corp should gain 1cr from Paywall Implementation (to help pay for a trap); mirror condition for Lamprey
  * Medium and Nerve Agent need to gain their token before accessing
  * John Masanori should draw a card (helps survive a Snare!)

mattlaschneider:
This is still broken—the delay doesn't seem to change access at all. This is an especially critical issue when using Fisk's ability on an R&D access, as you access the top card of R&D, then force the corp to draw that card. If you trash or steal that card, it then pulls it out of HQ, even though the runner shouldn't even have seen it, let alone be allowed to access it.

PR #673, opened 2015-09-06
Costs of installing/rezzing etc. are displayed in the log

queueseven:
bild 5

As you can see in the image, click and credit costs are displayed in the log window when installing/rezzing/playing cards.


queueseven:
Inspired by OCTGN, again ;-)

PR #674, opened 2015-09-06
implement New Angeles Sol

JoelCFC25:
The Corp is only prompted if there's at least 1 Current in Archives/HQ when an agenda is scored or stolen. If so, they get prompted to pick one location and then get a pick list of the Currents they can afford to play.


mtgred:
:+1:

PR #678, opened 2015-09-06
implement Security Chip

JoelCFC25:
Tip of the cap to @nealterrell for pointing me in the right direction. Instead of spamming the log with the names of all the boosted Cloud breakers, this just states how many got pumped.


mtgred:
:+1:

Issue #679, opened 2015-09-06
Character Assassination - choosing between duplicate cards


meepmeep13:
In a game the runner had 2 Liberated Accounts installed, one with 4 credits left (installed first) and one with 8 credits left. The Corp scored a Character Assassination, and wanted to trash the Liberated Accounts with the greater money still on it (obviously). However, only one option for Liberated Accounts appeared in the list, and on selecting it the 4 credit one was chosen by the server.

JoelCFC25:
That implementation is so old it probably pre-dates the card targeting system. It will be pretty easy to rewrite it to target a specific card instead of giving the prompt list.

JoelCFC25:
Fix submitted at #682.

JoelCFC25:
Should be live now.

PR #682, opened 2015-09-07
update Character Assassination to use targeting system

JoelCFC25:
Addresses #679 by making Character Assassination use the targeting system. This way the Corp can distinguish between specific copies of Liberated Account, Armitage Codebusting, etc. and trash the one that makes the most sense.


mtgred:
:+1:

Issue #85, opened 2015-01-30
Some odd behaviors for activated ability selection menus


anathomical:
A number of cards, when activated, bring up a target selection menu (Self-modifying Code and Clone Chip are examples of this).  If you change your mind, or clicked accidentally, there is no apparent way to dismiss this menu.  You must complete the action to get the menu to go away.

Further, clicking on the target card multiple times seems to bring the targeting menu multiple times, stacking them visually on top of each other (such that only one is visible at a time), but all triggered menus require resolution before access to the basic actions menu will be granted again (required for things like ending turns).

Note that it will actually break game-state in some ways as it resolves multiple times.  Allowing, for instance, the installation of two copies of a card that you only have one copy of as an available target.

mtgred:
It's indeed problematic. I will try to find a way to handle both issues.

JoelCFC25:
In the case of Clone Chip specifically, it has been changed in #863 to use the card targeting system, which will let you just hit Done with no effect if you change your mind after popping it.

Issue #32, opened 2014-12-16
Medium and Nerve agent do not ask how many cards you want to access


anathomical:
By the rules, I am permitted to access fewer than the maximum number of cards allowed by Medium or Nerve Agent if I wish, but that option is not given to me.

This is probably a pretty minor consideration as there aren't that many play-states where you want to access fewer than the maximum, but it does come up every so often (and accessing fewer could act as a decent work-around for Crisium grid not being implemented yet #31).

tekyro:
I know this is an edge case but it's making problems with Wendigo when you try to disable medium for a run...

I assume the best way is to ask with a drop down how many cards the runner want to access, I don't know how difficult it could be to implement though...

DominicKexel:
We need a way to bring the game into a state where it will wait for an answer of the non-active a player.

For example:

- At the start of the game, the game should pause for the corp until the runner has decided to mulligan or not
- When activating Jackson during a run, the game should pause for the runner until the ability is resolved by the corp player (yeah, I know, Jackson isn't implemented yet)
- When firing a trace, pause the game for the corp until the runner was finished boosting his/her trace

JoelCFC25:
Choice of how many extra accesses is implemented at #881.

josemgarcia:
I was playing with medium, and needed to use this feature (manually running using Jak Sinclair, which is not implemented yet), but it didn't work for me.

First, the message to warn the runner that he can choose fewer cards to access was shown before reaching the "successful run" prompt. Second, when I clicked on Medium, a virus counter was added (!), and when choosing the number of additional cards to access, the number I chose were actually added to the number of virus counters (!).

JoelCFC25:
My first reaction is that this sounds like a mix of the old and the new functionality--so either you or (more likely) your opponent had not done a forced refresh of the site (hold Shift while reloading in your browser) since the last server restart. This is really important to do every time you see that there's been a restart as indicated in the News header. 

The site currently compresses multiple steps of the timing chart into one for the sake of reducing player back-and-forth. The way the access code works, there is no way to block the execution of everything else on the `:pre-access` event trigger, so it's impossible to have this new Medium prompt come up after hitting "Successful Run" and still work. Since there is a pause after the Corp has hit "No more action" for the last time, doing the prompt and having the Runner set the number of additional accesses right before hitting "Successful Run" is basically the same thing. 

[...]

PR #686, opened 2015-09-08
Prevent cards from queueing the same prompt multiple times

queueseven:
related to #85

This prevents a card to create the same prompt multiple times (filtered by `:cid` and `msg`), so you can no longer use e.g. Self-modifying Code to search multiple programs or Executive Boot Camp to search for multiple assets.

### DIFF
--- a/src/clj/game/core.clj
+++ b/src/clj/game/core.clj
@@ -242,7 +242,11 @@
   ([state side card msg choices f] (show-prompt state side card msg choices f nil))
   ([state side card msg choices f {:keys [priority prompt-type show-discard] :as args}]
    (let [prompt (if (string? msg) msg (msg state side card nil))]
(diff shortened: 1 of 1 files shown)


mtgred:
:+1:

JoelCFC25:
@queueseven @mtgred 

This urgently needs to be reverted before it gets deployed--it breaks Medium and Nerve Agent multi-access, and I presume R&D Interface and HQ Interface too. I am working on addressing #32 and couldn't figure out how my code wasn't working. Then I figured out that even the existing Medium code wasn't giving extra accesses, you only see 1 card no matter how many virus counters it has when you access. I reverted this commit in my dev environment and Medium is working normally again.

queueseven:
Forgot a `nil?` check. Fix is here #726

Issue #636, opened 2015-08-29
Apex MU


PinkWarrior:
Though not hugely important as it's easily cured in game but I noticed a slight issue with Apex's MU, if you install a program face down it still takes the MU off of you.

steveklabnik:
I ran into this too :+1:

PR #687, opened 2015-09-08
Installing facedown should not cost MU mtgred/netrunner#636

queueseven:


queueseven:
Was broken with #614

mtgred:
:+1:

Issue #693, opened 2015-09-08
Bug with Gemini (or perhaps Rutherford Grid or Improved Tracers.)


StashAugustine:
I had two seperate games where, upon firing Gemini's trace sub, the runner would repeatedly receive prompts to boost the trace with no way out of the loop. Each trace would do damage, leading to an inevitable flatline. In both cases, the Gemini trace was boosted with money and with either Improved Tracers or Rutherford Grid. Other tracer ice (Ichi 2.0 and Flare were tested) worked fine with the agenda/upgrade boosts. I didn't test other constellation ice nor Gemini without agenda/upgrade boosts, but Gemini was consistently bugged each time the sub fired.

StashAugustine:
Update: Gemini is bugged even without the Corp boosting the trace or with Rutherford/Improved Tracers. Runner had to beat the trace to break the loop.

JoelCFC25:
I see the problem--there's a closing paren in line 229 in the wrong place. 

Current (wrong): `:effect (effect (damage :net 1) {:card card})`
Intended (right): `:effect (effect (damage :net 1 {:card card}))`

Issue #689, opened 2015-09-08
Bifrost array ability should be optional


josemgarcia:
The HB agenda "Bifrost array" ability is always executed, though the "when scored" ability is optional.

nealterrell:
Fixed in #694.

PR #694, opened 2015-09-09
fix Gemini, make Bifrost Array optional

JoelCFC25:
Fixes for #689 and #693, plus minor corrections for a few ability labels. Also added the new Jackson Howard method of multi-selecting from Archives/Heap to Trope so it can automate the return of cards to the Runner's Stack.


mtgred:
:+1:

PR #698, opened 2015-09-09
implement Copycat, Showing Off, Itinerant Protesters

JoelCFC25:
Fused together pieces of Howler and Bullfrog to do Copycat. The Runner needs to use Copycat after breaking an ice's subroutines but *before* advancing past it. 

Showing Off ended up being easier than I thought, now that I know how to visualize the data structures and transform them. 

Itinerant Protesters just needed a simple watch state.


mtgred:
:+1:

PR #703, opened 2015-09-10
implement An Offer You Can't Refuse

JoelCFC25:
Feel free to ignore this one if you want to wait until someone figures out how to initiate a run during the Corp's turn--I tried hard and couldn't get there. 

EDIT: I basically recreated the run definition inside the card to get it to work. Still not sure why `(effect (run serv))` and all the other combinations I tried didn't do the trick.


mtgred:
:+1:

PR #706, opened 2015-09-11
implement Rebranding Team

JoelCFC25:
Stealing from Mother Goddess. When scored, all assets gain Advertisement. In order to affect later asset installs, it listens for `:corp-install` and applies the same effect if the asset doesn't already have Advertisement. 

In full disclosure, a small problem is that if only 1 Rebranding Team has been scored and then gets forfeited (or Turntabled), existing assets won't lost Advertisement.


mtgred:
:+1:

PR #709, opened 2015-09-12
Fixes #668 - Doppelganger doesn't prompt to run again

### LINKED ISSUE #668: Doppelgänger is broken
Doppelgänger no longer seem to offer a free run after successfully running a server.

sodwyer:
Hi,

I'm still getting used to the language and dev environment, so don't hesitate to crack down on me if this fix is totally bogus :)

Doppelganger was checking if turn-events (:run) was empty (via first-event :run) but it already contained one for the current active run. I saw that quite a few cards were using (first-event :run) so I didnt want to modify that function in case of side effects. So I added a new function (has-one-event) and made Doppelganger use that instead. It seems a bit crappy though, so if there is a better way to fix this I'd be happy to re-do it.

As an aside, I saw you guys have some dev comms going on via slack. Would it be ok to get an invite for that please? It'd probably be better to ask questions on there instead of via pull requests.

Thanks

### DIFF
--- a/src/clj/game/cards-hardware.clj
+++ b/src/clj/game/cards-hardware.clj
@@ -141,7 +141,7 @@
    {:effect (effect (gain :memory 1)) :leave-play (effect (lose :memory 1))
     :events {:successful-run-ends
              {:optional
-              {:req (req (first-event state side :run))
+              {:req (req (has-one-event state side :run))
                :prompt "Use Doppelgänger to run again?" :player :runner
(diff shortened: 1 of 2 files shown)


queueseven:
Hey there. Drop mtgred a mail and I think he'll invite you :-)

sodwyer:
Hmm, not sure what happend but this is actually not working. I think I may have got it working with an earlier hack, then thought this was also working but I hadn't properly reloaded the code or something. I'll close this one for now.

PR #710, opened 2015-09-13
Add/improve cards that temporarily assign subtypes to ice

JoelCFC25:
Full implementations of Tinkering and Paintbrush, plus the addition of a barrier targeting requirement to Surfer. For Tinkering, we log the position of the ice and the server it's protecting to avoid confusion when targeting a piece of unrezzed ice. Sub Boost applies the barrier subtype to the ice it's hosted on (but doesn't remove it if that ice leaves play--I'll tackle that later in the unlikely event it gets to be an issue). 

Adds a new implementation of Rielle "Kit" Peddler, which the Runner can click during the encounter of the first rezzed ice of a turn to add *Code Gate* to its subtypes. It's of limited usefulness right now (will turn on auto-pumping ability for decoders), but who knows what might come in the future where it could come in handy.

### DIFF
--- a/src/clj/game/cards-events.clj
+++ b/src/clj/game/cards-events.clj
@@ -542,8 +542,29 @@
                :msg (msg "gain " (* 2 (count (:successful-run runner-reg))) " [Credits]")}}
 
    "Tinkering"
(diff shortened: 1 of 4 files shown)


mtgred:
:+1:

PR #713, opened 2015-09-14
implement Cyber Threat, Social Engineering, Shipment from MirrorMorph

JoelCFC25:
I feel like I hammered away on Cyber Threat for a month before finally getting it. Lessons learned from An Offer You Can't Refuse...had to manually create the run with a `swap! state` in the `:no-ability` to get it to work. 

Social Engineering steals from Tinkering to notify the Corp which ice was selected (ice position 0 = innermost). An event is registered to give the Runner the credits if that ice is rezzed anytime during the rest of the turn, then unregister when the turn ends. 

Credit to @queueseven for the recursive install function in Shipment from MirrorMorph. Without it you can't get a new remote you create with the first install to show up in the list for the rest of the installs.


mtgred:
:+1:

PR #715, opened 2015-09-15
Fisk bug workaround

chua-mbt:
Implementation of workaround suggested by @nealterrell to deal with #515.
- Clicking Fisk just before a successful-run event trigger forces the draw.
- Displays system message when Laramy Fisk can be used and describing how to the ability.

Drawback is that the player can jack out after using Fisk's ability, aborting the successful run.


mtgred:
:+1:

PR #716, opened 2015-09-15
implement Director Haas' Pet Project and The Twins

JoelCFC25:
Pet Project: Very similar to Shipment from MirrorMorph, except that we skip past the server selection prompt--on the first install, automatically make a new remote, then in the helper function we identify the outermost remote that was just created and keep installing there. 

The Twins: Enforces that the run is on this server, that the Runner has passed at least 1 piece of ice and that the most-recently-passed ice is rezzed, and that the ice you choose from HQ matches its title.


mtgred:
:+1:

PR #717, opened 2015-09-15
fix Howler

JoelCFC25:
The bioroid ice chosen is now installed rezzed and the run continues after the sub fires from the position after Howler.


mtgred:
:+1:

PR #722, opened 2015-09-17
implement False Echo

JoelCFC25:
Check that we're in a run, at least 1 ice has been passed and that the most recently passed one is unrezzed, then prompt the Corp. If the Corp doesn't have enough credits for the rez cost, only show the "Add to HQ" option, otherwise they get the choice to rez or bounce back to HQ.


mtgred:
:+1:

PR #727, opened 2015-09-17
Endless Hunger

queueseven:
- trashing is an ability cost, so you can't prevent it
- show [Subroutine] like on the printed card
- removed some whitespace


mtgred:
:+1:

PR #730, opened 2015-09-18
Revert the prompt preventing stuff

queueseven:
There was another issue and I think it's better to revert this entirely and come up with another approach to solve the underlying issue. 

The root cause is that when an effect shows a prompt, the cost is paid when the choice is done by the user, not directly after activating the effect/ability.


queueseven:
This will fix the current code to allow Shipment from MirrorMorph to work again, but I'll try to come up with a better and simpler solution

Issue #714, opened 2015-09-14
Hunting Grounds and Quicksand


yang573:
When the runner uses Hunting Grounds to prevent the "when encountered" ability on Quicksand, Quicksand still gains a power counter.

queueseven:
Yes, Hunting Grounds currently simply displays the message. All "when encountered" abilities are manual, expect Quicksand.

queueseven:
Makes me wonder if Quicksand works correctly with Femme Fatale and Inside Job. I guess not.

nealterrell:
Yeah, it's a sloppy implementation. Should probably make the Gain a Counter ability manual on card click.

nealterrell:
Joel submitted a fix in #732 to make the effect completely manual.

Issue #699, opened 2015-09-09
Quicksand


CptHilts:
Weyland barrier does not gain a power token on encounter but only when corp click no more action.

Issue #638, opened 2015-08-29
Cortez Chip does not print which ice was targeted


nealterrell:
So the Corp doesn't know which ice will cost more to rez unless the Runner is honest.

nealterrell:
Fix submitted by Joel in #732.

Issue #311, opened 2015-07-02
Ice carver, parasite, quicksand


MasterAir:
I'm fine if this never gets resolved, but I encountered it and it might be easy to solve.

A quicksand is installed and has 1 power counter on it. A parasite is installed on said quicksand with 0 virus tokens. Ice Carver is installed.

When the runner runs the quicksand ice carver should resolve reducing the strength of the quicksand to 0 and trash the quicksand before it gains a power counter.

In fact, in the jinteki.net implementation the quicksand gains a strength and is therefore not trashed.

nealterrell:
Same issue as #309 etc.; we don't deal well with simultaneous effects resolution. The order they resolve is largely random. I don't think we have a timetable for fixing this as it's very very complicated. Thanks for the report though!

JoelCFC25:
We made adding power counters to Quicksand a manual process, so the correct result (Parasite + Ice Carver insta-trashing it on encounter) should happen now.

PR #732, opened 2015-09-19
Cortez Chip target ID, Quicksand manual effect, discounted Monolith program installs, fix DaVinci

### LINKED ISSUE #735: Davinci won't fire for 0.
Had an issue in a game last night where I tried to fire a Davinci to install an SMC from hand. This should be doable as you can fire it for 0 tokens.

JoelCFC25:
The addition to Cortez Chip solves #638 in the same manner as Tinkering does now with unrezzed ice--just identify the server and the position (0 = innermost). 

Adding a power counter to Quicksand is now done manually by the Corp, which should hopefully alleviate some or all of the issues in #714, #699, and #311. 

The discounted program installs granted when installing Monolith are restored (got reverted some time ago erroneously), and work better now--the recursive function in Shipment from MirrorMorph is ported over so you can order the installs (e.g., maximizing Overmind counters by installing it first). Note: this function won't work until the reversion in #730 is merged. 

Added a nil check to DaVinci so it can install 0-cost programs when it has no counters. Fixes #735.

### DIFF
--- a/src/clj/game/cards-hardware.clj
+++ b/src/clj/game/cards-hardware.clj
@@ -89,10 +89,18 @@
                                        :effect (effect (play-instant target))}}}}}
 
    "Cortez Chip"
-   {:abilities [{:label "increase cost to rez a piece of ice by 2 [Credits]"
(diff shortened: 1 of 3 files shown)


mtgred:
:+1:

PR #736, opened 2015-09-19
Unit testing framework

nealterrell:
This is my attempt at getting a unit testing framework going. With the current lull in card development, I think now is a good time to get proper testing going using the convenience of our current circle-ci integration. Newer developers in particular can benefit from writing a few unit tests to better learn the core engine capabilities and the ways that card operations affect the game state. Using clojure's `clojure.testing` namespace, we can write small unit tests to ensure the validity of card operations, to validate things like "Using Sure Gamble should net 4 credits", or "Using Account Siphon on a corp with 5 credits should leave them at 0, give the runner 10cr and 2 tags". 

I will work on a more thorough documentation of the system, but this was my general approach:

1. Tests go in src/clj/test, in files mirroring the core game files.
2. test/core.clj, test/utils.clj, and test/macros.clj provide enough utility functions to make writing the tests as easy as possible. For example, a helper function `newgame` can quickly set up a new game with particular cards in each deck without having to go through the entire process manually.
3. For each card, define at least one test in an appropriate file. Use the utility functions to play the card, interact with its abilities, and then verify that the game state has changed as predicted.

Example test for Magnum Opus:

```clojure
(deftest magnum-opus-click
  "Magnum Opus - Gain 2 cr"
  (do-game
    (new-game (default-corp)
              (default-runner [(qty "Magnum Opus" 1)]))
    (take-credits state :corp)
    (play-from-hand state :runner "Magnum Opus")
    (is (= 2 (:memory (get-runner))))
    (is (= 0 (:credit (get-runner))))
    (let [mopus (get-in @state [:runner :rig :program 0])]
      (card-ability state :runner mopus 0)
      (is (= 2 (:credit (get-runner))) "Gain 2cr"))))
```

* `do-game` is a macro to give you a `state` variable as in the core.
* `new-game` starts a new game and returns its state, taking as arguments a deck to use for corp and runner.
* `(default-corp)` returns a map for a "default" corporation of HB Custom Biotics (no effect on game) with 3x Hedge Fund.
* `(default-runner)` returns The Professor with 3x Sure Gamble... unless you provide a vector of cards to use as the deck.
* `(qty _name_ _amount_)` loads a card with a given name and puts that many into a vector for the deck.
* After `new-game`, the game is started and on the corp's first click after mandatory draw
* `take-credits` ends a player's turns by taking a specified # of credits (or max if unspecified) until their turn ends, then starts the next turn.
* `play-from-hand` finds a card with a given name in the player's hand and calls `play` on it (what the UI calls to play a card when clicked)
* `is` enforces a condition as being true, and fails the test if it is false. Here, we make sure that Magnum Opus leaves 2 unused memory and 0 credits.
* We retrieve the installed version of Opus directly from the state.
* `card-ability` triggers an ability of a card based on its index, counting from 0 for the first ability. The first ability of Opus is the Click: Gain 2cr.
* We make sure the runner now has 2 credits.

If someone accidentally breaks Magnum Opus, perhaps by changing its `(gain :credit 2)` to `(gain :credit 1)`, then the `is` at the end of the test will fail and report the string "Gain 2cr".

Sometimes you need to interact with the state in other ways, particular for answering prompts. Here is a test for installing Atman at strength 2.

```clojure
(deftest atman-install-2
  "Atman - Installing with 2 power counters"
  (do-game
    (new-game (default-corp) (default-runner [(qty "Atman" 1)]))
    (take-credits state :corp)
    (play-from-hand state :runner "Atman")
    (core/resolve-prompt state :runner {:choice 2}) ; pay 2 credits
    (is (= 3 (:memory (get-runner))))
    (let [atman (get-in @state [:runner :rig :program 0])]
      (is (= 2 (:counter atman)) "2 power counters")
      (is (= 2 (:current-strength atman)) "2 current strength"))))
```
1. Start a new game with Atman in the runner's deck.
2. Pass to runner's turn, the install Atman. This puts a prompt into the state that we need to answer.
3. `core/resolve-prompt` calls the "resolve prompt" code from core, which requires a map of `{:choice _choice_}`; here the prompt is asking for credits, so we choose 2. Other prompts ask for strings, particularly optional prompts, which can be resolved with "Yes" or "No".
4. Make sure Atman cost memory, now has 2 counters, and has a current strength of 2.

I have included at least one test for each card type in this submission, as an example of the kinds of tasks that type of test might need.

----

I appreciate comments from other developers about this framework. If people like it, I would then propose that we assign test-writing to newer developers to learn the system before they start working on cards. If we all contribute 1 test a week then we will quite soon have a good coverage system to help avoid issues like the bugs plaguing the recent `:optional` prompt rewrite, or with Account Siphon not giving tags, to name two examples. We could also insist that every card submission be accompanied by appropriate unit tests.

Thoughts?


mtgred:
Awesome! :+1:

domtancredi:
This is great! Perfect for preventing regressions.

Suggestions: just to stay organized, can we have a list of tests we want to do and for each card? Is it one test per card? 

For example: *Research Grants:* I tested that the agenda score worked for Corp and Runner. I tested if it was scored by the Corp, and one existed, would it also score? Then tried again with two on the board. Could be around 4 tests in total for code coverage (not including default scores).

PR #740, opened 2015-09-21
Fix uniqueness. The unique flag should not be removed from cards when

queueseven:
creating the decks at game start.


mtgred:
:+1:

PR #741, opened 2015-09-21
STOP!/Thinking/BRB buttons

queueseven:
click them to display a message in the log


queueseven:
bild 11

queueseven:
I actually have a better idea to implement this.... Let's see if I find the time...

PR #743, opened 2015-09-21
fuller implementations: Pawn, Bishop, Bandwidth, Sage

JoelCFC25:
Bandwidth now takes away a tag when the run is successful. 

Once hosted, Bishop will now require bouncing back and forth between central servers and remotes as required, and will prompt the Runner accordingly. 

Pawn automatically advances to the next-innermost ice when it sees a successful run, and then trashes itself and kicks off the free Caissa install from Grip or Heap once it advances past the last ice. 

Sage checks for the correct ice subtypes now.


mtgred:
:+1:

Issue #552, opened 2015-08-18
Feature Request: Reverse the order of upgrades


jannn:
New cards installed in a server are added on the left side rather than the right, with the old ones being moved without any visual indication (if they're all unrezzed). This is counterintuitive, and the order of installation between turns is relevant knowledge for the runner.

Issue #551, opened 2015-08-18
Feature Request: Dynamic spacing of servers to avoid overlapping upgrades


jannn:
This is a prety major annoyance with decks that like to put 2 or sometimes even more upgrades on servers, for both sides. The third card in a remote server isn't clickable at all if there's another one to the right, and it just looks messy and confusing. Multiple upgrades on central servers are less common, but also very messy.

JoelCFC25:
This is a known bug--see #427.

jannn:
Searched a few keywords and couldn't find it - sorry! Still, I think it'd be nice to implement dynamic spacing for aesthetic/readability purposes as a more general solution, even if it's technically functional once fixed.

JoelCFC25:
Not a problem--when in doubt, report it!

Issue #512, opened 2015-08-08
Add a horizontal scrollbar when corp has lots of servers


SomeSketchyGuy:
In a game with two players on the corp side the cards will go off the bottom of the screen along with the menu for selecting the servers, making the game unplayable

SomeSketchyGuy:
I just updated chrome so it may have something to do with that since it was working fine before

queueseven:
Have you tried just zooming out?

Could you tell what version of chrome you're running and perhaps provide a screenshot?

SomeSketchyGuy:
Zooming out worked, I believe I'm running the latest version of chrome which is 44.0.2403.130

queueseven:
I have a small laptop and I can only play when setting the zoom level to 90%; that's a good workaround

mtgred:
I also believe it is a Chrome issue. Version 43 broke the game log. Version 44 pushes the bottom of the board outside the window when there are too many content. It also causes #427.

zaroth:
I can confirm that this issue exists in latest Firefox too, I was unable to play as/vs horizontal corp decks since zooming out in Firefox works differently than in Chrome - just the text (Server 0/1/2 etc.) got smaller, not the card graphics themselves. Only after installing Firefox plugin [Zoom Page](https://addons.mozilla.org/en-US/firefox/addon/zoom-page/) I was able to zoom out and see whole board properly.

I tried to dabble in the UI with Firebug to try and fix it, tinkering with *overflow* CSS property, but unfortunately my CSS skills weren't strong enough to make Firefox show the layout properly.

queueseven:
related #1383 

To get this behaviour right now, use a browser plug in that lets you add custom CSS to a website (I use `User CSS ` for Chrome), and add

```
.gameboard {
   overflow-x: auto;
}
```

to get a scrollbar.

Issue #427, opened 2015-07-23
Can't click on cards in servers with 2 upgrades and agenda


sperry023:
This occurred in two separate games for me -- I had a scoring server started (Server 0 in the image below), and then created Server 1. As the game went on and I added more upgrades to Server 0, Server 1 did not shift to allow space for it (there is a Server 0 upgrade under the PAD Campaign). Even though I can see the card, hovering over it did not show me anything in the preview window and clicking did not rez it. Zooming out did not make the cards shift. I am using Google Chrome on a MacBook Air.
![upgrades problem](https://cloud.githubusercontent.com/assets/13188311/8853468/cddc7b62-3128-11e5-9468-d7ba1423d591.png)

mtgred:
Thanks for reporting the issue. The last update causes a layout issue with upgrade positioning. When there are more than 2 cards in a remote, the cards on the right go are underneath the next remote server and they can’t be clicked anymore.

The code has been reverted on the server while this is being investigated.

mtgred:
It seems to be problem with the latest Chrome update (version 44). The problem is also present on Firefox on Mac OS though.

eedok:
![card-overlap](https://cloud.githubusercontent.com/assets/7716019/9767509/c9084d2c-56db-11e5-9214-7dd6b6391250.jpg)

Looks like it happens on the windows version of firefox as well, as a temp workaround maybe we can have it so we can manually drag servers around(or create empty servers and rearrange them by dragging the label)?

JoelCFC25:
This is more or less fixed, though remotes with 4 or more items might still be a little tricky.

PR #745, opened 2015-09-22
Let upgrades overlap in remote servers

queueseven:
this enables multiple cards in a remote server to overlap:

bild 13

so you can actually play with 3 (or 4) cards in a single remote.
In a perfect world the (CSS-)size of a remote would increase with more cards, but I think it's not worth implementing.

Also, the order of cards in a server is no longer reversed, so installing a new card in a server does not move the old cards to the side (so you can better keep track of the order in which the cards were installed).

Relates to #552, #551, #512, #427 and probably others, too.


mtgred:
Awesome! :+1:

queueseven:
I could be better, but I didn't get it to work like I want :-(
See my recent messages on Slack.

davide-ceretti:
Thanks a lot, this is awesome!

Issue #733, opened 2015-09-19
ICE strength needs to be updates after purging virus counters


queueseven:
Because of one of the most common virus programs, Parasite.

Issue #723, opened 2015-09-17
IQ does not work


queueseven:
It needs a `watch` to update its strength when the cards in HQ change.

JoelCFC25:
It doesn't change instantly when hand size changes, but whenever it gets encountered it will update to the correct strength according to the number of cards in HQ at that moment.

PR #747, opened 2015-09-22
Fix IQ to update strength dynamically, update all ice after a purge

JoelCFC25:
Fixes for #723 and #733.   EDIT: Added a fix for #749.

Add a watch state to IQ so its strength updates in real-time as the Corp's hand size changes. Purging viruses is now followed up by a call to `update-all-ice` so strengths get shown correctly right away instead of waiting until the next encounter. Messages for derezzing a card and forfeiting an agenda are also tweaked to match the present tense of other game action messages.


mtgred:
:+1:

Issue #198, opened 2015-05-19
Kaplan + Comet + Starlight Crusade Funding


osiridecascioli:
I was testing my deck and I think I found a bug; I made a test to be sure.
Given:
Identity: Kaplan,
Comet hardware on the Rig,
Starlight Crusade Funding on the Rig.
If as my first click I play an event, specially an event of subtype "double", I was expecting to play another event (maybe another double) all at 1 click cost.
Now, it cost me 3 click.

eah13:
Playing two doubles with Comet would cost 3 clicks because you must still pay the one click additional cost.  Playing a double and a normal event should cost only 2 clicks.

osiridecascioli:
Sorry but I disagree; Starlight C.F. should prevent any additional click costing, so, for example:
on click 1: Lucky find (1 click because Starlight C.F.) then Comet allow me to play 1 more event, Lucky Find, and no additional click costing because Starlight C.F.

eah13:
Oh you're exactly right.  Sorry, forgot about Starlight Crusade.  With that and Comet you can play two of any normal or double event for 1c.

I think your original issue then should be about Startlight Crusade not working correctly.  Sure enough, it's shown as [only partially implemented on the big ol spreadsheet](https://www.dropbox.com/s/rpkhxafgile5spp/Cards%20status.xlsx).

JoelCFC25:
Starlight Crusade Funding implemented at #752.

PR #752, opened 2015-09-23
fully implement Starlight Crusade Funding

JoelCFC25:
On install and at the beginning of every subsequent turn, put a key in `[:runner :register]` that `core\play-instant` checks before applying the extra click cost of a Double event. This takes care of #198.

Improved Power Nap to report how many credits were gained.


mtgred:
:+1:

Issue #753, opened 2015-09-24
Replacement effects on runs don't trigger success state in game (datasucker, data leak, notoriety, quest complete)


StalkerB:
If I run on archives and install Hades Shard when Successful Run comes up, it doesn't count as a successful run for the purposes of gaining sucker tokens or counting towards a successful run on that central.

JoelCFC25:
I think we can fix this by manually swapping some stuff into `[:runner :register]` and `[:run :successful]`. I'll give it a shot...

PR #754, opened 2015-09-24
allow free install of Shards from hand to count as successful runs

JoelCFC25:
Fix for #753. Some code in `(defn successful-run` is split off into a separate function which can then be called by the Shards when installing from hand. 

This will let Datasucker tokens accrue, update `[:runner :register]` for Notoriety/Quest Completed, etc. 

Added a fix for Executive Boot Camp's asset search message, and shifted Keegan Lane out of the assets file into upgrades.


mtgred:
:+1:

PR #755, opened 2015-09-24
Redo on user-selected skill levels for games

nealterrell:
Simplified and merged with the spectator server changes.

Example

If you want to set your own game title, go ahead -- it's still editable.

Otherwise select a game type (Private, Friendly, Serious, Tournament) and skill level (Beginner, Intermediate, Expert), and the title gets automatically filled. No other changes to the server or user experience.

The assumption is that the average Jinteki player is looking for a friendly game with intermediates. Therefore we don't say anything when Intermediate is selected, only calling attention if Beginner or Expert is. Game titles then turn into things like

* Friendly game -- the default title
* Serious game -- "Intermediate" skill implied
* Private game
* Friendly game for beginners -- call attention to the fact you're a beginner
* Serious game for experts -- likewise.

"Tournament" was meant to be an indication of top-tier play/decks. I think the community will naturally discover which labels are appropriate to which types of games, so we don't need 100% definitions for each.

### DIFF
--- a/src/cljs/netrunner/gamelobby.cljs
+++ b/src/cljs/netrunner/gamelobby.cljs
@@ -49,7 +49,9 @@
 (defn new-game [cursor owner]
   (authenticated
    (fn [user]
-     (om/set-state! owner :title (str (:username user) "'s game"))
+     (om/set-state! owner :title "Friendly game")
+     (om/set-state! owner :skill-level "Intermediate")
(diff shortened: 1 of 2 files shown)


queueseven:
I fully support this, since it's a common complaint of good/serious players that they don't have a better way of filtering out the noobs

PR #756, opened 2015-09-24
Fix for multiple Team Sponsorships. Testing iteration.

### LINKED ISSUE #748: Team Sponsorship timing
There's a small timing issue with Team Sponsorship. If you have two Team Sponsorships installed, it will prompt you to select both cards then prompt install both. This can cause trouble in NEH where you may want to install a card in a new remote, drawing a card, and then potentially install the new card.

nealterrell:
Rather than let each Team Sponsorship fire individually, we make the first one to fire do all the work by counting how many TS are installed and rezzed, then recursively resolving the TS ability once for each installed copy. This forces Corp to decide on a server for the first install before the next install location is selected.

Fixes #748.

### DIFF
--- a/src/clj/game/cards-assets.clj
+++ b/src/clj/game/cards-assets.clj
@@ -547,15 +547,22 @@
                  :msg "swap the positions of two ICE"}]}
 
    "Team Sponsorship"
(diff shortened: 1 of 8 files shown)


mtgred:
:+1:

PR #758, opened 2015-09-25
Guard against card name being displayed during grip-to-grip movement

### LINKED ISSUE #86: Runner accidentally reveal grip cards
When Runner (idk if Corp does as well) accidentally moves card in his/her grip, status board on the right shows the name of the moved card, revealing it to the Corp.

chua-mbt:
Fixes #86

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -11,7 +11,8 @@
                  [sablono "0.3.4"]
                  [environ "1.0.0"]
                  [com.novemberain/monger "3.0.0-rc2"]
-                 [org.slf4j/slf4j-nop "1.7.12"]]
+                 [org.slf4j/slf4j-nop "1.7.12"]
+                 [org.clojure/core.match "0.3.0-alpha4"]]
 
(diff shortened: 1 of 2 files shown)


queueseven:
Cool. Maybe you could also prevent the message when the corp moves a card in HQ? While it wont reveal the card name it's still confusing

chua-mbt:
You mean suppress the system message entirely for both corp and runner for hand-to-hand movement? Sure.

queueseven:
Yes. While you're on the run: it would be cool to display the previous zone also. So instead of `The runner moves a card to grip`, it would be cool to display something like `The runner moves a card from play to their grip` or `...from heap to their grip` and for the corp something like `The corp trashes a card installed in Server 2` etc. 

That's a bigger change, maybe you want to do it next :-)

chua-mbt:
Done and done. Tweak the messages as needed.

mtgred:
:+1:

PR #763, opened 2015-09-26
Use async version of bcrypt.compare

rluba:
Previously, the whole node server was blocked for a short period of time whenever someone logged in.

I couldn't find any tests for the node server. Are there any?


queueseven:
There are no Tests for the Node Server. We recently started writing Tests for cards and the game engine. If you could start writing some tests for the node server that would be awesome!

rluba:
I'll give it a try (as separate PRs). But don't let it keep you from merging this PR :-)

mtgred:
:+1:

PR #764, opened 2015-09-26
Feature/better password hash

rluba:
1. Deduplicate the `salt` parameter constant for `bcrypt.hash(…)` by extracting it to a function.
2. `3` is a _really weak_ setting for the number of rounds. See the [`brcrypt` readme](https://github.com/ncb000gt/node.bcrypt.js#a-note-on-rounds) for details. I've increased it to the default `10` which means `2^10` rounds.

This change will make login/registration marginally slower but the effect should be mitigated by #763 (moving the expensive login operation to an async worker).

### DIFF
--- a/server.coffee
+++ b/server.coffee
@@ -361,7 +361,7 @@ app.post '/reset/:token', (req, res) ->
     res.redirect('/')
 
 hashPassword = (password, cb) ->
-    bcrypt.hash password, 3, cb
+    bcrypt.hash password, 10, cb
 
 app.get '/messages/:channel', (req, res) ->
   db.collection('messages').find({channel: req.params.channel}).sort(date: -1).limit(100).toArray (err, data) ->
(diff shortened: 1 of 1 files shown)


rluba:
… of course this will only take effect for newly registered users and those who use the "forgot password" feature.

But we could extend the login function to rehash any weak password upon login.

zaroth:
I think if you provided migration script to batch rehash all current passwords from DB (so, 7 times?) up to the new standards this change would be way less 'icky' and easier to go through with.

rluba:
The first step would be to merge this PR so that new users benefit from the increased security. Old password hashes will **remain valid**, because bcrypt stores the number of rounds together with the hash.

Regarding @zaroth’s request: You can’t "upgrade" hashes without knowing the clear-text password because the password is part of the input for each round. [See this Stack Exchange answer for more information](http://security.stackexchange.com/a/15876). So patching the login function to gradually improve security is the only way forward for existing users.

zaroth:
Thanks for clarifying how it works, didn't know it's so flexible.

Tested it locally, could both login with the old users and create a new user and log him in - so that's a :+1: from me.

mtgred:
:+1: 

Thanks for the clarification

rluba:
NP, it’s great to see this finally merged! :tada:

Issue #765, opened 2015-09-26
Two bugs with Team sponsorship


heydevhey:
HI guys,

Awesome site!  I've noticed two bugs using the new team sponsorship:

1- I have two pad campaigns in the archives.  I score a breaking news while two team sponsorships are installed and rezzed.  I'm prompted to install a card from archives and click PAD Campaign both times.  I'm able to install the first time, but the second time the install fails and no cards are installed.

2- I have 2 cards in archives. One faceup pad campaign and one face up pad campaign.  When I score a breaking news with a Team sponsorship installed, i can choose to install a pad campaign, but am not able to choose weather to install the face down pad in archives or the face up one in order to make make the runner check a card.

Hope this makes sense.

Devin

nealterrell:
Thanks for the report! We have addressed the first issue in #756, but the second won't be addressed. I'll see how easy it will be; probably we'll make it repeat card titles for duplicate cards, and they'll be in the same order they appear in archives.

nealterrell:
Alright, that was easy. I am applying the logic to Interns, Crick and Architect as well.

HongkongKoma:
I can report another bug: Opponent has 2 Gang Signs installed, me 1 Team Sponsorship. When I score, the Gang Signs fire before I can choose a card to install. Since it's my turn, I might want to install an agenda from HQ first - before Gang Sign triggers.

JoelCFC25:
Yet another example of #309, I'm afraid--until blocking prompts are developed, there isn't a way to properly order multiple effects firing from a single triggering event. It's a priority!

JoelCFC25:
Both Gang Sign and Team Sponsorship are now manually triggered, so players can properly sequence everything when multiple copies of each are in play.

PR #766, opened 2015-09-26
Have Team Sponsorship prompt show duplicate entries

nealterrell:
... to address #765. Installing a card from Archives requires knowing more than just the card title, as we may want to choose a card that is specifically face up or face down. Without `:not-distinct`, we only get one entry for each card title even if there is more than one card in Archives; now, we choose which card we want based on the order it appears in Archives.

I have also applied this to other cards that install things from Archives: Interns, Crick (fixed to only show non-Operations), Architect.


mtgred:
:+1:

PR #772, opened 2015-09-29
add recursive installs to Accelerated Beta Test and Mass Install

JoelCFC25:
Extending @queueseven's recursive technique to ABT and Mass Install. This is especially needed for ABT so the Corp can order the ice installs, as well as allowing a new remote created with the first one to be available in the server choices for the rest of the ice. The cards are first moved to the play area and then can be selected for install from there.

### DIFF
--- a/src/clj/game/cards-agendas.clj
+++ b/src/clj/game/cards-agendas.clj
@@ -2,13 +2,21 @@
 
 (def cards-agendas
   {"Accelerated Beta Test"
-   {:optional {:prompt "Look at the top 3 cards of R&D?"
-               :yes-ability {:msg (msg (let [c (count (filter #(= (:type %) "ICE") (take 3 (:deck corp))))]
-                                         (str "install " c " ICE and trash " (- 3 c) " cards")))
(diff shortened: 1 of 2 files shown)


JoelCFC25:
Downside: this won't comply with [the ABT/Foundry ruling](http://ancur.wikia.com/wiki/Accelerated_Beta_Test_%26_The_Foundry_Interaction_Ruling), but the current implementation of ABT doesn't either. Not sure how big of a deal that is. Making the interaction of those 2 cards work right will be as insanely difficult as making Accelerated Diagnostics work.

queueseven:
I don't think it's a big deal, since you can always adjust the game state manually. Nice work, keep going!

JoelCFC25:
I put in a `trigger-event` for a rez since doing the `corp-install` with `:install-state :rezzed` doesn't result in anything that will fire The Foundry's ability. I can either add a `when` statement that generates a prompt only to the Corp if the ID is The Foundry, or leave it as-is and rely on the players to know the ruling and trash any remaining uninstalled ice from firing ABT if they opt to use The Foundry.

mtgred:
:+1:

Issue #767, opened 2015-09-27
Haarpsichord Studios: + Franchise City result in game crash


lorion42:
If the corp creates two new servers one with an agenda and one with Franchise City. On the runners turn he runs on the agenda server, while the corp rez Franchise City. Result: the runner can't click on the steal button. Franchise City is scored before the agenda could be stolen. There is no way to go to standard game flow again - the code seems to be in a loop. If the agenda is for exampe TGTBT, the runner gets a tag for each mouse click but he can't get TGTBT into his scoring area.
This happend 3 times in a row. There are turns where Franchise City works as it should, but with this bug a deck with both cards is unplayable.

nealterrell:
This will crash if the asset is in a lower numbered server than the agenda. Obviously we can't expect you to always put your cards down in one order, but we are aware of the issue and working on a fix. Thanks!

Issue #742, opened 2015-09-21
Security Testing on a remote checks the wrong remote if an earlier remote is destroyed


anathomical:
If you have Security Testing installed at the beginning of your turn, and select Server 1 as its target, then you run Server 0 which has no ice and trash whatever is in it, Server 1 falls into the position of Server 0.  If you run the new Server 0, Security testing will not trigger.  If you then run Server 1 (assuming that there was a Server 2 at the start of the turn to fall into this position), then Security Testing will trigger rather than allowing you access.

Basically, it looks like Security Testing ties itself to a server index when you select at the beginning of turn, but that index is not updated with the destruction of servers which may change the contents of the server at that index.

JoelCFC25:
Server destruction is definitely a problem. This same issue plagues Franchise City (or any asset/upgrade that might disappear or be trashed mid-run)--if you have one rezzed when an agenda is accessed in a higher-numbered remote, the Runner's steal prompt for the agenda is bugged because of collapsing servers and the steal cannot be completed.

queueseven:
I'll try to fix this tomorrow if I find the time (and fixing means removing the server destruction code. Too...many...problems....)

nealterrell:
Yeah, this was noted when ST was implemented, but I didn't know how to fix it at the time. Sounds like @queueseven is on it!

PR #773, opened 2015-09-29
Give remote servers a keyword

queueseven:
instead of storing them in a vector.

This fixes some bugs (mtgred/netrunner#742 mtgred/netrunner#767) and makes the code actually a little simpler.

Also, it's easier to keep track of the game state, since once a remote server is gone, the next remote will have a different number.


mtgred:
:+1:

Issue #381, opened 2015-07-17
Scheherazade misclick causes programs to disappear


finnbar:
Sorry, I've got two issues to report today! It turns out playing a Professor deck means you get to play with lots of different cards, so it's quite easy to find bugs with them.

If you misclick when using Scheherazade's "Host an installed program" action and ask it to host itself, it gets into a loop of some kind and all of the programs disappear. They're not trashed, removed from play, returned to your grip or whatever, they just disappear. This makes it especially hard if you're playing with Levy. Any newly installed programs also disappear, and your used MU stays the same which would suggest that the programs are still considered to be there, but aren't visible at all.

JoelCFC25:
I have an idea to fix this that I'll try tonight.

PR #774, opened 2015-09-29
Don't allow self hosting

### LINKED ISSUE #739: Leprechaun Self-Hosting
Today, I accidentally tried to host Leprechaun on itself due to lag. Leprechaun and a hosted card disappeared as a result. They were not in the heap, stack, or grip, and they weren't marked as removed from the game. Should it perform a check to prevent such an occurrence?
screenshot 1

queueseven:
Fixes #739 and #381.

It's not perfect as Scheherazade will still print its message, but adjusting that is a rather big change. Also, we have to way to check the `:cid` of the hostin `{:choices {:req ....`, so there's no easy way to prevent a card from selecting itself.

At least the cards don' disappear any longer.


mtgred:
:+1:

PR #779, opened 2015-09-30
fix for Laramy Fisk forced draw

JoelCFC25:
The new Fisk code in #715 doesn't actually result in the Corp drawing a card. This works. :)


mtgred:
:+1:

PR #783, opened 2015-10-03
fix Darwin to charge 1 credit when optionally boosted

JoelCFC25:
The `:msg` and `:cost` needed to be shifted into the `:yes-ability` block in order to charge the Runner 1 credit and display the message.


mtgred:
:+1:

PR #786, opened 2015-10-05
Send game list updates at most 1 per second

nealterrell:
Right now we send the entire game list (typically in the 10-50Kb range depending on traffic) any time a player:

* creates a game lobby
* joins a game lobby
* leaves a game lobby
* disconnects from a game in progress
* reconnects to a game in progress

This often happens multiple times per second, causing lots of data to be sent which will only be overridden a few milliseconds later. This change sets a JavaScript interval to 1000 milliseconds, and only emits the game list in this interval if something caused the list to change since the last interval. This should reduce the frequency of lobby updates which are a non-insignificant portion of the site's traffic. 

It should also help with the bug we're seeing re: rapid disconnects/reconnects, which cause tremendous lag precisely (I believe) because they cause so many lobby updates to be pushed rapidly.

### DIFF
--- a/server.coffee
+++ b/server.coffee
@@ -25,9 +25,13 @@ db = mongoskin.db(mongoUrl)
 # Game lobby
 gameid = 0
 games = {}
+lobbyUpdate = false
 
 swapSide = (side) ->
   if side is "Corp" then "Runner" else "Corp"
+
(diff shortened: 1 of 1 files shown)


mtgred:
:+1:

PR #787, opened 2015-10-05
implement Worlds Plaza and Awakening Center

JoelCFC25:
These cards caused a series of other tweaks to support them. The `gameboard.cljs` needed an addition to allow hosted Corp cards to be clickable so abilities would work (e.g., Private Contracts). The access helper for remotes in `core.clj` needed to account for all hosted cards in addition to server contents--this way you could first access and trash one of the scoreable hosted executives instead of just trashing Worlds Plaza and all the hosted assets with it. Unregistered S&W also needs to filter on hosted cards in case someone wanted to shoot a qualifying hosted card. 

Awakening Center uses a funky `doseq` toward the end to get any rezzed hosted ice to trash, but there should only ever be 1 rezzed at a time. There is probably a more artful way of doing this, but it seems to work fine.

### DIFF
--- a/src/clj/game/cards-assets.clj
+++ b/src/clj/game/cards-assets.clj
@@ -34,7 +34,7 @@
    "Blacklist"
    {:effect (effect (lock-zone (:cid card) :runner :discard))
     :leave-play (effect (release-zone (:cid card) :runner :discard))}
-                 
+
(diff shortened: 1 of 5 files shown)


mtgred:
:+1:

Issue #790, opened 2015-10-05
Certain cards that alter ICE strength don't trigger until after a run


thehunter518:
So far I have noticed this issue with the agendas "Superior Cyberwalls" and "Encrypted Portals", as well as the ice "Quicksand".

With the agendas, it seems that the strength increase to ice does not kick in until the runner has completed a run (sometimes the strength suddenly kicks in mid-run, and sometimes it only kicks in for the server currently being run on).

Quicksand's strength from new power counters doesn't kick in until after the runner is done encountering it. For example, if the runner encounters a fresh, newly rezzed Quicksand, you can click it to add a power counter, but its strength remains at 0 while being encountered. As soon as the runner clicks to continue (or jacks out), the strength suddenly correctly adjusts to where it should be for the power counter you added.

JoelCFC25:
These cards should call `update-all-ice`, which they currently are not doing. Should be an easy addition, thanks!

PR #791, opened 2015-10-05
Show updated ice strengths (Quicksand, Encrypted Portals) immediately

JoelCFC25:
Addressing the request in #790--these cards now call `update-all-ice` so you don't have to wait until the next encounter or Parasite install to see the new strength.


mtgred:
:+1:

PR #794, opened 2015-10-07
fix Franchise City, target and identify ice when installing Femme Fatale

### LINKED ISSUE #793: Franchise city doesn't respect Film Critic
A ressed franchise city gets auto-added to the score area before the runner has a chance to host on film critic and stop it.

### LINKED ISSUE #796: Forged Activation Orders doesn't account for discounts on space ice
Possibly other stuff as well.

JoelCFC25:
Fixes #793 by changing Franchise City to listen for a later trigger, allowing Film Critic to snipe the agenda first. Fixes #796 by using `rez-cost` on the targeted ice rather than its printed cost, to allow for bonuses from space ice, Xanadu, etc. 

Cortez Chip and Tinkering are refactored to use the new server name function, and Femme Fatale now prompts for and identifies the ice selected for bypassing in the log window.

### DIFF
--- a/src/clj/game/cards-assets.clj
+++ b/src/clj/game/cards-assets.clj
@@ -201,8 +201,9 @@
                  :effect (effect (trash card) (lose :bad-publicity (:advance-counter card)))}]}
 
    "Franchise City"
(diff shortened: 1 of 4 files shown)


mtgred:
:+1:

PR #795, opened 2015-10-08
Send diffs of game state to clients

nealterrell:
### State diffs
Rather than send the entire game state to each client after an action is performed, we send a much smaller diff structure that the client can apply to its copy of the state.

There are a few changes here:

1. The Clojure library `differ` calculates and applies diffs of the state map. I evaluated two different libraries for this and found `differ` to be best. It's not perfect -- it will send an entire vector if one element of the vector is removed, as when drawing a card (will send the entire new deck in the diff) -- but it consistently sent smaller diffs than the other library with no drawbacks.
2. In addition to storing the current game state for each game in progress, the server now also saves the "last state" sent to the clients of the game. Diffs are calculated from this last-state. This __effectively doubles__ the server's memory requirements for tracking game states, but historically that has not been an issue. If this becomes a concern, we can try making a duplicate of the state upon receiving a message and then using that to calculate the diff after applying the message command; the current implementation was just a little more simple.
3. In the game engine core, we send the full state when starting a game or receiving a "notification" (sent when a player joins or reconnects to the game -- necessary for observers and people with bad connections). Otherwise we send a diff relative to the last-state.
4. When the client receives a message from the server, it examines the message to see if it contains a state or a diff. If a state, it overrides its local state as before; otherwise, it uses `differ` to "patch" the local state with the diff from the server.

#### Bandwidth effect

A typical game starts with a state size of about 30Kb (prior to compression over the WebSocket from #731), and this size tends to grow over time due to log messages, new corp servers, etc. The size of a diff depends on what has changed, naturally. Brief experiments showed these approximate sizes:

* Drawing a card: 10Kb (depends on how many cards are left in deck/hand)
* Playing an instant: 0.5Kb
* Installing to Corp: 1.5Kb
* Installing to Runner: 1.3Kb
* Click for credit: 0.4Kb
* Adjusting player values (clicks, credits, points): 0.4Kb
* End turn: 0.5Kb
* New chat message: 0.4Kb
* Rez a card: 1Kb
* Each step of a run: 0.5Kb
* Accessing a card: 1Kb

Not bad :sunglasses:, since the "old" code would send the full 30Kb+ state for each of these.

### App version number

I took the time to add a version string to the coffee server, which gets appended as a query parameter to the JavaScript include for `app.js`. The string is pulled from environment variable `APP_VERSION`, or falls back to a default string; the string is sent to `layout.jade` for doing the `app.js` include. Browsers will see the query parameter and request a new version of the app script any time the version string changes. With this, we can force browsers to download changes to the client scripts without having to beg people in chat. This is especially important for the diff changes; without a refresh, they simply won't be able to play the game. @mtgred, you can decide if you want to update the environment variable yourself, or if you would like us to manually increment the string any time client code changes.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -12,7 +12,8 @@
                  [environ "1.0.0"]
                  [com.novemberain/monger "3.0.0-rc2"]
                  [org.slf4j/slf4j-nop "1.7.12"]
-                 [org.clojure/core.match "0.3.0-alpha4"]]
+                 [org.clojure/core.match "0.3.0-alpha4"]
+                 [differ "0.2.1"]]
 
   :profiles {:dev {:dependencies [[figwheel "0.3.7"]]}}
 
--- a/server.coffee
+++ b/server.coffee
@@ -69,7 +69,10 @@ requester.connect("tcp://#{clojure_hostname}:1043")
 requester.on 'message', (data) ->
   response = JSON.parse(data)
   unless response is "ok"
-    lobby.to(response.gameid).emit("netrunner", {type: response.action, state: response.diff})
+    if response.diff
+      lobby.to(response.gameid).emit("netrunner", {type: response.action, diff: response.diff})
+    else
+      lobby.to(response.gameid).emit("netrunner", {type: response.action, state: response.state})
 
 # Socket.io
 io.set("heartbeat timeout", 30000)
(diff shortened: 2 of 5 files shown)


mtgred:
Awesome! :+1:

PR #797, opened 2015-10-08
fully implement Demolition Run

JoelCFC25:
This is a weird one, but needed--it seems that a lot more people have been asking about it recently. The card is given abilities which are of course discarded upon play because of `desactivate`--but then we pull it back out to the play area and do a `card-init` to restore the on-click ability to trash at no cost. The Runner gets a prompt so they'll know what to do with it, and `gameboard.cljs` is modified to allow it to be clickable when in the play area. As each accessed card comes up, the Runner can access normally or trash by clicking Demolition Run. When the run ends, the card goes into the Heap!

### DIFF
--- a/src/clj/game/cards-events.clj
+++ b/src/clj/game/cards-events.clj
@@ -95,7 +95,18 @@
                                      :effect (effect (move target :hand))} card nil)))}
 
    "Demolition Run"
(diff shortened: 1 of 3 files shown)


JoelCFC25:
Trying to write a test for this so it'll validate in the testing framework...

mtgred:
:+1:

PR #799, opened 2015-10-09
Apply Kate discount to The Supplier install (#578). Tests for The Supplier, Franchise City, Security Testing.

### LINKED ISSUE #578: Street Peddler and The Supplier do not take Kate's install discount into account
Cards like Street Peddler and The Supplier do not take Kate's discount into account when displaying cards you are allowed to install from them. 
Example: If you have a Plascrete hosted on The Supplier and have no credits at the beginning of your turn you will not be prompted to install it even though it should cost zero to install with Kate.

This is due to the card filtering using ```(:cost %)``` to figure out what cards to allow the user to attempt to install which does not take any discounts into account.

nealterrell:
Fixes #578. Tests to validate this fix as well as #655 / #767 .


mtgred:
:+1:

PR #804, opened 2015-10-11
Feature css corp

domtancredi:
Minor change to take cover overlay issues on the corp side (i.e. using Knight)

Minor change to remove "am/pm" from localized content


mtgred:
:+1:

Issue #803, opened 2015-10-10
Snoop encounter


GrantZilla1979:
There is no reveal of the Runner's hand when they encounter Snoop currently.
In addition, the tracer subroutine is not listed as an option upon click - only "add hosted power counter" and the effect of the power counter.

JoelCFC25:
Thanks for the report. I don't think we want the reveal to happen automagically on encounter (look at the problems Quicksand had). But yes, this card needs some work--submitting shortly.

queueseven:
Yes, all the encounter abilities should be manually.

PR #807, opened 2015-10-12
improvements to Snoop and Architect

### LINKED ISSUE #363: You can install Operations into a server
`corp-install` does not check the card type, so it's possible that you can install operations via this function. It should check the card type and prevent installing operations.

I noticed it while working at #358; I'll try to fix both issues in one go, but it may take some time....

JoelCFC25:
Addresses #803 by adding the trace to Snoop and the on-encounter reveal (not automatic). 

To fix #363/#802, Architect removes operations from the "look at the top 5 of R&D" subroutine prompt choices, instead moving them into the text (the Corp will still want to know what was seen). Operations are also filtered out of the choices for the install from Archives/HQ abilities.

### DIFF
--- a/src/clj/game/cards-ice.clj
+++ b/src/clj/game/cards-ice.clj
@@ -596,8 +596,11 @@
                                 (handle-access state side [c])))}]}
 
    "Snoop"
-   {:abilities [{:msg "place 1 power counter on Snoop" :effect (effect (add-prop card :counter 1))}
-                {:counter-cost 1 :label "Look at all cards in Grip and trash 1 card"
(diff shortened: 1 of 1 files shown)


nealterrell:
Not a fan of the Architect change, since it does not say where those operations were relative to the others. A better fix would be to change the effect with an if that only applies when the target is not an Operation.

JoelCFC25:
I'll revisit this later today.

mtgred:
Is it ready for merging?

JoelCFC25:
Yep, I reworked it the way Neal said--it still shows the top 5 of R&D like it did before, but nothing will happen if the choice you click is an operation.

mtgred:
:+1:

PR #808, opened 2015-10-12
fully implement Immolation Script

JoelCFC25:
The system currently glosses over the access of ice in Archives, so we don't really need to distinguish between accessing or not accessing a piece there. Therefore, this event just listens for the conclusion of the successful run on Archives before listing all the ice and then using the title of the chosen one as the basis for selecting a rezzed copy to trash.


mtgred:
:+1:

PR #811, opened 2015-10-13
Fix cards that broke due to the restructuring of remote servers

JoelCFC25:
Several cards are no longer functioning or only semi-functioning due to the new way that remote servers are stored in the game state. 

This patch also contains a couple small items not related to that: a simplification to Social Engineering to use a newer, more concise core function and a fix for the wonky message of Cerebral Cast. 

Added fixes for Project Junebug, Trope, and Order of Sol (see #812).


mtgred:
:+1:

queueseven:
@JoelCFC25 Sorry that my change resulted in so much work for you :-( That's what you get if you code in a hurry. Thanks for fixing the broken cards, which should have been my task!

PR #813, opened 2015-10-14
proposed fix for Surfer

### LINKED ISSUE #789: Can't stop surfing
Attempted to use surfer (on a 4 deep remote), had the following issues:

* After swapping two pieces of ice (one ressed, one not), it didn't deduct cash
* the 'done' button did nothing - couldn't stop surfing!
* I was (for a while) unable to surf deeper - all I could do was swap the two pieces back and forth. After a while, I figured that click ice, click done click ice allowed me to surf one deeper (problems above remained), but when I tried to go deeper, it swapped me back with the outside (an illegal swap)

Further details of game state:
the ice was unrezzed as I approached it, it was hive. Opponent was blue sun, I was kit, had yog+corroder installed and plenty of cash.

JoelCFC25:
Although I can't actually vouch that this will fix #789 because I still can't reproduce it, I think this is worth a shot. As was pointed out by @dlrowb in that issue, `update-ice-strength` is being passed the wrong thing. In its place we can try just updating all ice (in case the one being swapped changes its strength, e.g. Curtain Wall). I also triggered `:approach-ice` so that any breaker being used will recalculate the cost needed to match strength.

### DIFF
--- a/src/clj/game/cards-programs.clj
+++ b/src/clj/game/cards-programs.clj
@@ -503,8 +503,9 @@
                                                       #(assoc % oldndx target))
                                                   (swap! state update-in [:run]
                                                       #(assoc % :position (inc tgtndx)))
-                                                  (update-ice-strength state side (cons :corp (:zone cice)))
-                                                  (update-run-ice state side)))}
+                                                  (update-all-ice state side)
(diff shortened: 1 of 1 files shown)


mtgred:
:+1:

Issue #820, opened 2015-10-15
Laramy Fisk ability doesn't work after a run on a remote server


Merlin8000:
If Laramy Fisk runs on a remote server before the first time running on central server, his ability doesn't activate.  

As a work around, the runner can jack out so the corp can take a click to draw and then run back in, with the corp allowing the runner back through to the same point at the successful run.

JoelCFC25:
Thanks for the report. I have this fixed and will probably submit it tonight. We keep uncovering more cards that were affected by the remote server restructuring (Gagarin Deep Space and Bishop will also be getting fixed).

JoelCFC25:
Should be deployed now.

Issue #801, opened 2015-10-10
You can play scavenge with no programs installed.


Jozrael:

JoelCFC25:
I have a fix for this and will submit it soon (collecting some other fixes first).

PR #821, opened 2015-10-15
More cards fixed to work with remote servers

JoelCFC25:
Some people in chat uncovered a couple more cards that don't work with the new remote server system. Gagarin Deep Space, Laramy Fisk (fixing #820), and Bishop use `is-central?` and `is-remote?` now. 

This also adds a requirement of at least 1 program installed to play Scavenge, which addresses #801.


mtgred:
:+1:

PR #827, opened 2015-10-19
fully implement Clot

### LINKED ISSUE #777: Advancement tokens should be removed if the runner steals an agenda

JoelCFC25:
When Clot is installed, it checks `[:turn-events]` for all Corp installs of agendas this turn and writes that list to the Corp's register, as well as continuing to listen for other agendas installed later. When the Corp attempts to score, it will fail if the `:cid` of the agenda is found in the list. 

Also fixes #777 by dissociating advancements when an agenda is stolen.

### DIFF
--- a/src/clj/game/cards-programs.clj
+++ b/src/clj/game/cards-programs.clj
@@ -62,7 +62,12 @@
    {:recurring 1}
 
    "Clot"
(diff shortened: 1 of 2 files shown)


mtgred:
Awesome :+1:

PR #825, opened 2015-10-19
adding run arrow to knifed and inside job; 

domtancredi:
Resolution for this Feature Request:

https://github.com/mtgred/netrunner/issues/729

PR #828, opened 2015-10-22
Update remaining run events to show card title above run arrow

JoelCFC25:
Finishes the remaining run events that don't yet work with the new feature in #825.


mtgred:
:+1:

PR #829, opened 2015-10-22
Identify the server of exposed cards

JoelCFC25:
Until there is a CSS solution for adding a colored halo to exposed cards, this identifies the server of the exposed card to reduce ambiguity for the Corp and spectators.


mtgred:
:+1:

PR #830, opened 2015-10-22
Stop activating cards installed facedown

### LINKED ISSUE #818: Facedown cards are getting activated when installed
This is presenting a big problem for Apex. It looks like `card-init` is being called on cards installed facedown, which should not be happening. As a result, you get things like: 

* facedown PrePaid Voicepad shows a recurring credit
* facedown Ghost Runner shows 3 credits
* facedown Makers Eye will yield 3 accesses the next time you run R&D
* if Technical Writer is installed, it will gain a credit every time a hardware or program is installed facedown

JoelCFC25:
Fixes #818 so people can actually play Apex decks. Facedown status has to be the factor determining if `card-init` is called at all, not just if it should also resolve.

### DIFF
--- a/src/clj/game/core.clj
+++ b/src/clj/game/core.clj
@@ -1427,9 +1427,11 @@
                (let [c (if host-card
                          (host state side host-card card)
                          (move state side card [:rig (if facedown :facedown (to-keyword type))]))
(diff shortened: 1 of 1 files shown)


mtgred:
:+1:

Issue #817, opened 2015-10-14
24/7 News Cycle doesn't seem to fire anymore


eedok:
The trashing an agenda part works but then the effect doesn't seem to happen

JoelCFC25:
It's missing a `:choices` entirely. Did this card ever work?

eedok:
maybe not, marked as finished in the document

JoelCFC25:
I'm stumped on this one. I tried adding `:choices` to effectively make it the operation version of scoring a Bifrost Array. Bifrost works just fine but 24/7 doesn't.

JoelCFC25:
I have it now.

PR #831, opened 2015-10-22
Batch of miscellaneous card fixes

JoelCFC25:
##### Bug fixes
Pawn must be hosted on ice in order to start responding to successful runs. 24/7 News Cycle was missing a `:choices` completely and probably never worked--it works now (fixing #817). 

##### Enhancements
* BlacKat ditches `auto-icebreaker` since there are two different ways to pump it, one of which was missing.
* People reported having confusion about how to work the revised Accelerated Beta Test, so some language was added to the prompt telling the player to look at the top of the play area.
* Thanks to the just-released FAQ, the Caissa can now be moved off Scheherazade. Bishop now checks to see if it's being hosted on ice.
* Tweaked the Quantum Predictive Model message to be in line with other cards that use `as-agenda`
* Added a message to the log when the Runner opts to use Global Security Clearance (it was happening silently before, which is confusing to the Corp)
* Rebranding Team is greatly simplified, after I got confirmation that it should apply Advertisement to all assets in R&D, Archives, HQ, and ones already installed. This provides support for Ad Blitz (hopefully coming soon...)

### DIFF
--- a/src/clj/game/cards-agendas.clj
+++ b/src/clj/game/cards-agendas.clj
@@ -2,7 +2,7 @@
 
 (def cards-agendas
   {"Accelerated Beta Test"
-   (let [abthelper (fn abt [n i] {:prompt "Select a piece of ICE to install"
+   (let [abthelper (fn abt [n i] {:prompt "Select a piece of ICE from the top of the play area to install"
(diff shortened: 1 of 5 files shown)


mtgred:
:+1:

Issue #785, opened 2015-10-05
Jinteki doesn't notice when corp is decked


DanSuperXeo:
I just decked out the corp with a virus mill deck, and Jinteki did not indicate that the game was over.

Merlin8000:
Congratulations for decking a corp.  You monster.

Issue #691, opened 2015-09-08
Comet and run events


unitled:
Comet asks if you want to play a free event as soon as you've used the previous event. It should wait until the event is resolved before it does this. This means a run event lets you play another run event right before the first one starts. It's not a big issue as you can manually take the click and play the event once you've finished your run!

JoelCFC25:
Duplicate of #322.

JoelCFC25:
Fixed by #854.

Issue #519, opened 2015-08-09
Multiple Gang Sign doesn't seems to work properly


Zulyen:
My opponent scores an agenda. My 2 Gang Sign triggers.
I have also HQ Interface.
First access : Howard Jackson + Howard Jackson. i trash them both.
Second access : Howard Jackson + EMP.
This case is impossible because i trashed 2 jacksons before and one is already used and out of the game.

later in the same game, a similar scenario is reproduced.
My opponent scores an agenda. My 2 Gang Sign triggers.
First access : Psychic Field + Neural EMP. I trash the Psychic.
Second access : Psychic Field + Scorched Earth. I trash the Psychic but nothing appears in Archives.

It's like the second access is done with the same cards in hand than the first.

JoelCFC25:
Known issue, and it's noted in the site guide on the wiki as a limitation. Ultimately it boils down to #189--multiple simultaneous effects can't be ordered at this time.

Issue #322, opened 2015-07-03
Comet triggers before the triggering event finishes resolving


anathomical:
Currently Comet asks if you want to play a second event as soon as you play the first event.  This causes problems if the first event played requires the player to further interact (via a run, or via other interaction prompts).

Comet sets up a parallel set of decisions to interact with, and those interactions alternate with the first set.  Here's an example:

1. Comet is installed.
2. Runner plays Test Run.  Is shown an interaction prompt asking if they want to play an additional event.
3. Runner selects "Yes".  Is shown an interaction prompt asking if they want to use Test Run to search their Stack or their Heap.
4. Runner selects "Stack".  Is shown an interaction prompt asking what additional event to play with Comet.
5.  Runner selects "Infiltration".  Is shown an interaction prompt to select which program to install from the stack with Test Run.
6. Runner selects "Femme Fatale".  Is shown an interaction prompt to choose to gain 2 credits or expose a card.

Now, in this particular example, this isn't a huge problem, but if the runner wants to play Scavenge at step 5, the Femme Fatale isn't yet installed.  And this gets REALLY messed up if one (or worse, both) of the events triggers a run.  In fact, if they're both run events, the second run event gets triggered in the middle of the first run event, disrupting the run.

Basically, it would be cleaner if we could tell when an event is "finished resolving" so that Comet could trigger then rather than on-play.  That seems complicated, so it may make sense to de-automate Comet so that rather than prompting to play a second event when you play an event, the user can click on Comet to trigger the second event when appropriate.

JoelCFC25:
This one is closely related to #183, but broader.

gandrasch:
Problem still occours (Comet with Inject as example, I have to choose another event before gaining the 4 cards from inject).

PR #834, opened 2015-10-23
manual trigger for Leela, Gang Sign, Comet, Team Sponsorship; Corp decked message; card fixes

### LINKED ISSUE #806: Gang Sign causes game-ending bug
IF a player with multiple Gang Signs in place has them trigger, all instances of gang sign check the opponents hand simultaneously.  If the first instance to resolve finds an agenda, and it is stolen, if other instances of gang sign find the same agenda, you are prompted to steal the agenda which you have already stolen, and you cannot advance the board state.

Multiple Gang signs need to fire sequentially, not simultaneously, as this allows players to use trash effects to take cards from hand or steal agendas from hand before the second instance of gang sign checks the now reduced hand.

JoelCFC25:
People have had it up to their eyeballs with multiple Gang Signs not working correctly, and I don't blame them. This changes each Gang Sign to a manually triggered effect using the technique from Laramy Fisk (print a message in the log informing the Runner what to do). Until there is a global fix for multiple simultaneous effects, this fixes #806 and #519. 

Leela Patel is also switched to work this way so a Runner can use Gang Sign(s) first before doing the bounce to HQ. Leela's click ability doesn't work right when you steal an agenda from a remote that also has 1 or more upgrades--you'll have to proceed with accessing all cards before clicking her. But the current implementation doesn't work right in this regard either, so we aren't going backwards. The Runner can ask the Corp to drag a remaining unrezzed card in the server back to HQ manually if that's what they want to bounce. 

Comet is changed to a manual trigger to allow a Runner to play a Run event and wait until it resolves before playing another event by clicking Comet. Addresses #322 and #691.  

Team Sponsorship with a manual trigger allows it to work correctly with a scored Accelerated Beta Test--instead of the prompts of ABT and TS getting interleaved, you can fully complete the ABT installs before firing Team Sponsorships. 

There's a new log message for when the Corp is decked, addressing #785. I fixed Diversified Portfolio, which was counting servers even if they only had ice but no server contents, and fixed Hemorrhage which wasn't charging the 2 counters correctly.

### DIFF
--- a/src/clj/game/cards-assets.clj
+++ b/src/clj/game/cards-assets.clj
@@ -512,6 +512,30 @@
    {:events {:runner-spent-click {:req (req (not (= (:server run) (:zone card)))) :once :per-turn
                                   :msg "gain 2 [Credits]" :effect (effect (gain :corp :credit 2))}}}
 
+   "Team Sponsorship"
+   {:events {:agenda-scored {:effect (req (system-msg state :corp
+                                            (str "can install a card at no cost from Archives or HQ by clicking on Team Sponsorship"))
+                                          (update! state side (assoc card :ts-install true)))}}
+    :abilities [{:req (req (:ts-install card))
+                 :label "Install a card from Archives or HQ"
+                 :prompt "Install a card from Archives or HQ?" :choices ["Archives" "HQ"]
+                 :msg (msg "install a card from " target " at no cost")
+                 :effect (req (let [serv target]
+                                (resolve-ability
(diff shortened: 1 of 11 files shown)


JoelCFC25:
Don't merge this yet, I'm working on Team Sponsorship too--just have to modify the tests for it.

JoelCFC25:
Closing this so I can split out the simpler card fixes into a separate PR with others I've been collecting.

PR #835, opened 2015-10-25
Implement Old Hollywood Grid; fixes for Red Herrings, Strongbox.

nealterrell:
Fix Strongbox and Red Herrings so their effect persists for the run in which they are trashed. This was broken when we moved to manually selecting access order.

Implement Old Hollywood Grid to prevent stealing agendas that are not in the runner's scored area. 

Tests for all three.


mtgred:
:+1:

PR #839, opened 2015-10-28
implement Turntable, undo effects of Corp agendas that are forfeited or Turntabled

JoelCFC25:
After what seems like months of pecking at Turntable off and on, I think this does it. This is a click-to-trigger effect after a steal, like what has been done to Gang Sign, Leela, Team Sponsorship, etc. 

The scored agenda chosen from the Corp's area will have any counters, effects, and abilities dissociated, and then will get `desactivate` as well--this is the part that will drop the Corp back down to 3 clicks per turn if the Runner Turntables away a Mandatory Upgrades, or restores the Runner's max hand size to 5 if Turntabling away a Self-Destruct Chips. 

Similarly, the act of forfeiting either of these two agendas doesn't currently reverse their effects, so `forfeit` and `desactivate` got small tweaks to support that in addition to the use of Turntable.

### DIFF
--- a/src/clj/game/cards-agendas.clj
+++ b/src/clj/game/cards-agendas.clj
@@ -254,7 +254,8 @@
                        :effect (effect (corp-install target nil {:install-state :rezzed}))} card targets))}
 
    "Mandatory Upgrades"
-   {:effect (effect (gain :click 1 :click-per-turn 1))}
+   {:effect (effect (gain :click 1 :click-per-turn 1))
+    :leave-play (effect (lose :click-per-turn 1))}
 
    "Market Research"
    {:req (req tagged) :effect (effect (set-prop card :counter 1 :agendapoints 3))}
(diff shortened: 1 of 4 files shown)


mtgred:
:+1:

src/clj/game/cards-agendas.clj
```diff
@@ -254,7 +254,8 @@
                        :effect (effect (corp-install target nil {:install-state :rezzed}))} card targets))}
 
    "Mandatory Upgrades"
-   {:effect (effect (gain :click 1 :click-per-turn 1))}
+   {:effect (effect (gain :click 1 :click-per-turn 1))
+    :leave-play (effect (lose :click-per-turn 1))}
```
adamnfish: Should the leave play not also lose a click? e.g. if the corp has to forfeit an agenda halfway through the turn, they should lose the 5th click. See also, [Rachel Beckman](https://github.com/mtgred/netrunner/blob/master/src/clj/game/cards-resources.clj#L441)

JoelCFC25: Spot on--thanks for the tip. I was also alerted that the Runner forfeiting a stolen Mandatory Upgrades will drop to 3 clicks per turn, so I've fixed that as well and changed `desactivate` to only do the `:leave-play` effect if the forfeited agenda came from the Corp's score area.

Issue #712, opened 2015-09-13
Recurring credits on hosted cards do not refresh


josemgarcia:
When programs that have recurring credits (e.g. Paricia) are installed hosted in another program (e.g. Scherezade or Leprechaun), the recurring credits are not refreshed when the turn begins.

I think it may also happen in other cases, so it may be actually worth to take into account for solving #362

JoelCFC25:
I've tested this both on the live server and in my dev environment and can't reproduce it so far. I tried a Scrubber hosted on Off-Campus Apartment and Paricia hosted on a Leprechaun--in both cases if credits had been taken on a previous turn from them, they refreshed to the correct amount at the start of the next Runner turn.

josemgarcia:
I was having the issue in two consecutive games with Scherezade, and I
believe it was only when hosting the program after installing it, that is
installing normally and then hosting on Scherezade.

José M. García

Balkmeister:
I also have seen this issue with Sahasrara installed on scherezade. The first one does not refresh but when i installed a second that one refreshed.

queueseven:
The issue is the following:

The `host` function removes all events from a card, does the hosting mumbo jumbo, and calls:

    (when-let [events (:events (card-def target))]
        (when installed
             (register-events state side events c)))

to re-register the events of the card. The problem is that the recurring credit reloading is also triggered with an event, but the event isn't in the card definition (from `card-def`), but is created on-the-fly in `card-init`.

When installing a card, after the `host` function is called in `runner-install`, `card-init` is called, so recurring credits work if a card is directly installed onto a host.

I think instead of re-registering the events, `host` should maybe call `card-init` with `resolve` = `false`.

justinliew:
I might be able to take this on, having been working with recurring credits.

JoelCFC25:
Fixed at #840.

PR #840, opened 2015-10-29
Fix refresh of recurring credits on cards hosted after installation

JoelCFC25:
Fix for #712. As @queueseven noted in that issue, if the card being hosted is already installed, we have to `card-init` without resolving instead of registering events. 

Tested with Scheherazade and Off-Campus Apartment as hosts for installed Pheromones and Compromised Employee to ensure that credits recur *and* they're listening for their event triggers.


mtgred:
:+1:

PR #843, opened 2015-10-31
Fix for unreported Red Herrings issue

nealterrell:
Well, it was reported by @JoelCFC25  :P.

Red Herrings, Strongbox, and Old Hollywood Grid will fire during a run on HQ if the card was discarded from the corp's hand without being played. Fun times. Test to verify the issue, and a fix for the issue.


mtgred:
:+1:

PR #845, opened 2015-11-02
Stop spectators from viewing facedown cards in Archives, fix for seen cards returned to R&D/HQ

### LINKED ISSUE #781: Facedown cards in Archives are visible to spectators
When spectating a game, clicking on Archives shows the cards in a similar fashion as it does to the Corp when they view their Archives. (Slighty greyed out)

JoelCFC25:
Fixes #781 so spectators can no longer view facedown cards in Archives as if they were the Corp. 

The `core.clj` fix is for the current flaw in using Jackson Howard or Archived Memories to return faceup cards in Archives to R&D or HQ. Right now faceup cards are retaining `:seen true` when moved to these destinations, which results in them being faceup again if later milled (e.g., by Noise or DLR), or named in the log if manually dragged to Archives--neither of which should happen.

### DIFF
--- a/src/clj/game/core.clj
+++ b/src/clj/game/core.clj
@@ -248,7 +248,9 @@
                             (not (:facedown c))))
                  (desactivate state side c) c)
              c (if (= dest [:rig :facedown])(assoc c :facedown true) (dissoc c :facedown))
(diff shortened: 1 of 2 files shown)


mtgred:
:+1:

PR #848, opened 2015-11-04
implement Capstone

JoelCFC25:
No one ever actually uses this card I'm sure...but there it was on the list of cards that weren't done yet. Not anymore!


mtgred:
:+1:

Issue #491, opened 2015-08-03
disappearing Parasite when installed from Street Peddler at 1 credit


JoelCFC25:
Reporting on behalf of someone in chat...I did reproduce it to confirm. 

If the Runner tries installing a Parasite off Street Peddler when sitting at 1 credit, the Parasite disappears altogether (though the log does report the install). Plenty of MU were available and there was a piece of rezzed ice ready to host it. Street Peddler and the other 2 cards hosted on it went into the Heap as expected.

ZomBagus:
I just had a similar experience with a mid run install of Faust disappearing from the game, with enough MU available. Again the log reported the install. Peddler and other 2 cards went to the heap. Card was not in heap or stack or grip. Corp could not see it either.

nealterrell:
Working on a fix.

PR #850, opened 2015-11-05
Fix for installing Parasite off Street Peddler with 1cr

nealterrell:
The way install-on-host works was incompatible with install cost bonuses, as the initial call to `runner-install` would trigger another call to `runner-install` with additional arguments to select the host. The first call would remove Peddler's install cost bonus before Parasite was successfully installed. Fix and test for #491.


mtgred:
:+1:

PR #854, opened 2015-11-06
change Comet to a manually triggered ability

### LINKED ISSUE #322: Comet triggers before the triggering event finishes resolving

JoelCFC25:
The first in a series of cards that will adopt the Laramy Fisk technique to become manually triggered, providing an interim solution for the widespread problem with the resolution of multiple simultaneous effects. Specifically with Comet, playing the first event of a turn unlocks the usage of the card ability. The Runner is prompted in the log how to activate it. This works with run events, allowing you to fully resolve them before triggering Comet. 

Fixes #322 and #691.

### DIFF
--- a/src/clj/game/cards-hardware.clj
+++ b/src/clj/game/cards-hardware.clj
@@ -99,14 +99,16 @@
 
    "Comet"
    {:effect (effect (gain :memory 1)) :leave-play (effect (lose :memory 1))
-    :events {:play-event
-             {:optional {:prompt "Play another event?"
-                         :req (req (and (first-event state side :play-event)
-                                        (not (empty? (filter #(has? % :type "Event") (:hand runner))))))
-                         :yes-ability {:prompt "Choose an Event to play"
-                                       :choices (req (filter #(has? % :type "Event") (:hand runner)))
-                                       :msg (msg "play " (:title target))
-                                       :effect (effect (play-instant target))}}}}}
+    :events {:play-event {:req (req (first-event state side :play-event))
+                          :effect (req (system-msg state :runner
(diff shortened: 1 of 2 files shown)


mtgred:
:+1:

PR #855, opened 2015-11-06
change Gang Sign to a manually triggered ability

### LINKED ISSUE #519: Multiple Gang Sign doesn't seems to work properly

JoelCFC25:
Gang Sign adopts the Laramy Fisk technique to become manually triggered. The Runner is prompted in the log to activate each Gang Sign. This works with multiple copies of Gang Sign and multiple copies of HQ Interface to get the correct outcomes--e.g. 2x Gang Sign and 1x HQI will yield: access 2 cards from HQ, return them, access 2 cards again (possibly seeing the same ones). Currently you get 4 accesses at once, which is wrong.

Leela Patel as a manually triggered effect is coming shortly, so Gang Signs will be able to be used before or after Leela's bounce to HQ ability. 

Fixes #519.

### DIFF
--- a/src/clj/game/cards-resources.clj
+++ b/src/clj/game/cards-resources.clj
@@ -166,9 +166,13 @@
                              (:agendapoints c) " agenda point" (when (> (:agendapoints c) 1) "s"))))}]}
 
    "Gang Sign"
-   {:events {:agenda-scored {:msg (msg "access " (get-in @state [:runner :hq-access]) " card from HQ")
-                             :effect (req (let [c (take (get-in @state [:runner :hq-access]) (shuffle (:hand corp)))]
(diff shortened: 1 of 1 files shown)


mtgred:
:+1:

PR #856, opened 2015-11-06
change Leela Patel's ability to be manually triggered

JoelCFC25:
Making Gang Sign use the Laramy Fisk technique to become manually triggered pretty much requires that Leela Patel use it as well. The Runner is prompted in the log on either of the triggering events to activate Leela by clicking her. 

Doing it this way ensures that the Runner can order Leela's bounce to HQ before or after Gang Sign(s) as desired.


mtgred:
:+1:

PR #858, opened 2015-11-08
Chronos

justinliew:
Chronos is manual. When the first Net damage is incurred, it will inform the corp they can use the ID to either pick a card and then it does N-1 net damage. There are 2 ID abilities: what was just described, or one where they can just do all the net damage without revealing, as it's a "may" ability.

### DIFF
--- a/src/clj/game/cards-identities.clj
+++ b/src/clj/game/cards-identities.clj
@@ -56,23 +56,24 @@
    "Chaos Theory: Wünderkind"
    {:effect (effect (gain :memory 1))}
 
-   "Chronos Protocol: Selective Mind-mapping"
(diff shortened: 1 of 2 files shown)


justinliew:
Joel gave me some pointers so I'm going to refactor the ability to be optional. Plus some slight messaging changes to make it more consistent.

justinliew:
Aite, this seems to work better. Same functionality but no ID clicking. Asks you on first net damage and then functionality is the same.

justinliew:
I've added :non-distinct so duplicates will show multiple times in order to be less confusing.

mtgred:
:+1:

PR #860, opened 2015-11-10
implement Dr. Lovegood

JoelCFC25:
The Runner needs to click Dr. Lovegood before hitting Start Turn, then the events and abilities of the targeted card will be unhooked for the turn. When `:runner-turn-ends` fires, it does a `card-init` to restore everything.


mtgred:
:+1:

Issue #180, opened 2015-05-03
Hayley install bug


eah13:
In the hayley install flow, if you install from grip by clicking a card instead of using the dialog it crashes the game withthese errors:

![screenshot 2015-05-02 at 10 11 32 pm](https://cloud.githubusercontent.com/assets/1702745/7443452/bdedd66a-f118-11e4-8a87-2d8f213d3fef.png)

JoelCFC25:
Adding this label not to shame anyone, but just that this and #167 remind me of the Smith and Dale "Dr. Kronkheit" sketch of the 1920s: 

SMITH: Doctor, it hurts when I do *this*.
DALE: Don't *do* that.

queueseven:
TBH, that really sucks and is a serious UI issue and should be fixed some day. Related: #587 and #290

aiannacc:
I'm guessing this is moot by now.

JoelCFC25:
Hayley uses card targeting now, so this one can't happen anymore.

Issue #167, opened 2015-04-24
Modded BUG


benticurus:
When playing Modded (the Shaper operation) the game freezes if the runner doesn't use the drop menu to choose which card to install through Modded; e.g., if the runner plays Modded but then clicks directly on the card that they want to install from hand, then the drop menu doesn't go away and the game freezes (or doesnt allow the runner to do anything else).

I think sometimes it gets fixed if the runner moves the intalled card back into hand, and then uses the drop menu from Modded; but this doesnt always work, and the game cannot be continued.

benticurus:
I recognize this is more an error on part of the player than the system; but if the player makes this simple mistake, the whole game may be lost.

JoelCFC25:
Modded uses the card targeting system now as of #863, so this interface glitch can't occur anymore.

PR #863, opened 2015-11-11
Change a variety of cards to use the card targeting system (allowing easier cancellation)

JoelCFC25:
Getting rid of the pop-up prompt with card titles and switching to card targeting will resolve the potential for UI errors as seen in #167 and #180. 

Also changed a number of cards to the card targeting system to allow more flexibility with cancelling the action once initiated--anything using the old card selection prompt doesn't allow a cancel option and causes varying degrees of headache to undo after the fact. Operations like Modded, Archived Memories, Career Fair, and Interns will still charge clicks and put the card in the discard, but the user won't face having to reveal anything if they change their mind after playing the event--they can just take clicks back and retrieve the card.

### DIFF
--- a/src/clj/game/cards-events.clj
+++ b/src/clj/game/cards-events.clj
@@ -385,8 +385,9 @@
      {:effect (effect (resolve-ability (mhelper 1) card nil))})
 
    "Modded"
-   {:prompt "Choose a card to install"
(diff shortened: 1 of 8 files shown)


mtgred:
:+1:

PR #864, opened 2015-11-11
implement Ad Blitz

JoelCFC25:
Changes the `:install-state` option of a `corp-install` from `:rezzed` to `:rezzed-no-cost` so that Ad Blitz (and any future cards that install and immediately rez normally) can use `:rezzed`. Accelerated Beta Test and License Acquisition are modified accordingly to use `:rezzed-no-cost`. 

Ad Blitz brings up Archives and lets the user recursively do an install and rez of the specified number of advertisements. Now all the Spark players can go crazy with Rebranding Team decks. ;)


mtgred:
:+1:

PR #865, opened 2015-11-11
stop Harbinger from installing if trashed from hand by the Runner

JoelCFC25:
Reported by a user--Harbinger was installing facedown if trashed by the Runner from hand when discarding down to max hand size.


mtgred:
:+1:

PR #869, opened 2015-11-12
another fix for Turntable

### LINKED ISSUE #867: Turntable & Agenda Counters
Here's what happened: 

The corp scored a House of Knives and used it once, so 2 counters remaining.  Then I stole a second HoK from R&D and swapped it with the corp's.  I got the HoK with 2 counters in my score area, but the new HoK in the corp's area has 3 counters when it should have none.

It seems when Turntable swaps agendas, it treats the agenda it pushes into the corp's score area as if it were just scored (or at least grants agenda counters as if just scored)?

JoelCFC25:
Fixes #867. 

It doesn't work to call `card-init` on agendas in the Corp's area when swapping one back to them, because it will make "when scored" effects fire--so it would replenish any spent agenda counters from AstroScript Pilot Program, Nisei Mk II, House of Knives, etc. 

But we still need to hook up events and abilities so things like a Gila Hands are usable by the Corp when getting swapped back--I stole the relevant bits of the `card-init` function to accomplish this.

### DIFF
--- a/src/clj/game/cards-hardware.clj
+++ b/src/clj/game/cards-hardware.clj
@@ -453,7 +453,11 @@
                                                      (gain-agenda-point state :runner (- swpts stpts))
                                                      (gain-agenda-point state :corp (- stpts swpts))
                                                      (doseq [c (get-in @state [:corp :scored])]
-                                                       (card-init state :corp c false))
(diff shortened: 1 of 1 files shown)


mtgred:
:+1:

PR #870, opened 2015-11-12
Global food initiative

justinliew:
Added a get-agenda-points function.
Changed all refs to :agendapoints to use this function.
GFI overrides :agendappoints-runner function to return 2.

Tested GFI, Turntable, Humans First.


mtgred:
:+1:

PR #871, opened 2015-11-12
Runner cards turned facedown keep :installed, do agenda scored effects before trashing a Runner current

### LINKED ISSUE #866: Chopbot 3000 won't chop facedown cards
Play Apocalypse, or use Apex's ability to install face down cards. 
At the beginning of your turn try to use Chopbot 3000's ability to trash the facedown installed card and the ability will not trigger. 
Still works fine for other installed cards as long as they are not facedown.

JoelCFC25:
Fixes #866. 

When the Runner plays Apocalypse, cards get moved to `[:rig :facedown]` and have `desactivate` called on them. This was wrongly dropping `:installed true`, leaving facedown cards later unable to be sacrificed to Chop Bot, Aesop's, Heartbeat, and other cards that were targeting with `(all-installed :runner)`. 

I was also alerted that Jinteki:PE scoring an agenda was dealing 1 net damage even with Employee Strike active. A Runner current in play shouldn't get trashed until after the `:agenda-scored` trigger.

### DIFF
--- a/src/clj/game/core.clj
+++ b/src/clj/game/core.clj
@@ -118,7 +118,7 @@
   ([state side card] (desactivate state side card nil))
   ([state side card keep-counter]
    (let [c (dissoc card :current-strength :abilities :rezzed :special :facedown)
-         c (if (= (:side c) "Runner") (dissoc c :installed :counter :rec-counter :pump) c)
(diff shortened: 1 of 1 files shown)


mtgred:
:+1:

PR #872, opened 2015-11-13
Using Hades Shard should not access upgrades in root of Archives

JoelCFC25:
Someone alerted me today that popping Hades Shard doesn't let you access upgrades in the root of Archives. I've done this one wrong just about forever, apparently:   [Lukas ruling on Twitter](https://twitter.com/RukasuFox/status/541656010075955201)


mtgred:
:+1:

PR #873, opened 2015-11-13
implement Archives Interface

JoelCFC25:
This uses the increasingly valuable "Laramy Fisk method" of activating the card ability when the Corp clicks "No Action" for the last time right before the run becomes successful. At that time, stuff a key into Archives Interface that unlocks its ability and prompt the Runner in the log that it can be used. 

Clicking the card turns everything in Archives face-up and gives a card selection prompt from which the Runner removes a card from the game (1 card per used Archives Interface), then they can hit Successful Run when finished to access the remainder of Archives normally.


mtgred:
:+1:

PR #874, opened 2015-11-14
fix Old Hollywood Grid, Red Herrings, and Strongbox

### LINKED ISSUE #868: Red Herrings Doesn't Protect TFP/R&D
I managed to steal The Future Perfect with a rezzed Red Herrings installed in the same server, without paying the Herrings fee. The psi game proc'ed on access and my opponent thought the Herrings would protect him, we both bid 0 and so I won it.

Other possible factors: 
TFP was accessed during an R&D dig of multiple cards using Medium.
I also had Turntable installed.

JoelCFC25:
Fixes #868. 

Red Herrings and Strongbox weren't charging the additional costs for stealing agendas when installed in the root of central servers--the `:req` only worked for installs in remotes.

None of Old Hollywood Grid, Red Herrings, or Strongbox were having their prevent/additional cost effects persist if the Runner accessed them first in the root of a central and trashed them before hitting the agenda.

### DIFF
--- a/src/clj/game/cards-upgrades.clj
+++ b/src/clj/game/cards-upgrades.clj
@@ -192,8 +192,9 @@
          un {:effect (req (swap! state update-in [:runner :register] dissoc :cannot-steal))}]
      {:trash-effect
       {:req (req (= :servers (first (:previous-zone card))))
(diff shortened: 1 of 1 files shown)


mtgred:
:+1:

PR #878, opened 2015-11-15
Restore :installed true for Runner cards turned facedown

### LINKED ISSUE #877: Parasite Gaining Counters
Parasite seems to be gaining multiple virus counters at start of the runner's turn.

What happened: 
My runner opponent installed Parasite onto a Swordsman. It was installed from the runner's grip, cost paid from their credit pool. No Grimoire. At the start of their next turn, the Swordsman was insta-trashed from strength 2.

JoelCFC25:
Fixes #877. 

Instead of handling the problem with cutting `:installed true` from cards turned facedown in the `desactivate` function (problematic with recurred Parasites), we just re-associate it in the `move` function if the destination is `[:rig :facedown]`.

### DIFF
--- a/src/clj/game/core.clj
+++ b/src/clj/game/core.clj
@@ -118,7 +118,7 @@
   ([state side card] (desactivate state side card nil))
   ([state side card keep-counter]
    (let [c (dissoc card :current-strength :abilities :rezzed :special :facedown)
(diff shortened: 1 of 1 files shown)


mtgred:
:+1:

Issue #876, opened 2015-11-15
Chronos Protocol and Hokusai Grid


Skeskrske:
I was playing Chronos Protocol, and when the runner hit a rezzed Hokusai Grid my ID's "reveal and trash" prompt appeared to them, not me. (After talking it through with my opponent, he described his hand and I chose which card to throw out - then the report appeared in the chat to show he'd used the Chronos ID dialog to trash it.)

Other cards that cause net damage during the runner's turn (eg. Fetal AI) do work as intended with Chronos, so I'm not sure if this problem is limited to just Hokusai Grid.

Additional info:
Hokusai Grid was installed in a remote server, no other upgrades or assets in that server.
Runner ID was Chaos Theory, no damage prevention abilities in play.

JoelCFC25:
This one will be easy...the `:prompt` needs `:player :corp` added.

PR #879, opened 2015-11-15
give Chronos Protocol prompt only to the Corp player

JoelCFC25:
Fix for #876. 

The prompt generated from the identity needs to be given explicitly to the Corp player if the ability kicks in on the Runner's turn.


justinliew:
Nice catch, thanks!

mtgred:
:+1:

PR #881, opened 2015-11-15
Allow choosing number of extra accesses with Medium/Nerve Agent

JoelCFC25:
Fix for #32. 

Medium and Nerve Agent will by default do the maximum number of additional accesses if the Runner does nothing, but both prompt in the log that they are clickable prior to hitting "Successful Run" if the Runner wants to choose fewer than all of them.


mtgred:
Awesome :+1:

Issue #875, opened 2015-11-14
Chakana + Cyberdex virus suite


DavePickles:
I advanced an NAPD 4 times whilst my opponent had 2 copies of chakana and a hive mind with 3 or more counters on it,rezzed and purged virus counters with the cyberdex virus suite, but was then unable to score the NAPD. No menu popped up when clicking the agenda to score it

JoelCFC25:
My first hunch is that Chakana is missing a `:purge` event hook that would fire `update-all-advancement-costs`. I'll dig into it.

PR #883, opened 2015-11-15
Chakana needs to update advancement requirements after a purge

JoelCFC25:
Fix for #875. 

Chakana was doing `update-all-advancement-costs` when it got trashed, but not after a purge of viruses. Also added a small improvement to Reversed Accounts so it reports how many Runner credits were lost in the game log.


mtgred:
:+1:

PR #885, opened 2015-11-16
revise Hades Shard fix to flip all cards in Archives face up

JoelCFC25:
Touch-up to the Hades Shard fix in #872..I didn't notice that the new version isn't flipping cards in Archives face up. Fixed!


mtgred:
:+1:

Issue #888, opened 2015-11-17
Chronos Protocol & Brain cage


timeh300:
I installed brain cage on my turn (runner) and the chronos protocol option appeared to look at the runners hand. While i cant be sure as well, i think after clicking "no" an additional point of net damage was dealt.

JoelCFC25:
Yep--Chronos Protocol is triggering on the first kind of _any_ damage, not just net. We're on it.

PR #889, opened 2015-11-17
fix Chronos Protocol to fire on net damage only

JoelCFC25:
Fix for #888. 

Chronos Protocol wasn't checking for net damage only, so this will stop it from firing on meat or brain. I also suppressed the portion of the log message about remaining net damage if there is none left over to do.


mtgred:
:+1:

PR #895, opened 2015-11-19
tweak for refresh of recurring credits on cards hosted after install

JoelCFC25:
Follow up to #840--the `card-init` must occur only if the hosted card has recurring credits, otherwise a Cache that gets Clone Chipped into play with Grimoire installed (getting 4 credits), will drop back down to 3 credits when hosted on Scheherazade.


mtgred:
:+1:

Issue #837, opened 2015-10-27
Server 10 appears to the left of Server 2, not to the right of Server 9.


Kiv:

nealterrell:
Ah, the classic string sort vs numeric sort. Paging @queueseven.

justinliew:
I have a fix for this incoming.

PR #896, opened 2015-11-19
Fix server ordering to be numeric

justinliew:
remote->num is an integer, and we use that to sort in get-remotes and
remote-list


justinliew:
Addresses https://github.com/mtgred/netrunner/issues/837

mtgred:
:+1:

PR #903, opened 2015-11-21
Report Gagarin ability result in the chat log

### LINKED ISSUE #899: Gagarin's ID ability working, but not reporting in chat
Pretty simple, Gagarin's 1credit tax ability is now working correctly (cheers on that, by the way), but it really needs to be reported in the chat log. I've already been witness to or a participant in a dozen arguments because of this (people assume if it didn't say it in chat, it didn't happen).

dersam:
Resolves #899 

Displays different messages depending on whether the runner was able to pay the cost to access.


JoelCFC25:
Nice job! This was really needed.

mtgred:
:+1:

PR #906, opened 2015-11-22
fix conditions for Medium and Nerve Agent activation messages

JoelCFC25:
When both Medium and Nerve Agent are installed and each has 1 or more counters, the log messages for both are triggering on either HQ or R&D runs due to a faulty `:req`. 

Also threw in a small fix to make Corporate Town's resource trash unpreventable.


mtgred:
:+1:

PR #908, opened 2015-11-23
Sort the games list so open games show first, then games in progress.

nealterrell:
Requested a few times in chat. Rather than sorting games list strictly on Creation Time, sort it to show unstarted games first, then started games, both sections sorted by Creation Time internally. No change to the UI itself, just a change to the sort order.

The `sort-by` function takes each game map, negates the `:started` key (we want started=0 to be "bigger" than started=1 for the sort), then takes the `:started` and `:date` keys and places them into a vector. Clojure sorts vectors in a lexicographic ordering, which enforces our "multi-column" sort criteria.

Should help people who feel like their games get buried when the server is very active.


mtgred:
:+1:

PR #909, opened 2015-11-23
partial implementation of Jak Sinclair

JoelCFC25:
Trying to temporarily disable all installed programs proved way too difficult and would prevent required conditional abilities that should still occur (like gaining Datasucker tokens). Now that Medium and Nerve Agent can be set to do 0 additional accesses, the two programs with the potential to create the most annoying undos can be worked around (and still gain counters like they should). 

For most other programs the Runner just needs to avoid clicking them during the run, or adjust at the end if anything was gained or lost that shouldn't have been.


mtgred:
:+1:

PR #910, opened 2015-11-23
Media Blitz Implemented.

justinliew:


justinliew:
In order to implement this, I refactored the auto-pump implementation to support multiple types of dynamic ability. 
We use that in this case to "copy" a card (events, abilities and appropriate effects).

mtgred:
:+1:

Issue #898, opened 2015-11-20
Public Agendas incorrectly "firing" when tokens are placed (not advanced)


thehunter518:
All 3 of the Weyland public agendas have special abilities that trigger when they are advanced, but these abilities should NOT fire when advancement tokens are placed on them through other means that don't specifically say "advance the agenda". Currently I've noticed this working incorrectly with Early Premiere + Oaktown Renovation and Hollywood Renovation, but it may also be similarly broken with Space Camp, Trick of Light, Matrix Analyzer, Shipment from SanSan, and Shipment from Kayuga.

JoelCFC25:
I think this could be fixed by modifying the `add-prop` function to accept an optional key `placed` and then not trigger the `:advance` event if `:placed true` is passed in...

PR #911, opened 2015-11-24
Distinguish "placing advancements" from advancing cards

JoelCFC25:
Fix for #898.

We refactor `add-prop` to accept an optional `:placed true` key, then use that to determine if the `:advance` event trigger should be fired. Then the whole list of cards that place advancement tokens are modified accordingly. 

A few cards got improved messages for the case when the card getting advancements is rezzed, and Oaktown Renovation gets a message to be more clear about how many credits are being gained when advanced conventionally.


mtgred:
:+1:

PR #912, opened 2015-11-24
Pass on Targeting/Prompt systems

nealterrell:
@JoelCFC25 started the work on this upgrade, I'm just following through.

Two goals:

1. Better consistency in whether we show Choice (a list of options as buttons) or Select (click on the target) prompts when resolving card abilities.
2. Allow both Choice and Select prompts to cancel their ability if the user messed up.

The decisions we made:

* If a card lets you tutor through your draw pile, then its targets should be shown as a Choice prompt with entries in sorted order, followed by a Cancel button. The Cancel button does nothing, except fire an optional `:cancel-effect` in the ability definition. (See Project Atlas.) Alphabetical order does not reveal the ordering of the cards in the deck, so no information is made available to the player which was not already available. 
* If a card directly affects another card on the board, in your hand, or in your discard, use a Select prompt to directly click on the desired card. If appropriate, use `:show-discard true` to automatically open the discard UI. Remove all old-style prompts that first ask "Install from _Hand_ or _Discard_?" in favor of direct selection.

Addresses the following concerns/issues:

1. Inconsistency. Daily Business Show makes you click the card in your hand to return to R&D, but Modded gives you a list of card names to Choose from. If you click a card while Modded's prompt is active, you will spend a click to install it normally. When you then try to select that card from Modded's prompt, the game will error. 

    Now any effect that targets a card from hand, discard, or game board uses Select prompts. The message to the user is "If you can see the card you want to effect, just click it."

2. Cancellable Choice prompts. There is currently no way to cancel a misclick on Self-modifying Code etc... once you click it, you must select a valid Choice to install, giving away information to your opponent. We now (generally) only use Choice prompts when tutoring through the draw pile, which is cancellable because we show that information in alphabetical (and not deck) order. 

3. Cancellable Select prompts. All Select prompts are cancellable through the "Done" button. If you click on Dinosaurus to host from hand but then decide against it, you can now select Done to abort. (Previously this was done with a Choice prompt that was uncancellable.)

@mtgred: this may cause some minor confusion at first. I recommend a News item along the lines of "We have changed how some cards select the targets of their effect, like SMC, Modded, and many more. If you can see the card you want to target, try clicking it!"


mtgred:
Awesome :+1:

mtgred:
I merged other pull requests and there are to be some merge conflicts with you branch. Can you have a look at it?

nealterrell:
Workin on it! :)

nealterrell:
@mtgred  done!

nealterrell:
I added a "targeting" icon when the active prompt is a Select prompt. I found the icon on flaticon.com and copied the attribution text to about.cljs.

mtgred:
:+1:

PR #915, opened 2015-11-25
Miscellaneous card fixes

### LINKED ISSUE #337: Bank Job
If I click on Bank Job out of my turn the ADD CREDIT window appears and it won't go away no matter what I do.

Tried everything.

### LINKED ISSUE #913: Philotic Entanglement for zero still triggers CP's ID
When playing as Chronos Protocol, scoring Philotic Entanglement while the Runner has no agendas in their score area will still trigger the Chronos Protocol ID, effectively doing 1 net damage although no net damage should be done.

Thanks much!

JoelCFC25:
* Expert Schedule Analyzer was missing `join` and not writing card titles from HQ to the log
* Cosmetic changes to Sealed Vault's menu and messages
* Only fire Philotic Entanglement's effect when damage will be > 0 (fixes #913)
* Fixed Bank Job to register a successful run and end the run if you take credits, and only be triggerable during a run (fixes #337)
* Yet more fixes for Old Hollywood Grid, Red Herrings, and Strongbox--if the Corp was voluntarily trashing them (even unrezzed) or if they got trashed by Drive By, their steal cost effects were persisting, which they shouldn't. The effects only persist now if trashed during a run.

### DIFF
--- a/src/clj/game/cards-agendas.clj
+++ b/src/clj/game/cards-agendas.clj
@@ -29,7 +29,7 @@
                  :choices {:req #(or (= (:advanceable %) "always")
                                      (and (= (:advanceable %) "while-rezzed") (:rezzed %))
                                      (= (:type %) "Agenda"))}
-                 :effect (effect (add-prop target :advance-counter 1))}]}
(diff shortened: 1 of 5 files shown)


mtgred:
:+1:

Issue #921, opened 2015-11-27
Street Peddler facedown cards acting like they're actively installed


thehunter518:
So far I've only noticed this with Adjusted Chronotype - the Noise player had a Wyldside installed, and a Street Peddler with an Adjusted Chronotype facedown under it. The Adjusted Chronotype was giving him back his lost click from Wyldside as if it was actually installed, and was reporting this in chat.

Haven't tested it with other facedown cards on Peddler (or with Apex), so this may be an isolated incident.

JoelCFC25:
We're aware of a major problem with at least one of the recent commits. Unfortunately it's a holiday weekend here in the States and some of us won't have much availability until early next week.

GordonsBeard:
https://github.com/mtgred/netrunner/issues/919

This looks like a related issue.

JoelCFC25:
Have submitted the fix for all of these issues in #923.

Issue #919, opened 2015-11-26
Multiple calls to callback functions?


finnbar:
Parasite seems to give -2 strength per token, and Medium seems to give two extra accesses per extra token. Furthermore, Underworld Contacts triggers twice per turn, and John Masanori gives two free cards on a successful run. Is there some place where callbacks are being run twice in the code by mistake? In the case of Parasite and Medium, neither of them gain more tokens than expected, they just seem to have a greater effect than usual.

(sorry if these issues are unrelated and I've filed them in the same Issue)

JoelCFC25:
It will be investigated, but it's a holiday weekend despite the fact that this one is almost certainly my fault, I can't get to it until Monday at the earliest.

finnbar:
That's OK, don't worry about it! Have a good weekend. :)
Also, while I remember, during these games all of the mentioned cards were hosted (Underworld and Masanori on Off-Campus Apartment, Medium on Djinn and Parasite on various ICE), so maybe both the host and the card is triggering or something? I'm not sure why I didn't mention that before, and I'm sorry that I seem to be throwing more and more random bits of information into a slightly incoherent report, but hopefully this is actually helpful and not just my hypothesising over little evidence. :P

JoelCFC25:
I have submitted a fix in #923.

PR #923, opened 2015-11-27
fix for facedown cards having events and hosted cards have doubled events

### LINKED ISSUE #918: Parasite trashes ice too early if the ice has additional strength (e.g. Meru Mati)
Steps to reproduce:

* Install Meru Mati on HQ, it has 1 strength + 3 because it's protecting HQ
* Install a Parasite on it, it still says it has 4 strength on the overlay
* Wait until the next runner turn, Parasite gains a virus counter, MM is immediately trashed even though it should have 3 strength still

JoelCFC25:
@mtgred: Urgent fix for the problems affecting facedown hosted cards and hosted cards having double event hooks. 

Fixes #918, #919, and #921.

### DIFF
--- a/src/clj/game/core.clj
+++ b/src/clj/game/core.clj
@@ -1477,7 +1477,8 @@
                            :previous-zone (:zone target))]
        (update! state side (update-in card [:hosted] #(conj % c)))
        (when-let [events (:events (card-def target))]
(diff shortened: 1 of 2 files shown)


mtgred:
:+1:

PR #924, opened 2015-11-29
Added partial support for Bug.

GordonsBeard:
Adds mostly-complete support for Bug. Does not work with multi-draw (Jackson Howard), only shows the last card drawn in those instances.


JoelCFC25:
Bug is only installable if there has been a successful HQ run--your alteration to the `:req` would allow it to be installed after a successful run on any central server.

GordonsBeard:
Apologies about that, a quick fix will limit that to just HQ.

mtgred:
:+1:

PR #928, opened 2015-11-29
Enforce Surge target rules

dersam:
This PR updates [Surge](http://netrunnerdb.com/en/card/02081) to only accept viruses that have had tokens added this turn as valid targets.

An `:added-virus-counter` key is added to the card, which is set to true if a virus counter was added this turn.  There is a special case in `runner-install` for cards like [Imp](http://netrunnerdb.com/en/card/02003) that come into play with virus tokens.

At the end of each turn, the flag is set to false on installed viruses; the corp turn is included in the reset for cases like [Gorman Drip](http://netrunnerdb.com/en/card/04005) so that Surge will only apply to tokens set on the runner turn.

This should work for all viruses without any additional modification on the viruses themselves.


dersam:
Updated with tests for the virus counter flag and playing Surge on various target scenarios.

JoelCFC25:
This is awesome! Since you are apparently taking to the code like a duck to water, any chance you want to tackle Architect immunity to trashing? :)

dersam:
Sure, I can do that. https://github.com/mtgred/netrunner/pull/939

mtgred:
I merged orther pull requests and there is a merge conflict with yours. Can you have a look at it?

dersam:
Resolved- conflict was in core.clj, lines 187-190, conflict with some refactoring that looks to have been done to fix facedown.

mtgred:
:+1:

Issue #893, opened 2015-11-19
Indication of the effect of a trace


R2devo:
When a trace is initiated the only indication of which effect it has is the base trace strength, this can be confusing with multiple trace effects of equal strength or in the same timing window. example; Uroboros or Viper on a last click run; a successful run on a server with a rezed Ash while Manhunt is active.

GordonsBeard:
I have an idea on how to get the trace's effect to print out, but I'm wondering if it's a case of showing too much in the logs.

Right now starting a trace will print out:
`[player] users [card] to initiate a trace with strength [total strength] ([base str] + [boost str]).`
![image](https://cloud.githubusercontent.com/assets/4015791/11459184/891bbd7c-9685-11e5-8a93-39c04ea7c1be.png)


I can change it to:
`[player] users [card] to initiate a trace with strength [total strength] ([base str] + [boost str]) ([trace effect]).`
![image](https://cloud.githubusercontent.com/assets/4015791/11459195/efb3144a-9685-11e5-9823-9cb2d307b6be.png)

or

`[player] users [card] to initiate a trace with strength [total strength] ([base str] + [boost str]) (effect: [trace effect]).`
![image](https://cloud.githubusercontent.com/assets/4015791/11459200/1f1a8db2-9686-11e5-9fee-9bfa40769632.png)

It seems this can be a global change, as long as the card has a `:msg:` defined, the `:trace` will pick it up. There are a few exceptions I've found that would require manual editing, as they are missing a `:msg` tag due to how they fire (Troll, Taurus, TMI, etc.). 

And there is also the Security Nexus, which wouldn't be able to list the part about the Runner bypassing.
![image](https://cloud.githubusercontent.com/assets/4015791/11459230/e2d146c4-9686-11e5-90ab-79475aefda6f.png)

GordonsBeard:
I also just noticed/ran into a problem regarding Data Hound, as it has a `:msg`, but this calculated based on the trace strengths; which (at the time of firing the trace) is unknown, causes a bit of barfo: 
![image](https://cloud.githubusercontent.com/assets/4015791/11459346/86b6f6ba-9689-11e5-8cc4-b252951b421a.png)

GordonsBeard:
I've got a PR in place that would fix this. My only question would be, what should show in the cases like Midseason Replacements, or Data Hound, where the trace will have a calculated cost? "Give the Runner X tags"?

PR #929, opened 2015-11-29
Traces list their effect in the chat log

GordonsBeard:
resolves https://github.com/mtgred/netrunner/issues/893

This will print the traces' effects at the same time as the trace's strength in the chat log.

image

It does this by modifying the message that `init-trace` in `core.clj` spits out. (https://github.com/mtgred/netrunner/commit/79be59225d71e11504eba16d71db5803e1e949f5)

There are a very small number of trace events that will have a calculated cost as part of their resolution. The current handling of traces will fire the associated `:msg` event at the time of resolution, meaning the calculated cost (final trace amounts from both Corp and Runner) will be known. If we try to print this `:msg` before we know these values, we get an error.
* **Archangel** *(add [card name] to grip)*
* **Data Hound** *(look at [x] cards)*
* **Midseason Replacements** *(give [x] tags)*
* **Invasion of Privacy** *(trash [x] resources/events from grip)*
* **Power Grid Overload** *(trash hardware of [x] cost)*

https://github.com/mtgred/netrunner/commit/509b38b439964ccf89e6dd4c212ca5ba215bc279 alters these problem cards' `:msg` values to just list "X" instead of the calculated cost, and include the calculated cost declaration as a `(system-msg)` in the cards' `:effects`. This gets around the problem, but adds some slight repetition to the chatlog.

*this is probably the worst-case scenario*

This duplication could be avoided by not ever listing the calculated cost (leaving it up to the reader to read both link/trace strengths declared in the chat log), but that seems like loss of functionality.

### DIFF
--- a/src/clj/game/cards-ice.clj
+++ b/src/clj/game/cards-ice.clj
@@ -13,8 +13,8 @@
                             :effect (req (system-msg state :corp "pays 3 [Credits] to force the Runner to encounter Archangel"))}}}
     :abilities [{:label "Trace 6 - Add 1 installed card to the Runner's Grip"
                  :trace {:base 6 :choices {:req #(:installed %)}
-                         :msg (msg "add " (:title target) " to the Runner's Grip")
-                         :effect (effect (move :runner target :hand true))}}]}
+                         :msg "add 1 installed card to the Runner's Grip"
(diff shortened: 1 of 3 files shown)


mtgred:
Nice :+1:

Issue #930, opened 2015-11-30
Medium not allowing multiaccess


MHGray:
So I had Hivemind with 1 counter on it, a Medium hosted on a Djinn and another Medium by itself. Ran R&D and was successful, so mediums triggered and got a counter each, so should have been accessing 3 cards - 1 base and 1 for each '2 counter' Medium thanks to Hive mind, but it only let me access 1 card. Chalked it up to weirdness. 

Ran on a later turn and before clicking successful run, moved virus counters from hivemind over to one of the mediums so I could at least access a couple cards. Medium had 4 counters, none on hivemind. Still only let me access 1 card. I don't even know why Hivemind has the ability to move counters onto other viruses. That actually wouldn't be a bad implementation if it didn't also lose the counter itself when you placed them.

JoelCFC25:
You're right, something's goofy here with Hivemind added to the mix. You aren't meant to have to actually use the ability to move Hivemind tokens here--I think that's for when you want to "spend" them like from Cache. 

When I replicate your scenario (Hivemind with 1, both Mediums initially with none), I accessed 1 card on the initial run when I should have accessed 3, as you say. When I run again, I correctly access 5 (Hivemind's 1 "added" to what becomes 2 counters on each Medium). I'll investigate.

JoelCFC25:
Thanks for the report--found the cause and fixed it. The `:req` for the `:no-action` event needed to use `get-virus-counters` instead of simply checking the card's own counters. I don't think I would have caught that one for a long time...

MHGray:
Glad I could be of some use and thank you for fixing it.

Issue #926, opened 2015-11-29
Neutralize All Threat access issues


ericboivin:
Neutralize All Threats works fine at first. Then as soon as you use Dr Lovegood on it, from now on and every turn after, you access only one card. 

For this part, I'm not too sure exactly of the sequence but I'm playing Adam a lot lately and I've had this happen many times. If later on Neutralize gets trashed, from now you don't access cards anymore on HQ runs.

JoelCFC25:
Thanks for the report. This is happening because I used `card-init ... false` on the assumption that we wouldn't want the card being "switched off" to act as if it's just been installed for the first time when it gets "switched back on". But it looks like most of the likely targets modify things via `:effect`, so they would need the `false` flag removed from `card-init` to resume proper behavior.

Issue #916, opened 2015-11-26
24/7 News Cycle


ihmcallister:
Just had a game where 24/7 was used on a Breaking News to trigger the same Breaking News. This is not allowable by the current rules. Fix would be appreciated.

Issue #809, opened 2015-10-12
Neutralize All Threats interacts poorly with point-reducing Ambushes


anathomical:
Neutralize All Threats triggers a trash as soon as a trashable card is encountered.  This causes weird interactions with Shi-Kyu and News Team.  In both cases the card is trashed, and then the runner is given the prompt regarding whether to take the non-point effect, or to put it in their score area as an agenda worth -1 point.

However, because the card has already been trashed, it apparently can't be found if the runner selects the -1 point option.  They still get the effect (their point total is reduced), but the card remains in Archives where it may trigger again.

anathomical:
Possibly related to #303

JoelCFC25:
Shi.Kyu works now in this situation as of #914, and I'll have a PR with the fix for News Team soon.

PR #932, opened 2015-11-30
Card fixes: Medium/Nerve Agent, 24/7 News Cycle, News Team, Dr. Lovegood

JoelCFC25:
Fixes for #930, #926, #916, and #809. 

* Medium and Nerve Agent have to use `get-virus-counters` so they'll see a Hivemind even when they have 0 counters
* 24/7 News Cycle switched to use card targeting so you can't trigger the agenda you just forfeited
* News Team receives the same fix that Shi.Kyu got so that Neutralize All Threats can take it as -1 agenda point from Archives
* Dr. Lovegood has to do a full card-init on its target at turn end so all the cards that modify hand size, HQ access, etc via `:effect` will turn back on properly

Also did a test for refresh of recurring credits on hosted cards, since it took me 4 tries to get it right...


mtgred:
:+1:

Issue #884, opened 2015-11-15
Invalid decks can be used


percomis:
I don't know if this is on purpose or something overlooked, but players can use invalid decks (out-of-faction agendas, less cards than required and overspending influence are the ones we tested).

JoelCFC25:
While I don't think we want to prevent starting a game when an invalid deck is loaded (it's incredibly useful for testing), I did manage to alter the game lobby code to do this: 

![image](https://cloud.githubusercontent.com/assets/10407969/11186660/45d0829c-8c48-11e5-9c16-de5ae0c0eb87.png)

What does everyone think of this idea? If you're playing an unknown opponent you can either use this as a sign to bail from the game or talk it through with them, but on the other hand you could still play draft format with someone you know.

percomis:
Sounds cool, I think it's a perfect solution.

sjohnc:
I'd prefer to keep invalid decks so that fun modes like draft and other fan made formats can be ran in this system.

PR #934, opened 2015-11-30
Alert players to invalid opponent decks in game lobby

JoelCFC25:
Following up to address #884. 

Add a check in the game lobby code to show the red "Invalid deck" text if applicable after a player has loaded a deck. This way you can have some more peace of mind when playing random people but still play testing decks or draft format games if you and the opponent are both aware of the situation and agree to proceed. 

screen shot 2015-11-30 at 12 05 52 pm


mtgred:
Nice :+1:

PR #935, opened 2015-11-30
implement Record Reconstructor

JoelCFC25:
We dare someone--ANYONE--to actually play this card! 

Added a fix for #936--Medium/Nerve Agent would not subsequently add counters if you hit a CVS when they were empty; also added a small change to prevent a Commercialization bug.


mtgred:
:+1:

PR #937, opened 2015-11-30
Identity Cyber-Cypher server selection with CSS overlay

JoelCFC25:
Feel free to adjust the color...I just picked something not being used and that had good contrast. It should go in the upper right I think to avoid clashing with strength boosts. Later on we can figure out how to extend this to Femme Fatale, which is slightly trickier because the overlay needs to remain even if the ice is derezzed, but be removed if either Femme Fatale is trashed or the targeted ice is trashed. 

screen shot 2015-11-30 at 2 35 38 pm

screen shot 2015-11-30 at 3 31 17 pm

### DIFF
--- a/src/clj/game/cards-icebreakers.clj
+++ b/src/clj/game/cards-icebreakers.clj
@@ -140,7 +140,10 @@
 
    "Cyber-Cypher"
    (auto-icebreaker ["Code Gate"]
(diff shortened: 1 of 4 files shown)


mtgred:
I merge some other pull request and there is a merge conflict on this one. Can you have look at it?

JoelCFC25:
Will resubmit...

mtgred:
Awesome :+1:

Issue #901, opened 2015-11-20
Apocalypse triggers "if trashed" effects for runner cards


thehunter518:
When Apocalypse fires successfully, it is supposed to trash all installed corp cards and turn all runner cards facedown. While the latter should cause the original cards to become uninstalled, it explicitly does NOT trash them. 

Certain cards like Gang Sign, Rolodex, and Tri-Maf Contact have significant penalties that occur when they are *trashed*, and these penalties are all currently firing off automatically when they are turned facedown. I'm not sure if this also occurs with effects that cause runner cards to leave the board through other means, such as Archangel, but it's worth investigating.

JoelCFC25:
It's because the `move` function is calling `desactivate` on cards headed to `[:rig :facedown]`, which in turn is firing the `leave-effect` if the card has one. We just need to block that for this case.

JoelCFC25:
I think I have a fix for this, but since I've been guilty of breaking things inadvertently a lot lately, I want to test it more thoroughly.

PR #938, opened 2015-11-30
Do not trigger leave-effect for runner cards put facedown

lostie:
:information_desk_person: This intends to prevent `leave-effects` from being triggered when a runner card is put facedown.

Solves https://github.com/mtgred/netrunner/issues/901

**NOTE:** I'm a newbie at `clj`, so I'm happy for you guys to give me some good feedback on what/how to do stuff, but I'm keen to improve `jinteki.net` and help the community :family: .

### DIFF
--- a/src/clj/game/core.clj
+++ b/src/clj/game/core.clj
@@ -118,7 +118,7 @@
   ([state side card] (desactivate state side card nil))
   ([state side card keep-counter]
    (let [c (dissoc card :current-strength :abilities :rezzed :special :facedown)
-         c (if (= (:side c) "Runner") (dissoc c :installed :counter :rec-counter :pump) c)
+         c (if (and (= (:side c) "Runner") (not (:facedown card))) (dissoc c :installed :counter :rec-counter :pump) c)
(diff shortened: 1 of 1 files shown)


JoelCFC25:
Unfortunately this won't work--when the `move` function makes the call to `desactivate`, the card being moved isn't technically in `[:rig :facedown]` yet and hasn't had `:facedown true` associated yet. Making this change and playing Apocalypse while you have Tri-maf Contact installed still results in taking 3 meat damage, which is what needs to be avoided.

lostie:
I have my development environment setup and running. 

Initially my implementation was in that direction using `(not (dest [:facedown :rig]))` but in this scope could not have access to the `dest` function.

Do you have any guides regarding how you design the tests? I was looking for tests for the Apocalypse card and could not find one.

JoelCFC25:
The unit testing framework is described in #736. I always have two browsers open side by side to play test games and see edits in action. Come join us in our Slack if you want--send an email to @mtgred to get an invite!

lostie:
Closing in favour of #942

PR #939, opened 2015-12-01
Enforce architect untrashable

### LINKED ISSUE #317: Architect can be trashed
As the title says, had a brain fart and tried to trash a rezzed Architect with a Parasite. My opponent was understandably surprised when it worked; guess sometimes I get rewarded for being dumb.

dersam:
Resolves #317.

Card definitions can now have a `:flags` map.  This map can be checked for a flag and its given value by using the `flag?` function.

Architect possesses the `:untrashable-while-rezzed` flag, which blocks trashing when a card is rezzed. Since the text of Architect is not in effect when it is derezzed (per [Lukas](https://twitter.com/rukasufox/status/487406756381749248)), it's effectively that, instead of when installed.  As an example (however unlikely), [Emergency Shutdown](http://netrunnerdb.com/en/card/02043) into [Apocalypse](http://netrunnerdb.com/en/card/09030) could trash an Architect.

Another new function added is `enforce-msg`. When passed a card and some message text, this displays an alert of the form "cardname message."
4eijjah

A test is included for attempting to trash architect when derezzed, rezzed, and from hand.

### DIFF
--- a/src/clj/game/cards-ice.clj
+++ b/src/clj/game/cards-ice.clj
@@ -22,7 +22,10 @@
                 trash-program end-the-run]}
 
    "Architect"
-   {:abilities [{:msg "look at the top 5 cards of R&D"
+   {:flags {
+            :untrashable-while-rezzed true
+            }
+    :abilities [{:msg "look at the top 5 cards of R&D"
                  :prompt "Choose a card to install" :priority true
                  :activatemsg "uses Architect to look at the top 5 cards of R&D"
                  :req (req (and (not (string? target))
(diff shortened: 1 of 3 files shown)


JoelCFC25:
Bravo! :+1:

JoelCFC25:
Would you be interested in joining our Slack instance? If so, send an email to @mtgred and he'll get you hooked up with an invite!

dersam:
Definitely, email sent. :)

mtgred:
:+1:

PR #942, opened 2015-12-01
stop Apocalypse from triggering leave play effects of cards turned facedown

### LINKED ISSUE #940: Tri-Maf Contact interaction with Apocalypse
If the runner has a Tri-maf Contact installed and then Apocalypses, they take meat damage as if Tri-maf had been trashed. However, the card is only being turned face-down, not being trashed. The meat damage shouldn't trigger in this occurrence. Obviously this isn't a high priority fix as Tri-maf isn't a super common card, but it accidentally flatlined my opponent earlier today.

JoelCFC25:
Fixes #940 and #901. 

In the `move` function, we have to move the `desactivate` trigger for installed and facedown cards further down to after the card has been associated with its destination zone. 

Then in `desactivate` we have to refine when we dissociate `:facedown` and make sure to only fire `:leave-play` effects if the card _isn't_ facedown.

### DIFF
--- a/src/clj/game/core.clj
+++ b/src/clj/game/core.clj
@@ -117,11 +117,12 @@
 (defn desactivate
   ([state side card] (desactivate state side card nil))
   ([state side card keep-counter]
-   (let [c (dissoc card :current-strength :abilities :rezzed :special :facedown)
-         c (if (= (:side c) "Runner") (dissoc c :installed :counter :rec-counter :pump) c)
+   (let [c (dissoc card :current-strength :abilities :rezzed :special :named-target)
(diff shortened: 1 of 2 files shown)


mtgred:
:+1:

JoelCFC25:
This is going to have to be reverted. As it stands, if the Runner has any programs in play consuming MU, cards that add link, or cards that modify max hand size, none of those will be reset to default values when everything gets flipped facedown by Apocalypse.  I think we'll have to add a new event trigger like `:turned-facedown` to distinguish these types of effects from actual `:leave-play` effects for things like the cards highlighted in #901.

PR #943, opened 2015-12-02
Enforce Blackmail no rez

dersam:
These changes enforce Blackmail preventing the rezzing of ICE.

This is done through use of a "current run" register on the game state. Within this register (`:register :current-run` on the game state), Blackmail sets the `:no-rez-ice` flag. This is checked by the `rez` function, which prevents rezzing if this is set.

To prevent any state pollution between runs, `handle-end-run` wipes this register at the end of every run.


dersam:
Withdrawn for now- refactoring these registers to use functions provided by the card definitions.

PR #944, opened 2015-12-02
improvements to Whirlpool and Clot, fix Bookmark and Pawn

### LINKED ISSUE #816: Bug with Whirlpool on an empty server
The Corp has Whirlpool rezzed on an empty server. I run the Whirlpool (in order to cause it to trash itself) and it fires and trashes. I cannot jack out, and the Continue button does nothing as the server no longer exists. (We resolved it by firing a Nisei mk2)

JoelCFC25:
Fixes #816. 

Only execute `prevent-jack-out` if it's a remote server and there are more cards than just Whirlpool in it so the server will continue to exist after Whirlpool trashes. Also added a bit to update the run position after Whirlpool trashes and the run is ongoing (because the server is still there). 

I was alerted that Clot needs a `:leave-play` for cases where the Runner installs it having missed the correct paid ability window--if pulled back to the Grip it was still preventing the Corp from scoring. 

* Bookmark no longer duplicates hosted cards when using the option to trash it and return all hosted cards to your Grip.
* Pawn no longer blocks the ability to make successful runs when hosted on non-ice and uses the card targeting system now for installs so you can do multiple installs when 2 or more Pawns advance past all ice simultaneously.
* Oversight AI tightened up to only allow targeting installed ice (could target ice in hand before)

### DIFF
--- a/src/clj/game/cards-hardware.clj
+++ b/src/clj/game/cards-hardware.clj
@@ -43,7 +43,7 @@
                       :effect (effect (rez :corp target))}}}
 
    "Bookmark"
-   {:abilities [{:label "Host 3 cards from your Grip facedown"
+   {:abilities [{:label "Host up to 3 cards from your Grip facedown"
                  :cost [:click 1] :msg "host up to 3 cards from their Grip facedown"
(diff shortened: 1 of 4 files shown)


mtgred:
:+1:

PR #947, opened 2015-12-04
Implement DDoS and Blackmail with turn/run flag registers

dersam:
This creates two new sets of automatically managed stacks of flagged conditions to be provided by card implementations. Both exist at the root of the game state. The first is a `current-turn` group- anything in `:stack :current-turn` is wiped when the current turn ends. The second is a `current-run` group- anything in `:stack :current-run` is wiped when the current run ends.  DDoS and Blackmail are implemented as demonstrations of these two stack groups.

The general idea of this is to provide hooks for conditions that are transient in nature, and generally provide negative conditions.  For example, both of the examples (DDoS and Blackmail) prevent rezzing ice temporarily.

To implement these cards, a "can-rez?" check was added to `core/rez`. This function is responsible for checking the `can-rez` stacks in current-turn and current-run groups.  If all of the conditions pass, it will allow the rezzing of the card.  In the case of DDoS and Blackmail, the functions they provide to the stack are coded to only trigger on ICE.

While only rezzing is currently using this system, it would be fairly trivial to provide hooks in other locations (such as run, install, etc).  `register-run-flag!` and `register-turn-flag!` both expect a keyword as a flag, but it can be anything.

Tests are provided for DDoS and Blackmail.

### DIFF
--- a/src/clj/game/cards-events.clj
+++ b/src/clj/game/cards-events.clj
@@ -34,9 +34,10 @@
     :effect (effect
               (resolve-ability (register-run-flag! state :can-rez-ice
                                              (fn [state side card]
-                                               ( (constantly false)
-                                                 (system-msg state side (str "is prevented from rezzing ICE on this run by "
-                                                                             (:title card))))
+                                               (if (has? card :type "ICE")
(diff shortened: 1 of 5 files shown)


mtgred:
Awesome :+1:

PR #950, opened 2015-12-05
Fix and test for installing Parasite facedown with Apex. 

### LINKED ISSUE #890: Parasite can't be installed facedown by Apex
If Apex chooses a Parasite in hand to install facedown when the turn starts, it brings up Parasite's install dialog (to choose a rezzed piece of ice) instead.

nealterrell:
Fixes #890 by not showing the "select a host" prompt when the card is being installed facedown.


mtgred:
:+1:

PR #952, opened 2015-12-06
Update remaining trace messages to be descriptive

GordonsBeard:
This updates the remaining trace messages to list their effect alongside the declaration of the trace.

* Assassin
* Burke Bugs
* Foxfire
* Hellion Alpha Test
* Punitive Counterstrike
* Snatch and Grab

Also adds a trace message to Assassin's second subroutine, which it did not have.


mtgred:
:+1:

PR #953, opened 2015-12-07
Card fixes and improvements

JoelCFC25:
_Fixes_
* Isabel McGuire no longer revealing unrezzed cards returned to HQ, added missing click cost
* Fixed Worlds Plaza hosting an asset bug (misplaced closing parens)
* Fixed redundant Wanton Destruction log message
* Cleaned up Rex Campaign message when taking the credits (lowercase was making it log [credits] instead of credit symbol)

_Improvements_
* Add instructional prompts to Indexing, Precognition, and Rolodex (users are asking all the time what to do when these cards are played). 
* Changed Unregistered S&W '35 to use card targeting


mtgred:
:+1:

PR #954, opened 2015-12-07
Fix Blackmail/TMI interaction

dersam:
Fixes the interaction between Blackmail and a rezzed TMI by removing the resolve-ability call in card defs and aligning the register calls with the function parameter order of most functions.


mtgred:
:+1:

PR #960, opened 2015-12-08
Improvements to Sundew, Fetal AI.

### LINKED ISSUE #301: Sundew grants credits even when running it first click

nealterrell:
This doesn't completely fix #301, but does improve it so that Sundew won't trigger on a click-run. Cards that trigger runs (Dirty Laundry) etc. will still grant 2credits if initiated on Sundew's server first click, but this is at least a minor improvement. I added a test for this scenario, and left one test commented out to verify the Dirty Laundry failure.

I attempted to validate a reported bug with Fetal AI dealing net damage when hosted on Film Critic, but my tests did not reproduce the issue. I did encounter an issue where choosing "Yes" to pay the 2cr to steal would cause the damage to not trigger if you did not have 2cr to pay. I fixed that and added tests.

### DIFF
--- a/src/clj/game/cards-assets.clj
+++ b/src/clj/game/cards-assets.clj
@@ -523,7 +523,7 @@
              :effect (effect (add-prop target :advance-counter 1 {:placed true}))}}
 
    "Sundew"
-   {:events {:runner-spent-click {:req (req (not (= (:server run) (:zone card)))) :once :per-turn
+   {:events {:runner-spent-click {:req (req (not this-server)) :once :per-turn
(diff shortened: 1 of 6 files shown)


mtgred:
:+1:

PR #963, opened 2015-12-08
Card fixes: 24/7 News Cycle, Mutate, Human First

### LINKED ISSUE #957: Mutate on Remote Broken
Details:  As playing as the foundry I rezzed a quandary in a remote server (only ice on server).  Later I mutated that ice and it went through the motions of looking through R&D but did not install card and kept asking for another target to mutate.  I was able to reproduce in testing.

When testing, Mutate appears to work on a central server just not on a remote server.  A side note: the cursor icon looks like targeting crosshair on a central ice, but does not on the remote ice.

### LINKED ISSUE #958: Global Food Initiative and Human First interaction
When the Corp scores GFI, Human First gives the runner 2 credits when it should give 3.

### LINKED ISSUE #961: Breaking News and 24/7 News Cycle not correct
If you fire Breaking News with 24/7 News Cycle, it shouldn't remove the tags at the end of the turn (since the turn you have scored breaking news wasn't this turn.) Additionally it shouldn't remove the tags if you forfeit the BN the turn you score it, but I haven't tested to see if that's implemented.

JoelCFC25:
* Fix #961: 24/7 News Cycle tags given from Breaking News remain when the Corp turn ends; updated unit test
* Fix #958: Human First now gives the Runner 3 credits when the Corp scores Global Food Initiative
* Fix #957: Mutate to use `:keep-server-alive` so it won't bug the game when you trash ice that is the only card in a server

Also added a small tweak to Sneakdoor Beta so its log message will read right when the run is successful.

### DIFF
--- a/src/clj/game/cards-operations.clj
+++ b/src/clj/game/cards-operations.clj
@@ -252,12 +252,13 @@
 
    "Mutate"
    {:req (req (seq (filter (every-pred rezzed? ice?) (all-installed state :corp))))
+    :prompt "Choose a rezzed piece of ICE to trash"
     :choices {:req (every-pred rezzed? ice?)}
     :msg (msg "to trash " (:title target))
     :effect (req (let [i (ice-index state target)
                        [reveal r] (split-with (complement ice?) (get-in @state [:corp :deck]))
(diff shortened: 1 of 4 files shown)


mtgred:
:+1:

Issue #962, opened 2015-12-08
'End Turn' button is not responding


rubyfell:
Using Chrome. After a turn or two End Turn button doesn't respond as expected. Console eventually shows that the turn ended, but the button is still highlighted and opponent states that he cannot start his turn. 

Here's the output from the game:

ruby_fell keeps their hand.

ruby_fell
hiya

eisentefel
glhf no bugs

eisentefel
hi
eisentefel keeps their hand.
eisentefel started their turn 1 with 10  and 5 cards in HQ.
eisentefel spends  to play Beanstalk Royalties.
eisentefel spends  to install a card in HQ.
eisentefel spends  to draw a card.
eisentefel is ending their turn 1 with 13  and 5 cards in HQ.
ruby_fell started their turn 1 with 5 and 5 cards in their Grip.
ruby_fell spends  and 5  to install Professional Contacts.
ruby_fell spends  to use Professional Contacts to gain 1  and draw 1 card.
ruby_fell spends  to use Professional Contacts to gain 1  and draw 1 card.
ruby_fell spends  and 2  to play Dirty Laundry.
eisentefel has no further action.
ruby_fell accesses 1 card.
ruby_fell uses Dirty Laundry to gain 5 .
ruby_fell is ending their turn 1 with 5  and 5 cards in their Grip.
eisentefel started their turn 2 with 13  and 5 cards in HQ.
eisentefel spends  and 10  to play Restructure.
eisentefel spends  to install a card in New remote.
eisentefel rezzes Jackson Howard.
eisentefel spends  to use Jackson Howard to draw 2 cards.
eisentefel trashes a card from their HQ.
eisentefel is ending their turn 2 with 18  and 5 cards in HQ.
ruby_fell started their turn 2 with 5 and 5 cards in their Grip.
ruby_fell spends  to use Professional Contacts to gain 1  and draw 1 card.
ruby_fell spends  to use Professional Contacts to gain 1  and draw 1 card.
ruby_fell spends  and 5  to play Sure Gamble.
ruby_fell spends  to install Paricia.
ruby_fell is ending their turn 2 with 11  and 5 cards in their Grip.

ruby_fell
bug

eisentefel
k

eisentefel
gg
ruby_fell is ending their turn 2 with 11  and 5 cards in their Grip.
eisentefel left the game.

GordonsBeard:
I can't seem to repro this on a local copy, the game server might just need a restart.

rubyfell:
[...]

PR #964, opened 2015-12-08
Checks if cards have subtype before re-find

GordonsBeard:
fixes https://github.com/mtgred/netrunner/issues/962

When all the cards have their subtypes checked, any cards without a subtype would return nil, which would lock the game up as you click "End turn".

This adds a small check to see if the card has a subtype. If not, then there is no need to check if it has the word "Virus" within the subtype.


GordonsBeard:
https://github.com/mtgred/netrunner/pull/965 will solve this in a better way

PR #965, opened 2015-12-09
Fix cards with no subtypes being unplayable

dersam:
Fixes the issue where a non-existent `:subtype` key caused an NPE that would unrecoverably crash games.  A test with Daily Casts is included as a regression test.

Also cleaned up some more files that were inadvertently commited (mongo binary files and an intellij project file).


dersam:
Resolves #962

dersam:
All instances of the problematic `re-find` call have been replaced with the more conventional `has?`.

mtgred:
:+1:

PR #969, opened 2015-12-10
Pull news items from Trello "announcements" card list

nealterrell:
To relieve @mtgred of the burder of having to insert news announcements into the database by hand, we can instead pull them from a card list on our Trello board. This changes `/data/news` to return a list of news items from the Announcements list on our Trello page. Anyone on the team can modify this list to then change the announcements on the main chat, allowing us to be more helpful when bugs come up or changes get pushed.

The announcements are shown in the same order they are sorted on Trello. The date shown is the "last modified date" -- there is no "creation date" field. The text shown for the announcement is the card's name.

__You must set an environment variable `TRELLO_API_KEY`__ to a key that you can get from here: https://trello.com/app-key . Shows a friendly message in the news section if the variable is not set.

The announcements are cached server side for 60 seconds.

Requires two new Node libraries, so make sure to `npm install` when updating.

image


mtgred:
Cool! :+1:

PR #966, opened 2015-12-09
Various test helper functions

dersam:
Defines a test helper function `score-agenda` in test.core.  Calculates the number of advancements required to score an agenda, awards the corp enough clicks and credits to score it, and then scores the agenda.  In the process, asserts expected advancements vs completed advancements.  Useful for writing agenda tests.

Adds tests for two agendas using this helper:
* Ancestral Imager
* Breaking News

PR #970, opened 2015-12-10
Tests for Scorched Earth, The Cleaners

nealterrell:
Damage has broken once or twice in the past, so I wrote some tests for Scorched Earth and The Cleaners.

I also fixed the testing framework's `load-card` method to close the database connection when finished. Whoops.

Needs @dersam's Pull Request #966 for the `score-agenda` method.


mtgred:
:+1:

PR #971, opened 2015-12-10
Add log check test helper

dersam:
Adds `last-log-contains?` function to `test.core`. Given a string `content`, it will check to see if the given string appears in the last log entry, and return true or false accordingly. `content` can also be a regular expression.

Test for Quetzal is included as a demonstration.


dersam:
Bundling these together with #966 .

PR #973, opened 2015-12-10
implement Gyri Labyrinth and Valley Grid

JoelCFC25:
Gyri Labyrinth has one small shortcoming in that it won't bump the Runner's max hand size back up in the somewhat unlikely event that it gets derezzed or trashed in between when the subroutine fires and the start of the next Corp turn. Possible causes would be death by Parasite, Emergency Shutdown, Crescentus, etc. In this case the Runner would need to adjust max hand size manually. 

Valley Grid, on the other hand, will store the number of times the Corp uses it and then increase the Runner's max hand size accordingly even if the Runner trashes it.


mtgred:
:+1:

PR #975, opened 2015-12-10
New file for mini-games and card combos in testing framework

JoelCFC25:
In addition to testing cards in isolation, we've agreed that we need to test larger pools of cards in combination with each other and using several core game subsystems at once. Since these "mini-games" or card combo tests don't fit neatly into the existing files we create a new one. 

The idea is to run more involved scenarios where turns pass back and forth and there is generally more interaction between the two sides. To kick things off, the first "mini-game" tests all of the following: 

* Traces (**SEA Source**)
* Net damage prevention (**Net Shield**)
* Resource trash cost increases (**Wireless Net Pavilion**)
* Ensure events work on hosted cards (**Off-Campus Apartment** hosting **Fall Guy**)
* Resource trash prevention (**Fall Guy**)

### DIFF
--- a/src/clj/test/cards.clj
+++ b/src/clj/test/cards.clj
@@ -10,3 +10,4 @@
 (load "cards-programs")
 (load "cards-resources")
 (load "cards-upgrades")
+(load "games-scenarios")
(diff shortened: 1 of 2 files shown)


mtgred:
:+1:

Issue #967, opened 2015-12-09
Forged Activation Orders doesn't tell you what server the ice is on


eedok:
becomes confusing when there's multiple copies of ice in play, it should say ice X at position Y on Server Z

JoelCFC25:
Thanks for the suggestion. I'll add this to my next PR when I collect a few more items.

PR #982, opened 2015-12-11
Prevent Dinosaurus/Djinn/Leprechaun from adding MU of trashed hosted items

### LINKED ISSUE #350: Trashing a hosted card triggers its "gain MU" leaves-play clause, even when it shouldn't
If a program is hosted on a card that reduces the number of MU the hosted card takes up (cards like Djinn, Leprechaun, Dinosaurus, etc), and the hosted card is trashed, you gain "back" the MU that the card would be taking.

So if Kate has Dinosaurus installed, puts Garrote on it, then trashes Garrote with no other programs installed, she'll gain the 2 MU that Garrote would have taken if it weren't hosted, going up to 6 MU available.

JoelCFC25:
Fixes #350. 

Dinosaurus, Djinn, and Leprechaun store the `:cid` of hosted cards and listen for the `:card-moved` event (would fire when you trash a hosted item from any of these cards). If the `:cid` of the moved card is found in the list, drop it from the list and deduct the MU of the program so there is no net change in MU. Tests for Dinosaurus and Leprechaun are included. 

There is also a minor tweak to Lamprey's log message and a small improvement to Forged Activation Orders to give the Corp precise information about which ice the Runner targeted (addressing #967). 

Lastly I added another mini-game that tests the following (and more):

* Rez cost reductions (**Breaker Bay Grid**)
* Psi games (**Caprice Nisei**)
* **Ash 2X3ZB9CY** access constraint
* HB: EtF ability
* Positional ice strength increase (**Turing**)
* Reduced agenda points for Runner when stealing **Global Food Initiative**
* Ice derezzing (**Emergency Shutdown**)
* Runner drip and run economy (**Data Folding**, **Dirty Laundry**)


mtgred:
:+1:

PR #983, opened 2015-12-11
Remove :seen flag from Corp cards when installing face-down.

### LINKED ISSUE #979: Installing a faceup card from Archives makes them retain :seen true
This bug should be fixed by adding :seen parameter card update in corp-install function in core.clj, however this function seems a bit too big for me to parse for now :^)

(Example card installing from Archives: Interns)

nealterrell:
Fixes #979, whereby a faceup card installed from Archives (via Interns etc.) will retain its `:seen true` flag. If said card is then trashed while unrezzed, it will go into Archives face up.

Includes tests.

### DIFF
--- a/src/clj/game/core.clj
+++ b/src/clj/game/core.clj
@@ -1647,7 +1647,7 @@
        (do (when (= server "New remote")
              (trigger-event state side :server-created card))
            (let [cdef (card-def card)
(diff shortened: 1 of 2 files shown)


zaroth:
Neat, thanks for taking care of it!

mtgred:
:+1:

PR #986, opened 2015-12-12
fix Vamp to allow tag to be avoided

### LINKED ISSUE #985: New Angeles City Hall doesn't ask me to avoid Vamp tags
and everytime my Rachel is trashed due to being tagged. this is more of a nuisance, but it has spawned a rules discussion with my opponent two times already.

JoelCFC25:
Fixes #985. Vamp needs to use `(tag-runner 1)` instead of the old way of applying tags.


zaroth:
on a related note, the Snatch & Grab tag should be fixed to be unavoidable (rules source: http://ancur.wikia.com/wiki/Snatch_and_Grab )

mtgred:
:+1:

PR #989, opened 2015-12-14
fix Snatch&Grab tag to be unpreventable

zaroth:
As per Lucas's ruling: http://ancur.wikia.com/wiki/Snatch_and_Grab


nealterrell:
I think a better implementation would be to use the `:unpreventable` optional argument to `tag-runner`. `tag-runner` does more than just increase the `:tag` count... it fires two important events that other cards rely on. Sure, currently no one listens for `:runner-gain-tag`, but several cards _should_ (Rachel Beckman), and we have future cards to consider as well.

zaroth:
ok, i'll cook up something then.

zaroth:
Changed the card to use `(tag-runner :unpreventable)`

Issue #334, opened 2015-07-06
structure of the run, rezzing just before access


prozz:
there is a need of new step when run is almost finished:
- runner should be asked: 'Access?' with answers 'Yes' or 'Jackout'
- when 'Yes' corp got window for rezzing DRT or Crisium Grid and confirms successful run.

right now there is a confusion, after passing all ices when i rez DRT runner still sees 'Jackout' option and this leads to misunderstandings.

if this issue can be sorted with current implementation, pls let me know. im happy to gather such things and write a faq for the game ;)

regards
PR

JoelCFC25:
Related to #131.

mtgred:
I prefer to avoid adding an additional step as it is rarely needed and would make all runs more tedious.

A solution would be to ask the Runner in the chat when he passes the last ICE if he want to access before clicking "No more action". It could look suspicious, but it could be a bluff as well. It's basically how you would do it in real life.

Another solution would be to disable the jack out button when the Runner decides to go pass the last ICE. But I prefer leave jack out button always available to allow the Runner to cancel the run in case something goes wrong.

prozz:
hmm. valid reasoning.

got another idea then, that comes along with 'thinking button' suggested in another issue.

how about a set of very common phrases that are built-in as keyboard shotcuts or even some cute buttons? examples would include:
- thinking
- access?
- afk/brb

id love to be able to use more keyboard for both that and some common operations like drawing or taking credits.

thehunter518:
I was going to say that asking the runner "do you want to access?" makes them immediately suspicious and spoils the trap, but mtgred makes a good point - that is literally how it works face to face. In real life I usually try to say something innocuous like "alright, so you're in?". I've never really considered doing it to full-on bluff someone, but it's a valid option.

I think if it was implemented on the site as an additional button prompt, the advantage would go to the corp, as the runner would quickly grow weary of pressing "Continue" multiples times during a run, and wouldn't ever think to jack out. I also see how it would make every single run in the game more tedious for the sake of one or two cards. 

I'm just going to start asking every time I guess, to get in their head. Whenever my friend plays Wayland (in-person), he constantly asks the runner how many cards they have in their hand. Same idea I think, just a bit more tedious to type it out.

aiannacc:
[...]

PR #990, opened 2015-12-14
Added new chat commands

zaroth:
- `/end-run` -> End the run (Corp only)
- `/jack-out` -> Jack out (Runner only)
- `/trace #n` -> Start trace with base strength `n` (Corp only)
- `/psi` -> Start a Psi game (Corp only)

Adding the "emergency jackout" levers mentioned by me in #334.

### DIFF
--- a/src/clj/game/core.clj
+++ b/src/clj/game/core.clj
@@ -1925,12 +1925,12 @@
                                                :choices {:req (fn [t] (= (:side t) (side-str %2)))}}
                                         {:title "/card-info command"} nil)
         "/counter"    #(resolve-ability %1 %2 {:effect (effect (set-prop target :counter value)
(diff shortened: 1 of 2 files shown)


zaroth:
Added more commands: 
- `/counter #n` -> set counters on a card to `n` (player's own cards only)
- `/adv-counter #n` -> set advancement counters on a card to `n` (player's own cards only)
- `/card-info` -> display debug info about a card (player's own cards only)

nealterrell:
Nice job, these are great. Can you add them to the Wiki (Jinteki.net Guide) under Console Commands? Then we can put up a news announcement to get people to read about them.

zaroth:
Updated the wiki as @nealterrell suggested.

Issue #734, opened 2015-09-19
Trashing ICE during a run should update the run position


queueseven:

JoelCFC25:
The self-trashing ICE part of this will be handled by #1016.

PR #991, opened 2015-12-15
Have ice that self-trash during runs modify the run position

JoelCFC25:
Adds a new core function to decrement the run position when ice self-trash (currently Data Mine, It's a Trap!, Lab Dog, Special Offer, and Whirlpool) during a run. Mostly addresses #734, but not for the case of Parasite killing ice. 

Test included for Special Offer.

### DIFF
--- a/src/clj/game/cards-ice.clj
+++ b/src/clj/game/cards-ice.clj
@@ -169,7 +169,10 @@
                                       )}}]}
 
    "Data Mine"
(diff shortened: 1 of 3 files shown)


JoelCFC25:
Don't merge this yet, cleaning up the core function shortly.

JoelCFC25:
Recreating this after I resolve conflicts.

PR #993, opened 2015-12-15
Tweak corp spend msg when installing ice

porterjamesj:
This is just a small tweak to change the system message when the Corp installs ice. Previously it would say "spends [Click] to install a card in {server}", which isn't really correct since ice are protecting servers, not in them. This makes it say "spends [Click] to install ice protecting {server}", which seems better to me, feel free to ignore if you disagree or this broke something else :grin:

Screenshot:

screen shot 2015-12-14 at 11 30 37 pm

### DIFF
--- a/src/clj/game/core.clj
+++ b/src/clj/game/core.clj
@@ -1661,10 +1661,11 @@
                      (system-msg state side (str "trashes " (if (:rezzed prev-card)
                                                               (:title prev-card) "a card") " in " server))
                      (trash state side prev-card {:keep-server-alive true})))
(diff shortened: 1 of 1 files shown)


JoelCFC25:
I really like it--it's much more informative and precise. Great job!

nealterrell:
We should be using official game terminology whenever possible :+1:

zaroth:
will tweak your addition in my next PR - "ice" should be "ICE" and there is no reason to locally define `is-ice` function when in included cards.clj is a perfectly good `ice?` function. Besides these, welcome to the *adding clarity to log squad*! :D

zaroth:
tweaked above in #1004.

PR #995, opened 2015-12-16
Fixes astroscript token placement

### LINKED ISSUE #984: AstroScript tokens can be placed on scored agendas
Minor issue, but nothing stops AstroScript's selection prompt from targeting a scored agenda, or even an agenda in HQ for that matter. Needs a check to see if the target is installed.

Saintis:
Fixes #984. Prevents Astroscript or Trick of Light from being able to
place advancement counters on Agendas that are not installed.

Tried using :installed but did not get it to work.


Saintis:
Potentially worth extracting two functions: "is-installed?" and "can-be-advanced?".

nealterrell:
Curious that :installed didn't work. What behaviour were you seeing?

Saintis:
Probably because I tried it before I got a hang on how the code works. Can give it a try again.

Saintis:
Had a look, seems like it is not going to work easy since it would involve nesting #() (compiler error when I tried).

nealterrell:
OK. Up to you if you want to continue on this. For nested anonymous functions, use `(fn [args] ...)` for one of the two functions, like this line from Mutate:

```clojure
(swap! state update-in [:corp :deck] (fn [coll] (remove-once #(not= (:cid %) (:cid newice)) coll)))`

We wanted the `(fn [coll])` to be a one-argument anonymous function but couldn't use `#` because of the nesting.

Saintis:
Decided to go deeper on this stuff. Created a new function `can-be-advanced?` and am now implementing it's use in all relevant cards ([14 in total](http://netrunnerdb.com/find/?q=x%3A%22that+can+be+advanced%22&sort=name&view=list&_locale=en)).

Will try to write some tests for the related cards as well to make sure everything works as intended.

nealterrell:
How about one more improvement? :D Forgive me, can't help it as a CS instructor.

When you do something like this: `:choices {:req #(can-be-advanced? %)}` you are creating an anonymous function of one parameter that calls another function with that parameter. You can instead just use the name `can-be-advanced` to pass that function to the `:choices`. It's similar to doing

`(filter even? [1 2 3 4 5])` instead of `(filter #(even? %) [1 2 3 4 5])`

Saintis:
All suggestions and help more than welcome!

I don't really know what is going on with Clojure to be honest. Just picking up stuff from what is going on in the code (first taste of Clojure was 3 days ago with this stuff :wink:).

I will go and remove all unnecessary anonymous function now. Also thanks for explaining `#()`, couldn't find anything with some quick googling!

zaroth:
@Saintis As one intermediate beginner to another, I can recommend http://www.braveclojure.com/ - after reading it a bit, more and more of jinteki.net's code starts to make sense :)

nealterrell:
ALMOST DONE :).

Just need to merge with the latest changes, which has a conflict with my minor change to Shipment from SanSan (use yours). With that merge, look at `card-is?` in `util.clj` and how I used it in Shipment to simplify your `can-be-advanced?` function. Then we're DONE :+1:

Saintis:
Merged stuff and prettified the function. Any last fixes? :smile:

zaroth:
I'd say that adding your own `declare` line right under a massive `declare` line is a bit weird (why not add to the existing one), but this time it could actually be better that way. In another PR I made changes to this huge declare line and us both changing it would lead to a merge conflict, and I think everyone hates those ;-)

Saintis:
Yeah. I think the proper solution would be to move the short `is-ice?` and `rezzed?` functions up to and remove the `declare` I added.

zaroth:
or even, as @nealterrell suggested to me, move all functions not requiring @state to `utils.clj`.

nealterrell:
This is ready, @mtgred. We'll do a cleanup of the utility functions after the next wave of merges.

mtgred:
There are conflicts with the master branch. Can you have a look?

Saintis:
Fixed :smile:

mtgred:
:+1:

PR #996, opened 2015-12-16
Allow player to pick side when creating new game

### LINKED ISSUE #609: Feature Suggestion: Make two buttons for a New Game (Runner/Corp)
I often want to play as a specific Side but and it happens ever so often that I join a Game that was advertised as Runner but the second I join it gets switched around.
A Simple Solution would be to allow the User to select witch side he wants to Play during Game Creation, hence two New Game Buttons one For a new Corp Game and one for a new Runner Game.

zaroth:
how does it look like on test server

Fixes #609.


mtgred:
:+1:

PR #1000, opened 2015-12-16
Have Kit update altered ICE and be able to match strength with an icebreaker

JoelCFC25:
When Kit uses her ability to change an encountered ice and make it gain Code Gate, we should update the run ice and update the ice strength (extreme corner case of the ice suddenly getting a strength boost from a scored Encrypted Portals!). 

This also adds a new event trigger so Kit can now successfully match strength with an installed Decoder after changing the ice.


mtgred:
:+1:

Issue #902, opened 2015-11-21
Haarpsichord and Hades Shard


yang573:
If the runner uses Hades Shard on the corp's turn, s/he can only access agendas if s/he stole one during his/her previous turn.

JoelCFC25:
I can't reproduce this. Hades Shard was modified recently--have you (or the opponent) done a forced refresh of the site since Wednesday. 

However, I was able to steal _multiple_ agendas by using Hades Shard against Haarpsichord, so that's an error that needs to be investigated.

JoelCFC25:
I reverted to the way Hades Shard was before I did any modifications to it and found that it was also capable of stealing multiple agendas from Haarpsichord then, too--so this has been a bug waiting to be discovered for quite awhile.

PR #1001, opened 2015-12-16
Enables access selection for Hades Shard

Saintis:
Using Hades Shard now brings up selection choice like normal Archives access, and correctly works with Haarpsichord (can only steal one agenda). (Fixes half of #902).

Tried to get stealing on Corp turn to work but couldn't find a work-around for prevent-steal disabling taking agendas.

### DIFF
--- a/src/clj/game/cards-resources.clj
+++ b/src/clj/game/cards-resources.clj
@@ -240,7 +240,7 @@
    {:abilities [{:msg "access all cards in Archives"
                  :effect (req (trash state side card {:cause :ability-cost})
                               (swap! state update-in [:corp :discard] #(map (fn [c] (assoc c :seen true)) %))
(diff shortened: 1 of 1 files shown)


cupello:
I am a noob in this program so forgive me if I am saying something stupid.

One way to think is putting a trigger in the "end-turn" method that reverts the state of the ":cannot-steal" - the problem with that is you are putting some card logic in the core.

Other way is using some trigger that at the end of each turn reverts the status of the ":cannot-steal", I dont know if there is something like that in the code already.

I was working on this bug but you have found a solution before me; so, after your commit is merged I will be pushing only my test cases.

Saintis:
Seems like that's how it should work, as in the :cannot-steal is reset at the end of each turn. Not sure how end-of-turn stuff is handled but can have a dig around.

mtgred:
:+1:

PR #1004, opened 2015-12-17
Make card descriptions in chatlog more... descriptive

zaroth:
Added a new function, `code-str [state card]` which is supposed to print strings for the use of chat log messages. Using it will provide the chatlog with necessary information to identify a card being written about. Example results:

- `Ice Wall protecting HQ at position 0`
- `Parasite hosted on Ice Wall protecting HQ at position 0`
- `Jackson Howard in Server 1`
- `Corroder` (I figured `in Runner's Rig` would be redundant, unless a Runner's card is hosted on something)
- `Cache hosted on Scheherezade hosted on Djinn` (nevermind pointlessness of such hosting)

So far it is used in Constellation Protocol (a card which effect was completely silent in chatlog) and targets of `/counter` command, although I recommend using `card-str` wherever a target card needs to be stated in chatlog. For greater uniformity and clarity!

Complete with test suite :)


mtgred:
:+1:

Issue #111, opened 2015-02-19
Clicking a card with only 1 card-action should show still show the menu


Agnate:
Clicking on a card with only 1 card-action should still show the menu of actions. Currently many Corp and Runner actions are unintentionally triggered because someone clicks on the card to see what actions are available and causes the only action to occur.

From a User Experience perspective, this is bad, as it makes it hard to tell which cards are actually implemented and working. From a game pace perspective, it wastes time because now the player needs to correct their board state.

dimarco70:
Seconded. Having a pop-up m for rezzing and derezzing would avoid several misclicks.

JoelCFC25:
We can now have a system for making Runner one-option cards cancellable, as well as #1034 for optionally doing the same to Corp cards.

PR #1005, opened 2015-12-17
Choice Popup for all actions

Saintis:
Even single choice actions now have a pop-up to use them.

This aims to solve the issues outlined in #111.

Maybe this would work as a user preference setting.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,5 @@ pom.xml
 /data/andb-*.json
 /data/img
 /.lein-failures
+
+/data/db
--- a/.gitignore
+++ b/.gitignore
@@ -21,5 +21,3 @@ pom.xml
 /data/andb-*.json
 /data/img
 /.lein-failures
-
-/data/db
--- a/src/cljs/netrunner/gameboard.cljs
+++ b/src/cljs/netrunner/gameboard.cljs
@@ -97,14 +97,20 @@
           (if (not rezzed) (cons "rez" %) (cons "derez" %))
           %))))
 
-(defn handle-abilities [{:keys [abilities facedown side] :as card} owner]
+(defn show-menu? [side show-menu]
+  (or (and (= side "Corp")
+           (not (= show-menu false)))     ; menu opt-out for scored corp cards
+      (and (= side "Runner") show-menu))) ; menu opt-in for runner cards
+
+(defn handle-abilities [{:keys [abilities facedown side show-menu] :as card} owner]
   (let [actions (action-list card)
         c (+ (count actions) (count abilities))]
     (when (not (and (= side "Runner") facedown))
-      (cond (or (= side "Corp") (> c 1)) (-> (om/get-node owner "abilities") js/$ .toggle)
+      (cond (or (show-menu? side show-menu)
+                (> c 1)) (-> (om/get-node owner "abilities") js/$ .toggle)
             (= c 1) (if (= (count abilities) 1)
-                          (send-command "ability" {:card card :ability 0})
-                          (send-command (first actions) {:card card}))))))
+                        (send-command "ability" {:card card :ability 0})
+                        (send-command (first actions) {:card card}))))))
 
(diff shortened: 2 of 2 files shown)


zaroth:
I don't think this should be a global setting, rather a per-card setting. Or maybe there should be a per-card setting to work otherwise. If no exceptions are made either way, player experience will worsen.

Imagine clicking Datasucker 10 times in 2 places instead of clicking 5 times quickly like currently to reduce a Parasited Tollbooth to noise. Or double clicking a clone chip. Or clicking a Mimic 10 times to break a Komainu. I wouldn't be surprised if players started clicking on **[-]** button next to the credits then instead of actually **using** the UI we're trying to give them.

zaroth:
Also, this would break a lot of cards who don't have `:label` defined, since their ability would be blank in the pop-up menu. Another argument for per-case implementation.

Saintis:
Using a per-card setting sounds like a very good idea. Not sure how I would implement that though, any pointers?

And since this would probably be most useful for the Corp (preventing accidental asset rez) it could also just be turned on for the corp side (and potentially opt-in on the runner side).

I'll start by changing it to corp-side only.

nealterrell:
Corp-only sounds like a good start. Datasucker is a great example from @zaroth, so is Cache, but I can't think of any equivalents on Corp side, since anything that gets rezzed with only one ability will already show the menu because of Derez.

Saintis:
I guess the only thing on corp side would be advancing non-scorable agendas. But that would be nice to not do accidentally as well.

zaroth:
Corp side example? Private Security Force, there you go :)

Saintis:
Yeah, any scored agenda with an effect would be affected.

.gitignore
```diff
@@ -21,3 +21,5 @@ pom.xml
 /data/andb-*.json
 /data/img
 /.lein-failures
+
+/data/db
```
Saintis: This does not need to be committed / pulled I guess. I did not intend it to sneak in...

nealterrell:
Thoughts on how to do this on a per-card basis.

If a card definition (where the abilities are written) has a `:data` field, the contents of the data's vector will be merged into the card's data when the card is activated. This is how Plascrete starts with 4 counters, for example. If you want a card itself to have a property stating that its ability menu should show a menu even if there's only 1 ability, you can add something like `:data [:show-menu true]` to the card's definition. The UI can then check for that key in gameboard.cljs to show the menu even if there is only 1 action.

GordonsBeard:
@zaroth I think it would be good to try and slow down (even a bit) the ability for people to spam click that fast. You might want to click PSF 3 times really quick because it's satisfying, but I've got any number of tools I can use between each click that might let me live.

zaroth:
There are certainly better ways for giving reaction windows (e.g. JS timers, hourglass, etc.) than UI clutter.

Saintis:
So now how it works is that menus are enabled on corp by default, and disabled on runner cards.

Corp menu can be turned off by adding `:show-menu false` to the data, then behaves as before, if this is added to an agenda it is only active while scored.

Runner menu can be turned on by adding `:show-menu true` to the data.

Haven't added anything to any cards. My suggested candidates include agendas that have an action that cannot be cancelled and runner cards that use up counters, or trash themselves (e.g. Imp, Shards).

nealterrell:
I think this PR is ready, @mtgred. Net effect is summarized in the post before this one. For now, the only visible change will be that installed agendas will take 2 clicks to advance, unrezzed cards will take 2 clicks to rez, and scored agendas with one ability will take 2 clicks to use the ability. I think this is a good change for user friendliness without requiring too many more inputs for the average game.  (1 extra click per rez, 1 extra click per advancement -- maybe 20 extra clicks per game?)

mtgred:
The only issue is for unimplemented corp assets and upgrades. The most elegant solution would be to disable the auto trigger of 1st ability when that ability is "Derez".

Saintis:
Resubmitting change that adds menu for "derez" action.

PR #1006, opened 2015-12-17
Utility to make testing some card properties easier

nealterrell:
Big discussion on this in the Slack channel, but in general, cards that use `:choices` to target another card have to account for both keywords and strings as property values because of the way JSON data is handled when sent back to the game server. This utility `card-is?` should make it easier to account for the difference... so rather than write code like `(or (= (:side %) "Corp") (= (:side %) :corp))`, you can write `(card-is? % :side :corp)` and the utility will do the work to determine if it's a string, needs to be converted to a keyword, etc.

Includes @JoelCFC25's test case for Shipment from SanSan, which was the inspiration, and some improvements based on @zaroth's work.


mtgred:
There is merge conflict. Can you have a look at it?

nealterrell:
Huh, thought I fixed it. One minute.

nealterrell:
Merged.

zaroth:
:+1: Definitely want this!

Issue #842, opened 2015-10-30
Display Warning messages


lostie:
When:
- a player has no credits to installed/rez a given card
- a runner is trying to install a program but doesn't have enough MU
- other examples where applicable

it would be a nice to have if the player got presented with a message which indicates the cause for him/her being prevented to take the action he/she is trying to do.

JoelCFC25:
With toasts we can now provide these notifications when it's useful. Let's handle new specific cases of needed notifications with their own issues!

PR #1007, opened 2015-12-17
Notification on unable to pay cost

Saintis:
(The next instalment in my UI bloating series…)

Prompts a warning / notification if player tries to do an action or ability that they cannot pay for.

Displays card that asked and the cost unable to pay.

Attempts to alleviate #842. Discuss! :smile:


Saintis:
Failing tests shall be fixed!

Saintis:
Still lots to do. Will fix and submit PR again.

PR #1010, opened 2015-12-19
Prevent Lotus Field strength from being lowered.

### LINKED ISSUE #1009: Ice Carver lowers Lotus Field's strength
In a game, Runner has Ice Carver in play, and runs a server with a Lotus Field. Lotus Field is listed as 3 strength, when you shouldn't be able to lower Lotus Field's strength.

nealterrell:
Fixes #1009. Includes a few tests for Lotus Field, Ice Carver, and Stronger Together.

Had to make a few minor changes to core, mostly to more frequently update ice strength (when starting a run, or when jacking out, for example).

### DIFF
--- a/src/clj/game/cards-agendas.clj
+++ b/src/clj/game/cards-agendas.clj
@@ -133,7 +133,7 @@
                                   0 (flatten (seq (:servers corp)))))
                     (update-all-ice))
     :events {:pre-ice-strength {:req (req (has? target :subtype "Code Gate"))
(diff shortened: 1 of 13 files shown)


zaroth:
Nice job! I wonder how often and in what way is `flag?` used, because judging just from the name it sounds like its default behavior could be checking for truthy value, i.e. `(flag? [card flag]) (flag? card flag true)`. Although that wouldn't work, it would just check for equality with `true`... Nevermind :)

mtgred:
:+1:

PR #1016, opened 2015-12-21
Have ice that self-trash during runs modify the run position

JoelCFC25:
Adds a new core function to update run position when a piece of ICE self-trashes during a run, and updates the 5 existing ICE that currently do this to use it. Includes a test for Special Offer.


mtgred:
:+1:

PR #1018, opened 2015-12-21
handle subtypes of Morph ICE

JoelCFC25:
We needed a small core modification and a new function to handle the flip-flopping subtypes of Morph ICE. They will now respond to conventional advancements and placed advancements (Shipment from Kaguya/SanSan, Trick of Light, etc) and thus be able to update their strength to take advantage of scored Superior Cyberwalls or Encrypted Portals, or whatever else comes in the future.

### DIFF
--- a/src/clj/game/cards-events.clj
+++ b/src/clj/game/cards-events.clj
@@ -640,13 +640,13 @@
                        stypes (:subtype ice)]
               (resolve-ability
                  state :runner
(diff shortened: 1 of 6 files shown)


mtgred:
There are merge conflicts with master. Can you have a look?

JoelCFC25:
I am pretty baffled by merges and resolving conflicts, so I'll just close this and resubmit. :)

PR #1029, opened 2015-12-23
Fixes Paige Piper.

### LINKED ISSUE #1013: Paige Piper prints the wrong card name when choosing to remove cards from Stack
Paige Piper installed (on Off-Campus Apartment in case it matters), then installing GS Striker M1.

```
shmop spends  and 4  to install GS Striker M1.
shmop uses Paige Piper to shuffle their Stack.
shmop trashes 1 copy of Paige Piper.
```

It was another GS Striker M1 which was trashed, not Paige Piper.

Saintis:
Changes the `pphelper` function to just take the card title. Clash happened in the use of the `req` macro overwriting the previous `card` definition (or something like that). 

Also made the trash `:unpreventable true`.

Fixes #1013.

### DIFF
--- a/src/clj/game/cards-resources.clj
+++ b/src/clj/game/cards-resources.clj
@@ -426,25 +426,26 @@
     :leave-play (req (remove-watch state :order-of-sol))}
 
    "Paige Piper"
-   (let [pphelper (fn [card cards]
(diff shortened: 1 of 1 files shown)


mtgred:
:+1:

PR #1030, opened 2015-12-23
handle subtypes of Morph ICE

JoelCFC25:
Resolved conflicts from the previous attempt...

We needed a small core modification and a new function to handle the flip-flopping subtypes of Morph ICE. They will now respond to conventional advancements and placed advancements (Shipment from Kaguya/SanSan, Trick of Light, etc) and thus be able to update their strength to take advantage of scored Superior Cyberwalls or Encrypted Portals, or whatever else comes in the future.


mtgred:
:+1:

PR #1032, opened 2015-12-23
Prevent Medium/Nerve Agent repeated clicking; fix click-trash advanceable assets

### LINKED ISSUE #1027: Medium "click on card to access less cards" does not work properly
It should be more clear when you can access less cards using medium. The phrase "Runner may click on medium to access less cards" is lacking "before runner starts accessing".

At the current state clicking on medium mid accessing causes delayed window which should only appear before runner starts accessing. Also clicking on medium mid accessing adds virus counters on it.

JoelCFC25:
Fixes #1027 by stopping Medium and Nerve Agent's "access fewer cards" ability from being clickable repeatedly or during access. 

While helping me debug a unit test last night, @nealterrell discovered that Reversed Accounts bugs if for some reason the Corp ever rezzed and then tried to click-trash it when it had 0 advancements. By extension the same problem could afflict Alix T4LB07, GRNDL Refinery, Project Junebug, and Thomas Haas, so these are all fixed too.

### DIFF
--- a/src/clj/game/cards-assets.clj
+++ b/src/clj/game/cards-assets.clj
@@ -22,8 +22,8 @@
    "Alix T4LB07"
    {:events {:corp-install {:effect (effect (add-prop card :counter 1))}}
     :abilities [{:cost [:click 1] :label "Gain 2 [Credits] for each counter on Alix T4LB07"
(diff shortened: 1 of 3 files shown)


mtgred:
:+1:

PR #1033, opened 2015-12-23
Bumper crop of new card tests

JoelCFC25:
We're trying to use the window of time before Mumbad starts to beef up the testing framework, so....

Here's a big batch of new card tests. I tried to hit a variety of things with mildly complicated effects. This also restores alphabetical order to a couple of the files, which I think we've agreed is the best organization.


mtgred:
:-1:

PR #1034, opened 2015-12-24
Adds pop-up choice menu for derez action.

Saintis:
Forces choice menu when only option is "derez".

Attempts to alleviate #111. See #1005 for discussion.


mtgred:
:+1:

PR #1045, opened 2015-12-28
Yet more card tests

JoelCFC25:
Nothing to add to my comments from the last batch of card tests--still trying to hit a wide variety of kinds of cards and types of effects.


mtgred:
:+1:

PR #1031, opened 2015-12-23
Fixes cards being draggable for spectators.

Saintis:
Turns off draggable for spectators.

Fixes #1028 and #1026

PR #1048, opened 2015-12-28
Spectator server-side fix.

### LINKED ISSUE #1028: Spectators messing with cards in new ways
Ko26 has been placing cards into the opposing player's deck- agendas on top of the stack, programs onto R&D, etc.  Apparently that's still allowed?  Weird oversight from the fix when spectators could trash cards.  Hopefully a quick and simple fix.

Saintis:
Creates new list of allowed actions for spectators. Passes user with `send-command` to check which list to use.

Intends to fix #1028 fully. See #1031 for client-side fix.

### DIFF
--- a/src/clj/game/main.clj
+++ b/src/clj/game/main.clj
@@ -13,6 +13,9 @@
 (def last-states (atom {}))
 (def ctx (ZMQ/context 1))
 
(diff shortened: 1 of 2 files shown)


mtgred:
:+1:

PR #1049, opened 2015-12-28
Fixes Brain Chip.

### LINKED ISSUE #1040: Brain Chip bonuses should floor at 0
From the clarification on page 12 of the official FFG clarification document, it states that brain chip will not reduce base MU and hand size.  Currently, if you have negative points, you will have 3 MU and a hand size of 1 less than it should be.

Saintis:
Resolves #1040.

Prevents negative effect of Brain Chip. Also adds test.


mtgred:
:+1:

PR #1051, opened 2015-12-28
Fix and test for triggering Posted Bounty with 24/7 News Cycle, #1043.

### LINKED ISSUE #1043: 24/7 having difficulty triggering Posted Bounty
Using 24/7 and then selecting posted bounty for use leads to the prompt getting stuck on the menu to choose whether or not to forfiet posted bounty.

Sometimes clicking on it, it will stay on the prompt but give the tag and the Bad pub...it also sometimes causes other wierd glitches

For instance 
1)used 24/7 and selected the agenda to forfiet to it
2) selected posted bounty
3) clicked yes to forfeit posted bounty (the menu then gets "stuck" at this point) 
4) the tag and bad pub are given (menu still stuck)
5) used scorch earth, target prevents 4 meat
6) used scorch earth again (for the kill) and the plastcrete somehow auto prevented

nealterrell:
Fixes #1043.

### DIFF
--- a/src/clj/game/cards-operations.clj
+++ b/src/clj/game/cards-operations.clj
@@ -10,7 +10,7 @@
                                            (= (first (:zone %)) :scored)
                                            (:abilities %))}
                       :msg (msg "trigger the \"when scored\" ability of " (:title target))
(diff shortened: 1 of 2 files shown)


mtgred:
:+1:

PR #1052, opened 2015-12-28
Fixes Chronos Protocol interaction with I've Had Worse.

### LINKED ISSUE #1038: Chronos vs IHW interaction
someone picked to net-damage i've had worse vs me, as the chronos ID. The ability then didn't fire - I didn't draw cards automatically.

Saintis:
Fixes #1038.


mtgred:
:+1:

PR #1055, opened 2015-12-29
Card fixes: Trope, Nisei Mk II, IT Department, Crisium Grid

### LINKED ISSUE #941: Crisium Grid effect lasting through subsequent runs, even after trashing
Crisium Grid, if activated once, seems to assume that no further runs on that server for the entire turn are successful, even if it is trashed and a subsequent successful run is made on the server.

An example from an actual game. Playing Chaos Theory, have removed a Hyperdriver from the game, have three Notoriety in hand that I plan to play with my last 3 clicks.

* Click 1: Run HQ with Legwork. Break the ice. Facedown upgrade is rezzed, it's Crisium Grid, Legwork fizzles but I access 1 card from hand normally. I then access and trash the Crisium.
* Click 2: Run HQ normally. Break the ice. Access 1 card from hand.
* Click 3: Run R&D with Maker's Eye. Break the ice. See three cards from R&D.
* Click 4: Run Archives. See nothing of consequence.
* Click 5: Try to play Notoriety. It cannot be played from hand, presumably because the Crisium run on HQ has created a game state where my subsequent successful run on HQ didn't count.

### LINKED ISSUE #1036: A bug of IT department's "add strength to a rezzed ICE"
When "add strength to a rezzed ICE" is used, the "Select a target for IT Department" shows up but nothing can be done. You can not either add strength or close the window by pressing the "Done" button. As the window does not disappear, you can not proceed the game.

JoelCFC25:
* Fixes #1036: IT Department was calling `ice-strength-bonus` but passing too few arguments. 
* Fixes #941: Crisium Grid was removing the wrong end of the vector of successfully run servers
* Fixes Trope to restore 1 MU to the Runner when it is removed from the game
* Adds a requirement that a run exist before spending a Nisei Mk II counter to end a run (mostly to prevent accidentally losing one from a misclick). 
* Updates Draco's strength immediately after choosing the number of counters

### DIFF
--- a/src/clj/game/cards-agendas.clj
+++ b/src/clj/game/cards-agendas.clj
@@ -282,7 +282,9 @@
     :advancement-cost-bonus (req (:bad-publicity corp))}
 
    "Nisei MK II"
-   {:data {:counter 1} :abilities [{:counter-cost 1 :msg "end the run" :effect (effect (end-run))}]}
+   {:data {:counter 1}
+    :abilities [{:req (req (:run @state)) :counter-cost 1 :msg "end the run"
+                 :effect (effect (end-run))}]}
 
    "Oaktown Renovation"
    {:install-state :face-up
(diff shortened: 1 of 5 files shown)


mtgred:
:+1:

Issue #204, opened 2015-05-22
Invasion of Privacy removes cards


McBEASTse:
Corp used Invasion of Privacy with Trace 5. In my hand was 3 events and 2 programs. All cards where removed (not in any space on screen) including programs.

mtgred:
Don't worry it's normal. Invasion of Privacy moves all the card in the Runner's hand to a zone visible only to the Corp. It's then up to the Corp to drag and drop cards back to either the discard pile or the hand of the Runner.

The card are not visible to opponent because it's the same mechanism used for Indexing or Precognition.

I will change the code to show only the back of the cards in that special zone to the opponent to avoid confusion.

PR #1058, opened 2015-12-30
Play-Area changes.

Saintis:
Makes the `:play-area` visible (with card backs) to players and spectators. Labels it as Temporary Zone.

Changes how a few cards (mainly Accelerated Diagnostics and Accelerated Beta Test) works in regards to the `:play-area`, and updates references to "the play area" to "the Temporary Zone".

Intends to alleviate #204.

Note that the way ABT and AD have been changed means that cards not installed / played are left in the Temporary Zone and have to be manually trashed / moved to archives.


mtgred:
:+1:

Issue #956, opened 2015-12-07
Mulligan should be corp first! 


Zach-ER:
Could you please change it so that it does not reveal the runner's mulligan decision until after the corp has chosen? That would be much appreciated!

nealterrell:
It's a low priority item, but it's on the list. As runner, you should just wait until the Corp reveals their decision, as you would when playing in person.

zaroth:
Connected to #329

nealterrell:
Fix submitted in #1059.

JoelCFC25:
Deployed!

Issue #907, opened 2015-11-23
Chronos Protocol Trash Menu Stuck


anyothername:
Ran into an issue with Chronos Protocol's ID menu getting stuck. The board state was I had a Hostile Infrastructure rezzed. The runner used Maker's Eye to run R&D, leaving four cards in hand. They hit a Snare, which I fired. Before I could choose the card to discard with Snare and Chronos' ability, they trashed it. This made another net damage fire because of Hostile and trash a card from their hand before I could use Snare's ability. After that, none of the button's on my menu worked. The runner tried adding back the cards to their hand, but I still couldn't trash anything to get out of the menu.

JoelCFC25:
Unfortunately there's nothing that can be done about this right now until there is a global solution for blocking Runner trashing until the Corp has concluded all optional abilities or made all the decisions generated by any given triggering event. The only way to get the correct outcome from this scenario is for the Runner to hold off on trashing the Snare! until the Corp is done--but there's no way to know in advance that you need to do this unless you've seen the problem crop up before. 

See #886 for a bit more discussion on this issue.

anyothername:
Got it, thanks so much! Just played out a second Chronos game and it went fine with me asking the Runner to hold off.

nealterrell:
Fix submitted in #1059.

JoelCFC25:
Deployed today!

Issue #576, opened 2015-08-21
Snare! doesn't wait for the corp to resolve before the runner can continue acting


thehunter518:
This is certainly not a game-breaking issue, especially for patient and polite players, but the corp's prompt upon the runner accessing Snare! (Pay 4 credits to deal 3 net damage and force the runner to take one tag?) doesn't actually wait for the corp's response - the runner can continue taking actions while you are deciding what to do. For example, if you have a Snare! on a remote server and a Dedicated Response Team rezzed, the runner can trash the Snare!, complete their run, and end their turn all before you've decided whether or not you'd like to fire it (this actually directly interferes with Dedicated Response Team working properly).

If this is happening on Snare, it's probably happening on other traps as well (although I'm pretty sure that Psi games DO correctly make the runner wait until you've revolved them).

JoelCFC25:
Yep, this is broadly problematic for lots of card interactions. See the duplicate of this at #423, but also #445, #309, and even #189 (sort of).

nealterrell:
Fix submitted in #1059.

Issue #423, opened 2015-07-23
Snare and Dedicated Response Team interaction issue


nealterrell:
Dedicated Response Team will not apply meat damage to a Runner who access Snare! if the Runner selects his option to trash Snare! for 0 credits before the Corp chooses his option to trigger Snare's tag/damage for 4 credits. This happens because a Run "ends" once there are no more prompts for the Runner to resolve; the Corp's prompts are not considered. 

I don't know if there would be side effects to requiring `[:corp :prompt]` to be empty for a run to end in `handle-end-run`, but it would potentially solve this issue. (It might make for a confusing UI, however, as the Runner will see his action area return to the default state, possibly taking a new action while the Corp is slowly deciding to trigger the Snare.

nealterrell:
Fix submitted in #1059.

PR #1059, opened 2015-12-30
Add "waiting on..." prompts to fix some card interactions

nealterrell:
This introduces the ability to show a "Waiting for ..." prompt to a user at the front of their prompt queue, preventing them from interacting with other prompts until some other card or function clears the waiting prompt. This is intended to fix several tricky card interactions like #576, #423, #907, and #289; plus some general game flow issues like #956, #329, and unreported issues with psi games and traces.

Example: Runner accesses a Snare! with a Dedicated Response Team rezzed. Corp sees the "Pay 4cr to use Snare?" prompt while the Runner simultaneously sees "Pay 0cr to trash Snare?". If the runner finishes his prompt first, then the run will be finished and "successful run ends" events will trigger. Dedicated Response Team will see that event but won't damage the runner because they are not tagged (yet). Corp then chooses to use Snare! to give the tag and damage, but it's too late for DRT to fire. This happens because the runner's ability to trash (and thus end the run) is allowed before the Corp has decided to trigger the Snare. Similar issues occur with Manhunt (the trace dialogs are still active when the runner accesses) and Ghost Branch... basically, any ability that needs to be resolved with a prompt by one player, but also needs to block further actions by the other player.

`show-wait-prompt` is a new function to add a "Waiting for..." prompt to a player's queue. The prompt has no buttons, as shown here:

image

The only way to dismiss the prompt is with `clear-wait-prompt` as part of a card or engine function. 

I have used this feature in the following ways:

* Mulligans: Runner sees "waiting for Corp to keep hand or mulligan" until that happens, then vice versa. Game then can start. (These are only shown if two players are in the game.) I moved the Keep/Mulligan stuff out of the UI (where it was hard-coded) and reimplemented it as prompts in the game engine.
* Traces: when a trace triggers, the Runner sees "waiting for Corp to initiate trace", preventing other actions like accessing. After initiating, the Corp sees "waiting for Runner to boost Link strength".
* Psi games: whomever bids first sees "[opponent] is choosing psi credits" until that is done. Prevents Runner from starting to access a second Future Perfect in Archives while a first FP's psi game is resolving.
* Snare! and Ghost Branch: because these are optional, they will block the Runner from acting until the Corp decides to trigger, which fixes the DRT issue outlined above.
* Chronos Protocol: runner sees "waiting for Corp to use Chronos Protocol" until the CP dialogs are complete. 

### Prompt priorities

I expanded (and documented) the `:priority` prompt system so that it's no longer binary. Priorities are now numerical, and prompts are inserted into the prompt queue in decreasing priority order. If two prompts have the same priority, then whichever was inserted first will remain first in the sort. `:priority true` is interpreted as `:priority 1`, and no priority is `:priority 0`. `show-wait-prompt` will by default insert a priority 1 wait prompt, but takes an optional parameter to change that priority as needed (Manhunt required trace prompts to be priority 2).

FINALLY, I added a `/close-prompt` chat command to remove the currently active prompt, just in case


mtgred:
There is a merge conflict. Can you have a look at it?

nealterrell:
Done!

nealterrell:
Extended functionality to Edge of World, with test.

mtgred:
:+1:

PR #1062, opened 2016-01-01
Allow Parasite to keep gaining counters on Architect; more card tests

### LINKED ISSUE #1057: Film critic message issue with agenda points
> Runner spends to use Film Critic to add Global Food Initiative to their score area and gain 3 agenda points.

The agenda points are added correctly, it's just log message that doesn't use the correct value.

JoelCFC25:
* Fixes #1057. Film Critic's message will report the correct number of agenda points gained if adding a hosted Global Food Initiative to the score area. 
* Modifies Parasite to check if the host ICE is untrashable before unregistering events and trying to trash it--this will allow Parasite to keep gaining counters and make Architect's strength go negative instead of stopping at 0. Includes tests to make sure standard Parasite ICE trashing and instant ICE trashing (with a loaded Hivemind) continue to work.
* Adds missing ability label for Panic Button
* Finally, another big batch of individual card unit tests. 

Happy New Year!

### DIFF
--- a/src/clj/game/cards-programs.clj
+++ b/src/clj/game/cards-programs.clj
@@ -433,7 +433,9 @@
              {:req (req (= (:cid target) (:cid (:host card))))
               :effect (effect (ice-strength-bonus (- (get-virus-counters state side card)) target))}
              :ice-strength-changed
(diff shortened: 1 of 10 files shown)


mtgred:
:+1:

PR #1066, opened 2016-01-03
Implement gene conditioning shoppe

proland:
Implemented Gene Conditioning Shoppe for the current set of Genetics cards.

### DIFF
--- a/src/clj/game/cards-resources.clj
+++ b/src/clj/game/cards-resources.clj
@@ -204,6 +204,13 @@
                                 (resolve-ability state :runner (choose-access c '(:hq)) card nil)
                                 (update! state side (dissoc (get-card state card) :access-hq))))}]}
 
+   "Gene Conditioning Shoppe"
+   {:msg "make Genetics trigger a second time each turn"
+    :effect (effect (register-persistent-flag! card :triggers-twice
+                                               (fn [state side card]
+                                                 (has? card :subtype "Genetics"))))
+    :leave-play (effect (clear-persistent-flag! card :triggers-twice))}
+
    "Ghost Runner"
    {:data {:counter 3}
     :abilities [{:counter-cost 1 :msg "gain 1 [Credits]" :req (req (:run @state))
--- a/src/clj/game/cards-resources.clj
+++ b/src/clj/game/cards-resources.clj
@@ -15,8 +15,10 @@
 
    "Adjusted Chronotype"
    {:events {:runner-loss {:req (req (and (some #{:click} target)
-                                           (empty? (filter #(= :click %)
-                                                           (mapcat first (turn-events state side :runner-loss))))))
+                                          (let [click-losses (filter #(= :click %) (mapcat first (turn-events state side :runner-loss)))]
+                                            (or (empty? click-losses)
+                                                (and (= (count click-losses) 1)
+                                                     (persistent-flag? state side card :triggers-twice))))))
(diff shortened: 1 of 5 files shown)


nealterrell:
Nice job. My instinct wants to avoid coupling the implementations of these cards so tightly together, but I also can't think of any other way of implementing this, so we'll roll with it.

Two suggestions for `installed-byname`:

1. Reorder the parameters to `state side title`; all core functions take `state` and `side` first, which allows the `effect` macro to auto-insert those parameters for free.
2. Small rework to the logic: it's perfectly effective now, but we can make the function more flexible by taking advantage of Clojure's "truthy" values: we don't have to return the literal `true` if we instead return anything that is not `nil` or `false`. `some` helps here:

    ```clojure
(some #(when (= (:title %) title) %) (all-installed state side))
```

    which will return the first card with the given title if there is one, or `nil` otherwise. That card value is truthy, so you still get to use `installed-byname` how you currently are using it... but the return value is now more useful (being the card itself). 
    And while you're at it... maybe rework `in-play?` in `core-misc.clj` to use your new function, since it's doing the same logic except using a card map instead of a string title. (For the record, the comment description of that function is wrong... it is returning `true` instead of a truthy card map.)

Finally, if you are interested in joining our developers Slack channel, shoot an email to mtgred@gmail.com for an invitation!

proland:
Thanks for the feedback.

I had also looked at implementing GCS with an effect that sets a flag `:genetics-trigger-twice` on the runner's state (similar to how Starlight Crusade and Doubles are implemented) But encountered a few issues:

1. `:register` seems to be the place for these sort of flags, but gets wiped out every turn. We could add an effect to GCS to set it at the start of each turn, but this could lead to a race with Adjusted Chronotype?
2. If there were another card that provided this effect, un-installing one of them would wipe the flag even if the other were still installed.

I will work on incorporating your suggestions.

nealterrell:
We have a better framework for flags in `core-flags.clj` (see `run-flag?` and `turn-flag?`, used for Blackmail and DDoS, respectively) that could be expanded to a `constant-flag?` or similar name. Up to you.

I just verified that `in-play?` is completely broken and that allows the Runner to install multiple copies of a unique card. Reworking that function would add #1071 to your victory sheet :).

proland:
Ok, I'll take a crack at adding a `persistent-flag?` in line with the functionality of the other two.

The changes to `in-play?` in [50690a5](https://github.com/proland/netrunner/commit/50690a51160114b654c8ccc367035292a0467ef6) should fix it by wrapping `(:side card)` with `(to-keyword)`

proland:
I think I'm pretty happy with this now.

nealterrell:
Looking good! Great first contribution :).

FYI (you don't need to fix this unless you want to), the `req` and `effect` macros declare implicit `do` blocks, so you don't need to `(req (do ...`)). 

@mtgred : provided there are no conflicts with any other PRs, this one is good to go.

JoelCFC25:
My only comment is about the `:msg` on Gene Conditioning Shoppe. As currently written it will output:

RunnerNameHere uses Gene Conditioning Shoppe to Genetics trigger twice

Usually `:msg` mimics the language of the card pretty closely, so an alternative could be: 

`:msg "make Genetics trigger two times each turn"`

proland:
@JoelCFC25 Good point. Fixed.

mtgred:
:+1:

PR #1070, opened 2016-01-04
Markdown support for news items

nealterrell:
Adds Markdown and Netrunner icon support to news items via Trello. The client does all the rendering of Markdown into HTML, so no server resources are used. We can also use our Netrunner icons like [Credit], [Click], [Link], etc.


mtgred:
:+1:

PR #1076, opened 2016-01-07
Fix for not affording optional abilities

nealterrell:
If a player says Yes to resolve an optional ability that has a cost but can't afford that cost, then we get this weird behavior where neither the `:yes-ability` nor the `:no-ability` of the optional gets resolved. This makes the "waiting for..." behavior introduced to Snare! and Ghost Branch difficult to implement. 

This core change will only trigger an optional's `:yes-ability` if the player can afford to pay its costs. Otherwise, the `:no-ability` will be triggered. Summary:

* If the player selected "Yes", and there is a `:yes-ability`, and the `:cost` of that ability can be paid, then trigger `:yes-ability`
* Otherwise, trigger `:no-ability` if there is one.

### DIFF
--- a/src/clj/game/core-abilities.clj
+++ b/src/clj/game/core-abilities.clj
@@ -128,11 +128,11 @@
 (defn optional-ability
   "Shows a 'Yes/No' prompt and resolves the given ability if Yes is chosen."
   [state side card msg ability targets]
-  (show-prompt state side card msg ["Yes" "No"] #(if (= % "Yes")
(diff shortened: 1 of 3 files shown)


nealterrell:
Added a fix for a chat-reported issue whereby simultaneous triggering Marcus Batty and Security Nexus could cause an unresolvable prompt sequence. (Corp sees "Waiting for Runner to play the psi game", Runner sees "Waiting for Corp to initiate trace".)

mtgred:
:+1:

PR #1077, opened 2016-01-07
Code Prettifying

Saintis:
Major attempt to make use of some (hopefully) more readable functions and idioms.

Does the following:
- Adds and makes use of `in-hand?` `is-type?` `has-subtype?`
- Makes use of `installed?` and `ice?` more extensive
- Splits several `and` and `or` expressions into several lines for readability
- Changes occurrences of `(if (not x))` to `(if-not x)`, same for `when`
- Makes use of `when` instead of `if` when relevant
- Makes use of `pos?` and `neg?`
- Refactors `resolve-ability` into several helper functions
- Removes unused `side-str` function
- Adds some docstrings
- Creates `card-name` that replaces `(if (:rezzed card) (:title card) "a card")` found throughout

Probably needs a lot of review as there are a lot of changes :smile:

### DIFF
--- a/src/clj/game/cards-resources.clj
+++ b/src/clj/game/cards-resources.clj
@@ -77,7 +77,8 @@
 
    "Compromised Employee"
    {:recurring 1
-    :events {:rez {:req (req (= (:type target) "ICE")) :msg "gain 1 [Credits]"
+    :events {:rez {:req (req (ice? target))
+                   :msg "gain 1 [Credits]"
                    :effect (effect (gain :runner :credit 1))}}}
 
    "Crash Space"
@@ -113,7 +114,7 @@
                            (register-turn-flag!
                              card :can-rez
                              (fn [state side card]
-                               (if (and (has? card :type "ICE")
+                               (if (and (ice? card)
                                         (= (count (get-in @state [:run :ices])) (get-in @state [:run :position])))
                                  ((constantly false) (toast state :corp "Cannot rez any outermost ICE due to DDoS." "warning"))
                                  true)))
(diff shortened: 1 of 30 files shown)


src/clj/game/cards-resources.clj
```diff
@@ -151,9 +152,9 @@
    "Eden Shard"
    {:abilities [{:effect (effect (trash card {:cause :ability-cost}) (draw :corp 2))
                  :msg "force the Corp to draw 2 cards"}]
-    :install-cost-bonus (req (if (and run (= (:server run) [:rd]) (= 0 (:position run)))
+    :install-cost-bonus (req (if (and run (= (:server run) [:rd]) (zero? (:position run)))
```
zaroth: I take it the `(and run` part takes care about this `zero?` having `nil` issues you talked about?

src/clj/game/core-abilities.clj
```diff
@@ -135,7 +209,7 @@
                                                     (resolve-ability state side no-ability card targets)))))
 
 
-; Prompts
+;;; Prompts
```
zaroth: so now we're going triple semicolon even? ;-)

src/clj/game/core-actions.clj
```diff
@@ -57,27 +57,31 @@
...
+                                 (= last-zone :hand)
+                                 (= server "Grip")))
+                       (or (and (= (:side c)  "Runner")
+                                (not (:facedown c)))
+                           (rezzed? c)
                            (:seen c)
                            (= last-zone :deck)))
-                (:title c) "a card")
+                (:title c)
+                "a card")
```
zaroth: the whole naming in here just asks to be replaced by an appropriate `card-str` call, maybe modified slightly. Ideally all card naming, if not very specialized for some reason, could be handled in that one place.

zaroth:
Great job! :+1: Had a few tiny concerns/comments, added them in line.

src/clj/game/cards-resources.clj
Saintis: Didn't realise the `nil` issues until after this I think. Will probably revert this change to be sure.

src/clj/game/core-actions.clj
nealterrell: :+1: to what @zaroth said about `card-str`.

Saintis:
Have fixed up the `card-str` stuff. Is ready for merge.

mtgred:

Saintis:
Should work now (may have the test failure issue solved in #1104)

Saintis:
Rebased changes on the fixes in #1104.

mtgred:

Saintis:
There will always be merge conflicts :wink:
Fixed now.

mtgred:
:+1:

Issue #955, opened 2015-12-07
Nerve Agent does not get virus counters when access is through card abilities


Merlin8000:
When making runs through sneakdoor beta with Nerve Agent installed (hosted on Djinn), the sneakdoor runs did not add virus counters to Nerve Agent as they should have.

JoelCFC25:
This one is caught up in the hairball of how the system treats steps 4.1 through 4.5 of runs right now. Because there is no way to make code execution "pause" at 4.4 and let "when successful" effects fully resolve before doing 4.5, the only way to make Nerve Agent (and Medium) allow the choice of fewer than all allowable accesses is to give the Runner the ability after the Corp has clicked "No more action" for the final time (and even that shouldn't technically be the final time, see #334). 

I think if I change line 345 of Nerve Agent (the `:req` for `:successful-run`) I can get it to add counters correctly, but it will never be able to let you choose fewer than the max. So, a partial fix but not a full one.

JoelCFC25:
I should add that the new counter adjustment console command is a workaround for this--right before you hit Successful Run while at the doorstep of Archives on your Sneakdoor run, type `/counter n` into chat, where n is the number of Nerve Agent counters you should end up with.

PR #1085, opened 2016-01-09
Added chameleon/london library tests

kevkcc:
I noticed Utopia Shard could not be installed through a Sneakdoor Beta run without accessing so I added a bit of code to allow this. I'm not 100% sure if this is the way it should be approached but as this is the only card that allows you to do this, I figure an extra check is simple enough to do. Comments are welcome.

Added Chameleon unit tests for my last commit as Joel requested. Also did one for London Library since it had issues with Chameleon in the past. Added one for Utopia Shard as well. Fixed a small typo in London Library print message as well.


zaroth:
Naming cards like this is usually awkward, but this case seems specific enough. Looks good to me :+1:

zaroth:
... But could get better ;-) Just read this in chat on jinteki.net:

> btw played a game yesterday and seems like nerve agent doesnt work with sneakdoor

Could you maybe try cooking up a more general solution which would also solve this case?

JoelCFC25:
See #955 for that one.

kevkcc:
Thanks for the feedback. I definitely agree that tying specific cards together is not ideal hence why I was unsure about my approach in the first place. Luckily I've since come up with what I believe to be a better solution that doesn't require changes to Utopia Shard at all (involves changing up the run timing on Sneakdoor Beta) so I have reverted that particular change.

Regarding Nerve Agent + Sneakdoor, I've thought about it and can pinpoint exactly where the problem is but cannot come up with a good solution for now. To my knowledge, #955 is not entirely accurate - Nerve agent works with run events like Dirty Laundry and Legwork, it's only because of the switch in the run from Archives to HQ on Sneakdoor that causes problems. But I'll keep thinking about this one.

mtgred:
There are some merge conflicts. Can you have a look?

kevkcc:
Fixed the merge conflicts. Unfortunately I won't have time to investigate Nerve Agent too much for next couple of weeks but for now Utopia and possibly some other weird Sneakdoor edge cases now work.

kevkcc:
Apologies but I thought of a new scenario last night and my code does not work with Crisium Grid. I have reverted the change because I won't have too much time in the coming weeks to investigate. But the tests I wrote for chameleon/london library should be fine.

zaroth:
More tests are always welcome. Looks good :+1:

mtgred:
:+1:

PR #1095, opened 2016-01-11
Card fixes: Parasite, Blue Sun, Chronos Protocol, DaVinci, Beale/Sleepers, Hunting Grounds

### LINKED ISSUE #1064: Fall Guy attempts to prevent Chronos Protocol Damage
I don't know if it's because the card is revealed and thus is visible to Fall Guy, but when applying targeted Net Damage with CP, it gave the option to prevent the trash using Fall Guy.

### LINKED ISSUE #1065: Log the Hunting Grounds trash effect in the chat.
If someone plays [Hunted Grounds](http://netrunnerdb.com/en/card/09035), the log says that three cards are installed facedown, but it doesn't explicitly say what card caused the effect in the chat, or that Hunting Grounds was trashed.

This might also occur for other cards that are trashed and cause effects, I'm not sure.

### LINKED ISSUE #1090: Davinci does not Proc Geist's Ability
Davinci (107 sansan) does not automatically draw geist a card upon use

### LINKED ISSUE #1094: Typo in blue sun ability text
When the corp uses Blue Sun's ability to add an ice back to HQ. Chat box says:
USERNAME add ICENAME to HQ
should be:
USERNAME adds ICENAME to HQ

JoelCFC25:
* Fix #1064: Make the specified card trashed by Chronos Protocol unpreventable so Fall Guy/Sacrificial Construct won't kick in to try and prevent cards lost from the Runner's Grip
* Fix #1065: Add a message when Hunting Grounds trash ability occurs so it doesn't happen silently.
* Fix #1090: Make DaVinci's self-trash trigger Geist
* Fix #1094: Change the tense of Blue Sun's log message
* Add `trash-ice-in-run` to Parasite so run position is automatically updated when ICE is destroyed--this will fix bugged runs that start with an immediate Ice Carver + Parasite instant ICE kill.
* Specify Runner agenda points of Project Beale and Domestic Sleepers so they will grant the Runner the correct number of points if ever Turntabled when they're worth increased points in the Corp score area.

### DIFF
--- a/src/clj/game/cards-identities.clj
+++ b/src/clj/game/cards-identities.clj
@@ -43,7 +43,7 @@
                               (let [cost (rez-cost state side target)]
                                 (gain state side :credit cost)
                                 (move state side target :hand)
-                                (system-msg state side (str "add " (:title target) " to HQ and gain " cost " [Credits]"))
+                                (system-msg state side (str "adds " (:title target) " to HQ and gains " cost " [Credits]"))
                                 (swap! state update-in [:bonus] dissoc :cost)))}]}
 
    "Cerebral Imaging: Infinite Frontiers"
--- a/src/clj/game/cards-identities.clj
+++ b/src/clj/game/cards-identities.clj
@@ -74,7 +74,7 @@
                                                                 (- (get-defer-damage state side :net nil) 1)
                                                                 " more net damage")))
                                                   :effect (effect (clear-wait-prompt :runner)
-                                                                  (trash target {:cause :net})
+                                                                  (trash target {:cause :net :unpreventable true})
                                                                   (damage :net (- (get-defer-damage state side :net nil) 1)
                                                                           {:unpreventable true :card card}))}
                                     :no-ability {:effect (effect (clear-wait-prompt :runner)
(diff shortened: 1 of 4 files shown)


src/clj/game/cards-identities.clj
```diff
@@ -74,7 +74,7 @@
                                                                 (- (get-defer-damage state side :net nil) 1)
                                                                 " more net damage")))
                                                   :effect (effect (clear-wait-prompt :runner)
-                                                                  (trash target {:cause :net})
+                                                                  (trash target {:cause :net :unpreventable true})
```
zaroth: Curious about this. Won't this make Chronos's net damage unpreventable too? Or does this part kick in already after the net damage prevention part?

zaroth:
Nice catch with Beale/Sleepers! :+1:

src/clj/game/cards-identities.clj
JoelCFC25: No, this part occurs after the Runner is asked about prevention. All the `damage-defer` and `get-defer-damage` stuff was put in just to make Chronos Protocol work. Although there is still the bigger problem that Chronos Protocol is still kicking in after the Runner performs prevention. :/

mtgred:
:+1:

PR #1100, opened 2016-01-13
Implementation and tests for two draft identities.

nealterrell:
Implement Wyvern: Chemically Enhanced and Strategic Innovations: Future Forward draft identities. Have to restore the `:faction` key to cards for these draft IDs to work.


mtgred:
:+1:

PR #1103, opened 2016-01-13
Allow rez to ignore just the rez cost, and not only all costs.

nealterrell:
Muertos Gang Member is the only card that allows something to be rezzed, ignoring just the _rez_ cost. Refactor `rez` to specify more exactly which costs should be ignored, if any. Muertos uses `:ignore-cost :rez-cost`, Oversight AI uses `:ignore-cost :all-costs`.


mtgred:
This seems to break tests according to Circle CI. Can you have a look at it?

nealterrell:
I have been talking to Saintis about it. He made a change in one PR that broke a test I wrote in another PR. He thought his Toastr fix PR would fix the issue, but it is currently breaking a different test.

PR #1104, opened 2016-01-13
Toastr fix.

Saintis:
Placed libs in wrong place. This fixes that. Run `bower install` after merge to get libs properly.


Saintis:
Also fixes the test failure for Snare!

Saintis:
Fixes persisten toasting for econ assets as well.

Saintis:
@mtgred This should fix the test failures on `master`, and optional ability error.

mtgred:
:+1:

PR #1109, opened 2016-01-14
Improvement to games list bandwidth.

nealterrell:
Here is my final idea for a large reduction in network traffic. Every change to a game (create a game, join a game, leave a game, swap sides, join as spectator, leave as spectator) causes the entire games list to be re-sent to all site visitors. We now collect a map of games that have changed since the last lobby update (still on 1 second intervals), and only send those games. A client receives the entire games list when first connecting to the server, and from then only receive game list diffs. 

I don't have many numbers to predict the traffic reduction, but it should be significant. Currently, a single person making a new game causes 40-100Kb of data to be sent to every single connected visitor. With this change, the new game will send 0.5-20Kb (?) (depending on the decks selected). Likewise for every time someone joins a game, swaps sides in a game, leaves a game. etc.


justinliew:
/minor - I'd have preferred this in 2 CLs, as we may want to take the memory leak change immediately.

nealterrell:
I can do that.

mtgred:
:+1:

PR #1110, opened 2016-01-14
Remove closed games from state-diff support structure.

nealterrell:
I believe I found the source of the server memory leak. When a game is over, the server correctly removes its state from memory. When I added state diffs, however, I introduced a second copy of the state -- used to calculate diffs -- and forgot to remove that structure on game end. So over the course of a week, with thousands of games played and closed, we are accumulating thousands of game state structures... yeah. One line fix.

### DIFF
--- a/src/clj/game/main.clj
+++ b/src/clj/game/main.clj
@@ -78,7 +78,8 @@
       (try
         (case action
           "start" (core/init-game msg)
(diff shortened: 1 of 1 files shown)


nealterrell:
The failing test is fixed in #1104, which I did not bother to merge into this branch.

mtgred:
:+1:

PR #1113, opened 2016-01-15
Fix for deckbuilder duplicate cards

### LINKED ISSUE #1112: Deckbuilder allows you to use 6 copies of a card and still shows as tournament valid
if you type in the cards and use a different number on each line the deck builder doesn't pick up on the duplicated cards, like on this totally tournament legal deck: https://dl.dropboxusercontent.com/u/168938/tournament-valid.png

zaroth:
One line fixed #1112. Forgot parentheses in one place essentially.


mtgred:
:+1:

Issue #1123, opened 2016-01-19
Game Data // Re-adding to new server


domtancredi:
Will have a PR for this soon

domtancredi:
Should be fixed in this PR: https://github.com/mtgred/netrunner/pull/1126

PR #1126, opened 2016-01-19
Save games to off-site database instance

domtancredi:
Will need environment variables set to the new database

### DIFF
--- a/server.coffee
+++ b/server.coffee
@@ -23,7 +23,9 @@ cache = require('memory-cache')
 # MongoDB connection
 appName = 'netrunner'
 mongoUrl = process.env['MONGO_URL'] || "mongodb://127.0.0.1:27017/netrunner"
+mongoGameDBUrl = process.env['MONGO_GAMEDB_URL'] || "mongodb://127.0.0.1:27017/netrunner"
 db = mongoskin.db(mongoUrl)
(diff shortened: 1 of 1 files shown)


nealterrell:
@mtgred this PR is to turn game statistics back on by offloading the data to an Azure instance that @justinliew and @domtancredi collaborated on. They can provide more details about the server in Slack, including the connection strings.

mtgred:
The stats code need some rework. Saving the full game states not only take a lot of disk space, it also make the aggregation queries run forever.

I'm currently learning Datomic and will add code to save summary game data in Datomic soon. The really cool thing about Datomic is that it has time travel built in. It's like using Git instead of only keeping the latest version of the code. The time travel will allow to analyse game stats overt time.

domtancredi:
That's very interesting and I'm reading up on Datomic. Very elegant separation of concerns model for data lifecycle. I like the idea of "facts" and allowing to persist that state of things via them. 

Should this branch stay alive? Is it of value? It'd be really great to store game data right now, at least, so we can present stats over Feb // March, etc. until Datomic is implemented.

What's the timetable for bringing in Datomic?

zaroth:
Connected to #1123.

nealterrell:
Closing this now, since we no longer have the coffee server.

Issue #1131, opened 2016-01-20
Cybernetics Court and Cerebral Imaging not playing nicely together


jmbostwick:
Hey there -- I've notice in multiple games that these two cards (Cybernetics Court and Cerebral Imaging) don't play nicely together, resulting in improperly-calculated maximum hand sizes routinely. The issue seems to be in regards to changing numbers of credits; once credits change, the hand size is recalcluated based on the new number of credits, but consistently does not include the additional bonus from Cybernetics Court.

JoelCFC25:
Thanks for the report--this is essentially the same fundamental issue as #1015. Max hand size adjustments aren't pooled together and applied to the base value.

PR #1133, opened 2016-01-20
Hand Size Split

### LINKED ISSUE #1015: Theo Bag & Brain Damage
Handsize for Theo recalculates every time your credit total changes; but, it doesn't account for Brain Damage and this leads to a lot of players having to reset hand sizes.

Reported by vinegarymink

Saintis:
Splits previous `:max-hand-size` variable into:
- `:hand-size-base` for cards that uses "hand size is" (Bagbiter and Cerebral Imaging)
- `:hand-size-modification` for all other cards that modify the hand size
Should fix #1015 and #1131.

### DIFF
--- a/src/clj/game/core-io.clj
+++ b/src/clj/game/core-io.clj
@@ -99,8 +99,7 @@
         "/tag"        #(swap! %1 assoc-in [%2 :tag] (max 0 value))
         "/bp"         #(swap! %1 assoc-in [%2 :bad-publicity] (max 0 value))
         "/link"       #(swap! %1 assoc-in [%2 :link] (max 0 value))
-        ;; removed until a good implementation has been though of
-        ;; "/handsize"   #(swap! %1 assoc-in [%2 :max-hand-size] (max 0 value))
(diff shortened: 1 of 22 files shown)


src/clj/game/core-io.clj
```diff
@@ -99,7 +99,8 @@
         "/tag"        #(swap! %1 assoc-in [%2 :tag] (max 0 value))
         "/bp"         #(swap! %1 assoc-in [%2 :bad-publicity] (max 0 value))
         "/link"       #(swap! %1 assoc-in [%2 :link] (max 0 value))
-        "/handsize"   #(swap! %1 assoc-in [%2 :max-hand-size] (max 0 value))
+        ;; removed until a good implementation has been though of
+        ;; "/handsize"   #(swap! %1 assoc-in [%2 :max-hand-size] (max 0 value))
```
zaroth: Quote from Slack:
> how about keeping the command syntax like it is and just setting `:hand-size-modification` to `(- value (:base-hand-size %2))`? Would that cause problems?
> of course `value` should be floored by 0 like it currently is

mtgred:
:+1:

PR #1139, opened 2016-01-21
some Kala Ghoda cards

JoelCFC25:
Don't merge this yet, it needs review--@zaroth  and @Saintis agreed to take a look at these. 

Implementations for Advanced Concept Hopper, Kala Ghoda Real TV, Mumbad City Grid, Chatterjee University, and Ramujan-reliant 550 BMI.

### DIFF
--- a/src/clj/game/cards-agendas.clj
+++ b/src/clj/game/cards-agendas.clj
@@ -33,6 +33,27 @@
                                                                              (resolve-ability state side (abt 1 n) card nil))}
                                                                card nil)))}}})
 
+   "Advanced Concept Hopper"
+   {:events
+    {:run
+     {:req (req (first-event state side :run))
+      :effect (effect (show-wait-prompt :runner "Corp to use Advanced Concept Hopper")
+                      (resolve-ability
+                        {:optional
+                         {:player :corp
+                          :prompt "Use Advanced Concept Hopper to draw 1 card or gain 1 [Credits]?" :once :per-turn
+                          :yes-ability {:prompt "Draw 1 card or gain 1 [Credits]?"
+                                        :choices ["Draw 1 card" "Gain 1 [Credits]"]
+                                        :effect (req (if (= target "Gain 1 [Credits]")
+                                                       (do (gain state :corp :credit 1)
+                                                           (system-msg state :corp
+                                                                       (str "uses Advanced Concept Hopper to gain 1 [Credits]")))
+                                                       (do (draw state :corp)
+                                                           (system-msg state :corp
+                                                                       (str "uses Advanced Concept Hopper to draw 1 card"))))
+                                                     (clear-wait-prompt state :runner))}
+                          :no-ability {:effect (effect (clear-wait-prompt :runner))}}} card nil))}}}
+
    "Ancestral Imager"
    {:events {:jack-out {:msg "do 1 net damage" :effect (effect (damage :net 1))}}}
 
--- a/src/clj/game/cards-assets.clj
+++ b/src/clj/game/cards-assets.clj
@@ -297,6 +297,19 @@
                                          (shuffle! state side :deck))}
                            card nil))}]}
 
+   "Kala Ghoda Real TV"
+   {:events
+    {:corp-turn-begins
+     {:optional
+      {:prompt "Use Kala Ghoda Real TV to look at the top card of the Runner's Stack?"
+       :yes-ability {:msg "look at the top card of the Runner's Stack"
+                     :effect (effect (prompt! card (str "The top card of the Runner's Stack is "
+                                                        (:title (first (:deck runner)))) ["OK"] {}))}}}}
+    :abilities [{:label "[Trash]: Trash the top card of the Runner's Stack"
+                 :msg "trash the top card of the Runner's Stack"
+                 :effect (effect (mill :runner)
+                                 (trash card {:cause :ability-cost}))}]}
+
    "Launch Campaign"
    {:effect (effect (add-prop card :counter 6))
     :derezzed-events {:runner-turn-ends corp-rez-toast}
--- a/src/clj/game/cards-hardware.clj
+++ b/src/clj/game/cards-hardware.clj
@@ -401,6 +401,17 @@
                                                      (shuffle! state :runner :deck)))}}} card nil))
     :leave-play (effect (lose :link 1))}
 
+   "Ramujan-reliant 550 BMI"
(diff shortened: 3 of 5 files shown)


src/clj/game/cards-upgrades.clj
```diff
@@ -175,6 +175,33 @@
...
+                                                       snew (assoc target :zone (:zone passed-ice))]
+                                                   (swap! state update-in (cons :corp ice-zone)
+                                                          #(assoc % fndx snew))
+                                                   (swap! state update-in (cons :corp ice-zone)
+                                                          #(assoc % sndx fnew))
+                                                   (update-ice-strength state side fnew)
+                                                   (update-ice-strength state side snew)))} card nil)
+                                 (system-msg state side (str "uses Mumbad City Grid to swap "
+                                                             (if (:rezzed passed-ice) (:title passed-ice) "a piece of ICE")
+                                                             " with "
```
zaroth: there are helper functions `rezzed?` etc. for that. Also, there is `card-str` for naming cards in chatlog, you may find it useful here.

src/clj/game/cards-upgrades.clj
```diff
@@ -175,6 +175,33 @@
...
+                                                             (if (:rezzed target) (:title target) "a piece of ICE")))))}]}
```
zaroth: This line for some reason didn't work when I swapped a rezzed Tollbooth with rezzed 1-advanced Ice Wall on HQ. The message I got:
> test1 uses Mumbad City Grid to swap Tollbooth with a piece of ICE.
Tollbooth was the just-passed ICE.

zaroth: I think I know why that's the case - the message was printed to the chatlog before even prompt selecting the ICE was finished. So the `target` could have just been `nil` at that moment...

src/clj/game/cards-assets.clj
```diff
@@ -297,6 +297,19 @@
                                          (shuffle! state side :deck))}
                            card nil))}]}
 
+   "Kala Ghoda Real TV"
+   {:events
+    {:corp-turn-begins
+     {:optional
+      {:prompt "Use Kala Ghoda Real TV to look at the top card of the Runner's Stack?"
```
zaroth: Two of these bad boys installed trigger a duplicate prompt that doesn't behave too pretty. Since you can't use multiples of KGRT to look at the top, trash it and look again (trashing is in a paid ability window, different from turn start), I think adding some kind of uniqueness to this prompt would help.

src/clj/game/cards-resources.clj
```diff
@@ -70,6 +70,18 @@
...
+   {:abilities [{:cost [:click 1] :label "Place 1 power counter"
+                 :msg "place 1 power counter on it"
+                 :effect (effect (add-prop card :counter 1))}
+                {:cost [:click 1] :label "Install a program from your Grip"
+                 :prompt "Choose a program to install from your Grip"
+                 :choices {:req #(and (is-type? % "Program") (in-hand? %))}
+                 :msg (msg "install " (:title target))
+                 :effect (effect (install-cost-bonus [:credit (* -1 (:counter card))])
+                                 (runner-install target)
+                                 (add-prop card :counter -1))}]}
```
zaroth: Either this or `add-prop` needs some sort of safeguard to not reduce counters below 0, I was able to install Femme Fatale for 10 with -1 counters on Chatterjee: [image](http://i.share.pho.to/4a9d108d_o.png)

src/clj/game/cards-agendas.clj
```diff
@@ -33,6 +33,27 @@
...
+                        {:optional
+                         {:player :corp
+                          :prompt "Use Advanced Concept Hopper to draw 1 card or gain 1 [Credits]?" :once :per-turn
+                          :yes-ability {:prompt "Draw 1 card or gain 1 [Credits]?"
+                                        :choices ["Draw 1 card" "Gain 1 [Credits]"]
+                                        :effect (req (if (= target "Gain 1 [Credits]")
+                                                       (do (gain state :corp :credit 1)
+                                                           (system-msg state :corp
+                                                                       (str "uses Advanced Concept Hopper to gain 1 [Credits]")))
+                                                       (do (draw state :corp)
```
zaroth: The same issue as with KGRT, multiples of this card scored can get multiple prompts in confusing order. Every first run.

Although the multiple prompts do serve a purpose here, opposed to KGRT, so they need to stay - I can't think of a better solution here. Maybe providing a single prompt with options dependent on number of ACH scored, example with 2 ACH:

* Draw 2 cards
* Draw card gain 1 cred
* Gain 2 creds
* Cancel

(I'm aware that this skips the options of just drawing 1 or just gaining 1 cred).

In fact, this may not be better at all, disregard this comment ;-)

src/clj/game/cards-hardware.clj
```diff
@@ -401,6 +401,17 @@
                                                      (shuffle! state :runner :deck)))}}} card nil))
     :leave-play (effect (lose :link 1))}
 
+   "Ramujan-reliant 550 BMI"
+   {:prevent {:damage [:net]}
```
zaroth: it prevents brain damage too.

src/clj/game/cards-hardware.clj
```diff
@@ -401,6 +401,17 @@
...
+    :abilities [{:effect (req (let [n (count (filter #(= (:title %) (:title card)) (all-installed state :runner)))]
+                                (resolve-ability state side
+                                  {:prompt "Choose how much net damage to prevent" :priority true
+                                   :choices {:number (req n)}
+                                   :msg (msg "prevent " target " net damage")
+                                   :effect (effect (damage-prevent :net target)
+                                                   (mill :runner target)
+                                                   (trash card {:cause :ability-cost}))} card nil)))}]}
+
```
zaroth: I'm confused on the rules what happens when there aren't enough cards in stack to trash, can the prevention still occur? Someone else may weigh in here.

src/clj/game/cards-agendas.clj
Saintis: Maybe a prompt for allowing Corp to trigger any ACH on click, with a "Done"  button to cancel action if they want to use fewer than they have scored. Like Team Sponsorship, except with a prompt?

zaroth: Or even better, just condense the current 2 prompts per ACH (use/don't use + card/credit) to just 1 prompt per ACH (card/credit/don't use). :)

PR #1143, opened 2016-01-22
Card fixes; improvements to Chimera, Deep Red, Woman in the Red Dress, Pawn

### LINKED ISSUE #1021: Multiple pawns on the same single ICE protecting a server don't work
In general, Exile Pawn Deep Red archetype needs some love and implementation. Currently the level of automation doesn't make the deck very playable. Example deck: http://netrunnerdb.com/en/decklist/29269/street-tech-chess

### LINKED ISSUE #1072: Multiple hosted pawns crash the game
Hey there!
I was playing some street chess Exile deck, and i noticed (i actually played a lot of games with it, to do a consistent issue report) that when there is 2 or 3 pawns hosted on a single ice and you make a succesful run, the game crashes for several seconds (not lag) and then you cant pass the "Successful Run-Jack Out" option panel, rarely, the game recovers and let you install one pawn, but seems to crash again when the second pawn has to trigger. Single pawn on board works good, the problem only occurs when 2+ are hosted on ICE (i tested it with all pawns in the same ICE).
Thanks for your attention!

PD: Another minor and not important related issue is that the just trashed pawn can return itself to the board while the rules say that this cannot occur.

### LINKED ISSUE #1106: Tollbooth takes credits even if the runner has fewer than 3
Tollbooth reads:

>When the Runner encounters Tollbooth, he or she must pay 3, if able. If the Runner cannot pay 3, end the run.

Expected behavior: If the runner has 3+ credits, firing the toll should take 3 credits. If they have 2 or fewer credits, it should take nothing and end the run.

Actual behavior: If the runner has 2 or fewer credits, Tollbooth sets them to 0 credits but does _not_ end the run.

To reproduce:
- Set your credits to 2 or 1 (I used 1)
- Make a run, encounter a Tollbooth
- Corp uses the "Force runner to lose 3 credits" option

### LINKED ISSUE #1116: Snoop power-counter ability reveals Runner's Grip even with 0 power counters.
To reproduce, install and rez a Snoop. Select the `Hosted power counter: Reveal all cards in Grip and trash 1 card` ability. Choices for the ability appears (the Runner's Grip). Upon selecting an option nothing happens (as there are no power counters on Snoop).

Incorrect reveal:
screen shot 2016-01-16 at 17 33 51

Compared to correct reveal:
screen shot 2016-01-16 at 17 34 35

### LINKED ISSUE #1124: Some  card interactions/bugs that shouldnt be
You can host on djinn already installed programms where this shouldnt be able

you can play cerebral cast ignoring the made a run cost

you can activate house of knives after accessing cards (psychic field ) and kill someone. that's illegal too

### LINKED ISSUE #1129: Datasucker strength lowering propagates to the next ICE after trashing Parasited ICE
Steps to reproduce:
- install 2 ICE on one server, rez both, in observed case outer was Next Silver, inner Ichi 1.0
- Parasite NEXT Silver with 0 counters
- Install Datasucker, get 1 virus counter on it
- run on the protected server, use Datasucker counter to trash NEXT Silver
- Ichi 1.0 now has strength reduced to 3, while it should keep it strength at 4

It would also be very nice to add this case to our Datasucker/Parasite unit tests ;-)

JoelCFC25:
* Fix #1129: Have `trash-ice-in-run` trigger the `:pass-ice` event to stop strength lowering from persisting
* Fix #1124: Adds missing prerequisite for play of Cerebral Cast
* Fix #1116: Require Snoop to have at least 1 counter before doing its look and trash ability
* Fix #1106: Change Tollbooth to use `pay` so Runner can't lose credits if they have only 1 or 2. 
* Fix #1072, fix #1021: Removes automation from Pawn entirely--it just proved too problematic. Users should now advance Pawns one by one, and trashing one will prevent the just-trashed one from being targeted for install.
* Fix Pheromones so unspent credits don't get zeroed out the turn after the Corp purges viruses (per Lukas ruling as seen on BGG)
* Fix Feint to access 0 cards only if the run is successful (fixes interaction with Crisium Grid)
* Add handling of subtype gain/loss and proper end-of-turn derezzing to Chimera
* Add Runner wait prompt and `:yes-ability` message to Woman in the Red Dress
* Add Deep Red ability to fire the click ability of a just-installed Caissa
* Add Runner toast to Bank Job (users frequently ask how to use it)
* Require self-trashing ICE to be currently encountered before they fire `trash-ice-in-run` (prevents a run arrow bug where neither player can end the run)
* Make Hades Fragment manually triggered, avoiding a "Corp is decked" loss being registered with an empty R&D and a scored copy
* Make Sub Boost do an `update-ice-strength` to account for a scored Superior Cyberwalls

### DIFF
--- a/src/clj/game/cards-agendas.clj
+++ b/src/clj/game/cards-agendas.clj
@@ -206,15 +206,13 @@
    {:abilities [{:cost [:click 1] :effect (effect (gain :credit 3)) :msg "gain 3 [Credits]"}]}
 
    "Hades Fragment"
-   {:events {:corp-turn-begins
-             {:optional
-              {:prompt "Add 1 card from Archives to bottom of R&D?"
-               :yes-ability {:effect (effect (resolve-ability
-                                              {:prompt "Choose a card"
-                                               :choices (:discard corp)
-                                               :effect (effect (move target :deck))
-                                               :msg (msg "add " (if (:seen target) (:title target) "a card")
-                                                         " to the bottom of R&D")} card target))}}}}}
+   {:events {:runner-turn-ends
+             {:effect (req (toast state :corp
+                                  (str "Click Hades Fragment to add 1 card from Archives to the bottom of R&D.") "info"))}}
+    :abilities [{:prompt "Choose a card to add to the bottom of R&D"
+                 :choices (req (:discard corp))
+                 :effect (effect (move target :deck))
+                 :msg (msg "add " (if (:seen target) (:title target) "a card") " to the bottom of R&D")}]}
 
    "Helium-3 Deposit"
    {:choices ["0", "1", "2"] :prompt "How many power counters?"
--- a/src/clj/game/cards-events.clj
+++ b/src/clj/game/cards-events.clj
@@ -219,7 +219,11 @@
     :effect (effect (move target :hand) (shuffle! :deck))}
 
    "Feint"
-   {:effect (effect (run :hq nil card) (max-access 0))}
+   {:effect (effect (run :hq nil card) (register-events (:events (card-def card))
+                                                        (assoc card :zone '(:discard))))
+    :events {:successful-run {:msg "access 0 cards"
+                              :effect (effect (max-access 0))}
+             :run-ends {:effect (effect (unregister-events card))}}}
 
    "Fisk Investment Seminar"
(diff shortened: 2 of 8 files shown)


Saintis:
:+1:

mtgred:
:+1:

Issue #1141, opened 2016-01-22
Sundew Problem


cjchristopher:
Sundew gives the corp two credits even when the runner uses their first click to initiate a run on the sundew server.

Observed so far when the run is ended by a piece of ice - not sure if it would still happen if the run were successful and/or the dew was trashed, but in any case, it shouldn't matter - simply initiating the run on the correct server should prevent Sundew from firing.

zaroth:
Connected to #1115 .

cjchristopher:
Seems so, but this was paying out on first click, not second.

mtgred:
Thanks for reporting the issue.

The problem has been solved. The fix will be deployed with the next server update.

zaroth:
There are still problems when a click is gained via Joshua B - Sundew fires on the second click then, not the first one.

JoelCFC25:
And if Wyldside is in play, Sundew won't pay out at all if the Runner does something other than run it with his first click.

mtgred:
This commit fceb19c4eb29050313d59fa05a74fe7cd2ef5eaa should normally fix all cases, including Wyldside and Joshua.

PR #1144, opened 2016-01-22
Fixes Sundew

### LINKED ISSUE #1115: Sundew pays on second click
If the runner spends first click on Sundew then it doesn't pay out (correctly). But if they're unable to trash, when they spend their second click elsewhere it pays out. I *think* that's wrong (but I've read the card so many times now I've no idea what it means any more).

Saintis:
Also allows abilities to turn of the message if the `msg` function returns a falsy value.

Should fix #1115 and #1141.


mtgred:
:+1:

PR #1145, opened 2016-01-23
Implemented Mumbad cycle alliances in deckbuilder

zaroth:
Currently works with all cards, just using a few hacks - ready for pull if we want the functionality now, prone to a bit of refactor in the future, possibly using `defn-` as @saintis suggested.

Comments on where to fix the no `:subtype` part in cards in deckbuilder would be appreciated. I'm talking about the commented out version of `has-alliance-subtype?` function.

For easier review, [prettified influence function](https://www.refheap.com/113975).

### DIFF
--- a/src/cljs/netrunner/deckbuilder.cljs
+++ b/src/cljs/netrunner/deckbuilder.cljs
@@ -174,14 +174,12 @@
         infmap (if (= (:code identity) "03029") ; The Professor: Keeper of Knowledge
                  (profhelper infmap)
                  infmap)
-        ;; dirty hack in place until the :subtype is figured out below
+        ;; checks card ID against list of currently known alliance cards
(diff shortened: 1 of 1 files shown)


zaroth:
After thinking it through, scratch what I said about current implementation of `has-alliance-subtype?` being an ugly hack. The alliance cards have to be listed in code by ID anyway AND `:subtype` has currently no use in deckbuilding outside of alliances. Pulling (in 99% of cases) unnecessary stuff from DB and sending it to the client for every card is probably uglier than hardcoding a few cards. 

Things may change, though, if new cards come out that make use of other subtypes in deckbuilding. Example would be "can't include barriers in deck". Then reworking above alliance function and pulling `:subtype` would be more justified.

mtgred:
Awesome :+1:

Issue #1056, opened 2015-12-29
Game List will sometimes scroll to top


domtancredi:
Please list the browser, operating system, platform, extensions, and any other variables to help replicate this issue.

sjohnc:
I've seen this on Chrome on Windows, Linux, as well as Mac OSX. It generally seems to scroll up when new games are created.

domtancredi:
Can you list steps to recreate the issue on each platform? Is it identical?

prozz:
chrome, mac osx

going down the list and it scrolls to top almost always. never been able to join older games. it may connected with refreshing game state list, not necessarily with new game creation.

ekayohlee:
chrome, mac osx (10.9.5) as well. doesn't happen to me every time, but it happens pretty often.

it jumps to the top when i scroll down and move the mouse over the "Watch" button.

![ick](https://cloud.githubusercontent.com/assets/16576475/12143514/35f6d02c-b44f-11e5-8a10-2c0594649f77.gif)

zaroth:
Arch Linux, Firefox 43.

The only scrolling up that happens for me in the game list is jumping up by 1 position when scrolled to the very bottom. To reproduce this, create a game in another browser while observing the very bottom of scrolled game list in the first one.

Not nearly as annoying as jumping all way to the top, but still could probably be improved upon.

domtancredi:
Interesting. I'm able to replicate it when you refresh the view, hover over the "Watch" button, and (before it turns into a "pointer" with the hand) it jumps to the top. 

I think it has to do with some event triggering on the `.lobby .gameline button` html element. I'll investigate.

If people can confirm, or review trying it. It's kind of interesting. Like, lightly mousing over the "watch" button after scrolling down, and you'll be able to "trigger" the scroll to top behavior.

domtancredi:
Possible solution after quite a bit of research:

http://stackoverflow.com/questions/30365818/react-js-stop-render-from-scrolling-to-top-of-page
http://jsfiddle.net/bengrunfeld/dcfy5xrd/
https://twitter.com/floydophone/status/608129119025561600

Will test "preventDefault()"

PR #1147, opened 2016-01-24
Possible solution for annoying game-list scroll issue

domtancredi:
Been investigating it heavily, and this may be the culprit for Issue #1056


mtgred:
:+1:

PR #1148, opened 2016-01-24
Allow fetch.coffee to only fetch released data packs

domtancredi:
You can run commands:
`coffee data/fetch.coffee dev` to get all sets

`coffee data/fetch.coffee production` or `coffee data/fetch.coffee` (default) to get "available" sets

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -10,7 +10,7 @@ dependencies:
     - sudo /etc/init.d/mongodb start
 
   post:
-    - cd data; coffee fetch.coffee
+    - cd data; coffee fetch.coffee dev
 
 test:
   override:
--- a/data/fetch.coffee
+++ b/data/fetch.coffee
@@ -44,15 +44,28 @@ cardFields = [
   "limited"
 ]
 
+availableSets = []
+
 baseurl = "http://netrunnerdb.com/api/"
 
 selectFields = (fields, objectList) ->
   ((fields.reduce ((newObj, key) -> newObj[key] = obj[key] if typeof(obj[key]) isnt "undefined"; newObj), {}) for obj in objectList)
 
+reduceBySetAvailability = (sets) ->
+  (set for set in sets when set.available isnt "")
+
+reduceByCardAvailability = (cards, sets) ->
+  (card for card in cards when card.setname in sets)
+
 fetchSets = (callback) ->
   request.get baseurl + "sets", (error, response, body) ->
     if !error and response.statusCode is 200
       sets = selectFields(setFields, JSON.parse(body))
+      if process.argv[2] isnt "dev"
+        sets = reduceBySetAvailability(sets)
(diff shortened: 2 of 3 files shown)


nealterrell:
@mtgred, how do you feel about this? NetrunnerDB has *all* spoiled cards in their database, which means they show in deckbuilder for us and lead to issues like #1178. This PR changes the card fetch script to only fetch from sets that are "available" -- this field is set by NetrunnerDB to the set's release date, on the day of the release. When operated in dev mode, however, all cards are fetched, to allow for early implementation work. 

Yes, you will have to run the fetch script on the server the day of a pack's release to enable it, but that's similar to how things had to be done during the SanSan Cycle. 

Let us know if you just flat out want all spoiled cards in deckbuilder, in which case this can be closed; or if there's an improvement that would make deployment/management easier.

bmatsuo:
Maybe I'm missing something. What is the intention here? Restrict the available cards to those that are tournament legal? I'm not sure this will still be a reasonable choice once cycle rotation starts.

As I see it I am somewhat in favor of the proposal in #1178, marking decks containing unreleased and other tournament illegal cards as being "Casual".

mtgred:
I prefer to fetch all the spoiled cards let people find them in the card browser and test new cards early.

The deck validation should mark the deck as casual and somehow indicate unreleased cards in the deck builder.

nealterrell:
@bmatsuo: historically, NetrunnerDB has added spoiled cards only when the entire pack gets released. They changed their behavior for Kala Ghoda, and we are seeing some frustration by users regarding the issue (as you see in #1178). The intention here, after a 5 minute conversation in Slack, was to not fetch unreleased cards into the database, making them unavailable for deckbuilding. 

@mtgred: thanks, we'll see if that can be done, since the release status of cards is not available in the CLJS code. Maybe @zaroth has some ideas.

domtancredi:
@mtgred If the decision for the product is to pull all cards and datasets, even on production and flag or highlight "released" cards // sets (as determined by NetrunnerDB), *should we delete this branch?* 

Ultimately, my opinion is that you're the ultimate product owner, so your decision on roadmap is the one we follow. 

In addition, I just want to note that an additional value-add to this feature-branch, which is to control the spread of "unreleased" or "spoiler" cards, may be perceived favorably from FFG.

zaroth:
Alright, now the approach we're going to take is clear, I'll try to take care of this. I wonder if grabbing http://netrunnerdb.com/api/sets/ and comparing the `available` date to server date is the better approach, or should we enable the sets manually.

zaroth:
actually the answer is simple, just parse the ID of a card and compare to the latest released (latest of KG has ID 10019). That way we won't have to send set information to the clients.

nealterrell:
Closed in favor of #1209.

PR #1154, opened 2016-01-26
Kala Ghoda cards

JoelCFC25:
I've put in a few of the fixes and safeguards suggested by @zaroth and @Saintis from the previous attempt. In particular, multiple copies of Kala Ghoda Real TV and scored Advanced Concept Hoppers should work OK.


mtgred:
:+1:

PR #1162, opened 2016-01-28
Implementation and test for Maya

nealterrell:
Like Imp and Film Critic, we expect the Runner to click Maya during the "You accessed ..." prompt to return it to the bottom of R&D.

I tested:

* Maya an Operation or ICE during the "You accessed ..." prompt.
* Maya an Upgrade or Asset during the "Pay X to trash?" prompt. If the asset is an ambush, using Maya before the ambush is fired does not affect its resolution. Some assets like Snare! even show a "Waiting for Corp..." prompt which prevents Maya's use until the Corp finishes the ambush. More ambushes are moving towards this style too.
* Maya an Agenda during the "You accessed ... Steal?" prompt. If this is chosen, the agenda's on-access effects (if any) are fired when Maya is clicked, because at this point they haven't been fired due to needing to support Film Critic.

### DIFF
--- a/src/clj/game/cards-hardware.clj
+++ b/src/clj/game/cards-hardware.clj
@@ -315,6 +315,22 @@
              {:player :runner :prompt "Choose a card" :msg (msg "add 1 card to Grip from Stack")
               :choices (req (:deck runner)) :effect (effect (move target :hand) (shuffle! :deck))}}}
 
+   "Maya"
+   {:effect (effect (gain :memoryunits 2)) :leave-play (effect (lose :memoryunits 2))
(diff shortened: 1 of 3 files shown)


zaroth:
Pasting this link for all Kala Ghoda cards, may be relevant: [Kala Ghoda UFAQ](http://ancur.wikia.com/wiki/Kala_Ghoda_UFAQ)

zaroth:
a bit of headache with Maya's explained timing according to new ruling: [image](http://i.share.pho.to/2feeada8_o.png)

mtgred:
:+1:

PR #1163, opened 2016-01-29
implement Dedication Ceremony and Heritage Committee

JoelCFC25:
New implementations for Dedication Ceremony and Heritage Committee. Also includes a small update to Chatterjee University to account for the pack-specific Unofficial FAQ that came out indicating you can use it to install a program when it has 0 counters.

### DIFF
--- a/src/clj/game/cards-resources.clj
+++ b/src/clj/game/cards-resources.clj
@@ -77,14 +77,15 @@
    {:abilities [{:cost [:click 1] :label "Place 1 power counter"
                  :msg "place 1 power counter on it"
                  :effect (effect (add-prop card :counter 1))}
-                {:req (req (> (get card :counter 0) 0))
+                {:req (req (>= (get card :counter 0) 0))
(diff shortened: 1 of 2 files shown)


src/clj/game/cards-resources.clj
```diff
@@ -77,14 +77,15 @@
    {:abilities [{:cost [:click 1] :label "Place 1 power counter"
                  :msg "place 1 power counter on it"
                  :effect (effect (add-prop card :counter 1))}
-                {:req (req (> (get card :counter 0) 0))
+                {:req (req (>= (get card :counter 0) 0))
```
nealterrell: This function is always true ;).

JoelCFC25: So I just need to take it out?

nealterrell: Yeah. In effect there is no requirement for using the ability.

mtgred:
:+1:

PR #1168, opened 2016-01-31
scroll to top for button bar on games list

domtancredi:
Issue #1056: Adding the fix specifically for button bar on game list; the previous fix applied to deckbuilder button-bar


mtgred:
:+1:

PR #1173, opened 2016-02-01
Eden Fragment implementation (manual trigger)

bmatsuo:
I was trying to play [Eden Fragment](http://netrunnerdb.com/en/card/06030) and realized its mechanics weren't implemented. The workaround requires a manual credit (edit: ~~click~~) adjustment to install ICE after it is scored.

I have added a 1-click ability to Eden Fragment that lets the corp install a piece of ICE for free. The implementation does not exactly match the card text, but the end result is essentially the same (as long as the players use it properly). I've implemented it this way because I am new to Clojure and there seem to be several problems automating the mechanics.

Screenshot of implementation:
screenshot

### DIFF
--- a/src/clj/game/cards-agendas.clj
+++ b/src/clj/game/cards-agendas.clj
@@ -154,6 +154,8 @@
 
    "Eden Fragment"
    {:abilities [{:cost [:click 1] :msg "install the first piece of ICE this turn at no cost"
+                 :req (req (empty? (let [cards (map first (turn-events state side :corp-install))]
+                                     (filter #(is-type? % "ICE") cards))))
                  :prompt "Select a piece of ICE from HQ to install"
                  :choices {:req #(and (:side % "Corp")
                                       (ice? %)
--- a/src/clj/game/cards-agendas.clj
+++ b/src/clj/game/cards-agendas.clj
@@ -154,12 +154,12 @@
 
    "Eden Fragment"
    {:abilities [{:cost [:click 1] :msg "install the first piece of ICE this turn at no cost"
-                :prompt "Select a piece of ICE to install from the HQ"
-                :choices {:req #(and (:side % "Corp")
-                                   (ice? %)
-                                   (= (:zone %) [:hand]))}
-                :effect (req (corp-install state side target nil
-                                           {:no-install-cost true}))}]}
+                 :prompt "Select a piece of ICE from HQ to install"
+                 :choices {:req #(and (:side % "Corp")
+                                      (ice? %)
+                                      (= (:zone %) [:hand]))}
+                 :effect (req (corp-install state side target nil
+                                            {:no-install-cost true}))}]}
 
    "Efficiency Committee"
    {:effect (effect (add-prop card :counter 3))
--- a/src/clj/game/cards-agendas.clj
+++ b/src/clj/game/cards-agendas.clj
@@ -152,6 +152,15 @@
                                 (gain-agenda-point state side 1))
                               (set-prop state side card :counter 1 :agendapoints 1))}]}
 
+   "Eden Fragment"
+   {:abilities [{:cost [:click 1] :msg "install the first piece of ICE this turn at no cost"
+                :prompt "Select a piece of ICE to install from the HQ"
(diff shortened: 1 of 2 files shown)


bmatsuo:
My first attempt was to model the card after the identity Kate "Mac" McCaffrey: Digital Tinkerer. But I ran into some problems doing that and a manual solution seemed appropriate for the time being. Following is a list of the implementation issues I encountered. They may deserve separate github issues. Otherwise if you could point me in the right direction I would appreciate it.

- It seems like the `pre-install` event is specific to installations performed by the Runner. AFAICT `pre-install` is not triggered before the `corp-install` action occurs. Naive attempts by me to hook that up have not worked yet. And I still don't get how that event could interact with the `no-install-cost` functionality for `corp-install`. Anyway, if this really needs to be changed than it likely deserves it's own github issue (and I couldn't find one in a search).

- Another potential problem with automation was that I didn't see how to track ICE installs before Eden Fragment is scored and activated. There is a small possibility that Eden Fragment is scored after an ICE has been installed by the corp player and that would complicate the card implementation. Following is example of this potential problems.

 - _Corp turn starts with 4 advancement tokens on Eden Fragment_
 - Corp installs ICE over HQ for 1 click and X credits
 - Corp advances Eden Fragment for 1 click
 - Corp scores Eden Fragment
 - Core installs ICE over HQ for 1 click and (X+1) credits
 - _Corp ends their turn_
 - ...
 - _Corp begins their turn_
 - Corp installs ICE over HQ for 1 click and **0** credits
 - Corp installs ICE over HQ for 1 cilck and (X+3) credits
 - ...

Let me know if anything there needs clarification.

JoelCFC25:
Nice work so far! You can filter on `[:turn-events]` in `state` to ensure that no previous ICE installs have occurred. The technique can lifted from Autoscripter--map `first` onto it and then `filter` for ICE!

bmatsuo:
Fastastic! Thanks for the tip, @JoelCFC25! I have been racking my brain thinking of similar cards. I didn't look too much on the runner side.

I believe I have correctly added the condition to the ability trigger in the latest commit. Please check it out and let me know if there is anything else I should tackle before this is merged.

nealterrell:
Your use of `turn-events` looks correct, so good job on that edge case. Completely automating the effect is possible, but would require some surgery in core. If you're feeling up to it, find in `runner-install` where `:pre-install` event is triggered, and `install-cost` is then called to calculate the final install cost. You can mirror that in `corp-install` to trigger a `:pre-corp-install` event, passing the card being installed and its target server to the event handler. That would then let Utopia Fragment handle the `:pre-install` event and, if it sees that the `target` is an ICE, count the number of ice already in the target server and give a `install-cost-bonus` of -X credits. 

It's OK if that's intimidating -- we can merge your work here to get 90% functionality into the game, then maybe you can work on playing with the rest.

bmatsuo:
Thanks for the feedback @nealterrell. It does sound similar to how I had started when trying to model the ability using Kate's identity as a reference. But the idea to make `:pre-corp-install` distinct from `:pre-install` sounds somewhat promising as risk of bugs would be minimized.

But, honestly, I don't feel comfortable adding something like that to the game's core at this point in time. I am not confident in my ability to write idiomatic Clojure or the tests I would like.

I think it would be best to just merge this for now to avoid potential stagnation. Then we can create a new issue regarding automation of the ability and making the required event modifications. After I get more comfortable with Clojure I will feel better about taking on such a task (or I would be happy to look at someone else's implementation). I still don't really understand what I'm doing here :wink: I am just reasonably good at groking code and copying techniques.

Cheers.

nealterrell:
Sounds good. Thanks for the contribution, and we hope to see more from you in the future :).

bmatsuo:
Can someone also remember to update the [doc](https://docs.google.com/spreadsheets/d/1ICv19cNjSaW9C-DoEEGH3iFt09PBTob4CAutGex0gnE/pubhtml) with the card status?

It seems like it should have "Partial" implementation and "Partial" testing.

mtgred:
:+1:

PR #1174, opened 2016-02-01
Special css class for virus counters and transparent counter colors

### LINKED ISSUE #522: Feature Request: Visual indicator for counter type (Power, Virus, etc)
When playing, the user normally has to check the card text to inspect which counter type is being represented on the card.  Some kind of coloring or visual indicator of counter type would be very helpful when deciding whether to take actions like purging virus counters

bmatsuo:
Fixes #522 

This change adds a css class for counters on **Virus** cards to give a more clear visual indication of which counters can be purged.

To make counters a little more consistent I added subtle transparency for the special counters (recurring, advancement, and now virus). 

Screenshot (see the [comment](https://github.com/mtgred/netrunner/pull/1174#issuecomment-179019900) below for a more recent shot):
screenshot

### DIFF
--- a/src/clj/game/cards-assets.clj
+++ b/src/clj/game/cards-assets.clj
@@ -2,7 +2,8 @@
 
 (def cards-assets
   {"Adonis Campaign"
-   {:effect (effect (add-prop card :counter 12))
+   {:data {:counter-type "Credit"}
+    :effect (effect (add-prop card :counter 12))
     :derezzed-events {:runner-turn-ends corp-rez-toast}
     :events {:corp-turn-begins {:msg "gain 3 [Credits]" :counter-cost 3
                                 :effect (req (gain state :corp :credit 3)
@@ -193,7 +194,8 @@
                          :effect (effect (trash-cost-bonus 1))}}}
 
    "Eve Campaign"
-   {:effect (effect (add-prop card :counter 16))
+   {:data {:counter-type "Credit"}
+    :effect (effect (add-prop card :counter 16))
     :derezzed-events {:runner-turn-ends corp-rez-toast}
     :events {:corp-turn-begins {:msg "gain 2 [Credits]" :counter-cost 2
                                 :effect (req (gain state :corp :credit 2)
@@ -313,7 +315,8 @@
                                  (trash card {:cause :ability-cost}))}]}
 
    "Launch Campaign"
-   {:effect (effect (add-prop card :counter 6))
+   {:data {:counter-type "Credit"}
+    :effect (effect (add-prop card :counter 6))
     :derezzed-events {:runner-turn-ends corp-rez-toast}
     :events {:corp-turn-begins {:msg "gain 2 [Credits]" :counter-cost 2
                                 :effect (req (gain state :corp :credit 2)
@@ -364,7 +367,8 @@
                                 card nil))}]}
 
    "Marked Accounts"
-   {:abilities [{:cost [:click 1] :msg "store 3 [Credits]"
+   {:data {:counter-type "Credit"}
+    :abilities [{:cost [:click 1] :msg "store 3 [Credits]"
                  :effect (effect (add-prop card :counter 3))}]
     :events {:corp-turn-begins {:msg "gain 1 [Credits]" :counter-cost 1
                                 :effect (effect (gain :credit 1))}}}
@@ -425,7 +429,8 @@
    {:recurring 3}
 
    "Private Contracts"
-   {:effect (effect (add-prop card :counter 14))
+   {:data {:counter-type "Credit"}
+    :effect (effect (add-prop card :counter 14))
     :abilities [{:cost [:click 1] :counter-cost 2 :msg "gain 2 [Credits]"
                  :effect (req (gain state :corp :credit 2)
                               (when (= (:counter card) 0) (trash state :corp card)))}]}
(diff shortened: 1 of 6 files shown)


JoelCFC25:
I love it. Any chance you want to tackle differentiating power counters like Virus counters have been and giving those a color too?

bmatsuo:
@JoelCFC25 I would be happy to try if you have a suggestion of how one might go about that. Also, to clarify, you specifically intend to differentiate power counters from, say, credits? I'm not sure what else needs differentiation. The only other counters I can think of are agenda counters (on agendas in the player score areas) but I think there is little ambiguity there.

I had thought about that but wasn't sure how to accomplish that without a change to the data model (e.g. an actual flag that says what kind of counters the card hosts). I was able to get around anything like that here because I was able to exploit the card subtype data. I'm happy to try something like more advanced like that if necessary. But I am still just dipping my toes into the codebase.

bmatsuo:
Perhaps I could try searching card text for the string "power counter" and assume that any hosted counters are power counters. Though it doesn't seem like card text propagates into the browser. And I'm not sure if doing so would cause any kind of noticeable performance impact. But it may be worth a try...

bmatsuo:
I could also inspect a card's paid abilities and assume that counters are power counters if the abilities have a counter-cost. This seems like it will work in many cases. But there will be false-negatives when counters are removed in response to an event (e.g. [Earthrise Hotel](http://netrunnerdb.com/en/card/06120)).

While I think that distinguishing power counters would be nice I'm not sure how important it is for the ux as cards hosting either power counters or credits have a low amount of interactivity with other cards and external actions. Cards like Bank Job, Earthrise Hotel, etc are all self contained and aside from being trashed the only thing that changes the number of counters on the card is the card's own abilities.

I am having a hard time remembering if there are cards like of [Trick of Light](http://netrunnerdb.com/en/card/02033) or [Hivemind](http://netrunnerdb.com/en/card/07042) for power counters (or for credits). I haven't found anything on netrunnerdb. But, if there are such cards they aren't really seeing a lot of play IMO. Cards that interact with counters on other cards (like those I mentioned) would make it more important to understand whether power counters or credits are on each card. Of course if it was possible to have both power counters **and** credits on a card at the same time then some distinction would be required. But as it is I don't think that case is possible.

bmatsuo:
The card [Personal Workshop](http://netrunnerdb.com/en/card/02049) has just come to mind. That is an interesting card in terms of power counters because the counters on the card are actually relating to the workshop's hosting abilities.

In fact, if Personal Workshop is used to host/install viruses under my patch in this PR the counters will appear as virus counters. This may be something that requires the implementation here to be thought out more carefully. Though I'm not sure I've seen someone use a workshop to install viruses, I think it is a definite possibility in some crazy Shaper decks.

bmatsuo:
I looked at how [Personal Workshop](http://netrunnerdb.com/en/card/02049) was handled elsewhere and am now checking if the card is `:installed` before declaring the counter a virus counter.

zaroth:
I second that it's a nice addition, just wanting to note that not all cards with virus counters have virus subtype. So if this PR gets added as-is, soon enough we will have a bug report that Virus breeding counters aren't viruses ;-)

Since the only card that is currently released and falls into this pit is VBG, I would be okay with just naming its card id (07052) in the function check.

bmatsuo:
I forgot that Virus Breeding Ground was a resource. Thanks for pointing that out! :smile: So, I ended up finding the `:counter-type` data attribute for cards. So I am using that to annotate "Power" versus "Credit" counters now and I've assigned CSS classes for all the kinds of counters. The new classes will be used when the `:counter-type` is known or can be inferred.

For VBG the type was already annotated. I annotated a couple others and have the following screenshot. My monitor isn't calibrated so I didn't tweak the colors very much. I think the `strength` and `credit` colors are a little to close to each other. And I wasn't sure what color to make agenda counters so I just reused the background color used for the grip, score areas, etc.

screenshot

bmatsuo:
OK.I think I have everything implemented. All the cards that use credit counters have been annotated, and power counters have been made the default counter if nothing else is inferred. There was a bug with the CSS class conflict so I changed the counter classes to be `*-counter` ("recurring-counter", "virus-counter", "credit-counter", etc). I tweaked the color of credits counters as well so they are a little more dark and distinct from strength indicators.

screenshot

mtgred:
What's happening to the orange button color?

In general I like keep the color palette as small as possible. It's better to keep a sober interface in my opinion and only add color parsimoniously to avoid having a rainbow of colours all over the place. I prefer to keep the original transparent black color for regular token.

bmatsuo:
Thanks for the feedback @mtgred. And thanks for pointing out the buttons. I did think they looked off but couldn't place it exactly. I accidentally overwrote the style variable `transparent-orange` with a brighter color.

I understand your concern. The palette got a little intense (the purple is probably more offensive to me). Just to clarify what you mean -- a "regular token" would be any non-virus tokens? Honestly I kind of like the gold token for credits. But I could really go either way. All I really wanted in the beginning was red virus tokens.

bmatsuo:
I have left all the classes in place so that someone with a good design sense can add style for `.advance-counter`, `.recurring-counter`, `.virus-counter`, `.power-counter`, `.agenda-counter`, and `.credit-counter`.

The only styles I have left are those for advancement, recurring and virus counters. The first two already existed but the names are now different and they have a slight transparency to them. Transparency has also been applied to the _effective strength_ indicator.

screenshot

mtgred:
:+1:

PR #1179, opened 2016-02-03
Minor fixes: Mutate, Turntable, Leela Patel

### LINKED ISSUE #1152: Turntable allows you to switch news team for a real agenda
The card news team is not stolen but placed in the score area as an agenda but still triggers turntable.
Have yet to test notoriety, fan site, shi.kyu or excutives for the same issues.

### LINKED ISSUE #1165: Mutate doesn't initialize the installed ICE
Wanted to prevent an Ice Wall from dying to a parasite by advancing it, but the only available option was to derez it.

### LINKED ISSUE #1169: Leela's ability will only work once if two agendas are stolen
I thought this was logged long ago but I couldn't find any record of it.

Anyway, probably pretty self-explanatory, but for completeness: If you multi-access and steal two agendas you will be told twice to click Leela to use her ability, but after you use it the first time it's no longer enabled. I've also tried firing Leela after the first steal but the dialogue box forces you to complete your accesses/steals first.

JoelCFC25:
* Fix #1169: Add `:priority true` to Leela's ability so it jumps to the front of the prompt queue in a multi-access situation (avoiding interwoven prompts)
* Fix #1165: Mutate needs to call `card-init` on the ICE swapped in from R&D to enable its abilities and event hooks
* Fix #1152: Ensure that the thing you are about to swap with Turntable is an agenda, and add safeguards to stop delayed usage of Turntable (dissociates the enabling key if the Runner spends a click to do anything else). Updates Turntable's test.
* Adds a message to the log when the Corp declines to trigger the forced Archangel encounter

### DIFF
--- a/src/clj/game/cards-hardware.clj
+++ b/src/clj/game/cards-hardware.clj
@@ -510,42 +510,43 @@
    "Turntable"
    {:effect (effect (gain :memory 1))
     :leave-play (effect (lose :memory 1))
-    :events {:agenda-stolen {:effect (req (toast state :runner "Click Turntable to swap for a scored Corp agenda." "info")
-                                          (update! state side (assoc card :swap true)))}}
-    :abilities [{:optional
-                  {:req (req (:swap card))
-                   :prompt "Swap for a scored Corp agenda?"
-                   :yes-ability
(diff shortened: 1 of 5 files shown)


mtgred:
:+1:

PR #1183, opened 2016-02-03
Improvement to agenda forfeiting; fix Turntable + Mandatory Upgrades interaction

### LINKED ISSUE #1142: Turntable doesn't switch off Mandatory Upgrades
Reproduction steps:

* score Mandatory Upgrades on corp turn
* on Runner turn install turntable and steal a Vitruvius, swapping it for ManUp
* on Corp's turn Corp still has 4 clicks

JoelCFC25:
Fixes #1142. 

Adds a couple new flags for verifying a card's presence in Corp or Runner scored areas, then uses them to only deactivate forfeited agendas if they came from the Corp area. The logic to make forfeiting a Mandatory Upgrades or Self-destruct Chips work while simultaneously allowing Turntabled agendas to be deactivated needed to go into `forfeit` instead of `deactivate`. 

Includes tests for forfeiting a Mandatory Upgrades as both sides and for using Turntable to swap a scored one away from the Corp.

### DIFF
--- a/src/clj/game/core-flags.clj
+++ b/src/clj/game/core-flags.clj
@@ -108,6 +108,16 @@
   [card]
   (= (:zone card) [:deck]))
 
+(defn in-corp-scored?
+  "Checks if the specified card is in the Corp score area."
+  [state side card]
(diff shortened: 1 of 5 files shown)


mtgred:
:+1:

PR #1185, opened 2016-02-03
Card Refactoring

Saintis:
Please review :smile: 

- Ice swap function
- `:end-effect` to prompts -- triggers at the end of the prompt (useful for waiting prompts)
- Adonis / Eve / Launch campaign refactor
- Ambush refactor
- Refactors the installing functions
- Adds `:in-play` key to cards for effects that are passive while in play -- such as memory and hand-size
- Refactors Icebreakers -- creates strength and break ability factory functions and breaker set functions (Dogs, B&E, GlobalSec)
- Tests:
 - Apocalypse interaction with installed consoles etc reducing MU correctly
 - Wyrm reducing strength of ice
 - Dog breakers -- specifically Rex, also for ensuring boosting and counter use works properly

NB: The new icebreaker functions take arguments in the order shown on cards: cost then effect (broken subs or strength boosted).

### DIFF
--- a/src/clj/game/cards-assets.clj
+++ b/src/clj/game/cards-assets.clj
@@ -627,24 +627,8 @@
                  :cost [:click 1]
                  :prompt "Select two pieces of ICE to swap positions"
                  :choices {:req #(and (installed? %) (ice? %)) :max 2}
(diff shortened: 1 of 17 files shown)


mtgred:
:+1:

PR #1187, opened 2016-02-05
implement Port Anson Grid

JoelCFC25:
Includes a test to ensure jacking out is prevented both when rezzed during a run and when it's already rezzed prior to the start of a run on the server.


mtgred:
:+1:

Issue #1192, opened 2016-02-07
Apocalypse duplicates hosted cards when it fires and does not properly regain MU


kevkcc:
Steps to reproduce:
1. Host something like a Corroder on a Scheherazade
2. Run the 3 centrals
3. Use Apocalypse
4. There will be a Corroder facedown and also in the heap (and you will still be at 3 MU instead of 4)

PR #1193, opened 2016-02-07
Apocalypse

kevkcc:
I noticed that Apocalypse had a few problems when I was playing with respect to hosted cards and memory (#1192). I have submitted a fix and a test for these issues. I also noticed during my testing that Leprechaun could also host 3 cards on it so I made a quick fix to that too.

### DIFF
--- a/src/clj/game/cards-events.clj
+++ b/src/clj/game/cards-events.clj
@@ -26,8 +26,15 @@
                                 (sort-by #(vec (:zone %)))
                                 (reverse))]
                    (trash state side c))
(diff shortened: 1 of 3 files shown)


kevkcc:
Has anybody looked at this?

zaroth:
Well, Apocalypse has always been a bit of a mess, so you'll see no opposition from me to fix the state of things. Also the changes are tested, so overall big :+1: from me.

nealterrell:
Looks good :+1:

mtgred:
:+1:

PR #1194, opened 2016-02-07
Fix grammar for game count when there are zero games

bmatsuo:
When no games are active the game count will display "0 Games" instead of "0 Game".

This is not going to be seen often by users.  But it may in the case when the game server is down.  It just bugs me when I'm running the service in development.

### DIFF
--- a/src/cljs/netrunner/main.cljs
+++ b/src/cljs/netrunner/main.cljs
@@ -50,7 +50,7 @@
     [:div
      [:div.float-right
       (let [c (count (:games cursor))]
-        (str c " Game" (when (> c 1) "s")))]
+        (str c " Game" (when (not= c 1) "s")))]
      (when-let [game (some #(when (= (:gameid cursor) (:gameid %)) %) (:games cursor))]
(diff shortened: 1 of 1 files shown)


mtgred:
I'm not sure i understand. "0 game" should be "0 games" ?

bmatsuo:
Yes. If you are going to say the number zero then "zero games" is correct, not "zero game". The pluralization might be more clear in a sentence.

> There are zero games.
> There is zero game.

English is weird =\ Like I said though. It's not a very big deal.

mtgred:
Ok I didn't know that. English is not my native language and in french zero is singular.

This should be correct and many more places in the code then.

bmatsuo:
I didn't realize the difference from french. In english any number other than 1 is plural (including -1) [[ref.](http://english.stackexchange.com/questions/38293/why-is-zero-plural)].

The only other case in the code I found easily had to do with displaying the number of spectators. But because that is only shown when there are >1 spectators it wasn't affected by this error.

PR #1197, opened 2016-02-07
Adds counter for Femme Fatale.

Saintis:
Creates functions and styling for having cards putting a "counter" on a different card.

Colour of counter has to be implemented in base.styl (red and blue so far).

Currently only used for Femme Fatale.

Counter is removed if the card that placed the counter (Femme Fatale) is trashed, or the card with the counter is moved to `:discard` or `:hand` as it's implemented now.

Cases above are tested in the new test `femme-counter`. Blue-Sun interaction works correctly but is not tested for.


mtgred:
:+1:

PR #1198, opened 2016-02-07
Toast on taking a tag.

Saintis:
Simple 1-line change.

Sorry for the short PR.

Suggested use of toasts to inform Runner when they have taken a tag. Lets the runner know additionally when they take a tag, and how many.


mtgred:
:+1:

Issue #1199, opened 2016-02-08
Trope won't add counters


domtancredi:
Issue is that the subtype check in javascript is thrown an indexOf error on "unknown" because Trope has no "subtype"

Issue #1196, opened 2016-02-07
Davinci makes the game "freeze"


Grimstringer:
2 times in two games my opponent used one time  maker's eye and the next one Dirty laundry and Davinci made the game freeze, he couldnt do anything or me so we just quit. 

no counters on it, no access, nothing

bmatsuo:
This sounds like it may be the same as #1199. Do you think so, @domtancredi?

domtancredi:
Program with counters but non virus? Sounds like the same issue exactly.

This PR should fix it (will need to test)

bmatsuo:
I replicated this problem running on an empty Archives with a newly installed DaVinci. The root cause appears to be the same as #1199.

domtancredi:
@Grimstringer The fix has been merged. Would you test and close the issue if it passes?

PR #1200, opened 2016-02-08
fix for counter subtype detection

domtancredi:
Fix for Issue #1199


bmatsuo:
This makes sense. And it fixes #1199 on my machine.

Thanks for tracking it down.

bmatsuo:
I have confirmed that this PR also fixes #1196.

mtgred:
:+1:

PR #1202, opened 2016-02-08
Minor fixes

### LINKED ISSUE #1201: R&D Multiaccess Giving one less access than it should
When accessing R&D, i was getting one less access than i should have been. With no RDI, I'd run R&D and get no access, with one RDI, i'd get one access and so on.  Not sure what was causing it, but it happened to games in a row (trashing an RDI off a street peddler maybe?).  I've attached a screenshot of the game state from the second game.

image

Saintis:
Fixes and adds test for `:in-play` interaction with Street Peddler.

Fixes and adds test for Surfer.

Fixes #1201.

### DIFF
--- a/src/clj/game/cards-programs.clj
+++ b/src/clj/game/cards-programs.clj
@@ -583,12 +583,12 @@
                                                   card nil))}]}
 
    "Surfer"
-   (letfn [(surf [cice]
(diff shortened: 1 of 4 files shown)


mtgred:
:+1:

PR #1204, opened 2016-02-08
fix Medium and Nerve Agent mid-run installation corner case

JoelCFC25:
This fixes a somewhat obscure bug--if Grimoire is installed, using Clone Chip/Peddler/SMC to install Nerve Agent or Medium at paid ability speed during a run **after** the Corp has clicked "No action" for the final time is leading to these cards not gaining a 2nd counter when you hit Successful Run.


mtgred:
:+1:

PR #1206, opened 2016-02-09
Pluralization fixes for player stats view

bmatsuo:
Follows up my previous PR #1194 after @mtgred pointed out additional occurrences of incorrect pluralization.

A few pieces of text were actually modified.  Mostly the cases under
which plural text is seen have been modified to include values less than
1.

Notable changes to text:

- **Brain Damage**; I don't think it makes sense to pluralize this
  quantity.  It is more natural to say "I have 5 Brain Damage." This
  is similar to the card Scorched Earth that says, "Deal 4 Meat Damage."

- **Bad Publicity**; Pluralizing this sounds wrong.  Similar to Brain
  Damage saying, "Take 2 Bad Publicity," sounds more correct within
  reason.

- **Link** also didn't make sense to pluralize.  But I believe that
  **Link Strength** is the canonical term.  So that is used instead to
  make it sounds more natural.

### DIFF
--- a/src/cljs/netrunner/gameboard.cljs
+++ b/src/cljs/netrunner/gameboard.cljs
@@ -605,17 +605,17 @@
     (let [me? (= (:side @game-state) :runner)]
       [:div.stats.panel.blue-shade {}
        [:h4.ellipsis (om/build avatar user {:opts {:size 22}}) (:username user)]
-       [:div (str click " Click" (if (> click 1) "s" "")) (when me? (controls :click))]
-       [:div (str credit " Credit" (if (> credit 1) "s" "")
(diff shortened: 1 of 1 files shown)


JoelCFC25:
I like this--especially the bad publicity one, which is definitely incorrect. These have bothered me for awhile but obviously never to the point of fixing them... :P

bmatsuo:
Bad Pub and Brain Damage are indeed abuses of the english language. Both should probably be referred to as "units" (i.e.. two units of Bad Publicity/Brain Damage). That is awfully cumbersome to say though.

mtgred:
:+1:

PR #1209, opened 2016-02-10
Unreleased cards and decks appear as casual in deckbuilder

### LINKED ISSUE #1178: Detect unreleased cards
NetrunnerDB now has unreleased cards public on their API, and it seems that people can play with unreleased cards on Jinteki.net. Some players seem to not realize that these cards are not yet released or legal. It would be nice if the legality-checker could detect cards that are not yet released and fail upon detection.

zaroth:
As per discussion in #1148. Fixes #1178.

### DIFF
--- a/src/cljs/netrunner/deckbuilder.cljs
+++ b/src/cljs/netrunner/deckbuilder.cljs
@@ -254,10 +254,19 @@
            (let [min (min-agenda-points deck)]
              (<= min (agenda-points deck) (inc min))))))
 
+(defn released?
+  "Returns false if the card comes from a spoiled set or is out of competetive rotation."
+  [card]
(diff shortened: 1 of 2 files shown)


Saintis:
Note that this will require a manual commit / change whenever a new set of cards are released.

zaroth:
@saintis Yes, but I decided it's a lesser evil than adding otherwise useless set info of every card to the data sent to players for every card. We're doing manual commits every new set anyway ;-)

Although I guess it would be possible to parse only the `10019` number from netrunnerdb or process it locally, and then send only this to clients. However, I'm not sure it's worth it, especially when rotations comes into play. Maybe if we moved the entire card validation server side...

mtgred:
:+1:

PR #1211, opened 2016-02-11
Add :pre-corp-install event to implement install-cost bonuses for the corp

### LINKED ISSUE #1177: Allow pre-installation event handling for corp cards
As discussed by @nealterrell and I in #1173 the [Eden Fragment](http://netrunnerdb.com/en/card/06030) cannot be completely automated using the existing event structure. The `:pre-install` event only fires for runner installs (and may pose some risk if corp installs are included? still undetermined). But a `:pre-corp-install` event was suggested and that seemed reasonable.

This issue is meant to track changes to events (and probably the completion of Eden Fragment's implementation as a proof of concept).

bmatsuo:
Fixes #1177

Cards can hook up to a `:pre-corp-install` event to deduct or add cost for installs using the `install-cost-bonus` function. This event is triggered just before install costs are calculated during `corp-install`.

I ended up also adding a `ignore-install-cost` function for cards like Eden Fragment. My reasoning was that Eden Fragment cannot assist in any **additional cost** that might be incurred for installing ICE. So, a straight deduction may not be accurate. I'm definitely open to any critique on that choice before merging.

I have fully automated/tested Eden Fragment.

eden fragment
(edit: I clearly didn't playtest it much. I've fixed the text in a later commit)

**Open questions**:

- Is that event name correct? Should it be `:corp-pre-install`?

- Did I over complicate things with the added `ignore-install-cost` functions? Are there other cards that place additional cost on ICE installs? I couldn't find any.

- The Eden Fragment message is modeled after Kate and shows up after the install. Should it show up before?

### DIFF
--- a/src/clj/game/cards-agendas.clj
+++ b/src/clj/game/cards-agendas.clj
@@ -162,7 +162,7 @@
                {:req (req (and (is-type? target "ICE")
                                (empty? (let [cards (map first (turn-events state side :corp-install))]
                                          (filter #(is-type? % "ICE") cards)))))
-                :msg (msg "ignore the install cost of " (:title target))}}}
+                :msg (msg "ignore the install cost of the first ICE this turn")}}}
 
    "Efficiency Committee"
    {:effect (effect (add-prop card :counter 3))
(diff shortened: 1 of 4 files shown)


bmatsuo:
Oh snap. Now I see that it is revealing an unrezzed card... Oh boy... ;)

nealterrell:
Nice job :).

mtgred:
:+1:

PR #1212, opened 2016-02-11
full implementation of Toshiyuki Sakai

JoelCFC25:
Peak jank has arrived--an armed and fully operational Toshiyuki Sakai! Includes a test, and though there is no written test for Toshi in a server with upgrades, I tested that in a variety of circumstances and it seems to be working OK. 

This also touches up Shattered Remains to have a more informative prompt and a log message listing all trashed hardware in the case of multiples.


mtgred:
:+1:

PR #1213, opened 2016-02-11
enable High-Stakes Job; small Turntable/Aesop's tweaks

JoelCFC25:
High-Stakes Job isn't working due to an adjustment in NRDB data (it was in there initially as High-stakes Job, so that's how @Saintis titled it). 

Made a small adjustment to Turntable to push its prompt to the front of the queue during multi-access just like was done to Leela Patel. Aesop's trashing is made unpreventable so Fall Guy won't interfere.


mtgred:
:+1:

PR #1047, opened 2015-12-28
Feature account random background

domtancredi:
Added an Account View, for managing the user's account + profile.
Added navigation to the view
Added updating the background image (visible in the gameboard when playing or spectating: had to update the opacity 1 instead of 0.7)

Could use some code review and possible refactoring

PR #1215, opened 2016-02-11
Split game lobby into casual/competitive rooms

### LINKED ISSUE #345: Suggestion: Game type prompt

zaroth:
Fact: platform is often pushing over 100 games at peak times currently and the game list gets pretty long sometimes. A some way of splitting that would be nice.

To do that, and to help players that are having mismatched expectations when coming into the game, lo and behold, a competitive room (for the beginning, it's super easy to add more):

http://i.share.pho.to/b1549035_o.png

If we want to do that, creating custom rooms using this logic would be trivial - just allow the room name to be a text field. That is a sort-of replacement for passworded games ;-)

Could be enhanced by merging #1047, so we finally have user profiles and user could select their default lobby in their profile. Currently the default is the casual room.

Since there are changes to `server.coffee`, its restart will be needed.

Fixes #345 (kind of).

### DIFF
--- a/server.coffee
+++ b/server.coffee
@@ -120,7 +120,7 @@ lobby = io.of('/lobby').on 'connection', (socket) ->
     switch msg.action
       when "create"
         gameid = uuid.v1()
-        game = {date: new Date(), gameid: gameid, title: msg.title, allowspectator: msg.allowspectator,\
+        game = {date: new Date(), gameid: gameid, title: msg.title, allowspectator: msg.allowspectator, lobby: msg.lobby,\
                 players: [{user: socket.request.user, id: socket.id, side: msg.side}], spectators: []}
         games[gameid] = game
         socket.join(gameid)
--- a/server.coffee
+++ b/server.coffee
@@ -120,7 +120,7 @@ lobby = io.of('/lobby').on 'connection', (socket) ->
     switch msg.action
       when "create"
         gameid = uuid.v1()
-        game = {date: new Date(), gameid: gameid, title: msg.title, allowspectator: msg.allowspectator, lobby: msg.lobby,\
+        game = {date: new Date(), gameid: gameid, title: msg.title, allowspectator: msg.allowspectator, room: msg.room,\
                 players: [{user: socket.request.user, id: socket.id, side: msg.side}], spectators: []}
         games[gameid] = game
         socket.join(gameid)
--- a/src/cljs/netrunner/deckbuilder.cljs
+++ b/src/cljs/netrunner/deckbuilder.cljs
@@ -255,7 +255,7 @@
              (<= min (agenda-points deck) (inc min))))))
 
 (defn released?
-  "Returns false if the card comes from a spoiled set or is out of competetive rotation."
+  "Returns false if the card comes from a spoiled set or is out of competitive rotation."
   [card]
   (let [cid (js/parseInt (:code card))]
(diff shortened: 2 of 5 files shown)


zaroth:
The concept of having games separated by skill using rooms kind of obsoletes #755 IMHO, although I'll leave the decision up to others.

zaroth:
Changed usage of 'lobby' to more correct 'room', since arguably lobby is a collection of rooms in this case.

zaroth:
as per @mtgred suggestion, replaced select with tabs: http://i.share.pho.to/3b18e7e0_o.png

mtgred:
It's better to have tabs instead of a select box: 1 click instead of 2 and you directly see all the options instead of hiding them.

A split based a "competitive" label isn't very meaningful. Anyone can just join a "competitive" game with jank.

What should be done is ranked and unranked games. The ranked games should count toward an ELO rating, with a player profile page and a leaderboard. When there is a stake, people won't play with jank. It's basically what's happening with the Stimhack League. It's the same with play money poker vs real money poker.

Jinteki.net is modelled after Battle.net. In the old Battle.net for Warcraft 2 and Starcraft 1, players create a ranked game and anyone can join them. The problem with that approach is people can game the system by creating fake account to play against their main account. An auto matchmaking has been introduced in Battle.net for ranked games in Warcraft 3 and Starcraft 2. This prevents people from playing against fake accounts and it allows players to play with other people of around the same level. So beginners or people wanting to play with jank can still enjoy ranked games and will be matched against people of their level. And strong players will have challenging games.

The main problem with ranked games on jinteki.net is to determine the winner. There might be rule disputes and there's no way to automatically determine the winner in case of a disagreement.

I was thinking about tournaments as a solution. People would register with deck lists and play swiss rounds like a tournament in real life. There would be a present during the tournament to resolve disputes. Dan also wanted to organise tournament on Jinteki.net so the organisation would be in collaboration with Stimhack.

zaroth:
Select vs tabs: good point, I replaced that: http://i.share.pho.to/3b18e7e0_o.png .

As I replied in the discussion, a ranked system is a nice goal, but we're not quite there yet and the room split is a nice solution to help players with similar *expectations* to find each other. Obviously, a jank player can join a competitive room. But will he want to, knowing that he will just get beaten over and over by Fastro and DLR? Probably not, he's more likely to stay in the Casual room, where he can win more often (and find players who are not angry at him for playing jank).

Again, it isn't about stopping players with malicious intent, it's about matching expectations. When we get the profiles, ranking system, etc. we can easily revisit this into something that makes more sense later.

zaroth:
The rooms could also be easily named Public/Private, so we don't have the main lobby littered with 4-5 private games waiting for their partner at any time.

mtgred:
I'm testing your branch. Game don't seem to appear in the game list. Can you have a look at it?

screen shot 2016-02-14 at 14 08 27

zaroth:
@mtgred I just tested it, it works fine here. From the symptoms you're having in the screenshot it looks like `server.coffee` wasn't restarted - it needs to recognize the new `room` field in every game data structure. Please try again.

mtgred:
:+1:

PR #1216, opened 2016-02-12
added identity check to legality deck check

zaroth:
Comment made by @dodgepong made me realize that I forgot to check identity for its released status. So decks with Nero Severn or Harishchandra would be marked as TL - this is fixed.

As for marking draft IDs likewise, I'm hesitant - I don't want to make draft players any more neglected than they already are, since there actually are a few draft format A:NR tournaments in my area. This mistake is also instantly recognized, unlike ninja-Brainstorm ;-) So I hope it won't cause too much issues.


mtgred:
:+1:

PR #496, opened 2015-08-04
basic support for touch devices

queueseven:
tested on Windows 10 and Android

PR #1217, opened 2016-02-12
Drag & drop for touch devices

zaroth:
Refreshed touch support from #496 by @queueseven. Tested using [excellent Chrome's touch screen simulator](https://developer.chrome.com/devtools/docs/device-mode) - drag & drop seems to be mostly working. Certainly a big step ahead from not working at all :)


mtgred:
:+1: 

The drag and drop seems to work on iOS devices too.

There's still no way to decrease to zoom level on iOS devices (as far I know) so the layout is kind of broken.

PR #1219, opened 2016-02-13
Extends toast functionality

Saintis:
Allows passing in options to toast specifying the time-out duration, whether it should have a close-button and whether duplicate toasts should be prevented.

Create a constant toast using `:time-out 0` - recommend that `:close-button` is also on so the toast can be (visibly) closed.

Turning off prevent-duplicates could be useful for toasting on taking tag, in case multiple tags are taken within close succession.


mtgred:
:+1:

PR #1222, opened 2016-02-14
UI fixes for start-turn changes.

nealterrell:
Fix runner seeing "Take Clicks" while corp is in his Start of Turn phase.

Fix the order of corp's "No Action" / "Action before access" buttons for consistency (No Action always in the same vertical position).

Add log messages for Daily Casts, Data Folding, Underworld Contacts when used in Step 1.2.

Give the Step 4.3 "Take actions, then click Done" prompt for corp a negative priority, so all other prompts hide it. (Like Jackson Howard.)

Use @Saintis's new toast functionality from #1219 to give a stronger warning when corp rezzes during Step 1.2, as a warning toast that does not go away until clicked.

### DIFF
--- a/src/clj/game/cards-resources.clj
+++ b/src/clj/game/cards-resources.clj
@@ -113,36 +113,33 @@
                  :effect (effect (trash card {:cause :ability-cost}) (damage-prevent :meat 3))}]}
 
    "Daily Casts"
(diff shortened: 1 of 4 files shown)


mtgred:
:+1:

PR #1225, opened 2016-02-14
Feature version check

domtancredi:
This is really close and should be tested. It checks for version (in a new version endpoint), refactors toast into its own script, and has a small component for comparing version and displaying a toast if they don't match.

There's some issue with how to check app-state with the version number in the jade file so that's something to look into.


domtancredi:
Based on the feature branch made before a bunch of updates to master, it may make more sense to drop and re-build the feature branch from a new place in master

zaroth:
yeah, I would hate to lose the sweet upgrades made to toasts that @saintis coded today.

Saintis:
How easy would it be to bring this PR up to speed with `master`?

domtancredi:
Not too bad. The changes were minor, on the template level as well as the server-side area (didn't touch core game engine).

domtancredi:
Done. Could use a review @Saintis

Saintis:
It doesn't look as though the toast function correctly allows the toaster-options to be passed in. 

Also, does the version number have to be manually updated or can it be done automatically somehow?

src/cljs/netrunner/main.cljs
```diff
@@ -28,6 +31,15 @@
 (.setPathPrefix history "")
 (.setEnabled history true)
 
+(go (while true
```
nealterrell: This loop is currently in `gameboard.cljs` -- why is it here too?

nealterrell:
RE: version number, he is loading it server-side from `app.locals.version`, which is read from the environment variable `APP_VERSION`. @mtgred updates this by hand when deploying a UI update.

How exactly does this work, @domtancredi? When does the warning about bad version show up?

nealterrell:
We can revisit this with the new Clojure server anytime, but the actual code here is dead I think.

PR #1228, opened 2016-02-15
Help page update and CSS consistency fix for room tabs

zaroth:
Pretty self explanatory, made hover behavior of room tabs consistent with the rest of the site.

Help question about 4.3 run window is removed thanks to @nealterrell's implementation :)


mtgred:
:+1:

PR #1231, opened 2016-02-15
Deck collection "Loading..." message

zaroth:
Hopefully should cut down on the number of people complaining about losing their decks.

Tested and it's working, but since it displays only for a short moment, it's tricky to screenshot.


mtgred:
:+1:

PR #1232, opened 2016-02-15
Adam deckbuilding fixes

zaroth:
- a persistent toast message when starting a game without enough directives
- validity check for directives in deckbuilder and minimum 48 cards for Adam's decks
- warning messages when the constraints aren't met

### DIFF
--- a/src/clj/game/cards-identities.clj
+++ b/src/clj/game/cards-identities.clj
@@ -8,7 +8,7 @@
    {:effect (req (let [directives (-> (:deck runner) (concat (:hand runner)) (get-directives) one-of-each)]
                    (when (not= 3 (count directives))
                      (toast state :runner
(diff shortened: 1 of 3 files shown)


mtgred:
The fact that the 3 directives have to be added manually to each Adam deck has always bothered me. An improvement would be to add directives on the board at the beginning of the game without requiring them to be in the deck.

zaroth:
@mtgred: When @nealterrell was asked about this, he replied:

> the game engine doesn't have the ability to load cards from thin air; the web server sends the entire decklist to the game server when creating a game, and those are the only cards the server knows about. This will need to change for the spoiled card Rebirth

I could do this, but with my current knowledge I'd be a bit out of my depth, so I'm waiting until I get more experienced with the codebase.

PR #1238, opened 2016-02-15
Implemented some cards from Mumbad

zaroth:
Picking some low-hanging fruit:
- Brainstorm
- Panchatantra
- Advanced Assembly Lines
- Zealous Judge (no tagged rez restriction yet)
- Vanilla
- Merger
- Improved Protein Source
- Pālanā Foods

Since some of these cards were added to NRDB very recently, a `coffee fetch.coffee` may be needed to get them.


mtgred:
:+1:

PR #1242, opened 2016-02-16
Fix deck collection formatting issue with long ID names

zaroth:
Primarily to solve this issue:

capture

Solved by changing format of displayed date for decks to ISO format, e.g. 2016-02-16. Hour probably isn't very important in this case, so it was stripped.


mtgred:
The ISO date format looks like a technical thing. I prefer to keep the current date format which look more user friendly.

But you are right that the hour is not very relevant. So i stripped it here: cb0a0967910679a40c109861e7cb9ada544da30a

PR #1243, opened 2016-02-16
Implemented some more cards from Mumbad

zaroth:
Currently in this branch:

- Palana Foods name fix (after NRDB updated)
- rez requirement for Jealous Judge
- Sports Hopper
- Diwan (purge+label), install cost manual


mtgred:
:+1:

PR #1245, opened 2016-02-17
Added deck validity tooltip with more info on hover

zaroth:
status-tooltip


nealterrell:
Sexy.

mtgred:
Nice :+1: 

Some remarks on the styling: the shadow is too strong, there is no contrast between the blue popup with the blue background and the dark blue border is not consistent with the border of other panel. 

I will correct that.

PR #1251, opened 2016-02-18
Full subliminal messaging implementation

kevkcc:
This is one of the most annoying cards to play currently (at least for me) since you have to manually drag back to HQ. This is a first attempt to fully automate this. Since Subliminals stack, I made it so there are three scenarios:

1 Subliminal in Archives: Yes, No prompt
2 Subliminals: Choice between getting 0, 1 or 2 out at once
3 Subliminals: Choice between getting 0, 1, 2 or 3 out at once

Since this card is not active when the effects occur, I have implemented it another way. I am unsure about this method so I would appreciate a thorough code review.

Also I'll write a test for it after some feedback.


zaroth:
I don't think putting that much logic specific for a single card into core is the way to go. Further discussion on slack.

kevkcc:
Thanks for the feedback. I definitely agree that it wasn't the way to go but couldn't think of anything better at the time. Based on the feedback, this is my next attempt. I'm finding this card fairly challenging and unfortunately, it doesn't work when Subliminal isn't played (i.e. Noise milled it last turn). Any suggestions?

kevkcc:
I'm pretty happy with the state of this now. Also added a new event hook :mill-effect to check for a mill. Not sure how useful it would be but I found I needed something like that for this card.

zaroth:
This certainly looks much cleaner than when I last looked at it :+1:. I like keeping it simple, so the approach with yes/no prompts per card gets a thumbs up. Haven't tested it yet locally, but a question - what happens if subliminal gets shuffled back into R&D/removed from the game? (Jackson Howard/Archives Interface). Does the prompt keep appearing?

kevkcc:
The prompt won't keep appearing but it's not fully ideal at the moment either since the :corp-turn-begins event isn't properly cleared when subliminal gets jackson'ed. I actually asked about this on Slack as well but never got a response. If I could get a bit of help to resolve this, I think it would the implementation could be even better yes.

queueseven:
Hey @kevkcc I suggest you close this PR and create a new one with a single commit, so it would be easier for use to actually see what changes are made. I don't think this PR has a chance to be merged right now.

PR #1264, opened 2016-02-21
Add several Runner event messages; fix Hacktivist Meeting

JoelCFC25:
Follow-up to previous PR that added messages for Corp operations--several Runner events will now print their effects to the game log instead of occurring silently. 

This also fixes the Hacktivist Meeting problem (responding to rez of ICE) introduced in #1243. Tallie Perrault is refactored to correctly be optional for the Runner. Added a little stub for Run Amok to at least let it initiate a run and print to the log what can happen when the run ends.


mtgred:
:+1:

PR #1267, opened 2016-02-22
Minor fixes

### LINKED ISSUE #1054: Same Old Thing Freezing Game
I was playing MaxX against RP and tried to use SoT to legwork.  I clicked legwork and it started a legwork run but the SoT menu asking me to select an event to play remained.  I was unable to close the menu and had to concede the game.

I tried adding clicks and clicking a different event as well as hitting 'done' without selecting an event.

It is possible that I clicked SoT twice due to a small amount of lag.

zaroth:
- use properly capitalized `maxLength` so React recognizes it
- fix trace log messages not using `msg` when available
- SOT fix #1054

### DIFF
--- a/src/clj/game/cards-resources.clj
+++ b/src/clj/game/cards-resources.clj
@@ -656,7 +656,9 @@
 
    "Same Old Thing"
    {:abilities [{:cost [:click 2]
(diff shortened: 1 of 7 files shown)


mtgred:
:+1:

Issue #1260, opened 2016-02-19
Turntable used to give Mandatory Upgrades to Corp doesn't add clicks-per-turn for Corp


zaroth:
Apparently Joel in #1183 didn't foresee that someone might actually want to GIVE someone mandatory upgrades via turntable (and I don't blame him for that, it's a super obscure case). In the case where ManUps are turntabled away for another copy of ManUps, the clicks are properly taken away for the stolen ManUps, but they aren't given back for the one that's given in exchange.

JoelCFC25:
It's actually worse than you describe. First, giving the benefit of a ManUps swapped back to the Corp is probably the same bug as #1125. 

Next, I kind of punted on deactivating the agenda swapped away from the Corp and did a `doseq` deactivate on *everything* in the Runner's score area. So if you either steal or swap a ManUps first and then do more Turntabling later in the game, ManUps (and each copy if there are multiple) will drop the Corp's clicks per turn every time. So this needs to narrow things down to find the precise `:cid` of what was just swapped.

Issue #1125, opened 2016-01-19
Turntable Turns off 24/7 News Cycle


ANRecho:
During recent games I have noticed that if turntable is used to swap an agenda out of the corp's score area, 24/7 news cycle is unable to target any agenda in the score area for the remainder of the game.

JoelCFC25:
Thanks for the report. This is due to the gymnastics in the Turntable code where I attempted to get events and abilities hooked up to the Corp agenda newly swapped into the score area by Turntable (e.g., if the Corp gets a Gila Hands swapped back to them, it needs to subsequently be clickable to gain 3 credits for 2 clicks). 

For agendas like Breaking News that have no click abilities, it doesn't seem to be working, because I see the card data changes to `:abilities nil` in the game state after swapping a stolen Breaking News back to the Corp. 24/7 News Cycle needs something to be in `:abilities` in order to be able to target it.

Right now I have no idea what to do on this one...

JoelCFC25:
I will have a partial fix for this soon. It still most likely won't work to target a specific card previously Turntabled back to you with 24/7, but cards you scored normally will no longer be rendered untargetable.

JoelCFC25:
I'm going to close this for now, as it's been 95% fixed. As noted above, 24/7 will still not be able to target an agenda that was previously Turntabled to the Corp, but that can become it's own issue if someone ever manages to get in that situation.

PR #1277, opened 2016-02-24
Assorted fixes: Turntable, Sneakdoor, Film Critic, 24/7 News Cycle

### LINKED ISSUE #955: Nerve Agent does not get virus counters when access is through card abilities

### LINKED ISSUE #1042: Trashing Director Haas scores it even if not accessed
if Director Haas is trashed from HQ without being accessed, it will still be scored as an agenda.

### LINKED ISSUE #1158: Sneakdoor beta +nerve agent
i did sneakdoor run, nerve got 1 counter, ok so far, 2nd run with sneakdoor, i got 1 acess only, nerve didnt get any counters, i tried again, 1 access, i tried again, 0 cards accessed??

should have been 1,1+1,1+2

### LINKED ISSUE #1181: Discarding Chairman Hiro after hosting an agenda on Film Critic
the corp player can´t discard chairman hiro without automatically putting it into the runners score pile.

### LINKED ISSUE #1276: Public support is repeatedly scored at the beginning of corp turn after turntabling
Public support scored. Agenda also turntabled later in game. Beginning of corp turn, Pub Support is scored again at the beginning of every corp turn.

JoelCFC25:
I've collected another strange mish-mash of fixes. We're going deep into the weeds on some of these...

* Fix #1181, fix #1042: After using Film Critic, `:access true` was incorrectly lingering on in the state, resulting in the "as agenda" executives teleporting to the Runner score area and granting them 2 points even if voluntarily trashed by the Corp. Includes a test.
* Fix #955, fix #1158: This is a bit of a hack, but triggering a fake `:no-action` in Sneakdoor Beta will allow Nerve Agent to gain counters. The Runner can't use the "choose fewer accesses" option even though they'll still see a toast about it, but this is better than nothing. Includes a test.
* Fix #1276. I can only partially improve the situation for #1260 and #1125. I wrote a new function to find a single `:cid` so that Turntable can activate/deactivate *only* the 2 cards that are swapped instead of everything in each score area. 24/7 News Cycle still won't be able to target an agenda that has previously been swapped back to the Corp. 
* Speaking of 24/7 News Cycle, it should be able to "top up" agenda counters on things like House of Knives, Geothermal Fracking, etc. This is probably pretty rare, but it doesn't work at all with these agendas currently using the form `:data X`. Changing to `(effect (add-prop card :counter X))` allows it.

### DIFF
--- a/src/clj/game/cards-agendas.clj
+++ b/src/clj/game/cards-agendas.clj
@@ -59,7 +59,7 @@
    {:events {:jack-out {:msg "do 1 net damage" :effect (effect (damage :net 1))}}}
 
    "AstroScript Pilot Program"
-   {:data {:counter 1}
+   {:effect (effect (add-prop card :counter 1))
     :abilities [{:counter-cost 1 :msg (msg "place 1 advancement token on "
                                            (card-str state target))
                  :choices {:req can-be-advanced?}
@@ -187,7 +187,8 @@
                                 :effect (effect (ice-strength-bonus 1 target))}}}
 
    "Executive Retreat"
-   {:data {:counter 1} :effect (effect (shuffle-into-deck :hand))
+   {:effect (effect (add-prop card :counter 1)
+                    (shuffle-into-deck :hand))
     :abilities [{:cost [:click 1] :counter-cost 1 :msg "draw 5 cards" :effect (effect (draw 5))}]}
 
    "Explode-a-palooza"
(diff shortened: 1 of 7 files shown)


mtgred:
:+1:

PR #1280, opened 2016-02-26
Fix Edward Kim; add toast reminder for which Jinteki Biotech version was chosen

### LINKED ISSUE #1107: Edward Kim doesn't have his ability disabled after accessing operations in Archives
There is some code in Kim's ID that would suggest such an interaction with trashed operations was intended:
> `:events {:access {:once :per-turn`
> `:req (req (= (:type target) "Operation"))`
> `:effect (effect (trash target))`
> `:msg (msg "trash " (:title target) (if (some #{:discard} (:zone target)) ", but it is already trashed."))}}}`

It doesn't work though, since currently cards in archives aren't accessed per se - only agendas and specific traps firing from archives are.

JoelCFC25:
* Fix #1107: Uses the turn flag system to disallow Edward Kim's ability if the first operation accessed in a turn was in Archives. Includes a test, but admittedly there is a somewhat-remote corner case not accounted for--if you pop Hades Shard and there are operations in Archives, the ability would still work if you subsequently accessed one in HQ or R&D. 
* Add an ability to Jinteki Biotech to toast yourself a reminder of which version you picked. I've fielded a few requests for this from players who forget over the course of a long game. 

image

### DIFF
--- a/src/clj/game/cards-identities.clj
+++ b/src/clj/game/cards-identities.clj
@@ -184,8 +184,17 @@
                               :choices ["[The Brewery~brewery]" "[The Tank~tank]" "[The Greenhouse~greenhouse]"]
                               :effect (effect (update! (assoc card :biotech-target target))
                                               (system-msg (str "has chosen a copy of Jinteki Biotech for this game ")))}}
-    :abilities [{:cost [:click 3]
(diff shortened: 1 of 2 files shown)


mtgred:
:+1:

PR #1283, opened 2016-02-26
Fix #1273, Globalsec Security Clearance not causing a lost click.

### LINKED ISSUE #1273: Globalsec Security Clearance
It doesn't appear to subtract a click from the Runner.

nealterrell:
W/ test.


mtgred:
:+1:

PR #1284, opened 2016-02-26
Fix recurring credits not refilling during Step 1.2.

nealterrell:
These credits should refill during Step 1.2 before the user can use any other abilities. Without this fix, they refill after Step 1.2 is finished, and so aren't available to use during that step if they were used on the previous turn.


mtgred:
:+1:

Issue #1275, opened 2016-02-23
Suggestion - Need a way to bypass rez requirements


Sonnydk:
It would be nice with a rez command that would ignore cost and just rez a card.

Had a game where my opponent used a Datasucker too early and trashed my Archer. I wanted to rez a Cyberdex suite and purge during approach, preventing the trashing during the encounter phase.

I couldn't find a way to restore the gamestate as I didn't have an agenda to forefeit. Had to call it a draw. Bummer..

nealterrell:
Good idea, we'll add this. This may not have applied, but if you had sacrificed an agenda to rez Archer earlier, that agenda would have been in the "Removed from game" zone -- you can always drag that back to your scored zone manually so you can sacrifice it again. (I think... I know you can at least drag it to your hand, then install and score it, but that could have bad side effects.)

Sonnydk:
When it happened, the only agenda I had scored was a Chronos Project, so it would have a rather large impact on the game if I were to score it again artificially.

While you're implementing this feature you can also consider a "Rez all cards" option. when the game is over, sometimes you want to show your opponent what cards you had played.

zaroth:
Dragging a card to score area doesn't trigger their "when scored" effects (re: chronos project).

nealterrell:
Coming in #1285.

PR #1285, opened 2016-02-26
Add /rez and /rez-all chat commands as per #1275.

nealterrell:


zaroth:
The new commands could use a little documentation in the `help.cljs` :)

nealterrell:
You got it :). :+1:

mtgred:
:+1:

Issue #1270, opened 2016-02-22
Hostile Infrastructure + Apocalypse should be once damage chunk


Krystman:
Hostile Infrastructure triggers wrong on multiple simultaneous trashes (eg. Apocalypse). The net damage should be dealt in one chunk. Right now, the net damage is dealt separately. Which means I've Had Worse triggers during an Apocalypse + Hostile Infrastructure. See here

https://boardgamegeek.com/article/15843899#15843899

nealterrell:
Could probably extend `trash` to take a parameter that suppresses the `:runner-trash` event, then in Apocalypse/Singularity trigger a single `:runner-trash` with a long list of targets trashed. Hostile would then need to check how many targets were trashed to calculate its damage amount.

nealterrell:
Coming in #1288.

zaroth:
I've Had worse will trigger anyway, since nothing is stopping it right now from firing when flatlining. See #1050.

nealterrell:
Sure, but that's fine. The issue here is that 5 individual 1-net damages won't flatline a Runner with 4 cards including IHW; and also that a non-flatline amount of damage can trigger IHW on the first damage, then cause the cards drawn by IHW to be discarded for the next damage. #1288 fixes that.

Saintis:
The previous interaction is correct, so the change should be reverted, see Jako's relevant comment on the progress report [here](https://www.reddit.com/r/Netrunner/comments/49jakn/jintekinet_quarterly_progress_report_winter_2016/d0shs1p).

Krystman:
Sorry for the confusion guys. I was under the impression that "simultaneously" meant it would resolve like Psychic Field. I apologize.

PR #1288, opened 2016-02-26
Fix interaction between Hostile Infrastructure and Apocalypse/Singularity

nealterrell:
See #1270. These cards now cause one bulk `:runner-trash` event to trigger by suppressing the event for each individual card that gets trashed. Hostile Infrastucture looks at how many cards were trashed by the event to calculate the damage.

Also fixes Singularity to be a mandatory on-access effect, instead of prompting for "Use card ability".


mtgred:
:+1:

Issue #1252, opened 2016-02-19
Deus X and multiple counts of net damage


PhantomOfAres:
While playing against Personal Evolution, if the runner has Deus X on the board, and accesses Fetal AI, the runner is presented with the option to prevent the first 2 net damage, from Fetal.  If the runner trashes Deus X, it prevents not only the damage from Fetal, but ALSO the damage from Personal Evolution.  This seems to happen with all amounts of net damage (ie multiple Komainu fires stacked up on one another.)  The only interaction I see it as a particular problem is Fetal + PE, since in other cases, it may behoove you to wait to prevent until the final one is fired, anyway.

nealterrell:
Thanks for the report. This would require a rewrite of Film Critic, as that card is why Fetal AI technically is not accessed until the runner chooses to Steal, so the two damage events are happening in the same window rather than in two different windows. Until then, the `/take-net 1` command can help.

PR #1289, opened 2016-02-26
more Mumbad spoilers (seeking review)

JoelCFC25:
Some notes on card choices...

**Councilman**: Based on early feedback, this doesn't automatically throw an optional use prompt at the Runner for every qualifying rez to avoid slowing the game down. Instead the Runner gets a toast only if they can afford to pay the cost, and then they just click to use. However, to avoid snafus it doesn't do tight enforcement of what you can target (beyond that it's a rezzed asset/upgrade), so players need to be accountable to each other.

**Guru Davinder**: Corp players need to slow down when applying multiple damage subroutines (e.g. Komainu) and let the Runner decide to whether or not to pay to keep it in play (could add a wait prompt for the Corp here). This will probably have a bad interaction with Gemini and have the same problems as Deus X highlighted in #1252.

**Salem's Hospitality**: The Runner gets a prompt that will perform the reveal of their Grip, but they should wait to use it until the Corp has named a card in game chat. Then any required trashing can be verified.

### DIFF
--- a/src/clj/game/cards-events.clj
+++ b/src/clj/game/cards-events.clj
@@ -428,6 +428,11 @@
    {:msg "gain 9 [Credits]"
     :effect (effect (gain :credit 9))}
 
+   "Making an Entrance"
+   {:msg "look at and trash or rearrange the top 6 cards of their Stack"
+    :effect (req (toast state :runner "Drag remaining untrashed cards from the Temporary Zone back onto your Stack" "info")
+                 (doseq [c (take 6 (:deck runner))] (move state side c :play-area)))}
+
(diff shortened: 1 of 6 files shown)


mtgred:
:+1:

mtgred:
Something really cool for Salem Hospitality would be an input box with autocomplete (like in the deckbuilder) to select the card. That input box could be used for Targeted Marketing too.

JoelCFC25:
...and Panchatantra.

nealterrell:
And Rebirth!

Saintis:
:+1:

PR #1290, opened 2016-02-26
Fix Sneakdoor Beta interaction with Crisium Grid, #1229.

### LINKED ISSUE #1229: Crisium Grid doesn't fire vs. Sneakdoor Beta
As the title reads, more or less! Crisium grid installed and rezzed on Archives should negate the "if successful, treat as a run on HQ" clause on Sneakdoor Beta, but does not.

nealterrell:
Don't fire a card's `:successful-run` ability (which is NOT an event that could've been suppressed by Crisium Grid, but is instead encoded into the `[:run :run-effect :successful-run]` state) if the `:successful-run` event would have been suppressed by some card source (namely Crisium Grid). Fixes #1229. Includes test.

### DIFF
--- a/src/clj/game/core-runs.clj
+++ b/src/clj/game/core-runs.clj
@@ -387,8 +387,10 @@
 (defn- successful-run-trigger
   "The real 'successful run' trigger."
   [state side]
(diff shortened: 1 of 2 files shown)


mtgred:
:+1:

PR #1294, opened 2016-02-29
Server run restrictions

Saintis:
Implements Replicating Perfection and Off the Grid. Prevents runs on blocked servers, and removes these servers from the run list (when clicking "Run" button).

Adds tests for Off the Grid and Replicating Perfection (and their interaction).

Note that the implementation of this flag system means that multiple cards can block the same server without the blocking being overridden. (Think RP and Off the Grid interaction).

Screen shot:
screen shot 2016-02-29 at 23 29 20

NB: Have not tested what happens in run selection for run events.


mtgred:
Awesome :+1:

Saintis:
That was quick :smile: 

Also, seems test is breaking... Will investigate.

PR #1298, opened 2016-03-02
Runs from card effects + Replicating Perfection; fix Off the Grid persistence

JoelCFC25:
This is a follow up to the Jinteki: RP and Off the Grid implementations in #1294. 

Added `runnable-servers` to the `req` macro so all cards that initiate runs (mostly run events plus a couple pieces of hardware and Jak Sinclair) can request a valid list of server targets. Off the Grid also was not persisting beyond the turn in which it was rezzed, so it fires back up at the start of every Runner turn now (updated the unit test accordingly).


mtgred:
:+1:

PR #1299, opened 2016-03-02
Add score/rez restrictions to Mushin No Shin and Dedication Ceremony targets

### LINKED ISSUE #1119: Mushin No Shin doesn't restrict rez or score restriction
I wouldn't put this as a high priority, but after thinking about the reasoning for #1118, it's probably good to enforce the restriction as well

JoelCFC25:
Fix #1119.

This extends the flag system to checking for ability to score agendas. Includes a Mushin No Shin test that validates both the same-turn scoring and rezzing restrictions on the installed card.


mtgred:
:+1:

PR #1300, opened 2016-03-03
fix High-Stakes Job identification of valid servers

### LINKED ISSUE #1297: Adjusted Chronotype does not prevent losing a click with Troll encounter
If I am not mistaken, Adjusted Chronotype should prevent losing a click when the on encounter ability of Troll fires and the runner loses the trace.

Now, I ran on my last click to R&D, while having a Starlight Crusade Funding and a Gene Conditioning Shoppe, so I already lost a click, but could have used Adjusted Chronotype again. I clicked on the "Lost 1 click" prompt, but it ended the run instead saying that I couldn't lose a click.

JoelCFC25:
The edit made to High-Stakes Job in #1298 resulted in an empty set of valid server targets and the event just going to the Heap with nothing else happening. Oops! 

Fix #1297: Runner should gain back a click if Adjusted Chronotype is installed and they lose a click from Troll.

### DIFF
--- a/src/clj/game/cards-events.clj
+++ b/src/clj/game/cards-events.clj
@@ -282,8 +282,8 @@
    "High-Stakes Job"
    {:prompt "Choose a server"
     :choices (req (let [unrezzed-ice #(seq (filter (complement rezzed?) (:ices (second %))))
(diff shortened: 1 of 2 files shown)


mtgred:
:+1:

PR #1303, opened 2016-03-04
implement Genetics Pavilion

JoelCFC25:
Seeking review on this one from the usual suspects...

I added a couple new draw-related functions and modified `draw` itself. Now the system will both track the number of cards drawn in a given turn and--if a maximum has been set--the number of remaining draws allowed. Then `draw` checks if the number passed to it needs to be reduced or not. 

Includes a test for the Corp rezzing Genetics Pavilion both prior to the Runner turn starting and after they've already drawn 2 or more cards to ensure that it accounts for prior draws when enforcing the maximum. I also tested interactions with I've Had Worse being hit by damage, allowing more than 2 Runner draws during the Corp turn (e.g., Sports Hopper), and that the maximum is lifted immediately if the Runner trashes GP.

### DIFF
--- a/src/clj/game/cards-assets.clj
+++ b/src/clj/game/cards-assets.clj
@@ -292,6 +292,14 @@
                       :msg "add it to their score area and gain 1 agenda point"
                       :effect (effect (as-agenda :corp card 1))}}}
 
+   "Genetics Pavilion"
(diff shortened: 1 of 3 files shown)


mtgred:
Awesome :+1:

PR #1305, opened 2016-03-04
Two fixes for Leela

### LINKED ISSUE #1271: Leela and HQ Multiaccess with bounce effects.
If there are less cards in the Corporations HQ/hand prior to accessing than the total amount Leela is allowed to access (HQI's, Legwork) and if an agenda is stolen, Leela cannot access additional cards up to the allowed maximum after a bounce.

eg. Corp has 2 cards in hand, Leela has 2 HQI's installed on the board.
Successful run on HQ results in accessing cards, Leela steals an agenda on the first access, bounces a card to HQ. A second access is triggered, but a third is not.

Run/access on HQ prematurely ends.

nealterrell:
1. If Leela has a multi-access on HQ and accesses an agenda first, she must immediately bounce an unrezzed card to HQ. That card is supposed to be eligible for the remaining access attempts, but the access code selects all the cards to be accessed the moment accessing begins; the "next card from HQ" UI is just a trick, as the cards to be accessed have been pre-determined, and any card added to HQ after that list is created would never be accessed. I reworked HQ access to pick a random card from whatever is currently in HQ with each access attempt.

    For this to work best, I added a new flag to Leela: `:slow-hq-access`. When this flag is true, then any HQ access will always show the "Select a card to access" prompt, even if the runner can only access one card. (Without this flag, if only one card is going to be accessed, it will be accessed immediately with no UI interaction.) Leela needs this to force the UI prompt to slow down access, allowing her to bounce whenever necessary. I think I can port this to Gagarin so that a runner can turn down accessing a single-card remote.

    Fixes #1271.

2. There is a specific interaction with Gang Sign by which Leela should be able to bounce more than one card during one access phase. If Leela has Gang Sign and HQ Interface installed, and the corp scores an agenda, Leela gains +1 bounce but can choose to Gang Sign first. If during Gang Sign she accesses another agenda, she must immediately bounce a card, but the original bounce that has been saved is currently being lost. I changed the card logic to keep track of a # of bounces allowed, rather than a boolean yes/no.

    Fixes a bug reported on reddit.

### DIFF
--- a/src/clj/game/cards-identities.clj
+++ b/src/clj/game/cards-identities.clj
@@ -266,17 +266,17 @@
    {:events {:agenda-scored
              {:effect (req (toast state :runner
                                   (str "Click Leela Patel: Trained Pragmatist to add 1 unrezzed card to HQ.") "info")
-                           (update! state :runner (assoc card :bounce-hq true)))}
+                           (update! state :runner (update-in card [:bounce-hq] #(inc (or % 0)))))}
              :agenda-stolen
              {:effect (req (toast state :runner
                                   (str "Click Leela Patel: Trained Pragmatist to add 1 unrezzed card to HQ.") "info")
-                           (update! state side (assoc card :bounce-hq true)))}}
-    :abilities [{:req (req (:bounce-hq card))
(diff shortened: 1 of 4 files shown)


mtgred:
:+1:

PR #1306, opened 2016-03-04
Allow runner to deny accessing a 1-card remote vs Gagarin

### LINKED ISSUE #1003: Gagarin forcing access one single-card remotes
When I run against Gagarin, it seems that if there is just one card in a remote, the game forces me to access that card. My understanding of the rules is that I can decline to access because there is an additional cost, but still get credited with a successful run. This is important for things like Desperado and Dirty Laundry, where you may just be in it for the money.

If there is more than one card, then I can access them piecemeal as I choose, which does seem right.

nealterrell:
Runner should be able to avoid accessing a 1-card remote vs Gagarin, but current access code speeds up remote access vs. a 1-card remote. This slows down remote access vs. Gagarin, using the framework from #1305. Fixes #1003.


mtgred:
:+1:

PR #1308, opened 2016-03-04
refactor Haarpsichord Studios and Old Hollywood Grid

### LINKED ISSUE #1091: Haarpsichord Studios + Gang Sign interaction is flawed
Haarpsichord Studios prevents a steal if the Runner has stolen an agenda on his last turn but then manages to steal one during the subsequent Corp turn (most probably by Gang Sign, but there are other ways).  

The `:req` on Haarpsichord is simply looking to see if `:stole-agenda` exists in `[:runner :register]`, a value that obviously needs to persist for things like Punitive Counterstrike. It needs to be tightened up somehow or made to look for a different condition that is then reset at the end of the Runner's turn.

JoelCFC25:
Fixes #1091. 

The old Haarp implementation was carrying over steal prevention into the Corp's turn because it was going into the Runner's register. Haarp is refactored to use the turn-flag system, and as a result of changing `handle-access` Old Hollywood Grid gets refactored to a much simpler form as well. 

This also makes Gang Sign use a toast instead of the now old-fashioned "print to the log" effect.

### DIFF
--- a/src/clj/game/cards-identities.clj
+++ b/src/clj/game/cards-identities.clj
@@ -131,8 +131,14 @@
    {:effect (effect (gain :credit 5 :bad-publicity 1))}
 
    "Haarpsichord Studios: Entertainment Unleashed"
-   {:events {:pre-steal-cost {:req (req (:stole-agenda runner-reg))
-                              :effect (effect (prevent-steal))}}}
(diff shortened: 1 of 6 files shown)


mtgred:
:+1:

Issue #1161, opened 2016-01-28
Dragging parasite from hosted ICE back to hand puts ICE in runner's hand


shosuroyokaze:
If I host parasite on a piece of ice but then want to take it back (because I can never keep track of my mem, it seems....) It will put the ice in my hand and the parasite in my heap. I seem to be pretty good at reproducing it since I've done it twice now.

zaroth:
This one could be prevented by making Runner unable to drag & drop `:corp` cards and vice versa, similar to what was done for the spectator exploit. What do you think, @saintis, are there any reasons not to do it?

Saintis:
Should be ok.

queueseven:
> This one could be prevented by making Runner unable to drag & drop :corp cards and vice versa

Beware of cases where the Runner has to actually move `:corp` cards around (e.g. Indexing).

PR #1315, opened 2016-03-07
WIP Drag and drop permissions

meatcompute:
Strangely, when I have set `{:draggable false}` [here](https://github.com/mtgred/netrunner/compare/master...meatcompute:drag-and-drop-permissions?expand=1#diff-36aa547a3c22ed5b22e8b3491f0ec7ceR445) with the changes in this branch, drag and drop is still enabled for cross-side interactions when I tested this branch locally. This could be an impl. mistake and I'd be interested in feedback and more testing of this branch outside my local network.

However, if {:draggable false} isn't sufficient, it looks like spectator interactions are primarily disabled [here](https://github.com/mtgred/netrunner/blob/master/src/cljs/netrunner/gameboard.cljs#L158) where clicks for spectators are not handled at all.

I hesitant to add more cases `handle-card-click` for runners and corps, as the simple case would break cards like  and it sets us up to introduce complexity on the order of every card in the game.

Hopefully I'm missing something simple with regard to setting `{:draggable false}` and #1161 can still be achieved without massive additions.

### DIFF
--- a/src/cljs/netrunner/gameboard.cljs
+++ b/src/cljs/netrunner/gameboard.cljs
@@ -86,25 +86,48 @@
         (reset! lock false)
         (case (:type msg)
           ("do" "notification" "quit") (do (swap! game-state (if (:diff msg) #(differ/patch @last-state (:diff msg))
-                                                                             #(assoc (:state msg) :side (:side @game-state))))
+                                                                 #(assoc (:state msg) :side (:side @game-state))))
                                            (swap! last-state #(identity @game-state)))
           nil))))
 
 (defn send [msg]
   (.emit socket "netrunner" (clj->js msg)))
 
-(defn not-spectator? [game-state app-state]
-  (#{(get-in @game-state [:corp :user]) (get-in @game-state [:runner :user])} (:user @app-state)))
+(defn spectator? [game-state app-state]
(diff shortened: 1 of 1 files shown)


Saintis:
If I remember correctly, `{:draggable true}` turns on the card following mouse while dragging behaviour, and doesn't affect any logic.

It's probably best (or simpler) to disable card moving from the wrong side / player on the server-side.

meatcompute:
Agreed, i'll start looking around

nealterrell:
So what exactly is going on in this PR?

meatcompute:
This PR is in the process of addressing the bugged interaction in #1161. Following the fixes for the spectator exploit mentioned, I made some additions to the frontend that were insufficient. I put this up for feedback on how to address unintended cross-side dragging of cards.

I'm going to be looking at a way to resolve this server-side, considering that the {:draggable} key on cards components is not sufficient for fixing the behavior listed in the issue.

Saintis:
Any progress on this?

meatcompute:
Sadly no, I haven't had time to work on it. I'm happy to pass it on and/or close out this pr.

Saintis:
I'll take over this then.

Saintis:
Should have now been implemented in #1743. Thanks for the start!

PR #1316, opened 2016-03-07
Enable Browser Console Logging by Default in Development

meatcompute:


nealterrell:
OK, so this lets `print`-like functions in ClojureScript to print to `console.log` in JavaScript, instead of doing nothing? Cool.

meatcompute:
Yeah, I mostly use this to `println` intermediate values. Found myself adding it to feature branches I was working on so I threw it up in its own pr

mtgred:
:+1: 

`prn` is useful for displaying Clojure data structures btw.

PR #1320, opened 2016-03-08
Business First cards

JoelCFC25:
Adds 9 new cards from Business First and a fix for Palana Foods not giving the Corp a credit on the Runner's first draw of a turn. There is also a minor tweak to `corp-install` that I discovered when testing New Construction to install Mumbad Construction Co. and rez it for free--derezzed events should not be registered if the installed card is immediately rezzed. 

I'll double back and add some unit tests soon...


mtgred:
:+1:

PR #1324, opened 2016-03-09
Added 'flipping archives faceup' effect to /rez-all command

zaroth:
In the most common use case of the command you just want to reveal the whole info after the game ends for easier post-game discussion. This change facilitates that.


nealterrell:
Nice :+1:

zaroth:
Nice catch, done!

mtgred:
:+1:

Issue #1325, opened 2016-03-09
Legwork run shows 'Runner accesses OK' as additional log


tolaasin:
Legwork run from Geist, chat says
runner accesses Enigma
runner accesses Assassin
Runner accesses OK.
Runner accesses Adonis Campaign.

tolaasin:
4 cards in HQ at the time

JoelCFC25:
Thanks for the report. We have this in already at #1247.

tolaasin:
Ok. I did search for legwork in the issues list first and it didn't come up.

JoelCFC25:
Yeah sorry, I forgot that the earlier one wasn't specific to Legwork, but the root cause I think is the same--multi-access of HQ. I'm hoping both of these will be taken care of by #1328.

JoelCFC25:
I am pretty confident this is fixed now, but please report it if you see it again!

zaroth:
Main issue in #1662.

Issue #1247, opened 2016-02-18
Accessing "OK" when accessing HQ


zaroth:
A bug spotted in a spectated game, the second time I saw it happen with Leela:

Corp advances a card in Server 1.
Corp advances a card in Server 1.
Corp advances a card in Server 1.
Corp uses AstroScript Pilot Program to place 1 advancement token on a card in Server 1.
Corp scores Global Food Initiative and gains 3 agenda points.
Runner can access cards in HQ by clicking on Gang Sign.
Runner can access cards in HQ by clicking on Gang Sign.
Runner can access cards in HQ by clicking on Gang Sign.
Runner uses Gang Sign to access 2 card from HQ.
Runner accesses Sweeps Week.
Runner accesses Swordsman.
Runner uses Gang Sign to access 2 card from HQ.
Runner accesses Swordsman.
Runner accesses Data Raven.
Runner uses Gang Sign to access 2 card from HQ.
Runner accesses Sweeps Week.
**Runner accesses OK.**

zaroth:
It may be very hard to track, so if anyone can produce reliable reproduction steps, it'd be most welcome.

JoelCFC25:
Keep an eye out for this one--I have a strong suspicion that #1328 fixed it, though.

zaroth:
Main issue in #1662.

PR #1328, opened 2016-03-09
PRIORITY: fix HQ access with rezzed upgrades;  implement CBI Raid

### LINKED ISSUE #1332: Mumbad Construction Corp + Dedication Ceremony
When you use Dedication Ceremony on Mumbad Construction Corp and then move the tokens you just put onto a new agenda, Jnet refuses to let you score the agenda that turn. Presumably it links the DC "can't-score" status to the advancement counters instead of the card given those counters.

### LINKED ISSUE #1333: gabe and crisium grid
hello opponent had crisium grid installed in hq, i had gabe with security testing, john masonori, and can't recall if desperado. when ran hq, asked to access cris or hand, clicked crisium grid and something like an infinite list of unrezzed card in hq came up. thank you

### LINKED ISSUE #1334: Lakshmi SmartFabrics
Hi, first of all, I'd like to thank you for being so dedicated to Jinteki.net. You guys are great!

Moving on to the issue, I just finished up a game as Gagarin with two Lakshmi Smartfabrics on the board, one with 9 counters, one with 7, and I could use neither of them on the runner's turn to prevent them from accessing Project Atlas in a remote. I had Ice Wall and Project Atlas in hand ( 
ejwgdju
 ) and Atlas is worth 2 points, so I think this might be a bug. 

Clicking Lakshmi and selecting "Spend X counters" didn't change the state, including the removal of counters. 

Thanks for taking a look at this!

JoelCFC25:
The access helper function for HQ has a bunch of problems from @nealterrell's work on getting Leela bounce w/ multiaccess working. He diagnosed it in chat this afternoon. This could also potentially take care of #1325 and #1247, but we'll have to wait and see.  

CBI Raid is fairly straightforward--decided to give the Runner a wait prompt that's only cleared when the Corp signals they're done putting everything from their hand onto R&D. 

Fixes #1332, fixes #1333, fixes #1334. Dedication Ceremony had a turn-flag that was too broad--preventing *any* agenda from being scored that turn, even if it was not the card targeted by the operation. Added a few small Business First touchups: New Construction shouldn't be able to install ICE, Lakshmi gets a counter for rezzing itself and had an agenda targeting goof, Product Recall needs to factor in trash cost modifiers, Mr. Li missing event trigger to combo with Palana Foods.

### DIFF
--- a/src/clj/game/cards-identities.clj
+++ b/src/clj/game/cards-identities.clj
@@ -370,7 +370,8 @@
                               :req (req (has-subtype? target "Virus"))}}}
 
    "Pālanā Foods: Sustainable Growth"
-   {:events {:runner-draw {:msg "gain 1 [Credits]"
+   {:events {:runner-draw {:req (req (not= 0 (:turn @state)))
+                           :msg "gain 1 [Credits]"
                            :once :per-turn
                            :effect (effect (gain :corp :credit 1))}}}
 
--- a/src/clj/game/cards-agendas.clj
+++ b/src/clj/game/cards-agendas.clj
@@ -340,6 +340,7 @@
                :prompt "Install a card from HQ in a new remote?"
                :yes-ability {:prompt "Choose a card in HQ to install"
                              :choices {:req #(and (not (is-type? % "Operation"))
+                                                  (not (is-type? % "ICE"))
                                                   (= (:side %) "Corp")
                                                   (in-hand? %))}
                              :msg (msg "install a card from HQ" (when (>= (:advance-counter (get-card state card)) 5)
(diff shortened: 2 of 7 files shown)


src/clj/game/cards-identities.clj
```diff
@@ -370,7 +370,8 @@
                               :req (req (has-subtype? target "Virus"))}}}
 
    "Pālanā Foods: Sustainable Growth"
-   {:events {:runner-draw {:msg "gain 1 [Credits]"
+   {:events {:runner-draw {:req (req (not= 0 (:turn @state)))
```
zaroth: I don't think putting an exception for mulligan turn is the right approach. It'd probably be more desirable to make mulligan draws not trigger any effects, since taking new cards due to mulligan isn't really a draw in normally used game terms.

JoelCFC25: So maybe a `take`, `doseq` and `move` in the mulligan function then?

mtgred:
:+1:

PR #1329, opened 2016-03-10
Casting Call fix

### LINKED ISSUE #1128: Casting Call + Improved Tracers
Playing the combination showed +1 str bonus on tracers as if the agenda was already scored even though it was sitting in the server. The runner also had 2x tracers in their scoring area.

JoelCFC25:
Fixes #1128. 

Adds some logic to `corp-install` to only initialize events on cards being installed faceup if the actual card definition says they get installed that way (as opposed to on the fly at install time via Casting Call)--otherwise just flip them faceup and do nothing else. 

Includes a full test for using Casting Call on a Public agenda that should have working events and one that shouldn't (Improved Tracers as in the initial report).

### DIFF
--- a/src/clj/game/core-installing.clj
+++ b/src/clj/game/core-installing.clj
@@ -153,8 +153,10 @@
                  (when (= install-state :rezzed)
                    (rez state side moved-card))
                  (when (= install-state :face-up)
-                   (card-init state side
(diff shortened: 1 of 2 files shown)


nealterrell:
Nice, simple fix.

mtgred:
:+1:

Issue #1160, opened 2016-01-27
Double Hosted Icebreakers do not reset strength after encounter


Taneru:
This was the situation: I had a Corroder and Multithreader hosted on a Leprechaun, which was itself hosted on an Omni-Drive.
I boosted the strength up to 4 during a run to break through a Meru-Mati, and after passing that ice, the strength stayed at 4, rather than dropping back down to 2. I later boosted the strength a single time again in order to see if it was just a display error, and the strength further increased from 4 to 5.

Below is the /card-info details of the icebreaker itself, then the Leprechaun, and finally the Omni-Drive.

# Corroder
`Taneru shows card-info of Corroder hosted on Leprechaun hosted on Omni-Drive: {:faction "Anarch", :limited 3, :pump {:encounter 3}, :breaks ["Barrier"], :zone (:onhost), :cid 2723052, :type "Program", :abilities [{:label "Break 1 barrier subroutine", :cost ["credit" 1]} {:label "Add 1 strength", :cost ["credit" 1], :pump 1}], :title "Corroder", :counter 0, :uniqueness false, :host {:faction "Shaper", :limited 3, :hosted-programs [2723060], :zone ["onhost"], :cid 2723047, :type "Program", :abilities {:label "Install a program on [Leprechaun", :cost ["click" 1]} {:label "Host an installed program on Leprechaun"}], :title "Leprechaun", :counter 0, :uniqueness false, :host {:faction "Shaper", :limited 3, :zone ["rig" "hardware"], :cid 2723079, :type "Hardware", :rec-counter 1, :abilities [{:label "Install and host a program of 1[Memory Unit] or less on Omni-Drive", :cost ["click" 1]} {:label "Host an installed program of 1 or less on Omni-Drive"} {:label "Take 1 "}], :title "Omni-Drive", :counter 0, :uniqueness false, :host nil, :code "03039", :side "Runner", :cost 3, :installed true, :advance-counter 0, :subtype "Gear", :previous-zone ["hand"]}, :code "06019", :side "Runner", :facedown nil, :cost 2, :installed true, :advance-counter 0, :memoryunits 1, :subtype "Daemon", :previous-zone ["hand"]}, :strength 2, :code "01007", :side "Runner", :current-strength 5, :facedown nil, :cost 2, :installed true, :advance-counter 0, :memoryunits 1, :subtype "Icebreaker - Fracter", :previous-zone ["hand"]}.`

# Leprechaun
[...]

PR #1331, opened 2016-03-10
fixes for all-installed: hosted Corp cards; include Runner cards hosted on ICE

### LINKED ISSUE #1081: Testrun + Parasite
Had an issue with Testrun where I got Parasite out of my Heap onto a piece of ice. Parasite did not bounce back to the top of the Stack after the turn was completed.

### LINKED ISSUE #1227: Director Hass extra click wouldn't go away when trashed
I had a Director Hass rezzed on World's Plaza for at least one turn. The runner ran and trashed Hass, putting it in their score pile. For the remainder of the game, I still received 4 clicks per turn.

### LINKED ISSUE #1233: Fisk + Daily Business show interaction: only last draw can be put back
As per Joel's observation:
> FIS + DBS: the Corp is correctly given 4 cards, but only the final 2 are available to select as the one to put on the bottom of R&D.

JoelCFC25:
Worlds Plaza has generated a couple of fairly obscure issues. The current `all-installed` function is not counting Corp cards being hosted on things like Worlds Plaza, Awakening Center, or the upcoming Full Immersion RecStudio. As a result, copies of Daily Business Show getting hosted on Worlds Plaza aren't counting toward the extra number of cards to be drawn by the Corp. 

The other WP issue is that `:leave-play` effects aren't firing for hosted cards that get trashed. The `deactivate` function is checking the card itself to see if it's installed--I cannot for the life of me come up with a way to do this using the `installed?` function for the case of hosted Corp cards. So instead I am checking to see if the card's `:cid` is found in the now all-inclusive list from `all-installed`. 

Fixes #1233. Use the target from the `:corp-draw` event to help identify which cards in the Corp hand are eligible to be put back into R&D. 

Fixes #1227. Includes a test for the Director Haas on Worlds Plaza issue and for the new modification to `all-installed`. 

Fixes #1081: While I was mucking around with `all-installed`, I realized we should be scanning Corp ICE for things like Parasite and Caissa which weren't being counted. This led to a fix for the Test Run issue--includes a test for returning installed programs that subsequently got hosted on Runner cards or ICE. This will cover the vast majority of cases, but if `all-installed` is further modified as being discussed in #1160 to search multiple levels of nested hosting we'll have a 100% solution.

### DIFF
--- a/src/clj/game/cards-assets.clj
+++ b/src/clj/game/cards-assets.clj
@@ -180,8 +180,8 @@
               :once-key :daily-business-show
               :req (req (first-event state side :corp-draw))
               :effect (req
-                        (let [dbs (->> (:corp @state) :servers seq flatten (mapcat :content)
-                                       (filter #(and (:rezzed %) (= (:title %) "Daily Business Show")))  count)
+                        (let [dbs (count (filter #(and (rezzed? %) (= (:title %) "Daily Business Show"))
+                                                  (all-installed state :corp))) 
                               newcards (take dbs (:deck corp))
                               drawn (conj newcards (last (:hand corp)))]
                           (doseq [c newcards] (move state side c :hand))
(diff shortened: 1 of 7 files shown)


mtgred:
:+1:

PR #1337, opened 2016-03-11
Add ability to request paid ability window at end of opponent's turn

nealterrell:
Brings functionality similar to the Step 4.3 code to Step 2.2/3.2 End of Runner/Corp's Turn.

At any point during your opponent's turn, a player may click a UI button titled "Action Before Turn Ends". If this is clicked, we silently wait until the opponent clicks End Turn; at this point we follow the same sequence as Step 4.3, in which:

* the opponent player (the one ending the turn) sees a "Waiting for opponent to take an action" prompt
* the user gets a prompt to take actions before opponent's turn ends
* the opponent gets one final window to act after the user indicates he is done
* after the opponent is done, the turn ends. Events are fired, cards react, etc.

If the Action Before Turn Ends button is not clicked, the turn ends as usual.

Our UI then corresponds to the Turn Timing steps in this approximate way:

* prior to pressing Start Turn: Step 1.1 (Window before turn starts, but "my" turn for ability conditions)
* press Start Turn: Step 1.2 into Step 1.3
    * if Step 1.2 is necessary, show Mandatory Draw/Take Clicks button to end Step 1.2. 
* after Start Turn is pressed: "start of turn" event triggers. I take clicks until none are left. Show "End Turn" button.
* press End Turn:
    * if opponent requested Action Before Turn Ends, then ability window 2.2/3.2 (Window before my turn ends)
* after End Turn is pressed: "end of turn" event triggers. It is now my opponent's turn.

There are only a few interactions that *really need* this code, most notably a Haley Kaplan who wants to trigger her ability at the end of the opponent's turn, and then again at Step 1.1 (window prior to "Start of Turn") of her own turn in order to install a card with a Start of Turn trigger, like Daily Casts. 

Example: at end of corp's turn, use Clone Chip to install Multithreader from heap, and Hayley install another Multithreader from hand. Use the recurring credits during opponent's turn to pump Study Guide. After corp turn ends, pop Street Peddler to install a resource, and Haley install Daily Casts from hand. Hit Start Turn: Multithreaders refresh their credits, Daily Casts ticks down.

This requires an explicit window to act while it is still considered the corp's turn, and another window to act in 1.1 before Start of Turn happens.

### DIFF
--- a/src/clj/game/core-runs.clj
+++ b/src/clj/game/core-runs.clj
@@ -429,6 +429,7 @@
   (if (get-in @state [:run :corp-phase-43])
     ; if corp requests phase 4.3, then we do NOT fire :successful-run yet, which does not happen until 4.4
     (do (swap! state dissoc :no-action)
+        (system-msg state :corp "wants to act before the run is successful")
         (show-wait-prompt state :runner "Corp's actions")
         (show-prompt state :corp nil "Rez and take actions before Successful Run" ["Done"]
                      (fn [args-corp]
--- a/src/clj/game/core-turns.clj
+++ b/src/clj/game/core-turns.clj
@@ -135,7 +135,9 @@
                  "info")
       (end-phase-12 state side args))))
 
-(defn end-turn [state side args]
+(defn end-phase-32
+  "End step 2.2/3.2 (window before end of turn) and end the turn."
+  [state side]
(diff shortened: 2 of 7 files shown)


mtgred:
:+1:

NoahTheDuke:
Breadcrumb so I remember to check on this at some later date.

PR #1338, opened 2016-03-11
Two damage fixes

### LINKED ISSUE #1252: Deus X and multiple counts of net damage

nealterrell:
Revert fix in #1270 as per Reddit discussion with Jakodrako.

Fix #1252 by giving prevention prompts a higher priority (10) than all other prompts, forcing them to the front so they don't get inter-weaved with other prompts.

### DIFF
--- a/src/clj/game/cards-events.clj
+++ b/src/clj/game/cards-events.clj
@@ -26,9 +26,8 @@
     :effect (req (let [allcorp (->> (all-installed state :corp)
                                     (sort-by #(vec (:zone %)))
                                     (reverse))]
-                   (apply trigger-event state side :runner-trash allcorp)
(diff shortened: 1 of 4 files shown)


mtgred:
:+1:

PR #1339, opened 2016-03-11
allow Apex to install a second console facedown

### LINKED ISSUE #1326: Heartbeat
Two problems that are definitely new. It worked before. 

1) If you have Heartbeat installed, you cannot facedown install another copy. I had this happen in two games in a row.
2) The only card Heartbeat seems to accept to pay its prevent ability is itself (neither faceup or facedown cards work). This may or may not be connected to Chronos Protocol somehow, that's the only time I used Heartbeat to prevent damage this week (and it worked just fine a few weeks ago)

Earlier when you could facedown install Heartbeat, you'd get the usual loss of 1 MU (which also occurs with Q-Coherence Chip, and also hand size -3 when you facedown Brain Cage).

JoelCFC25:
Fixes #1326. 

`runner-can-install?` is modified so that the second `or` statement can come back true in the case of a 2nd console getting installed facedown when one is already installed faceup and active. Includes a test!

### DIFF
--- a/src/clj/game/core-installing.clj
+++ b/src/clj/game/core-installing.clj
@@ -167,7 +167,7 @@
    Checks uniqueness of card and installed console"
   [state side {:keys [uniqueness] :as card} facedown]
   (and (or (not uniqueness) (not (in-play? state card)) facedown) ; checks uniqueness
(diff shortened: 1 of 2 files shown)


mtgred:
:+1:

PR #1340, opened 2016-03-12
fix Troll and Sealed Vault bugs

JoelCFC25:
I goofed on the previous fix to Troll and now it's docking 2 clicks when the Runner has 2 or more remaining. Also, a scored Improved Tracers should not boost Troll's trace because it is not from a subroutine. 

Sealed Vault isn't deducting credits from the Corp total when doing the storage ability.


mtgred:
:+1:

PR #1351, opened 2016-03-15
Modified all-installed to handle nested hosted cards

Tondo:
As described in #1160, all-installed was not returning cards that were hosted on already hosted cards. Now the function searches recursively for any hosted cards, and should handle any level of nesting.

Though #1160 only mentioned it as a bug for the runner, the problem also seemed to exist for the corp when I tested it, so I made the same changes to the corp-side as well.

I should mention that this is the first time I'm writing clojure code (though I've written some functional programming code before), so please review the code and let me know if there's anything I should fix up. 

Edit: Also added a test for all-installed for the runner-side.

### DIFF
--- a/src/clj/game/core-misc.clj
+++ b/src/clj/game/core-misc.clj
@@ -69,8 +69,7 @@
   (if (= side :runner)
     (let [top-level-cards (flatten (for [t [:program :hardware :resource]] (get-in @state [:runner :rig t])))
           hosted-on-ice (->> (:corp @state) :servers seq flatten (mapcat :ices) (mapcat :hosted))]
-      (concat top-level-cards
-              (filter #(= (:side %) "Runner") hosted-on-ice))
+      (concat top-level-cards (filter #(= (:side %) "Runner") hosted-on-ice))
       (loop [unchecked top-level-cards installed ()]
         (if (empty? unchecked)
           (filter :installed installed)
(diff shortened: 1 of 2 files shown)


Tondo:
Sorry, wasn't aware about the integration test. Will look into the failed test.

Tondo:
Alright, I think I found the issue. It seems that when Knight is hosted onto an ICE, it loses its :installed field, which I was filtering by before returning the all-installed seq. Are programs that get hosted on ICE supposed to lose their :installed field?

JoelCFC25:
Knight should definitely be considered installed when hosted on ICE. If this isn't happening with Knight, then it's probably not happening for Pawn, Bishop, or Rook either.

Tondo:
Alright, I'll confirm if this is happening for the other Caissa programs as well, then look into fixing this issue.

Tondo:
Well, I'm just super wrong. I just rechecked and all the programs look fine, including Knight. Sorry for the confusion.

Tondo:
all-installed is passing the integration test again. I was just making some elementary functional programming mistakes. On another note, do we want to add some tests for all-installed that tests nested hosting?

Saintis:
Moar test!

Feel free to add something easy that checks all-installed somehow. Maybe B&E suite on daemons stacked 7 layers deep.

JoelCFC25:
Great job!

Tondo:
Added a test for all-installed on the runner side that checks if it works on:
- programs hosted on ICE
- nested hosted runner cards
- cards hosted on runner cards that shouldn't count as installed (ex. Personal Workshop)

I put the test in core-game.clj as it seemed like the right place to put it, but let me know if it should be somewhere else.

mtgred:
:+1:

PR #1355, opened 2016-03-15
Automates Caprice.

Saintis:
Also adds test, and some new test macros.

Keeps ability to enable trigger in case of mistakes or such.


zaroth:
These prompt testing functions are truly great! :+1:

mtgred:
:+1:

PR #1356, opened 2016-03-16
Work-around for using Nisei MK II during 4.3.

### LINKED ISSUE #1336: Using Nisei Mk II counter at Corp-requested step 4.3 doesn't end the run
Steps to reproduce: 

1. Score a Nisei Mk II
2. Have the Runner initiate a run
3. As the Corp, hit "Action before access" to request the 4.3 window
4. Runner hits "Successful run"
5. Corp gets the "Rez and take actions before successful run" prompt and spends the Nisei counter...but the run does not end.

Saintis:
Adds test, and additional test function.

Intends to fix #1336.

Root cause is that 4.3 is a prompt window, hence `handle-end-run` does not trigger all the ordinary end-run procedures and simply marks the run as `:ended true`. This does not actually end the run, or prevent a successful access. Additional work-around needed to fix the issue involving Net-Ready Eyes.

### DIFF
--- a/src/clj/test/cards-agendas.clj
+++ b/src/clj/test/cards-agendas.clj
@@ -227,6 +227,24 @@
         (core/score state :corp {:card (refresh napd)})
         (is (= 2 (:agenda-point (get-corp))) "Scored NAPD for 2 points after 5 advancements"))))
 
+(deftest nisei-mk-ii-step-43
+  "Nisei MK II - Remove hosted counter to ETR, check this works in 4.3"
+  (do-game
+   (new-game (default-corp [(qty "Nisei MK II" 1)])
+             (default-runner))
+   (play-from-hand state :corp "Nisei MK II" "New remote")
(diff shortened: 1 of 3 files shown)


src/clj/test/cards-agendas.clj
```diff
@@ -227,6 +227,24 @@
         (core/score state :corp {:card (refresh napd)})
         (is (= 2 (:agenda-point (get-corp))) "Scored NAPD for 2 points after 5 advancements"))))
 
+(deftest nisei-mk-ii-step-43
```
zaroth: So it's a step or a phase then ? ;-)

nealterrell: It's not his fault. I originally wrote the code as "phase 1.2" / "phase 4.3", but then discovered the real terminology from FFG is "Step 1.2". At that point I didn't feel comfortable changing all the code references.

mtgred:
:+1:

Issue #1344, opened 2016-03-13
Aesop's Pawnshop can sell itself


Kiv:
It states 'another' installed card, so it shouldn't be able to target itself. Commonly, players click it twice by mistake and it instantly sells itself.

Tondo:
If no one's working on this already, I'll look into fixing this.

PR #1357, opened 2016-03-16
Made Aesop's not able to target itself

Tondo:
Fix for #1344.

Modified Aesop's Pawnshop to not be able to target cards with the title "Aesop's Pawnshop". Since this card is unique, this should guarantee that if Aesop's ever targets a card on the board with the title "Aesop's Pawnshop"  it is in fact trying to target itself, which is invalid.

I would have preferred to use the cid for this check, but I wasn't able to figure out how to retrieve the currently used card's cid in the :choices field of the card definition. If anyone knows how, or knows another card that does something similar, I'd love to know so I can make this less sloppy.

Edit: 
Thanks to @nealterrell and @JoelCFC25, the code is now properly using the card's cid to determine whether it's targeting itself.

Also added a test function that tests Aesop's used on another runner card and Aesop's used on itself.

### DIFF
--- a/src/clj/game/cards-resources.clj
+++ b/src/clj/game/cards-resources.clj
@@ -24,7 +24,7 @@
    "Aesops Pawnshop"
    {:flags {:runner-phase-12 (req (>= 2 (count (all-installed state :runner))))}
     :abilities [{:msg (msg "trash " (:title target) " and gain 3 [Credits]")
-                 :choices {:req #(and (= (:side %) "Runner") (:installed %))}
+                 :choices {:req #(and (= (:side %) "Runner") (:installed %) (not= (:title %) "Aesop's Pawnshop"))}
                  :effect (effect (gain :credit 3) (trash target {:unpreventable true}))}]}
 
    "Always Be Running"
--- a/src/clj/game/cards-resources.clj
+++ b/src/clj/game/cards-resources.clj
@@ -26,7 +26,7 @@
     :abilities [{:effect (req (resolve-ability
                                 state side
                                 {:msg (msg "trash " (:title target) " and gain 3 [Credits]")
-                                 :choices {:req #(and (= (:side %) "Runner") (:installed %) (not= (:cid card) (:cid %)))}
+                                 :choices {:req #(and (card-is? % :side :runner) (installed? %) (not (card-is? % :cid (:cid card))))}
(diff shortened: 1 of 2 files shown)


nealterrell:
Yeah, unfortunately the only parameter to a targeting function is the card being targeted; you have no access to the game state in your `:req` for the targeting filter. It's a drawback that we have been able to somewhat hack around using closures (in Clojure!). The idea is that the card's ability doesn't contain the targeting code directly; instead it uses `resolve-ability` inside its own `:effect` to resolve an anonymous ability with the targeting filter and the end effect. Something like this:

```clojure
{:abilities [{:effect (effect (resolve-ability 
                                {:choices {:req #(not= (:cid card) (:cid %))}
                                 :effect (Aesop's effect)}
                                card nil))} ...
```

the `(effect` captures the state, side, and card data for the remainder of the effect. You then resolve an anonymous ability to do the targeting, and in that targeting function use `card` which was captured earlier. Voila! You can also use `state` and `side` if you need to call other state-reliant functions, like `all-installed`.

nealterrell:
Also, @Saintis has done some great work with adding utility functions for checking things like a card's side or its installed status. Take a look at `card-is?` (Hostile Infrastructure) and `installed?` (Scavenge).

Tondo:
Awesome, thanks @nealterrell! I'll try checking out closures and the functions you mentioned and see if I can clean up the code a bit.

JoelCFC25:
I'll have to remember "closure"...I've been informally using the term "wrapper". Nonetheless, Neal explained it more fully than I was going to attempt to do. Political Operative is another card that uses this technique, so you can spare yourself from looking at Turntable and potentially getting a migraine!

nealterrell:
Looking better. Let's get `card-is?` and `installed?` incorporated -- I know it sounds silly, but using `installed?` over `:installed` allows us to easily implement future rules that modify what it means to be "installed". Similarly, `card-is?` solves a few edge cases involving the automated testing framework.

Speaking of which, can I insist that you write a simple test for this improvement? :smile: :smile:

Tondo:
Will do! :+1:

nealterrell:
Nice job :+1:. @mtgred, this is ready to merge.

mtgred:
:+1:

PR #1360, opened 2016-03-16
implement Corporate Scandal (seeking review)

JoelCFC25:
Corporate Scandal adds a new bad publicity key to the Corp state in similar fashion to how Paparazzi was done for the Runner. 

Includes a test to validate the non-combo with Raymond Flint, that rezzing Elizabeth Mills doesn't reduce this bad publicity, the Runner gaining a run credit, and updated bad publicity requirements of Activist Support and Investigative Journalism. 

The gameboard is updated to show a blue alert in the stats area when the Corp has bad publicity.

### DIFF
--- a/src/clj/game/cards-events.clj
+++ b/src/clj/game/cards-events.clj
@@ -42,7 +42,7 @@
                        (gain state :runner :memory (:memoryunits c))))))}
 
    "Blackmail"
-   {:req (req (> (:bad-publicity corp) 0)) :prompt "Choose a server" :choices (req runnable-servers)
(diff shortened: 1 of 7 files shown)


nealterrell:
This looks good, Joel. My only feedback is to consider removing the warning icon for bad publicity. Unlike tags, there's generally little to nothing a Corp can do about BP, so the warning is going to be unactionable for 90% of users and lead to UI clutter (IMO).

JoelCFC25:
Thanks--done!

mtgred:
:+1:

PR #1361, opened 2016-03-16
Fixes: Test Run + Scavenge, Dracō, Product Recall

### LINKED ISSUE #1345: Scavenging an Imp doesn't seem to reset the once-per-turn limitation
Play Imp, use it to trash something, then Scavenge it into itself, then run on another trashable card. It didn't give me the option to use the Imp the 2nd time.

### LINKED ISSUE #1354: Test Run + Scavenge Combo still returning program to stack when turn ends
Test Run/Scavenge combo returns program to top of stack when Runner's turn ends.

• Has occurred with Femme and Battering Ram
• Runner typically has tags, and has Magnum Opus installed - no other board state commonalities observed
• Occurs regardless of whether actions are taken in between the two programs.
• Outside of the return to stack, the other card functionalities appear to be correct.

JoelCFC25:
* Fix #1354: Restore working Test Run + Scavenge combo (includes a test)
* Fix #1345: Trashing a card should remove it from `[:per-turn]` so you could Scavenge or Clone Chip once-per-turn cards like Imp and use them again all within a single turn (includes a test)
* Fix Product Recall to swap out trash cost bonus after use so further plays of the card don't yield too many credits
* Fix Dracō to add counters instead of setting them in case it ever gets derezzed and re-rezzed. 
* Update Heartbeat and Monolith damage prevention ability prompts to numerical values that will bring them to the front of the queue

### DIFF
--- a/src/clj/game/cards-events.clj
+++ b/src/clj/game/cards-events.clj
@@ -682,7 +682,7 @@
                                              ((if (= target "Heap") :discard :deck) runner))))
                       :effect (effect (runner-install (assoc-in target [:special :test-run] true) {:no-cost true}))
                       :end-turn
-                      {:req (req (find-cid (:cid target) (all-installed state :runner)))
+                      {:req (req (get-in (find-cid (:cid target) (all-installed state :runner)) [:special :test-run]))
                        :msg (msg "move " (:title target) " to the top of their Stack")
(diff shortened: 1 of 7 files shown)


mtgred:
:+1:

PR #1370, opened 2016-03-18
Prevent cards rezzed by Executive Boot Camp from receiving start-turn event

### LINKED ISSUE #1346: Executive Boot Camp allowing illegal triggers
At the beginning of the turn, if you use Executive Boot Camp to rez a card, then take your mandatory draw, that card will trigger its start-of-turn abilities. This should not happen, since the triggering condition had already passed by the time the card was rezzed. It would appear that cards with start-of-turn triggers are treating the mandatory draw as their triggering condition.

nealterrell:
Includes test. Fixes #1346.


mtgred:
:+1:

Issue #1313, opened 2016-03-07
Employee Strike functionality


Repulze:
Initially I thought this card hasn't been fully implemented. But the card automation spreadsheet says full functionality. So I'm unsure whether if this is a bug or not, but figured I'll just report it.

As I play Employee Strike a lot, I've had to resolve a lot of effects manually. The most prominent example is Haarpsichord Studios.
1 - Runner makes a run and steal an agenda
2 - Employee Strike is played
3 - Runner accesses another agenda (same turn) but "you've access but cannot steal" prompt is shown

Other examples include:
Cybernetics Division - hand size has to be manually added.
Tennin Institute - corp still gets prompt when their turn begins.

It works fine with identities like Spark Agency, Weyland: Building a Better World, and Jinteki: PE.

titonosfe15:
Also don't work in RP. Causing Fatal error when the runner try and inside job in a remote.

GrantZilla1979:
Another RP bug -

Runner plays Employee Strike - should be able to then run all remotes.  Only centrals are available to run when command is clicked.  Issue persists when current is manually trashed.  However seems to return to normal functionality at beginning of next Corp turn but only if ES remains out of play.

nealterrell:
I can't replicate (heh) the Inside Job issue, but I have identified and fixed Employee Strike and Cerebral Static interactions with:

* Tennin, Blue Sun, MaxX: disable their special flags related to start-of-turn effects
* Cerebral Imaging, Chaos Theory, Cybernetics Division, Harmony Medtech, and NBN: TWIY (card text that doesn't always tie to in-game events, usually a one-time thing)
* Haarpsichord Studios, Replicating Perfect (special cases)

Repulze:
Encountered this against SYNC. Paid 3 credits instead of 2 for removing tags.

PR #1372, opened 2016-03-18
Employee Strike and Cerebral Static rework

nealterrell:
Addressing concerns in #1313 and my own observations.

1. Clean up identity implementations so that things which are permanent (link strength, draft identity agenda points, Andy starting hand) are moved out of the identity's `:effect`, which then only contains effects that actively affect the game state (Chaos Theory memory, Cerebral Imaging hand size, etc.).
2. Add `:leave-play` effects to identities that kept a `:effect` flag, to disable whatever effect they have on the game when Cerebral Static / Employee Strike is played.
3. Add `:leave-play` to more complicated cards like Haarpsichord and Replicating Perfection to disable the effect of their flag registers.
4. Tweak Cerebral Static / Employee Strike to call `:leave-play`on the opponent's identity when played, and call `:effect` on the ID when the current is trashed.
5. Fix a few identity flags to take into account whether the ID has been disabled. Blue Sun/Tennin/MaxX won't cause Step 1.2, Gagarin won't force slow remote access.

### DIFF
--- a/src/clj/game/cards-events.clj
+++ b/src/clj/game/cards-events.clj
@@ -177,9 +177,8 @@
 
    "Employee Strike"
    {:msg "disable the Corp's identity"
-    :effect (req (unregister-events state side (:identity corp)))
-    :leave-play (req (when-let [events (:events (card-def (:identity corp)))]
-                       (register-events state side events (:identity corp))))}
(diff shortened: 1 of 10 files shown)


mtgred:
:+1:

PR #1373, opened 2016-03-20
Fixes: Mushin, Populist Rally, Neutralize All Threats, Sec Testing, Snatch and Grab

### LINKED ISSUE #1359: Neutralize All Threats not disabled by running Archives containing cards with trash costs
Neutralize All Threats needs to be reworked. The Runner should be able to turn off NAT by running Archives first when assets or upgrades are in there to take care of the "the first time you access" effect, then run elsewhere and __not__ be forced to trash a trashable card they access.

### LINKED ISSUE #1366: Muslin No Shin prevents scoring of agendas other than the one chosen
I was playing a fun game of Beale or No Beale earlier. I had an agenda I had originally placed with a mushin left out for a few turns in server 1. In a later turn, I mushined an ambush into a new server, used my third click to advance the agenda in server 1 and attempted to score. I was unable to "due to Mushin No Shin" indicated in an orange alert in the top right of the screen. I should be able to score. Perhaps a leftover tag from the original install is being recalled by the newer Mushin play? Hope you can fix it. Thanks!

### LINKED ISSUE #1368: Populist Rally gives corp clicks
Populist Rally gives the corp a permanent addition click every turn after the turn on which it was resolved. That is, on the first corp turn after Populist Rally is played they have 2 clicks, the second they have 3 clicks, the third they have 4 clicks, the fourth they have 5 clicks, and so on.

I was playing Silhouette vs, Jinteki black tree.

### LINKED ISSUE #1369: Security Testing's 'gain 2 credits' triggers on corp turn
More specifically, the corp played An Offer You Can't Refuse, nominating the same server that the runner had nominated as the Security Testing server. When the run was successful 'gain 2 credits' triggered.

N.b. I did check with FFG if Security Testing was meant to trigger on the corp's turn and Damon Stone replied 'No, "this turn" refers back to the first part of the effect, choosing a server at the beginning of your turn.'

### LINKED ISSUE #1375: Snatch and Grab prompt won't close
Multiple times playing SnG and my opponents tell me that selecting the "do not prevent" option doesn't close the prompt or continue the action.
From my side what I see is: I play SnG they don't beat the trace, then nothing happens for a while and eventually get a bunch of messages that the runner declined to prevent the trash, and that the other player used a /close-prompt command.

JoelCFC25:
* Fix #1375: Snatch and Grab has a malformed `:effect` when the Runner declines to take a tag
* Fix #1369: Confine Security Testing replacement effect to Runner's turn only
* Fix #1368: Unregister events at the end of the Corp turn so they don't keep gaining a click per turn indefinitely. 
* Fix #1366: Fix turn flags on Mushin so only the installed card is prevented from rezzing or being scored and not everything.
* Fix #1359: Use `:per-turn` with NAT and set it to true if the Runner accesses something with a trash cost on an initial run on Archives to "switch off" the forced trash ability. 
* Reclamation Order isn't supposed to be able to target copies of itself
* Jinteki Biotech was able to be flipped more than once per game

### DIFF
--- a/src/clj/game/cards-events.clj
+++ b/src/clj/game/cards-events.clj
@@ -494,7 +494,8 @@
     :effect (effect (lose :corp :click-per-turn 1)
                     (register-events (:events (card-def card))
                                      (assoc card :zone '(:discard))))
-    :events {:corp-turn-ends {:effect (effect (gain :corp :click-per-turn 1))}}}
+    :events {:corp-turn-ends {:effect (effect (gain :corp :click-per-turn 1)
+                                              (unregister-events card))}}}
 
    "Power Nap"
    {:effect (effect (gain :credit (+ 2 (count (filter #(has-subtype? % "Double")
--- a/src/clj/game/cards-hardware.clj
+++ b/src/clj/game/cards-hardware.clj
@@ -2,7 +2,7 @@
 
 (def cards-hardware
   {"Akamatsu Mem Chip"
-   {:effect (effect (gain :memory 1)) :leave-play (effect (lose :memory 1))}
+   {:in-play [:memory 1]}
 
    "Archives Interface"
    {:events {:no-action {:effect (req (toast state :runner "Click Archives Interface to remove 1 card in Archives from the game instead of accessing it" "info")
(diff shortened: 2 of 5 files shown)


mtgred:
:+1:

PR #1376, opened 2016-03-21
Placed discard popup above card counters

### LINKED ISSUE #1363: Counter visibility on subwindow open
example:
I'm playing as corp
the runner, Geist, install some cloud icebreaker, like 2 shiv, so either have a counter with "2" on it.
Opening the Heap of the runner, the strenght counter on the 2 installed icebreaker is still visible and hide a little bit the card in the Heap.

Was Nice if the counter is hided by the subwindow

zaroth:
Fixes #1363.

### DIFF
--- a/src/css/base.styl
+++ b/src/css/base.styl
@@ -1309,7 +1309,7 @@ nav ul
     bottom: 85px
     width: 590px
     min-height: 92px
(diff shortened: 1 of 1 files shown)


mtgred:
:+1:

PR #1382, opened 2016-03-23
mark cards from Democracy and Dogma as released

zaroth:


mtgred:
There is actually a release date in the Sets collection in MongoDB.

It would be better to change the release logic to use that date to avoid having to change the code every month to mark a set a released.

zaroth:
Sure, if someone wants to step in and code it so it's pulled from API, they're more than welcome to do so.

I kinda went for simplicity on this one, since a pack being released needs implementing new cards anyway, two-line change in a deckbuilder isn't a big deal I think  :-)

zaroth:
Or JS could ask the server for current date, I guess - that just occured to me. Still a lot of complications for not that much gain.

mtgred:
The deckbuilder has been updated to check the release status of cards based on the set release date.

It is no longer need to update a hard coded value in the code every month.

PR #1383, opened 2016-03-24
added overflow-x to trigger horizontal scrolling when necessary

SimonS:
I've had this implemented for a few weeks in [jankteki](https://github.com/SimonS/jankteki), and it seems to work nicely in managing horizontal asset spam decks. On the occasions when you don't want this effect (ie, you're hovering a card on the left that you want to see on the right), the old zoom fix still works.

### DIFF
--- a/src/css/base.styl
+++ b/src/css/base.styl
@@ -1130,6 +1130,7 @@ nav ul
   display: none
 
 .gameboard
+  overflow-x: auto;
   position: absolute
   top: 31px
   bottom: 0


zaroth:
Huh, I was toying with adding this parameter to just the server row waay before, but somehow didn't think about adding it to the `.gameboard`. :-) Could you post a screenshot?

SimonS:
Because I'm a shameless self-promoter, I'm way ahead of you ;) (the bit you care about comes at about 3 minutes) :

https://www.youtube.com/watch?v=TtLmbbjWYec

Or to play with it for real, install jankteki on Chrome and opt in to it in options.

SimonS:
(my initial attempts were server-row based, but they screwed up other things, the flexbox layout was tricky to navigate around).

zaroth:
It's a shame that chat still gets pushed off the screen, but it's better for it to be pushed off and you are able to scroll to it, than pushed off and then you can't. ;-)

queueseven:
I think we should accept this. I use this css patch myself all the time. I looked into the cljs code to find a way to improve this, but the way the components of the gameboard is organized makes it impossible, so this is the best we can get ATM IMHO.

mtgred:
👍

PR #1389, opened 2016-03-25
Genetics Pavilion fixes

### LINKED ISSUE #1374: Genetics Pavilion interacts with Fisk Investment Seminar poorly.
When I have Genetics Pavilion in play, and Fisk is played, the Runner draws the number of cards that he/she is supposed to, but the Corp draws no cards. More intriguingly, when I attempt to rectify this, by manually adding clicks/drawing on the Corp's behalf, the Corp is still unable to draw. It seems that GP limits ANY draw by either of the two players to just two during the Runner's turn presently.

JoelCFC25:
Fixes #1374. 

Genetics Pavilion gets a little more precision and some core function support to cover all the bases. In particular we need to reset `:drawn-per-turn` and `:cannot-draw` in the player's register at the end of their turn to allow additional draws during the opponent's turn. The included tests demonstrate Genetics Pavilion compatibility with the following scenarios:

* Limit Runner's draws to 2 during the Runner's turn
* Allow unlimited Runner drawing during the Corp's turn (e.g., Sports Hopper, multiple trashes using Geist)
* Don't restrict the Corp's drawing (the Fisk Investment Seminar case)
* Lift the Runner draw restrictions immediately if GP is trashed or gets derezzed

### DIFF
--- a/src/clj/game/cards-assets.clj
+++ b/src/clj/game/cards-assets.clj
@@ -302,9 +302,9 @@
    "Genetics Pavilion"
    {:msg "prevent the Runner from drawing more than 2 cards during their turn"
     :effect (req (max-draw state :runner 2)
-                 (when (= 0 (remaining-draws state side))
-                   (prevent-draw state side)))
-    :events {:runner-turn-begins {:effect (effect (max-draw 2))}}
+                 (when (= 0 (remaining-draws state :runner))
+                   (prevent-draw state :runner)))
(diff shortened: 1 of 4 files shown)


mtgred:
There is a merge conflict. Can you have a look?

JoelCFC25:
I just don't know how to do all that stuff. Will close and recreate.

PR #1390, opened 2016-03-25
some Democracy and Dogma cards

JoelCFC25:
Implements the four Political assets from the new set with a small helper function for the "not protected by ICE" condition. Also includes Akshara Sareen and "Freedom Through Equality".


Saintis:
All these "if unprotected" cards are begging for a refactor.

mtgred:
:+1:

PR #1392, opened 2016-03-25
Voting Machine Initiative

nealterrell:
I believe this is the first card with an optional ability that triggers at opponent's start of turn. Rather than incorporate a UI window for this, I just give VMI an optional ability that will wait to trigger until `:runner-turn-begins`. The Corp can click VMI at any point during his turn to cause a delayed ability that causes the click loss only once the runner turn begins, without giving away the corp's intention.


mtgred:
:+1:

PR #1395, opened 2016-03-26
Update figwheel to 0.5.2.

cwgreene:
Necessary due to changed API between clojurescript versions.


mtgred:
Hi, we had the same idea concurently. The Figwheel version was already bump 2 hours earlier. :)

PR #1406, opened 2016-03-31
implement Mumbad City Hall

JoelCFC25:
This only fails CircleCI because of the broken Hayley test. 

Surat City Grid was completely wrong based on my initial misread of the card text--you get to rez any other card, you're not limited to SCG's server.


mtgred:
:+1:

PR #1407, opened 2016-03-31
Prevent install-and-host copies of unique cards in play

### LINKED ISSUE #523: Hosting cards does not check for uniqueness
e.g. you can install multiple copies of Kati Jones on Off-Campus Apartment.

Workaround: Just-don't-do-it™

JoelCFC25:
Fixes #523. 

The existing 6 Runner cards that have the ability to spend a click and host something from the Grip are modified to wrap that ability inside a `resolve-ability` along with the cost. This allows the use of `runner-can-install?` in the targeting that in turn checks uniqueness, and also ensures that the click is not deducted unless an install succeeds. Future Runner cards that can host in this fashion will need to use this method. 

Includes a test (again, this is only failing due to the faulty Hayley test that will be removed soon).

### DIFF
--- a/src/clj/game/cards-hardware.clj
+++ b/src/clj/game/cards-hardware.clj
@@ -381,12 +381,13 @@
 
    "NetChip"
    {:abilities [{:label "Install a program on NetChip"
-                 :cost [:click 1]
                  :req (req (empty? (:hosted card)))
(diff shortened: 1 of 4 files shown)


mtgred:
:+1:

PR #1418, opened 2016-04-03
Fix bug during use of damage prevention (PRIORITY)

### LINKED ISSUE #1415: Chat log gets filled when preventing damage with plascrete
http://pastebin.com/MBK6CjDH

Title, as you can see there will be lot of text. Wall of code came out whenever runner prevented meat dmg.

### LINKED ISSUE #1417: Chat box reports lots of code when preventing damage with Feedback Filter
Not sure if it's helpful to post this here; I assume it's similar to the Plascrete Carapace bug submitted earlier.  

EDIT: should have said - it was when preventing damage caused by the new card Bio-Ethics Association and yes, I should have trashed it sooner...

JoelCFC25:
Fixes #1415, fixes #1417. 

Turns out it's not a good idea to fire a function in the middle of a `system-msg`. Sorry for the goof!

### DIFF
--- a/src/clj/game/core-rules.clj
+++ b/src/clj/game/core-rules.clj
@@ -141,11 +141,12 @@
                state :runner nil (str "Prevent any of the " n " " (name type) " damage?") ["Done"]
                (fn [choice]
                  (let [prevent (get-in @state [:damage :damage-prevent type])]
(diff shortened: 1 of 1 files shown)


mtgred:
:+1:

PR #1433, opened 2016-04-08
Card fixes: Political assets, NeoTokyo Grid, PAD Factory, Housekeeping, Cell Portal

### LINKED ISSUE #1432: PAD Factory/Mumbad Construction Company Combo Not Allowing Score
When you use PAD Factory to place Advancement Tokens on Mumbad Construction Company, then use MCC to move the tokens onto a face up agenda, the system will not allow a score of that agenda, stating the error: "Cannot score due to PAD Factory".

Below is a screenshot of the error:
screenshot 2016-04-07 16 54 19

and here is a copy of the log:
mcc-padfactory error

### LINKED ISSUE #1437: Surat City Grid's rez discount only works on cards installed in the same server
When using Surat City Grid's ability to rez other cards with a discount whenever a card is rezzed in or protecting the same server as Surat City Grid, I ran into the following issue:

- The card works fine when rezzing a card in the same server. I get the following message in my chat log:
Rjayz spends 2 to rez Ash 2X3ZB9CY.
Rjayz uses Surat City Grid to rez Heimdall 1.0, lowering the rez cost by 2 .
Rjayz spends 6 to rez Heimdall 1.0. (Heimdall 1.0 normally costs 8 credits)
The Heimdall 1.0 is located in the same server as the Surat City Grid and Ash.

- Whenever I rez a card in a different server, I do not get the discount, and get the following chat messages:
Rjayz spends 2 to rez Ash 2X3ZB9CY.
Rjayz uses Surat City Grid to rez Eli 1.0.
Rjayz spends 3 to rez Eli 1.0. (So no discount)

The Eli 1.0 was located on HQ, the Surat City Grid and Ash in a remote server. I also tested reproducing the issue with ICE, assets and upgrades, and in each case if the card was installed in the same server I received the discount, and if it wasn't I didn't.

JoelCFC25:
* Tidy up the Political assets using a new macro instead of a `no-ice?` function. Much cleaner, and now accounts for these assets being hosted on Worlds Plaza or the upcoming Full Immersion RecStudio.
* Fix NeoTokyo Grid to listen for advancements being placed by Shipment from SanSan/Kaguya, Tennin, etc
* Fix #1437: Surat City Grid discount should apply everywhere
* Fix #1432: Add precision to PAD Factory's turn flag, as was done to Dedication Ceremony previously.
* Fix Housekeeping case of a Runner trashing a resource from hand when they have Fall Guy installed (don't want Fall Guy triggering for prevention).
* Fix Cell Portal, which is currently teleporting the Runner to position 0 of the run instead of to the outermost position.
* Fix Universal Connectivity Fee to update the run position if it trashes itself, and also fire the second way when the Runner is tagged from Paparazzi
* A few other small card message tweaks

### DIFF
--- a/src/clj/game/cards-assets.clj
+++ b/src/clj/game/cards-assets.clj
@@ -91,7 +91,7 @@
                                                 (lose :runner :credit 1))}}}
 
    "Bio-Ethics Association"
-   (let [ability {:req (req (no-ice? state (second (:zone card))))
+   (let [ability {:req (req unprotected)
                   :label "Do 1 net damage (start of turn)"
                   :once :per-turn
                   :msg "do 1 net damage"
@@ -137,7 +137,7 @@
    "Clone Suffrage Movement"
    {:derezzed-events {:runner-turn-ends corp-rez-toast}
     :flags {:corp-phase-12 (req (and (some #(is-type? % "Operation") (:discard corp))
-                                     (no-ice? state (second (:zone card)))))}
+                                     unprotected))}
     :abilities [{:label "Add 1 operation from Archives to HQ"
                  :prompt "Choose an operation in Archives to add to HQ" :show-discard true
                  :choices {:req #(and (is-type? % "Operation")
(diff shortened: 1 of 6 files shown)


mtgred:
:+1:

Issue #1362, opened 2016-03-16
Upgrades on centrals should not be considered "in the server"


JoelCFC25:
In the just-released [Business First UFAQ](http://ancur.wikia.com/wiki/Business_First_UFAQ) came the surprising (to me at least) ruling that upgrades in the root of a central server are not considered to be __in__ that server. 

As a result, the following running list of cards need to be updated so their effects don't work on upgrades in central servers:
- [x]  Breaker Bay Grid 
- [x]  Unregistered S&W '35
- [x]  Surat City Grid _(trigger condition)_
- [x]  Back Channels
- [x]  Diwan
- [x]  Oaktown Grid

nealterrell:
Add Oaktown Grid to the list. So does this mean that no cards are "in" a central server? Or are cards in your hand part of "HQ central server"? Would Oaktown Grid on HQ or R&D increase the trash cost of cards accessed from that server? Madness.

PR #1439, opened 2016-04-12
Test-a-palooza: Large batch of unit tests

JoelCFC25:
This is about a 33% increase in our number of card tests all in one fell swoop, plus a couple new test suite functions for more concisely retrieving a specific program/hardware/resource from the Runner's rig. 

There is only one change to actual cards, which is the Breaker Bay Grid case from #1362. I made the edit to support the unit test.


queueseven:
Can't have enough tests

mtgred:
:+1:

PR #1449, opened 2016-04-14
Small batch of fixes

### LINKED ISSUE #1445: Surat City Grid should only fire if there are unrezzed non-agenda cards...
...and also you should not be able to target agendas when you activate Surat City Grid

### LINKED ISSUE #1446: Corp trashing installed unrezzed Director Haas triggers leave-play effects
Steps to reproduce: 

1. Install a copy of Director Haas. 
2. Manually trash it, or just overwrite it with another asset/agenda
3. You immediately lose a click if you have any remaining, and lose 1 click per turn. 

This is a problem in [lines 46-48](https://github.com/mtgred/netrunner/blob/master/src/clj/game/core-installing.clj#L46-L48) of `deactivate`--it needs to add a check that if it's a Corp card it must be rezzed.

### LINKED ISSUE #1459: Chronos Protocol + 0 cards in hand doesn't end the game.
If the runner has no cards in hand and I play something that does 1 net damage (such as Neural EMP), the prompt asks me to select a card even though there are no cards and doesn't finish the game.

JoelCFC25:
I don't expect this will keep growing into a mega-patch, but you never know! 

* Fix #1446: Only lose `[:in-play]` of a Corp card when `deactivate` is called if the card is rezzed.
* Fix #1445: Don't fire Surat City Grid when only agendas remain, and don't allow targeting an agenda with it
* Entirely prevent the play of Run events when `:cannot-run` is in Runner's register after an Excalibur or Uroboros subroutine successfully fires.
* Turn off Chronos Protocol when Employee Strike is active so that net damage can occur normally. Also, don't do its ability when the Runner has 0 cards in hand (fixes #1459).
* Remove Titanium Ribs effect if it gets trashed
* Add the "can't rez the rest of the turn" restriction to Councilman's target. Also alerts the Corp player with a toast when they rez an asset/upgrade with Councilman in play so they will be less likely to plow ahead with more actions/abilities before the Runner has decided to use Councilman.
* Add missing `:leave-play` to Rolodex for milling the Runner 3 times.
* Move the ability to pay requirement for Tori Hanzō to the outermost level to avoid a state that stops net damage from working

### DIFF
--- a/src/clj/game/cards-hardware.clj
+++ b/src/clj/game/cards-hardware.clj
@@ -640,7 +640,8 @@
      :damage-chosen {:effect (effect (enable-runner-damage-choice))}}
     :effect (effect (enable-runner-damage-choice)
                     (system-msg (str "suffers 2 meat damage from installing Titanium Ribs"))
-                    (damage :meat 2 {:card card}))}
+                    (damage :meat 2 {:card card}))
+    :leave-play (req (swap! state update-in [:damage] dissoc :damage-choose-runner))}
 
    "Turntable"
    {:in-play [:memory 1]
--- a/src/clj/game/cards-hardware.clj
+++ b/src/clj/game/cards-hardware.clj
@@ -324,8 +324,10 @@
    "Logos"
    {:in-play [:memory 1 :hand-size-modification 1]
     :events {:agenda-scored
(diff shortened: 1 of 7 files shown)


mtgred:
👍

Issue #948, opened 2015-12-04
Trashing [Hellion Alpha Test] prints incorrectly in chat


GordonsBeard:
When Ed Kim accessed Hellion Alpha Test from R&D it was automatically trashed by the Runner, and then printed this to the chatlog:

![image](https://cloud.githubusercontent.com/assets/4015791/11582188/3b21cfa2-9a00-11e5-959a-60b41723c3a1.png)

GordonsBeard:
This happens any time the card is printed, even playing the card from hand will cause the incorrect text.

nealterrell:
@queueseven is going to have to chime in here, he did the card name highlighting.

Saintis:
Problem lies in the `add-image-codes-impl` function. The regex finds Alpha inside Hellion Alpha Test and links that card instead. So you get `[Hellion [Alpha~04087] Test~04031]` instead of `[Hellion Alpha Test~04031]`.

zaroth:
Also tried looking at this, but didn't have enough willpower today to perserve through and pick it apart. For brave souls looking at this, the regex in question is [here](https://github.com/mtgred/netrunner/blob/0aa976f2e3e21a2bbde33713b0d84f03c36353d9/src/cljs/netrunner/gameboard.cljs#L233)

zaroth:
Fixed by #1456.

PR #1456, opened 2016-04-19
Fix for #948 (and some clean up)

queueseven:
When looking for card names in the text message, it's not enough to check
if the found card name does not end with ~ (the then used seperator for
the card text tokens), we also have to make sure that no ] (the then used
closing tag) follows; so we don't the regex to match

  Alpha

in this text

  Someone played [Hellion Alpha Test~01110] or something.

because it's already a card text token.

I also replaced the used [ ] and ~ characters with some characters you
don't find on a keyboard and extracted them into a variable.


zaroth:
Tested, it's legit: http://i.share.pho.to/37b59644_o.png

I move that a Regex Wizard magic hat is awarded to @queueseven :tada:

mtgred:
👍

PR #1463, opened 2016-04-22
Some animation for the log pane

queueseven:
New messages fade in. I used this some time now with a custom css plugin
and it makes it easier to follow the log when a lot of stuff happens,
since the animation of the text draws your attention IMHO.

Try it; if you don't like it, np.


mtgred:
👍

PR #1474, opened 2016-04-28
Some fixes for Progenitor

### LINKED ISSUE #1470: Virus hosted on Progenitor still counts against memory limit
This should work just like Leprechaun or Dinosaurus, ignoring the memory cost of any hosted program. I see that Progenitor is marked as partially implemented, so I assume this just hasn't made the cut yet.

queueseven:
fixes #1470 
added unit test


mtgred:
👍

PR #1479, opened 2016-04-28
use textContent instead of innerHTML

### LINKED ISSUE #60: R&D Interface- Cannot add via Add Card Tool
Cannot add the card R&D Interface to my runner deck. No matter how you type in the card and click it, then click add card, it does not add. I can add it through the decklist, however not through add card tool.

queueseven:
fixes #60


mtgred:
Awesome ! 👍

PR #1481, opened 2016-04-28
Fix chameleon/llds interaction

### LINKED ISSUE #1475: Chameleon loses LLDS Processor strengh when choosing a subtype after installing

queueseven:
fixes #1475


mtgred:
👍

PR #1482, opened 2016-04-28
Resizable log pane

### LINKED ISSUE #583: Feature Request: make chat window on right-hand side resizable
For many resolutions, there's a lot of unused horizontal real estate between the chat window and the cards, and the chat window frequently has wrapped text.  It would be nice if this had a resize bar to make it possible to change the horizontal size of the chat pane.

queueseven:
fixes #583

you can make the log pane bigger and smaller (but it's not a solution to
the asset spam problem)

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -3,6 +3,7 @@
   "version": "0.0.1",
   "dependencies": {
     "jquery": "~2.0.3",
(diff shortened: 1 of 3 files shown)


mtgred:
👍

PR #1488, opened 2016-04-30
Implements Salsette Slums

ilyanep:
I'm not super familiar with Clojure, so I'd appreciate any pointers y'all have, especially when it comes to the "what isn't quite perfect" section below. I was trying to make this card be as automatic as possible, since I'm not a huge fan of having to click cards to manually trigger abilities.

(As a side note, I think your Mongo for the prod server might be missing this card, since it won't let me add it in deck builder)

What works:
- Successfully prompts to remove cards from the game upon trashing
- Does not do this when a runner trashes their own cards
- Successfully removes the cards from the game

What isn't quite perfect:
- Multiple Salsette Slums in play will prompt the player multiple times (and if they
  hit yes multiple times, they'll accidentally disable all of their Slums for that
  turn).
- The prompt says "Use Salsette Slums to remove from the game?" rather than referencing
  the card the runner is removing. This isn't ideal because it's inconsistent with
  other prompts in the game, but I'm not sure how do use this ```req``` closure/macro
  to do what I want.


ilyanep:
(Also I'm not sure why the Psycho test is failing for me -- I didn't touch anything that should be relevant?)

queueseven:
Thanks for your PR. Don't worry about circleCi failing. Sometimes it just happens.

Salsette Slums cant trigger automatically, it has to be a manual activation. The reason for this is the event system. 

Say the corp has three rezzed Hostile Infrastructures in play. Now you access one and are asked if you want to trash it. This is the point in time when the remove-instead-of-trash effect should happen. When you wait for the :runner-trash event you already suffered the three net-damage before you had a chance to answer the prompt.

Since Salsette Slums is a replacement effect, the :runner-trash effect must not fire if used. Take a look at Maya. That's how it should implemented IMHO.

JoelCFC25:
The Psychographics test failure is due to a change in the source NRDB data--the cost used to be stored as 0 but is now null.

ilyanep:
Oh dang, I had forgotten that it replaces accessing. I guess I'll change it to an ability after all.

ilyanep:
Cool, I think this should do it, at least as far as the card working as intended. Let me know if I missed any edge cases or if I can fix up the code style somehow.

queueseven:
Looks good so far, but I didn't test it.

Maybe you want also to try to write a unit test? That would be awesome

ilyanep:
Got it. Turns out writing a test was easier than expected. 

I noticed that my test is a bit more complex than other ones (I set up more than one hand of cards to avoid the corp decking itself, and tested over sever turns), so I hope that's okay.

queueseven:
Cool, I try to test this tomorrow

queueseven:
I tested this and it seems to work fine.

Some note for the test: you can also just use the `gain` function to get clicks and credits, like `(core/gain state :corp :click 3)` or `(core/gain state :runner :credit 18)`, so you don't have to play `Hyperdriver` ;-)

ilyanep:
Oh cool, I wasn't sure if using the gain functions was legit in tests, since I didn't see it anywhere else when I checked. I was able to simplify the test a bit this way.

JoelCFC25:
Just FYI, you will get doubled periods at the end of all the `:msg` in the game log (the system supplies those elsewhere in the code), so I would recommend dropping them.

queueseven:
Sharp-Eye-Joel, spotting everything....

ilyanep:
Is this ready to go?

mtgred:
👍

PR #1489, opened 2016-05-04
Implement Deva icebreakers

queueseven:
Most work was done by @JoelCFC25.

The only thing does not currently work is auto-pump after swapping after encountering, but I think that's a minor issue.


mtgred:
👍

PR #1491, opened 2016-05-04
implement Jeeves Model Bioroids

JoelCFC25:
This has to be manually activated for the time being since tracking how clicks were spent is way too hairy. At least we can ensure that the Corp has spent at least 3 clicks before it can be used.

### DIFF
--- a/src/clj/game/cards-assets.clj
+++ b/src/clj/game/cards-assets.clj
@@ -428,6 +428,12 @@
                                          (shuffle! state side :deck))}
                            card nil))}]}
 
+   "Jeeves Model Bioroids"
+   {:abilities [{:label "Gain [Click]"
+                 :req (req (< 2 (- (:click-per-turn corp) (:click corp))))
+                 :msg "gain [Click]" :once :per-turn
+                 :effect (effect (gain :click 1))}]}
+
    "Kala Ghoda Real TV"
    {:flags {:corp-phase-12 (req true)}
     :abilities [{:msg "look at the top card of the Runner's Stack"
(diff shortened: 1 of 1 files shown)


zaroth:
I'm not sure your `req` will work. Further comments inline.

src/clj/game/cards-assets.clj
```diff
@@ -428,6 +428,12 @@
                                          (shuffle! state side :deck))}
                            card nil))}]}
 
+   "Jeeves Model Bioroids"
+   {:abilities [{:label "Gain [Click]"
+                 :req (req (< 2 (- (:click-per-turn corp) (:click corp))))
```
zaroth: Subtracting clicks from max clicks isn't a good universal method to get spent clicks. This causes problems with i.e. Biotic Labor. Example situation where it fails: I have 3 max clicks, I play Biotic, Shipment from SanSan - I should be able to get a click back from Jeeves now to be able to play second Shipment from SanSan, but with the `req` above I can't. (This could be a nice unit test ;-) )

queueseven: Also the common combo with Jeeves is Sublimal Messages with Shipment from SanSan, which would also not work with this. Maybe we should just remove the req for now and leave it to the player to decide when to use Jeeves to get the click? I don't know how much work it would be to fully automate Jeeves

JoelCFC25:
Thanks--pulled that part out for now. Trying to filter `[:turn-events]` for `:corp-spent-click` and counting those is another approach, but probably more trouble than it's worth.

mtgred:
👍

PR #1500, opened 2016-05-10
trigger the active player's events first

### LINKED ISSUE #309: Replicator vs. Housekeeping order of activation
Opponent has Housekeeping active
Play Replicator or another hardware with Replicator out
The Housekeeping prompt activates before the Replicator prompt, causing Runner to discard a card before receiving their Replicated card, which Runner may want to discard for Housekeeping instead.

queueseven:
fixes #309


mtgred:
👍

PR #1502, opened 2016-05-10
Fix X-strength ice breakers (Darwin)

### LINKED ISSUE #1498: Darwin causes game to become unresponsive on successful run
Once Darwin is installed by the runner a successful run will cause the game to become unresponsive at the 'Successful Run' / 'Jack out' prompt. Clicking successful run will trigger the correct successful run actions, however the run will never end and the prompt will not close. This happens regardless of whether Darwin was actually used during he run.

NOTE: this may also happen on unsuccessful runs, has not been tested.

In addition, as soon as Darwin is installed, any interaction with Darwin will cause significant delay, giving the appearance of nothing having happened, usually until another item is interacted with.

queueseven:
fixes #1498

### DIFF
--- a/src/clj/game/core-installing.clj
+++ b/src/clj/game/core-installing.clj
@@ -68,7 +68,8 @@
          recurring (:recurring cdef)
          abilities (ability-init cdef)
          c (merge card (:data cdef) {:abilities abilities})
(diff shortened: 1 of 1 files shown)


mtgred:
👍

PR #1507, opened 2016-05-10
Split tests into namespaces to (hopefully) address Circle-CI memory failures

nealterrell:
This requires #1504 from @queueseven.

Splits the test files into proper namespaces so they can be run individually. Configures Circle-CI to run each namespace one at a time, which will hopefully avoid the memory issues we've been seeing. This does cause the build to take a little longer on Circle-CI, but I don't know if we can avoid that.

This changes how tests are run locally. To run one specific test file, simply `lein test test.cards.agendas` or appropriate. To run all files, I added a special namespace "all": `lein test test.all`.


mtgred:
👍

Issue #1469, opened 2016-04-26
Astrolabe posts draw message even though the draw is prevented


Balkmeister:
After drawing two cards already with a rezed genetics pavilion on the table i ran into a crick. It fired and installed a sundew from archives. The astrolabe i had installed posted to the log that i drew a card but the actual card draw was prevented.

 In short: Don't post to the log that a card is drawn if the draw is prevented.

JoelCFC25:
There are a _lot_ of cards that will be similarly afflicted. I don't know of a way to suppress the log message at this time (or just a portion of the message if the card draws and does something else, such as Lawyer Up).

Balkmeister:
prehaps make the message conditional? Like first check the number of cards, then do the action then check again. if the number of cards in hand is the same dont output the log message?

NBHS:
Problem is that this is a much more wide-ranging bug than just this circumstance. Log messages are also displayed for bugged (non-functional) damage in the case of #1453

Saintis:
Should be fixed in #1509.

PR #1509, opened 2016-05-10
Notify players when a draw is prevented.

Saintis:
Posts a `system-msg` when a draw is prevented (assumed to be due to the opponent).

Intended to help with #1469.

Screenshot:
screen shot 2016-05-10 at 22 35 25


mtgred:
👍

PR #1510, opened 2016-05-10
Adds "Gain subroutines" abilities to relevant pieces of ICE.

Saintis:
Allows for players to announce number of subroutines on ICE that have variable number of subroutines, like Komainu, Tour Guide and Ashigaru.

Also works with the advanceable only while rezzed ICE.

Does not account for 1 subroutine not being plural, so if that would irritate anyone they are more than welcome to fix it. I felt it was too much of a hassle.

Note that this is only a cosmetic change, no game state is changed, the number is just printed to chat.

Screenshot:
screen shot 2016-05-10 at 23 18 00

### DIFF
--- a/src/clj/game/cards-ice.clj
+++ b/src/clj/game/cards-ice.clj
@@ -211,7 +211,9 @@
                  :msg (msg (corp-install-msg target))}]}
 
    "Ashigaru"
-   {:abilities [end-the-run]}
+   {:abilities [{:label "Gain subroutines"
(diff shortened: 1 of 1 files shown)


JoelCFC25:
You can strike the "to " inside the `:msg`, since it gets automatically added--otherwise you end up with "uses Komainu **to to** gain 5 subroutines".

Saintis:
Fixed.

mtgred:
👍

PR #1512, opened 2016-05-11
implement Brahman, Upayoga, Temple of the Liberated Mind, Subcontract

JoelCFC25:
Brahman's post-encounter effect (add an installed non-virus to the top of the Stack) is left to be manual.


mtgred:
👍

Issue #402, opened 2015-07-20
Source code license?


Kiv:
It would be nice to know what license the code is available under - both for contributors to know what happens to the copyright of their contributions, and for anyone that might want to make a derivative work such as using part of the rules engine for another card game.

nealterrell:
@mtgred ?

mtgred:
This needs to be addressed indeed. I don't have a good knowledge about the differences the various open source licenses. I need to do some homework first.

aaronthewhite:
@mtgred 
This is a good tool for narrowing down the licenses:
http://oss-watch.ac.uk/apps/licdiff/

A site that lists and explains the most popular licenses:
http://choosealicense.com/
They also have a list of licenses in a format that is easier to compare:
http://choosealicense.com/licenses

This site is good for explaining a license in plain english:
https://tldrlegal.com/

Ultimately, it just depends on what you want to allow.

steveklabnik:
Came here to ask about this. Until you pick a license, this code is technically "all rights reserved". It also means that the contributions of other people are technically not under that license either...

> Ultimately, it just depends on what you want to allow.

Yup, this is it. @mtgred if you can explain a bit about what specifically you want with regards to sharing, I'd be happy to help explain some licenses.

IMHO, the biggest quesiton is "Are you okay with someone taking this code, using it in something closed source, and then distributing it to others?" Many are, many aren't.

jhli:
If you want share back:
AGPL, ALGPL
http://stackoverflow.com/questions/2127246/difference-between-affero-gpl-and-gplv3
If you don't want that: MIT.

zaroth:
As to add to the discussion, I'd like to note that just submitting code to open repository on Github subjects it to some [Terms of Service](https://help.github.com/articles/open-source-licensing/#what-happens-if-i-dont-choose-a-license), so forking and viewing is allowed.

queueseven:
Finally

PR #1516, opened 2016-05-12
Add MIT license

nikolasavic:


zaroth:
I compared it with the MIT License notice jQuery uses: https://github.com/jquery/jquery/blob/master/LICENSE.txt

I don't think this part relates to us:

     The following license applies to all parts of this software except as
      documented below:

JoelCFC25:
Fixes #402!

mtgred:
👍

jhli:
Cool! I could have lived with gpl too but mit dies the job as well.

PR #1519, opened 2016-05-13
Finish remaining "in this server" fixes

### LINKED ISSUE #1362: Upgrades on centrals should not be considered "in the server"

JoelCFC25:
Fixes #1362. 

This caps off the fixes needed to comply with the Business First ruling that upgrades in the root of a central server are not considered to be "in" that server.

### DIFF
--- a/src/clj/game/cards-hardware.clj
+++ b/src/clj/game/cards-hardware.clj
@@ -692,7 +692,8 @@
                                (has-subtype? % "Clone")
                                (has-subtype? % "Executive")
                                (has-subtype? % "Sysop"))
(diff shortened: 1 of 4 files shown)


mtgred:
👍

PR #1521, opened 2016-05-13
Fix doubled successful run prompt (seeking review)

### LINKED ISSUE #1114: Silhouette triggers successful run twice
Example: when I run on HQ with a security testing it triggers the two credits, asks me for an expose, then triggers again. I've found this causes double viruses on pheromones and hemorrhage. This does not occur when I press "Done" instead of accessing. It may be related to my use of Eater, not sure.

### LINKED ISSUE #1122: Turntable + R&D multiaccess
When running R&D with multiaccess (only seen with Medium and R&D Interface, but there may be others) -- if an agenda is accessed, you can click turntable to choose to swap that agenda.  After that agenda is swapped and the run is over, instead of ending the run, there's a prompt for "successful run / jack out".  (instead of having the normal menu of options)

Hitting "jack out" gets around this, but I (and other people I was spectating) have accidentally hit successful run and accessed more cards than we should have.

I'll upload a screenshot if I can catch it doing this error again.

JoelCFC25:
Fixes #1114, fixes #1122. 

I'd definitely like some others on the team to pull this and test it out. I spent a _lot_ of time tracking this issue's cause down and trying various approaches. Ultimately the problem stemmed from [this line](https://github.com/mtgred/netrunner/blob/master/src/clj/game/core-runs.clj#L471) in `handle-end-run`--it was returning false and never processing the rest of the function. 

Splitting it into two parts allows "normal" runs to conclude without a hitch while also allowing time for the Silhouette + SecTesting and Retrieval Run + Femme/Parasite cases to play out fully.

### DIFF
--- a/src/clj/game/core-runs.clj
+++ b/src/clj/game/core-runs.clj
@@ -1,6 +1,6 @@
 (in-ns 'game.core)
 
-(declare clear-run-register!
+(declare clear-run-register! run-cleanup
          gain-run-credits update-ice-in-server update-all-ice
          get-agenda-points gain-agenda-point optional-ability
          get-remote-names card-name can-steal?)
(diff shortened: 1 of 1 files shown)


Saintis:
Tested this. I'm happy this works.

mtgred:
👍

PR #1522, opened 2016-05-13
fix Parasite and Faerie

### LINKED ISSUE #1460: Faerie implemented incorrectly?
screen shot 2016-04-22 at 12 32 49 pm

Faerie shouldn't be breaking "any number of subroutines" for free, it should break each individual subroutine for free. Typically, this difference doesn't matter, but it does in interactions with Midway Station Grid. Faerie should be breaking each sub individually for free, as a separate break. Not sure if it's actually doing that, based on the screenshot pasted here.

### LINKED ISSUE #1464: Runner trashed Parasite, host ICE strength was not updated.
See log / screenshot (the first Parasite was on the Tollboth)

image

JoelCFC25:
Fix #1464: Updates all ICE if Parasite is trashed (voluntarily or not) so strengths will show correct values. 

Fix #1460: Cosmetic only, but a more accurate message when using Faerie.

### DIFF
--- a/src/clj/game/cards-icebreakers.clj
+++ b/src/clj/game/cards-icebreakers.clj
@@ -286,7 +286,7 @@
 
    "Faerie"
    (auto-icebreaker ["Sentry"]
-                    {:abilities [{:msg "break any number of sentry subroutines"
+                    {:abilities [{:msg "break a sentry subroutine"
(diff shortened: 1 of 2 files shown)


mtgred:
👍

PR #1525, opened 2016-05-14
disable SYNC when Employee Strike is played

### LINKED ISSUE #1513: Employee Strike does not disable SYNC

JoelCFC25:
Fixes #1513. 

The ID's side has to be checked to see which effect to reverse, then a corresponding `:effect` is added to allow the ID to be re-enabled (has a turn check so it won't activate at game start).


mtgred:
👍

PR #1529, opened 2016-05-16
implement Personality Profiles

JoelCFC25:
After some discussion in Slack, I'm not sure we have a better way of identifying when the Runner searches their Stack other than manually inserting a custom event trigger. Therefore, that's what this patch does to [this list of cards](https://netrunnerdb.com/find/?q=x%3A%22search+your+stack%22&sort=set&view=list&_locale=en).


mtgred:
👍

Issue #769, opened 2015-09-28
Feature Request: Referee Mode


thehunter518:
Similar to how spectator mode works, this would probably be a second checkbox in game creation "Allow Referee(s)". 

A referee would be like a spectator, but could see player hands and unrezzed cards, and perhaps could interact with the board (dragging cards, etc) as if they were a player. While this could be used for a more serious tournament setting, I imagine it being mainly used as a more intimate spectator mode (between friends, for example) or as a helpful tool for teaching other players the game.

Merlin8000:
At first glance I didn't care for this idea but after a second read I have to +1 this assuming that it included required approval from both players.

Saintis:
This is partially possible by allowing spectators to view the players' hands.

JoelCFC25:
There is now an option to let spectators see hidden cards, so I think that's close enough to call this fulfilled. Not really wild about ever letting spectators move things around and interact with the board state.

PR #1532, opened 2016-05-17
implement Harishchandra Ent. (seeking review)

JoelCFC25:
Tip of the cap to @queueseven for getting me over the finish line with the gameboard modifications! 

Harishchandra has a watch state very similar to Rachel Beckman's. I've tested the case of a tagged Runner playing Employee Strike and it all works--the Runner's Grip is immediately concealed and then revealed again if they're still tagged when Employee Strike leaves play and the identity is switched back on. Not sure if `core-rules.clj` is the best place for the reveal/conceal functions, arguably they could go in `core-misc.clj` but it probably doesn't matter _that_ much. 

harishchandra

A lot of people are clamoring for a player option to reveal their hand only to spectators (and if so, muzzle them so they can't grief the game using chat). Maybe this is a jumping off point for that feature?

### DIFF
--- a/src/clj/game/cards-identities.clj
+++ b/src/clj/game/cards-identities.clj
@@ -206,6 +206,21 @@
    {:events {:pre-ice-strength {:req (req (and (ice? target) (has-subtype? target "Bioroid")))
                                 :effect (effect (ice-strength-bonus 1 target))}}}
 
+   "Harishchandra Ent.: Where Youre the Star"
+   {:effect (req (when tagged
+                   (reveal-hand state :runner))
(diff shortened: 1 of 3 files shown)


zaroth:
If you're talking about referee mode #769, it very well may be a starting feature for that :)

Saintis:
Tested and looks good. Although I feel some kind of indication to the runner that their hand is currently revealed could be useful, so that they don't forget. Or know that Employee Strike works.

This could be done with either a message to the log, or with some sort of indicator text on the Grip saying it's revealed.

mtgred:
👍

Issue #1341, opened 2016-03-12
Label influence dots with numbers for huge influence values in deckbuilder


zaroth:
One picture worth a thousand words:

http://i.imgur.com/aueCFEO.png

To prevent it from happening, make it so for total influence values above a certain threshold influence points aren't displayed as individual dots anymore. Use numbers instead, e.g. '22 o (red), 11 o (violet)', etc.

Tondo:
If no one's looking into this already, I'll take a crack at it. Looks like a nice intro into the front-end code.

domtancredi:
If this is merged and deployed, can this issue be closed?

PR #1536, opened 2016-05-19
Fix large quantities of influence dots overflowing the display

Kiv:
Thought I would tackle issue #1341 for my first attempt at contributing.

image


queueseven:
Looks good. But personally I would have used `(> num 20)` instead of `(<= 20 num)`, since I find it easier to read, but that's just my opinion ;-)

mtgred:
👍

PR #1547, opened 2016-05-24
Some small card fixes

### LINKED ISSUE #1508: Interaction between Sensie Actors Union and Daily Business Show not captured in log
Cenedra rezzes Sensie Actors Union.
Cenedra uses Sensie Actors Union to draw 3 cards.
Cenedra uses Daily Business Show to draw additional cards.
Cenedra uses Daily Business Show to add 1 card to bottom of R&D.

If I understand this, four cards get drawn, two get put back on the bottom of R&D? The automation was right (I think) but the log doesn't show that.

### LINKED ISSUE #1534: Jackson needs a :activatemsg
otherwise, it seems like he was just trashed

### LINKED ISSUE #1539: Activist Support/Paparazzi is not giving a tag
Activist Support's condition is "if you have no tags," but the game seems to be treating it as "when not tagged." It's rarely important, but it is slightly wrong.

### LINKED ISSUE #1540: Cortez Chip should be able to target rezzed ICE
The `:req` is wrong--there aren't _that_ many occasions where you'd want to do this, but to match the card text, Cortez should be allowed to target any installed ICE, not just unrezzed pieces.

JoelCFC25:
* Fixes #1508: add message to Sensie Actors Union
* Fixes #1534: add activation message to Jackson Howard
* Fixes #1539: corrects Activist Support tag condition
* Fixes #1540: Cortez Chip should be able to target rezzed ICE

### DIFF
--- a/src/clj/game/cards-assets.clj
+++ b/src/clj/game/cards-assets.clj
@@ -417,6 +417,7 @@
    "Jackson Howard"
    {:abilities [{:cost [:click 1] :effect (effect (draw 2)) :msg "draw 2 cards"}
                 {:label "Shuffle up to 3 cards from Archives into R&D"
+                 :activatemsg "removes Jackson Howard from the game"
                  :effect (effect
                           (move card :rfg)
                           (resolve-ability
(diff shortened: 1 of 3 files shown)


mtgred:
👍

PR #1550, opened 2016-05-25
Consulting Visit

mharris717:
Not sure if the effect needs a when-completed or something similar. Didn't believe I needed to have a more complex piece of code using resolve-ability, but I'm a beginner in the codebase so please let me know.


mtgred:
👍

PR #1557, opened 2016-05-27
implement Puppet Master

JoelCFC25:
This adds the new Puppet Master agenda and modifies all the ambush assets so they'll work in the case of the Corp targeting them for the placed advancement at the last second before the Runner accesses it. Aggressive Secretary got further additions to mimic the better prompt and message that Shattered Remains has in the case of multiple trashes.


mtgred:
👍

PR #1559, opened 2016-05-27
Rebirth Implementation

mharris717:
I send all rebirth-able identities as part of the JSON payload for a deck in server.coffee. Opinions on whether this is a good idea or not are welcome. 

## Major Issues
- Passive once per turn abilities (Kate, Gabe, etc) will fire if their condition is met after a Rebirth, even if the condition was met with a different action before the Rebirth. I added a message when Rebirth is played to notify players of this. 

- Recurring credit display is a bit wonky. If you go Whizzard -> Someone Else, his ability no longer works but the 3 recurring credits visually appear at the start of turn.

## Minor Issues
- The message when Rebirth is played doesn't say what ID was picked. This isn't an actual issue to me, since you can obviously see the new ID.

- Not positive, but I think how I removed the card from the game was a bit hacky. Levy works by moving the card from play-area to rfg, but I have to move the last card in the discard pile. Pretty sure this has to do with how I used final-effect instead of a different macro, but I don't really understand what's happening, since I was just cargo culting when I chose final-effect

---
This change is [Reviewable](https://reviewable.io/reviews/mtgred/netrunner/1559)

### DIFF
--- a/src/clj/game/cards-events.clj
+++ b/src/clj/game/cards-events.clj
@@ -596,7 +596,7 @@
                                              (not (= "Draft" (:setname c)))
                                              (not (= (:title c) (-> @state :runner :identity :title)))))
                         swappable-ids (filter is-swappable @all-cards)]
-                        (cancellable swappable-ids)))
+                        (cancellable swappable-ids :sorted)))
 
      :effect (req
                (move state side (last (:discard runner)) :rfg)
--- a/src/clj/game/cards-events.clj
+++ b/src/clj/game/cards-events.clj
@@ -588,6 +588,38 @@
     :choices {:req installed?} :msg (msg "access " (:title target))
     :effect (effect (handle-access targets))}
 
+   "Rebirth"
+   {:msg "change identities"
+    :prompt "Choose an identity to become"
+    :choices (req (let [is-swappable (fn [c] (and (= "Identity" (:type c)) 
+                                             (= (-> @state :runner :identity :faction) (:faction c))
+                                             (not (= "Draft" (:setname c)))
+                                             (not (= (:title c) (-> @state :runner :identity :title)))))
+                        swappable-ids (filter is-swappable @all-cards)]
+                        (cancellable swappable-ids)))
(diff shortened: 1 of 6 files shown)


mharris717:
Properly handles link now

nealterrell:
I am working with @mharris717 on resolving some of the issues here. Part of the rework will be a tweak to the web server so that it sends a full list of all cards in the database to the game server on launch, so game logic will have access to a list of all cards in the game, not just the cards in the players' decks.

mharris717:
Pushed an update

mharris717:
My brain is fried on Clojure, I can't do any more today. I likely won't be able to touch anything till next week. 

Even with the 2 minor things, I'd consider this ready to release. If anybody wants to test, I'd certainly appreciate that.

I went a bit overboard on adding convenience stuff to the tests. If anyone feels strongly about what I added I can remove anything. It was just too much of a hassle to break up the commits in advance. @JoelCFC25

nealterrell:
Tested a bit, only issue I saw was when Rebirthing as Kate, the first entry in the selection prompt is a blank button:

image

That can be tackled later, I think this is good for now (@mtgred).

queueseven:
> maybe we should consider only showing up to the colon to fix the spacing issues

Then we should take that this does not break the card preview feature, since it just looks up the text of the button.

mharris717:
@nealterrell adding the sorting now. I had them sorted originally and it got lost when I did V2. I'll take a brief look at why there's a blank but see if it's super obvious.

mharris717:
Added sorting. That blank is Chaos Theory, displays correctly for me. Unicode issue cause of the ü?

image

mharris717:
Doubt the Whizzard issue will come up much, since who would Rebirth OUT of Whizzard

nealterrell:
Ok, forget the colon thing. My computer has always had issues with weird Unicode characters, so that must explain CT.

mtgred:
Awesome 👍

PR #1560, opened 2016-05-28
Surat City Grid not triggering on rez of an upgrade in the same remote

benhosp:
…rezzed.

### DIFF
--- a/src/clj/game/cards-upgrades.clj
+++ b/src/clj/game/cards-upgrades.clj
@@ -356,8 +356,7 @@
    {:events
     {:rez {:req (req (and (= (card->server state target) (card->server state card))
                           (not (and (is-central? (card->server state card))
-                                    (= (card->server state target) (card->server state card))
-                                    (is-type? target "Upgrade")))
+                                    (= (card->server state target) (card->server state card))))
                           (not= (:cid target) (:cid card))
                           (seq (filter #(and (not (rezzed? %))
                                              (not (is-type? % "Agenda"))) (all-installed state :corp)))))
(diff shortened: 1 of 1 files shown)


JoelCFC25:
I see now that something is still wrong with the `:req` after the last (8th?) time I futzed with this card--indeed in its current form it's not triggering on the rez of an upgrade installed in the same remote. However, with your change it's not triggering on the rez of ICE protecting its remote either. 

I probably got myself all contorted with `not` and `and`...the attempt was to ensure it responds to rezzes installed in SCG's server when it's a remote, but not when it's a central.

benhosp:
Wait, but it *should* be triggered by rezzing other upgrades in its server even when it's on a central server, right?

benhosp:
You're right though that the rest of that conditional got me confused. But the point is that it should be triggered by rezzing (1) Ice on its server, (2) Asset in its server, (3) Upgrades in its server, none of which actually depends on whether it's in a central or remote (because if you could somehow install an asset in a central, rezzing it would still trigger SCG).

Meaning that rolling back the rest of your changes from that commit should fix it?

queueseven:
Surat City Grid should not trigger if it's installed in the root of a central server and another upgrade in the root of the same central server is rezzed. Because upgrades in a central server are not *in* the server, but *in the root* of that server.

JoelCFC25:
Correct, and [confirmed in the Business First UFAQ](http://ancur.wikia.com/wiki/Business_First_UFAQ/Question_4).

benhosp:
Huh. Well that certainly makes things more difficult.

benhosp:
You should probably kill this pull request because I don't understand clojure enough to fix the "doesn't trigger on upgrades" problem without causing the "does trigger on upgrades in central servers" problem.

JoelCFC25:
I will leave it open but retitle it, since there is still a legit bug that you helpfully identified--SCG is not triggering on the rez of an upgrade in the same remote server.

JoelCFC25:
Ah yes, this is a PR instead of an issue. I guess you're right, I'll create a new issue instead.

PR #1562, opened 2016-05-28
Fixes Ixodidae interaction with psi games. Adds a test.

### LINKED ISSUE #1546: Ixodidae gives credits for psi games
Currently Ixodidae gives credits to the runner when corp spend money during psi games. Spend is not the same as lose though, so this is wrong.

Saintis:
Changes the `lose` function in `resolve-psi` to `gain` and makes corp / runner gain negative amount instead. This will not trigger `:corp-lose` or `:runner-lose`.

Fixes #1546.


mtgred:
👍

PR #1567, opened 2016-05-30
fix The Turning Wheel to work on Sneakdoor runs

### LINKED ISSUE #1558: Can't use Turning Wheel power counters during a Sneakdoor run
When on a sneakdoor run, turning wheel will not allow you to remove power counters.

JoelCFC25:
Fixes #1558. 

Instead of setting the `access-bonus` directly as a result of spending 2 counters, register a `:successful-run` event contingent on HQ/R&D success. The card's events get unregistered and re-registered in `:run-ends` so that further standard HQ/R&D runs don't keep accessing an extra card.


mtgred:
👍

PR #1572, opened 2016-05-31
Load all cards into the game server on launch

nealterrell:
This adds a variable `all-cards` to the Clojure game server that is available to all game logic. That variable contains a list of all cards in the database, and can be filtered, mapped, etc. according to game needs. The data for the list is actually loaded by the web server when it first starts, and is sent as a ZMQ message to the game logic server once and only once. The game server then stores the list as an atom.

This should help @mharris717's implementation of Rebirth. It should also help with automating Targeted Marketing (but that will require UI work for a general text box prompt). It can also be used to validate decks server-side, by replacing the deck data sent by the client with authentic card data from the game server.


queueseven:
Other cards than can benefit from this: Adam and Salem's Hospitality

mtgred:
👍

PR #1584, opened 2016-06-03
Fix trash prompt, Salsette Slums, Surat City Grid

### LINKED ISSUE #1553: Salsette Slums + Trashing Crisium Grid = Crisium Grid PermaActive?
Had this happen in a game today, had Crisium Grid on HQ rezzed. Runner ran HQ and trashed Crisium using Salsette Slums thus removing it from the game. From then on whenever he ran HQ successfully he did not get his Datasucker virus tokens. Haven't tried this with other 'on successful run' effects but would suspect the same.

### LINKED ISSUE #1581: Surat City Grid not triggering on rez of an upgrade in the same remote
Surat City Grid is correctly not triggering when in the root of a central server and another upgrade in that same server root is rezzed, but as a side effect it's not triggering on upgrade rezzes in the same remote now. 

As identified in #1560 by @benhosp.

JoelCFC25:
* Fixes #1553: Any card removed by Salsette Slums needs to be deactivated so its effects don't persist. Also tweaked the toast on this card so it won't toast at all when accessing cards that have no trash costs.
* Fixes #1581: This time Surat City Grid really, _really_ is fixed--and has a test to prove it!
* Tidy up the trashing prompt in `access-non-agenda` so you don't get gibberish by referencing an undeclared variable: 

image

### DIFF
--- a/src/clj/game/cards-resources.clj
+++ b/src/clj/game/cards-resources.clj
@@ -818,14 +818,16 @@
    "Salsette Slums"
    {:events {:runner-install
              {:req (req (= card target))
-             :effect (effect (update! (assoc card :slums-active true)))}
+              :effect (effect (update! (assoc card :slums-active true)))}
              :runner-turn-begins
              {:effect (effect (update! (assoc card :slums-active true)))}
              :pre-trash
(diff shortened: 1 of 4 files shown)


mtgred:
👍

PR #1587, opened 2016-06-04
Fix Snitch for rezzed/no ice

kevkcc:
I noticed that Snitch works on rezzed ice as well as when there is no ice. This is a small fix to patch this. Includes a test.


mtgred:
👍

PR #1588, opened 2016-06-04
Information Sifting; improvements to when-completed

nealterrell:
Already merged this with Joel's #1584.

Implementation and test for Information Sifting. We ask the corp to select up to (n - 1) cards for the first pile, the implicitly gather the second, and shuffle each. Runner chooses which pile to access, then uses the familiar "Access 1 card from pile" routine until all cards are accessed. Thanks to #1582, each access fully resolves, including damage/access/steal events, before the next access occurs.

Included a few minor improvements to the when-completed framework, particularly for abilities whose `:req` is false and thus don't get resolved... these abilities will immediately trigger `effect-completed`. Noticed this when accessing Psychic Field from hand.

Also adds a priority to Chronos Protocol's prompt so it can't get buried under other prompts.

### DIFF
--- a/src/clj/game/cards-events.clj
+++ b/src/clj/game/cards-events.clj
@@ -377,6 +377,48 @@
                                        {:msg "gain 2 [Credits]" :effect (effect (gain :credit 2))})
                                      card nil))}
 
+   "Information Sifting"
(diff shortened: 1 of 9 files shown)


mtgred:
👍

PR #1589, opened 2016-06-04
allow Archangel trace to be fired during access

JoelCFC25:
This is a semi-frequently requested feature--currently if the Runner hits an Archangel in R&D or HQ, there is no way to fire its trace, forcing players to use the console command to initiate a trace. 

With this addition, the Runner is given a prompt to initiate Archangel's trace after the Corp has chosen to pay to force the encounter. They should respond "No" if they are going to break, and "Yes" if they can't (or won't) break the subroutine. The Runner also sees a wait prompt while the Corp is deciding whether or not to pay the 3 credits.

screen shot 2016-06-03 at 8 29 07 pm

### DIFF
--- a/src/clj/game/cards-ice.clj
+++ b/src/clj/game/cards-ice.clj
@@ -171,12 +171,23 @@
 ;;;; Card definitions
 (def cards-ice
   {"Archangel"
-   {:access {:optional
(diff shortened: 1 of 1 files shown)


mtgred:
Awesome 👍

PR #1591, opened 2016-06-05
Out of the Ashes; /rfg command

### LINKED ISSUE #1590: Feature Request: Remove From Game console command
It would be useful to have a console command to remove a card from the game (and possibly to return it to the game?). 

The specific case I'm thinking of was Out Of The Ashes. There's no way to manually resolve it at the moment.

nealterrell:
Implements Out of the Ashes and a `/rfg` console command (fixes #1590). Requires #1588.

Out of the Ashes registers a `:runner-phase-12` event when it ends up in the Heap for any reason. To do this, I added a `:move-zone` effect for card-defs that will fire when the card moves zones. (@kevkcc may find this helpful for Subliminal Messaging). If Ashes sees that it's now in `:discard`, it will hook an event to prompt at runner's turn start to invoke the removal effect. If the runner answers yes, an instance of Ashes will be moved to `:rfg`, and a run will be invoked. _Once that run ends_, another prompt will be shown if there are additional Ashes in the heap. The whole process ends once there are no Ashes or the runner says no.

The Ashes prompt has a `:priority` of -1 so that it will be hidden by any other prompts that activate because the runner clicked a card during Runner Step 1.2.

This prompt will show every turn while there is at least one Ashes in the heap. We don't have a way of clicking a card in the heap to activate an ability, so this will have to do.

I extended the `when-completed` framework to include `run`, so now a card can wait for a run to complete before triggering a followup effect... in this case, showing the next Ashes prompt.

### DIFF
--- a/src/clj/game/cards-events.clj
+++ b/src/clj/game/cards-events.clj
@@ -537,6 +537,40 @@
     :effect (effect (as-agenda :runner (first (:play-area runner)) 1))
     :msg "add it to their score area and gain 1 agenda point"}
 
+   "Out of the Ashes"
+   (letfn [(ashes-run []
+             {:prompt "Choose a server"
+              :choices (req runnable-servers)
+              :delayed-completion true
+              :effect (effect (run eid target nil card))})
+           (ashes-recur [n]
+             {:prompt "Remove Out of the Ashes from the game to make a run?"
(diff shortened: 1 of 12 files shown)


mtgred:
👍

PR #1592, opened 2016-06-06
Add a Beginner room.

newobj:
Reasons to support a beginner room:

* Beginners often play much more slowly than experienced players, and it's irritating for experienced players.
* Beginners don't learn by getting creamed by experienced players.
* Experienced players don't learn by creaming beginners.
* Beginners should/may want to limit to Core for simplicity's sake and this won't be fun for experienced players.

### DIFF
--- a/src/cljs/netrunner/gamelobby.cljs
+++ b/src/cljs/netrunner/gamelobby.cljs
@@ -244,7 +244,8 @@
                             roomname " (" (count-games room) ")"])]
             [:div.rooms
              (room-tab "competitive" "Competitive")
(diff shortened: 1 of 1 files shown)


Saintis:
I think this could be a good idea, however I would say for this to work well we also need to have an indication of number of open games in each room to encourage people to play with beginners, so people aren't just sat there forever.

zaroth:
It would be a good idea if we had a bigger userbase, as it stands, the beginner room would just stay empty most of the time and split the attention of users / potentially increase waiting time... Casual room with a name containing word "beginner" is the recommended solution for now.

PR #1596, opened 2016-06-08
User-determined resolution of simultaneous triggers on agenda-score/agenda-steal events

nealterrell:
As before, this needs #1591.

This is my prototype for allowing the player to choose the order of resolving simultaneous triggers. ~~I don't think this should be merged for another day or two to give some others time to test the changes.~~ I think this is pretty good now and am up for merging it, thanks to some testing with Joel.

### Simultaneous event resolution

I introduce a replacement for `trigger-event` called `trigger-event-simult`. Using the `when-completed` code and some lessons learned from my recent card implementations, `trigger-event-simult` will gather all listeners for a given event and break them into two groups based on the player who owns the card receiving the trigger. We then:

1. Resolve all triggers for the active player.
  0. Inform the opponent that the active player is resolving triggers.
  1. Present a menu to the user listing each card that is listening for the trigger.
  2. Upon menu selection, trigger the given card and wait for its ability's full resolution.
  3. Repeat until all listeners are resolved.
2. Resolve all triggers for the opponent following the same steps.
3. The event has now been resolved, and we can return to the code that triggered the event.
4. This is all recursive: if the resolution of a listener triggers another event, and that event is hooked to `trigger-event-simult`, then all listeners for that trigger must be resolved before going back to any remaining listeners for the original trigger. ("Cascading triggers" on ANCUR.)

A few techniques make this less tedious for players:

1. If there is only one unresolved listener (left) for the event, that listener is resolved immediately. Example: if I am Leela and have Logos installed, on agenda score I can choose (from a menu) to resolve Leela first. Once that ability is resolved, Logos will immediately fire without another click.
2. If there are many unresolved listeners left and none of them require user interaction, they will be automatically resolved. In this case, "user interaction" is indicated by an ability flag `:interactive` which is a function returning either true or false. If all listeners lack this flag or their flag returns false, then none of them care about the order in which they are resolved. Example: if I have three Fan Sites installed, does it matter in which order they are added to my scored area?
3. The two rules above apply recursively. If I am Leela and have three Fan Sites installed, I will get a menu to choose between 4 different listeners on agenda score because Leela is `:interactive`. If I resolve Leela first, I now have 3 non-interactive abilities that resolve immediately.

### User interface

Currently, the menu for selecting a trigger to resolve only displays the receiving card's name:

image

This could be changed, by giving abilities a `:label` describing what they do (in the image above, "Remove heap from game" and "Do 1 net damage"). Or maybe someone has a prettier solution in mind.

### Applications
I have implemented and tested simultaneous event resolution __only__ for agenda score/stolen triggers. These have historically been the most important interactions requiring manual ordering, so I started here. Highlights:

1. Jinteki: PE can deal net damage before resolving a scored Chronos Project, so the card trashed from the net damage is removed from the agenda score. Likewise, can deal 1 net damage before resolving Philotic Entanglement, increasing the % chance for a kill past I've Had Worse.
2. Team Sponsorship, Leela Patel, and Gang Sign have been cleaned up considerably now that the user can select their order of execution.
  3. In particular, on agenda score, Leela with Logos and 2x Gang Sign can trigger Logos first to retrieve a card, then trigger Leela to bounce a card, then Gang Sign to access, with the bounced card valid for access. If that results in a steal, then a nested agenda-stolen trigger must be resolved (requiring a Leela bounce) before the second Gang Sign can be activated. If THAT second Gang Sign results in a steal, then yet another Leela bounce must happen before the original `:agenda-scored` event is "finished", and the code in `core/score` can continue.

### Future

It won't be too much work to expand simultaneous resolution to other events, but it will take some testing and reworking of some abilities. We'll identify the most pressing issues and leave the rest until an important new card effect mandates the extension of the system. I do believe that this system will lead us to __100% rule enforcement__, as we now (for the first time) have the ability to enforce manual resolution of simultaneous triggers, as well as "cascading triggers" rules.

### DIFF
--- a/src/clj/game/cards-agendas.clj
+++ b/src/clj/game/cards-agendas.clj
@@ -24,23 +24,24 @@
               :effect (req (corp-install state side target nil
                                          {:no-install-cost true :install-state :rezzed-no-cost})
                            (trigger-event state side :rez target)
-                           (when (< n i)
-                             (resolve-ability state side (abt (inc n) i) card nil)))})]
+                           (if (< n i)
+                             (continue-ability state side (abt (inc n) i) card nil)
+                             (effect-completed state side eid card)))})]
      {:optional {:prompt "Look at the top 3 cards of R&D?"
                  :yes-ability {:effect (req (let [n (count (filter ice? (take 3 (:deck corp))))]
-                                              (resolve-ability state side
-                                                               {:msg "look at the top 3 cards of R&D"
-                                                                :effect (req (doseq [c (take 3 (:deck corp))]
-                                                                               (move state side c :play-area))
-                                                                             (resolve-ability state side (abt 1 n) card nil))}
-                                                               card nil)))}}})
+                                              (continue-ability state side
+                                                                {:msg "look at the top 3 cards of R&D"
+                                                                 :effect (req (doseq [c (take 3 (:deck corp))]
+                                                                                (move state side c :play-area))
+                                                                              (resolve-ability state side (abt 1 n) card nil))}
+                                                                card nil)))}}})
 
    "Advanced Concept Hopper"
    {:events
     {:run
      {:req (req (first-event state side :run))
       :effect (effect (show-wait-prompt :runner "Corp to use Advanced Concept Hopper")
-                      (resolve-ability
+                      (continue-ability
                         {:player :corp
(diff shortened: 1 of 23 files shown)


mtgred:
Amazing ! 👍

PR #1600, opened 2016-06-09
Fix Mumbad City Hall's "install a card" ability.

nealterrell:
Whoops.

The build is failing because the test in the Parasite/Builder PR didn't account for the counter rework.


mtgred:
👍

PR #1608, opened 2016-06-09
Fixes for latest engine upgrades; Sol fix, Leela fix

nealterrell:
Expands the simultaneous-trigger completion code to accessing from servers other than HQ (previous tests were focused on Gang Sign interactions).

Fixes the timing of when a current is trashed when an agenda is stolen or scored.

Fixes lingering prompt issues on successful runs with Leela, Silhouette, etc.

Fixes a few core issues to enable the previous fixes.


mtgred:
👍

PR #1612, opened 2016-06-09
Subliminal messaging full implementation (resubmission)

kevkcc:
Finally had time to finish this. Was pretty tricky to handle all the edge cases but I'm pretty confident I got em all finally. Thanks to @nealterrell for the new :move-zone which simplified a bit of the code. Added a new :mill-effect event hook for when a card gets milled (naming convention follows :trash-effect). Contains 5 tests.


nealterrell:
Out of curiousity, why was the `:mill-effect` necessary?

PR #1613, opened 2016-06-10
Fix for Fan Site, Posted Bounty

nealterrell:
Make sure a card still exists before dispatching an event to it. Make sure the agenda is still there (hasn't been forfeited) when granting agenda points.


mtgred:
👍

Issue #1447, opened 2016-04-14
The runner board is too big in spectator mode


queueseven:
see hand drawn circle below.

![image](https://cloud.githubusercontent.com/assets/3126597/14525724/22532a1e-023f-11e6-98fe-93f12253481f.png)

The cards of the runner are not on top of the screen, but in the center.

zaroth:
It is like this since the "fix for bouncing cards in Chrome" from #1176 and #1189 by @domtancredi.

domtancredi:
Is this at a specific zoom level?

queueseven:
100%

orieenav:
Has to do with margin-top: auto for first-child in the runner-board. Not quite sure why it's there to push everything down... It's a quick fix, though I want to test it.

```
.gameboard .mainpane .board .runner-board > div:first-child {
    /* margin-top: auto; */
```
http://puu.sh/pmB7j/afbbdac359.png
http://puu.sh/pmB7z/1f90e7e94f.png

domtancredi:
Has to do with margin-top: auto for first-child in the runner-board. Not quite sure why it's there to push everything down... It's a quick fix, though I want to test it.

.gameboard .mainpane .board .runner-board > div:first-child {
    /* margin-top: auto; */
http://puu.sh/pmB7j/afbbdac359.png
http://puu.sh/pmB7z/1f90e7e94f.png

domtancredi:
Related to this (I put in place that the "opponent" class is part of the runner to fix layout. However, the spectator doesn't see this, so the position of cards is off.

https://github.com/mtgred/netrunner/commit/2fd1e020041b9c05eb6ad80481646d5de9d85e38#diff-c2267bf25b4d05ad13046306f53b57c6

JoelCFC25:
Is this one safe to close now?

domtancredi:
It's ready for re-test. If it's no longer an issue (from a spectator and corp perspective), then it's ready to close.

PR #1616, opened 2016-06-11
changing conditional to output opponent class for spectator; this fix…

domtancredi:
…es the display of css for spectators since they view as a corp

For Issue #1447


mtgred:
👍

PR #1621, opened 2016-06-11
fix Improved Protein Source

### LINKED ISSUE #1617: Improved Protein Source Gives 4 Credits to Whoever isn't adding it to their score area.
The trigger that pops up says it's giving it to the runner, but if the runner scores it, the corp gains 4 instead.  Assuming trigger is set wrong.

JoelCFC25:
Fixes #1617.


mtgred:
👍

PR #1622, opened 2016-06-11
Patron fix

### LINKED ISSUE #1575: Patron should be optional
Patron is currently set up to be a mandatory like Security Testing, but the card text says "may" so it should be an optional ability instead.

kevkcc:
Fixes #1575

### DIFF
--- a/src/clj/game/cards-resources.clj
+++ b/src/clj/game/cards-resources.clj
@@ -667,7 +667,8 @@
                                                                 (vec)))
                                                  card nil))}}})
    "Patron"
-   (let [ability {:prompt "Choose a server for Patron" :choices (req servers)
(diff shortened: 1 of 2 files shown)


nealterrell:
I'd honestly rather add a "No server"  button to the Patron list. I know @queueseven suggested this course of action but I think it's over-engineering. We go from one mouse click (choose the server or No server) to 3 (click Patron; click server; click Take Clicks to start turn since we forced Step 1.2) if a server is desired (most common case) or 1 (Take Clicks) if no server is desired.

queueseven:
Yeah, good idea, Neal. Less clicks are better.

kevkcc:
Sure that's cool. Made the change.

mtgred:
👍

PR #1626, opened 2016-06-12
More fixes

nealterrell:
Fix ordering of parameters in The Brewery's net damage.

Add `eid` parameter to `:base` function call for cards that need dynamic trace base strengths. 

Fix infinite recursion in `trigger-event-simult` from Slack.


mtgred:
👍

PR #1629, opened 2016-06-13
fix NeoTokyo Grid

### LINKED ISSUE #1627: NeoTokyo Grid - gives credits when it shouldn't
NeoTokyo grid gives a credit whenever a card **in** a server is advanced, but currently on Jinteki.net is paying out if an advancable ice is advanced, such as Ice Wall.  This is incorrect.  It only works for assets, upgrades, and agendas.

JoelCFC25:
Fixes #1627. Beefed up the test a bit, too.


mtgred:
👍

PR #1639, opened 2016-06-13
Fix Film Critic, Maya

nealterrell:
Refactored the code that closes access prompts with non-standard effects (Imp, Film Critic, Maya). Now correctly triggers that the access is over.


mtgred:
👍

PR #1640, opened 2016-06-13
Fixes for Turning Wheel, Ramujan-Reliant, Liberated Chela, Bio-Ethics Association; rework /counter command

nealterrell:
`/counter n`: attempts to infer the type of counter to place on target card. If the card already has counters, use that type. If it is an advanceable card, add advancement counters. If it is a scored agenda, use agenda counters. If it is a virus subtype, use virus counters. Otherwise, player must use:

`/counter type n`: specify the type of counter to add, as `agenda`, `advancement`, `credit`, `power`, or `virus`. Can abbreviate the type by specifying enough letters to distinguish it from the other types, e.g., `ag`, `ad`, `c`, `p`, `v`. Really, anything that starts with `ag` will be an agenda counter, so you could live life on the edge and type `/counter aggrivated-assault 10` to ad 10 agenda counters.

Adds a new ability handler key `:silent true` which will hide an event handler from a simultaneous-trigger ordering prompt. Use when a card needs to handle an event for which simultaneous-triggers have been implemented (right now, only agenda steal/score), but the event handler isn't actually part of the card's printed text, but only used for internal logic reasons. (Example: The Turning Wheel uses `:agenda-stolen` so it can know that an agenda was stolen this run. This is not an ability that the player needs to interact with because it is not an ability actually printed on the card.)

Video on fixing multiple Bio-Ethics Association damages being prevented: https://www.livecoding.tv/nealpro/videos/z8o6Y-jintekinet-resolution-of-simultaneous-triggers

### DIFF
--- a/src/clj/game/cards-assets.clj
+++ b/src/clj/game/cards-assets.clj
@@ -95,6 +95,7 @@
 
    "Bio-Ethics Association"
    (let [ability {:req (req unprotected)
+                  :delayed-completion true
                   :label "Do 1 net damage (start of turn)"
                   :once :per-turn
                   :msg "do 1 net damage"
--- a/src/clj/game/cards-hardware.clj
+++ b/src/clj/game/cards-hardware.clj
@@ -506,9 +506,11 @@
 
    "Ramujan-reliant 550 BMI"
    {:prevent {:damage [:net :brain]}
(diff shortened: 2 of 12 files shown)


mtgred:
👍

PR #1648, opened 2016-06-15
Fix CVS on Archives; fix Replicator + Bazaar interaction

nealterrell:
Extended the simultaneous trigger resolution to the runner-install event, to allow manual ordering for interactions like Replicator + Bazaar. Slightly reworked Haley to work better with this system. Tweaked many cards so they don't show up in the manual resolution prompt (like Kate). 

Fixed CVS on Archives breaking further accesses.


mtgred:
👍

PR #1651, opened 2016-06-16
Expose prevention

nealterrell:
1. Introduce `:pre-expose` event and `expose-prevent` function for reacting to an attempt to expose.
2. Zaibatsu Loyalty: an `:unrezzed-event` to handle pre-expose and show a prompt to rez Loyalty before the expose goes through.
3. Underway Grid: automatically prevents exposing cards in/protecting server.
4. All expose cards have been reworked to only reveal the target exposed if the expose effect resolves.

A lot of work for 2 unused cards! However, this system did require a useful upgrade to the `when-completed` framework which allows async effects to __return__ values when they indicate that they have finished. This is necessary for Drive By, which must know whether the expose completed before triggering the trash. The await handler of a `when-completed` block now has access to a variable `async-result`, which can be optionally provided by a function when it triggers `effect-completed` by associating `:result _____` onto the effect's eid.

### DIFF
--- a/src/clj/game/cards-assets.clj
+++ b/src/clj/game/cards-assets.clj
@@ -1,5 +1,7 @@
 (in-ns 'game.core)
 
+(declare expose-prevent)
+
 ;;; Asset-specific helpers
 (defn installed-access-trigger
   "Effect for triggering ambush on access.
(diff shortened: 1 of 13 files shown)


mtgred:
Awesome! 👍

PR #1652, opened 2016-06-16
add Knight and Rook re-hosting restrictions

JoelCFC25:
A lot of lines for very minor additions, but nonetheless...I give you fully rules-compliant Knights and Rooks!


mtgred:
Nice 👍

PR #1653, opened 2016-06-16
Implement Magnet.

nealterrell:
The `:delayed-completion true` is just future-proofing. On rez, prompt to host a program that is hosted on ice, if there is one. `host` the target, unregister its events, and remove its `:abilities`.

### DIFF
--- a/src/clj/game/cards-ice.clj
+++ b/src/clj/game/cards-ice.clj
@@ -671,6 +671,23 @@
    "Lycan"
    (morph-ice "Sentry" "Code Gate" trash-program)
 
+   "Magnet"
+   {:delayed-completion true
+    :effect (req (let [magnet card]
+                   (continue-ability
+                     state side
+                      {:req (req (some #(some (fn [h] (card-is? h :type "Program")) (:hosted %))
+                                       (remove-once #(not= (:cid %) (:cid magnet)) (all-installed state corp))))
+                       :prompt "Select a Program to host on Magnet"
+                       :choices {:req #(and (card-is? % :type "Program")
+                                            (ice? (:host %))
+                                            (not= (:cid (:host %)) (:cid magnet)))}
+                       :effect (req (let [hosted (host state side card target)]
+                                      (unregister-events state side hosted)
+                                      (update! state side (dissoc hosted :abilities))))}
+                      card nil)))
+    :abilities [end-the-run]}
+
    "Mamba"
    {:abilities [(power-counter-ability (do-net-damage 1))
                 (do-net-damage 1)


FarCryFromHuman:
I'm still learning to read the code, but what happens in the following scenario:

Magnet is rezzed with a Knight on a different piece of ICE.
Knight is chosen as the target, and is hosted on Magnet.
Knight has all events deregistered and all abilities removed.

The runner uses Self-modifying Code to install Parasite on Magnet.
The runner uses Datasucker counters to reduce the strength of Magnet to 0.
Magnet is trashed, and is placed in the corp's Archives.
Parasite and Knight are trashed, and placed in the runner's Heap.
**The runner uses Clone Chip to play Knight from Heap onto the next ICE.**

I'm worried that last bit will break, as Knight has had all of its abilities scrubbed. I'd also worry that hosted cards aren't properly trashed when Magnet is, if they are trashed via registered events.

JoelCFC25:
A couple comments--the part in bold cannot in fact work due to the rules. Clone Chip can certainly resurrect Knight, but it would install back into the Runner's rig, not directly onto a piece of ICE. 

As for the rest, I can confirm it all works just fine, having tested the code. As a general engine feature, a card being trashed automatically results in the trashing of all cards it hosts--the hosted cards don't have their own registered events to have this work. If you follow the code trail, the `trash` function ultimately results in the use of the `move` function, which sequentially trashes all hosted cards if the destination of the `move` is the discard pile.

nealterrell:
Good questions.

First, I am fairly confident that your gameplay example doesn't work. Magnet has a "constant ability" to blank its hosted card, thus the instant Parasite is hosted on Magnet, Parasite's text goes blank. There is no window in which Datasucker tokens or any other triggered (conditional or paid) ability can be used. Now I'm not sure what happens if Parasite already has 3 tokens on it, or if some other shenanigans with constant ability priority (Ice Carver is one) reduce Magnet to 0 strength before Parasite is installed... as far as I'm aware there are no constant ability conflicts like this in the game, so we'll probably need a FAQ ruling.

Your game engine question is still relevant for things like Spooned or trashing in general. Two answers:

1. `:abilities` is actually two different things. In a card's behavior definition, `:abilities` is a vector of ability maps that are triggered by mouse clicks or menu selections on a card in the UI. But these ability maps aren't part of a card's _data_. Instead, when a card is initialized (installed, rezzed, scored, etc.), its data definition is given an `:abilities` vector that consists only of strings. Those strings are the labels shown by the card menu UI. When a menu label is clicked, its index is passed to the server, so the server is basically told "menu item 0 on this card was clicked"; the server then loads the card's behavior definition (`card-def`) and invokes the corresponding index in the `:abilities` vector of that definition. So removing `:abilities` from a card's data only has the effect of removing the menu of abilities from that card in the UI, and only for the one instance of the card.

2. As Joel mentioned, event handlers are used by cards to respond to in-game engine happenings. The engine does all the driving; the cards simply react to certain events to invoke their effects. Importantly, most game rules are not tied to card handlers, but are implemented as core functions. `trash` is one such function, which moves a card to `:discard` and also moves any hosted cards to the appropriate zones. `runner-install` is another, which installs a runner card and initializes its abilities.

Thanks to your question, I did realize a flaw in this implementation: if a runner card is hosted onto Magnet outside of the Magnet rez ability, that card will not be blanked. So if a Pawn advances onto Magnet, or the runner directly installs Knight onto Magnet, the card will not be blanked properly. I'll have to improve those edge cases.

JoelCFC25:
Apparently the [UFAQ will address what happens](https://www.reddit.com/r/Netrunner/comments/4n0woh/post_your_fear_the_masses_questions_here_for/d3zwz4s) with constant abilities when you pop a Parasite onto a Magnet at paid ability speed.

FarCryFromHuman:
Ah, I didn't consider that the text box of Parasite would be blanked upon installing, but that's absolutely how I'd rule it as well. Thanks for the clarifications on the underlying architecture, it will help with bug reports in the future. Glad I could help!

mtgred:
👍

PR #1654, opened 2016-06-17
implement Fear the Masses

JoelCFC25:
The event, not the whole pack!


nealterrell:
Can we condense the two optional prompts into one? If they don't want to reveal any copies, they can choose 0?

JoelCFC25:
I'll try that later today...

mtgred:
👍

PR #1660, opened 2016-06-19
"Report error" button when an exception occurs in game state

nealterrell:
This adds an error toast when an exception occurs in the game engine when applying a user's command. The toast asks the user to report the issue, and a handy button takes the user straight to our GitHub issues page, with a pre-filled issue body containing the stack trace from the exception.

netrunner-error

netrunner-error-github

### DIFF
--- a/src/clj/game/core-io.clj
+++ b/src/clj/game/core-io.clj
@@ -1,6 +1,6 @@
 (in-ns 'game.core)
 
-(declare ice-index parse-command)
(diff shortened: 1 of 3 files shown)


queueseven:
Maybe the complete game state could also be added (zipped and base64'd)? Then we would know the exact circumstances of the error.

mtgred:
Cool 👍

PR #1666, opened 2016-06-21
Fix Puppet Master; implement Dedicated Neural Net; add "card-title" prompts for Targeted Marketing, Salem's Hospitality

### LINKED ISSUE #1661: Puppet Master locks runner after failed trigger
If Puppet Master is scored and if there are no legal targets for Puppet Master, when the corp clears the prompt after successful run the runner's "waiting" prompt does not clear.

nealterrell:
Fix #1661. Gives selection prompts the ability to define a `:cancel-effect` that runs if Done is clicked with no selected targets.

@JoelCFC25's implementation of Dedicated Neural Net. Required enforcing `:once` keys for psi/trace abilities (not sure how that one was missed).

### DIFF
--- a/src/clj/game/cards-agendas.clj
+++ b/src/clj/game/cards-agendas.clj
@@ -468,11 +468,13 @@
 
    "Puppet Master"
    {:events {:successful-run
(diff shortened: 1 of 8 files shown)


mtgred:
:+1:

PR #1668, opened 2016-06-21
implement The Black File

JoelCFC25:
If the Corp is on a game-winning number of agenda points while this is in play, they should immediately win if they manage to trash it or when it hits 3 power counters at the start of a Runner turn. 

Includes tests!


mtgred:
:+1:

PR #1684, opened 2016-06-26
Runner side sub panel

### LINKED ISSUE #1599: Feature Request : Add a "Fire" button...
...in the runner's run menu. When the Runner clicks on it, it adds a text in the chat to inform the Corp that the Corp can fire the subroutines of the encountered ice. ie : "The Corp can resolve the unbroken subroutines of the encountered piece of ice."

Saintis:
Adds a panel to the runner's side where they can ask / indicate for specific or all subs to be fired.

Also adds a new thing to ICE: `:runner-abilities` that are abilities only the runner can trigger. Currently these are used to implement Bioroid ice clicking, Negotiator and Wotan.

The list of subs to fire is pulled from the `:abilities` list and done client-side. Hence ice with non-subroutine abilities will have extra entries (like Builder and Data Raven).

Like for the corp, this panel is openable always, not just during runs.

Screenshots:
screen shot 2016-06-26 at 17 52 06
screen shot 2016-06-26 at 17 52 32

Intends to fix #1599.


Saintis:
Now also allows runner to choose tag or etr on Data Raven. CSS is a bit poor though. Maybe someone with so strong CSS-fu can do something nice? 😄

Saintis:
DO NOT MERGE

Have ICE subroutine structure rewrite coming as well, just need to test locally.

Saintis:
Mostly ready to merge now. 

Splits ICE abilities into `:subroutines` (for subroutines on the ice) and `:abilities` (for other abilities, e.g. Builder and Tollbooth). Updates tests to work with this. Only scrapes `:subroutines` for runner's list.

Have tested that Bioroids 1.0 & 2.0, Turing, Wotan, Negotiator, Data Raven and Pop-up Window work.

Only issue is that the css of the runner-panel is odd. Can only select buttons on the left hand side of the buttons. And their layering is odd. Unfortunately my CSS-skillz are not strong enough to figure out what is going on.

Final notes: this is a minor rewrite on how ice works so some mild review could be useful.

mtgred:
There are merge conflicts. Can you resolve them?

I can take care of the CSS if necessary.

Also, instead of notifying the corp to execute the subroutines, we could perhaps let the Runner execute them directly the subroutines directly. What do you think?

Awesome job as usual. 👍

Saintis:
I will have a look and fix it this evening. 

As for runner firing subs it was discussed in the issue. Maybe worth as an extension in a different PR but I think it's better to let the corp fire things. They might want to do abilities when runner passes. And some subs would have to be rewritten or extended to allow runners to use them (trashing programs etc). 

Probably just easier to notify the corp. The sub that could be implemented would be "End the run" maybe.

Saintis:
Will resolve conflicts this weekend.

Saintis:
This now ready to merge again.

Some CSS issues though with the server `div` to the right of the server run on lying on top of the panel preventing clicks.

screen shot 2016-07-24 at 17 29 25
The red area is not clickable.

Saintis:
Should be ready for merge once again ;)

Saintis:
Ready for merge

Saintis:
Should be ready for merge again.

nealterrell:
Let's do it.

Saintis:
Will fix this afternoon.

Saintis:
Ready to merge again.

This PR actually does 3 things:

1. Splits `:abilities` into `:subroutines` for proper subs and `:abilities` for other things (like announcing # of subs, Tollbooth tax, Builder click-ability or Himitsu-Bako 2 credits to return to hand).
2. Adds a panel for the runner to specify `:subroutine` for Corp to trigger (or "All")
3. Adds `:runner-abilities` that the runner can use (like clicking Bioroids, paying for Pop-up Window, choosing ETR / Tag on Data Raven)

2 depends on 1 to only list the subroutines (and not the Builder ability for example), and 3 depends on 2 in the sense that the `:runner-abilities` are listed in the same panel.

Maybe if I extract out 1 into a separate PR and get that merged first we can deal with 2 and 3 (and the CSS issues) separately. This could be useful since 1 could be a start for better auto-breaking of ICE. Side benefit would be ability to rebase changes rather than merge with master.

Thoughts?

Saintis:
I can have a look. Either case it is probably best (and probably easier) to do that change in a PR after this is merged.

nealterrell:
👍

PR #1688, opened 2016-06-27
fix New Construction, Freelance Coding Contract, Cerebral Imaging, Qianju PT

### LINKED ISSUE #1410: Card interaction bug: Qianju PT +  Joshua B. should prevent tag
Using Qianju PT should prevent tag from Joshua B. Right now the runner gets a tag at the end of turn (Qianju PT doesnt trigger)

### LINKED ISSUE #1681: New Construction problem on installing an agenda.
New Construction panicked on installing an agenda on its fourth advancement. I don't know whether it's a bug with agendas or the fourth advancement or whatever as I don't normally run it (it was for a draft).

Stack Trace:
```clojure
#error {
 :cause nil
 :via
 [{:type java.lang.ClassCastException
   :message nil
   :at nil}]
 :trace
 []}
```

### LINKED ISSUE #1682: Freelance Coding Contract trashes are prevented by Sacrificial Construct
Sacrificial Construct should only prevent installed programs/hardware from being trashed anyway, so I'm not sure why this is triggering.

JoelCFC25:
* Fix #1681: Drop the `eid` parameter to `corp-install` so New Construction can successfully install something when advanced.
* Fix #1682: Make FCC trash all selected programs sequentially with the `:unpreventable true` parameter. 
* Fix #1410: Qianju PT had a misplaced closing brace that was keeping it from preventing a tag, and also wasn't losing the click if used.
* Fix watch state removal error when Employee Strike disables Cerebral Imaging, and also corrects base hand size adjustments when CI is disabled and later re-enabled.

### DIFF
--- a/src/clj/game/cards-agendas.clj
+++ b/src/clj/game/cards-agendas.clj
@@ -403,10 +403,10 @@
                              :msg (msg "install a card from HQ" (when (>= (:advance-counter (get-card state card)) 5)
                                        " and rez it, ignoring all costs"))
                              :effect (req (if (>= (:advance-counter (get-card state card)) 5)
-                                            (do (corp-install state side eid target "New remote"
+                                            (do (corp-install state side target "New remote"
                                                               {:install-state :rezzed-no-cost})
                                                 (trigger-event state side :rez target))
-                                            (corp-install state side eid target "New remote")))}}}}}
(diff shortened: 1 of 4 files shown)


mtgred:
👍

PR #1689, opened 2016-06-27
Don't trigger I've Had Worse if Runner flatlines

### LINKED ISSUE #1050: I've Had Worse triggering after a flatline
I've Had Worse shouldn't trigger when there is more damage dealt to the runner than he has hands in his grip (e.g. Scorched Earth vs hand of 3 with IHW and no other meat damage prevention).

JoelCFC25:
Fixes #1050. 

When resolving damage, we check if the amount of damage exceeds the Runner's hand size (e.g, Scorched Earth against a Runner with 3 cards). If so, flatline the Runner and discard all their cards unpreventably, but do nothing more. If the Runner survives, also trash cards unpreventably, but add in the `:cause` parameter, which is what I've Had Worse is looking for so it will draw 3.


mtgred:
👍

PR #1690, opened 2016-06-27
Only do trash effect of a Corp card if it's rezzed

### LINKED ISSUE #1485: Trashing an installed, unrezzed Research Station lowers Corp's max hand size by 2
During one game, twice I trashed an unrezzed Research Station from the root of HQ and the Corp's max hand size was changed to 3 both times.

JoelCFC25:
Fixes #1485. 

I uncovered an error in an existing test (Red Herrings) while fixing this--it was giving the Red Herrings steal penalty without the card even being rezzed. Before resolving a trash effect ability, check that it's either a Runner card or a _rezzed_ Corp card. Without this, a Runner trashing an unrezzed Red Herrings or Strongbox before accessing an agenda was wrongly seeing the effect of those cards when trying to do a steal.


mtgred:
👍

PR #1691, opened 2016-06-27
Require ambush assets to be advanced before effects fire

JoelCFC25:
Confirmed a report that the Corp can trash a piece of hardware with Shattered Remains even if the Runner hits one that has 0 advancements. Aggressive Secretary has a similar problem and the existing test didn't catch it. 

This updates all advanceable ambush assets to require at least 1 advancement before they'll do anything even if the Corp pays for the effect when they're empty. This will avoid triggering any prevention/avoidance from the Runner if they get 0 tags or are dealt 0 net damage, etc--or bonus effects from Corp cards like Defective Brainchips after an Overwriter deals 0 brain damage.

### DIFF
--- a/src/clj/game/cards-assets.clj
+++ b/src/clj/game/cards-assets.clj
@@ -67,7 +67,8 @@
                                       (= (:side %) "Corp"))}}]}
 
    "Aggressive Secretary"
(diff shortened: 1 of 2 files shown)


mtgred:
👍

Issue #1347, opened 2016-03-14
Chat log: improving message on creating new server


osiridecascioli:
When the corp create a new server the message is:
"${USER} spends  to install a card in New remote."
It would be useful writing the server number.

zaroth:
I agree, however it's useful to keep an indication of the fact that it's a new server to be able to easily differentiate from installing in an existing one. So the text should be IMHO added, not replaced, like _(...) install a card in New remote - Server 8_.

queueseven:
Well, it's always the server with the highest number... :-)

JoelCFC25:
Implemented at #1692!

PR #1692, opened 2016-06-27
Improved message when installing a card in/protecting a new remote server

Devencire:
Implementation of #1347 - creating a new remote server results in a message including both the new server's name and the fact it's new, e.g. "User spends [click] to install a card in Server 17 (new remote)."


mtgred:
👍

PR #1696, opened 2016-06-28
fix Jak Sinclair interaction with Replicating Perfection

### LINKED ISSUE #1693: Replicating Perfection won't let you run subs
I ran each of the centrals and it still didn't give me the option to run remotes.

For reference, my first run was via Jak Sinclair, so I thought that might be relevant at first. But doing 'normal' runs didn't make any difference.

JoelCFC25:
Fixes #1693. 

Extends Jinteki RP to listen for `:runner-phase-12` in the event that the Runner has Jak Sinclair and uses him to make a run before taking clicks. Jak Sinclair is modified to use `:runner-phase-12` so the Runner can trigger him manually before taking clicks, or just let it ask at `:runner-turn-begins` if it hasn't already been used.

### DIFF
--- a/src/clj/game/cards-identities.clj
+++ b/src/clj/game/cards-identities.clj
@@ -308,11 +308,7 @@
 
    "Jinteki: Replicating Perfection"
    {:events
-    {:runner-turn-begins {:req (req (let [runs (turn-events state side :run)]
-                                      (empty? (filter #(is-central? %) runs))))
-                          :effect (req (apply prevent-run-on-server
-                                              state card (map first (get-remotes @state))))}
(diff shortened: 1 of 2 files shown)


nealterrell:
You could probably simplify this by removing the `:runner-turn-begins` completely. `:runner-phase-12` will trigger just before turn-begins anyway. No big deal though.

JoelCFC25:
You mean remove it from Jak Sinclair?

nealterrell:
Sorry, no -- from RP. Just the `:runner-phase-12` event would probably do it.

JoelCFC25:
Cool, that worked--thanks!

mtgred:
👍

Issue #1675, opened 2016-06-24
Game state error when attempting to rez a Zealous Judge


Makai:
More specifically, I tagged the runner with SEA Source on click 1, Trashed a Resource on click 2, and installed Zealous Judge on click 3 and attempted to rez him right away. Instead of rezzing the judge I got a pop-up mentioning a game state error had occurred. Playing with SYNC as the ID.

nealterrell:
Got it, will fix. Not an error with Zealous Judge code, it's with `can-rez?`.

PR #1697, opened 2016-06-28
Many fixes

nealterrell:
1. Fix arity exception preventing the rez of Zealous Judge, #1675.
2. Fix arity exception in Showing Off, from server error log.
3. Fix arity exception when testing the `:req` of optional on-access effects in Archives, from server log.
4. Fix /counter command with scored agendas.
5. Remove the `[:run :ices]` vector from the state. As far as I can tell this is an artifact of earlier design decisions, and we've had to hack around it for ages, but I believe it can be removed now. Ice data should be pulled directly from a server's `:ices` vector, not from a vector that was being initialized once at the start of the run but not keeping updated to other changes to ice (strength changes, positional changes, etc.). 
6. Add another line of logging to `main.clj`. I don't know what has caused the last few server crashes, and I don't know why they aren't being logged, but this may help.
7. Compile game.* namespaces separately from test.* namespaces to address out-of-memory issue when compiling on Circle CI. @mtgred: you may want to look into excluding the test namespaces from the `lein uberjar` build, since they're unused in live deployment. I briefly looked for a solution but couldn't find one.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -13,6 +13,7 @@ dependencies:
 
 test:
   override:
+    - lein compile game.main
     - lein test test.core
     - lein test test.cards.agendas
     - lein test test.cards.assets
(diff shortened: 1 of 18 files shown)


mtgred:
👍

Issue #468, opened 2015-07-30
Fast clicking leads to getting more than 3 Credits from Cache


queueseven:
In a recent game my opponent got 7 credits from Cache. Some sort of timing issue?

nealterrell:
Yes. There is a general issue in that spamming a command will cause multiple messages to be sent to the server, and no checking is done server-side to distinguish between them. You commonly see this with Cache because people click it quickly.

DominicKexel:
OCTGN has it's <kbd>CTRL</kbd> +<kbd>C</kbd> for multiple credits, we have this :-)

DominicKexel:
The question is if this is worth fixing.

mtgred:
There is a lock mechanism to prevent this from happening. When a client sends an action to the server, a lock is set. When a game state comes back from the server the lock is removed. While the lock is on, any actions sent from the client is ignored. 

This is prevent a card from being played twice when double clicking on it. (It happened very often as playing a card on OCTGN requires a double click). It normally prevents taking too many credits from Cache. A possibility is that your opponent did an action while you were clicking very quickly on Cache. If a game state is received from the opponent's action the lock is unset too.

A more elaborate lock mechanism could prevent this from happening.

There is also another issue with the current simplistic lock mechanism: When the action causes an serious exception from the game engine, no game state is sent back and the client is locked forever. This gives the impression that the client disconnected although the connection is still there. The only way to unlock it is waiting for the opponent to type in the chat or do any other action to receive a game state.

A timeout on the lock could solve this issue. But it could also lead to an inconsistent game state.

queueseven:
> A possibility is that your opponent did an action while you were clicking very quickly on Cache. 

No, it was my opponent that got the credits, and I didn't do any actions in between. He just clicked very fast.

zaroth:
I just tested this, trying a runner with just 3 Caches in deck. I was able to get 4 credits from the first cache I installed, while every next Cache (including ones I dragged from heap back to my hand) I tried gave me only 3 creds. I was able to reproduce this result next game.

zaroth:
To quote a neal's guess on the cause of the issue:

> I see that in -main (which is the production main) two threads are started to listen to the ZMQ socket, but in dev only one thread is going. I wonder if that's why I've never reproduced this on dev.

ouroboros8:
fwiw I experienced a similar issue recently where clicking "successful run" twice in rapid succession triggered two successful runs (creds from desperado twice, accessed card in server twice)

PR #1705, opened 2016-06-30
Refactor main to run on single thread

nealterrell:
As per discussion with @mtgred, refactors `main` to run on a single message-handling thread instead of 2. Expected non-impact on server CPU. With no per-state locking mechanism, a single thread should help weird race conditions like the suspected cause of #468.

Also reworks the message-handling logic to be exception-safe. Should fix the server crashes we've been seeing, in which an exception caused when showing an error report toast wasn't being handled, crashing the server thread.

### DIFF
--- a/src/clj/game/core-io.clj
+++ b/src/clj/game/core-io.clj
@@ -215,10 +215,11 @@
 
 (defn show-error-toast
   [state side]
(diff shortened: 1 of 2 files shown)


mtgred:
👍

PR #1706, opened 2016-06-30
Fix for Dedicated Neural Net, Salsette Slums.

nealterrell:
The req on psi games goes inside `:psi`, not the main ability, but psi games also weren't respecting `:once` keys.


mtgred:
👍

PR #1712, opened 2016-07-01
Clean up AccDiag interactions.

nealterrell:
As reported in Stimhack Slack. My first iteration of the "when-completed" stuff worked well with AccDiag, but when I generalized it I added cleaner code that made AccDiag break again. This fixes that.

Note to self/all: if an ability wants to use `continue-ability` (most should, in the place of `resolve-ability`), then it should be marked `:delayed-completion true`.


mtgred:
👍

PR #1722, opened 2016-07-05
Jack out prevention implementation + other prevention upgrades

kevkcc:
This is an attempt to do jack out prevention along the lines of expose/damage prevention/etc. Only one card uses this right now (Labyrinthine servers).

I added a :pre-jack-out event hook along the lines of the other prevention mechanisms which is not necessary for Labyrinthine servers but decided to keep it in in case future cards may want it.

Also tries to make the prevention mechanisms already within the game more consistent by adding wait prompts to the opposing side when necessary (i.e. wait prompt for corp while runner is preventing tags).

### DIFF
--- a/src/clj/game/cards-agendas.clj
+++ b/src/clj/game/cards-agendas.clj
@@ -352,9 +352,14 @@
                               :effect (effect (init-trace-bonus 1))}}}
 
    "Labyrinthine Servers"
-   {:effect (effect (add-counter card :power 2))
-    :abilities [{:counter-cost [:power 1]
-                 :effect (effect (prevent-jack-out))
+   {:prevent {:jack-out [:all]}
+    :effect (effect (add-counter card :power 2))
+    :abilities [{:req (req (:run @state))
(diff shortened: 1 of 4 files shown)


nealterrell:
Nice job!

Saintis:
Though this was implementing An Offer You Can't Refuse and the like, turns out I just didn't know what Labyrinthine Servers actually did.

Good stuff! An extension task might be a soft-implement on "cannot jack-out" that informs users they are not allowed to jack-out and asks for confirmation when they hit the "Jack out" button.

Additionally I think that the prevention code is looking a bit wet, maybe time to bring out the DRYer. Will add to my ever-growing list.

mtgred:
There are merge conflicts. Can you fix resolve them?

kevkcc:
Yep I'm out at the moment but I can take care of them tonight

nealterrell:
I got yer back, kevkcc, since it's my code that caused the conflict.

PR #1727, opened 2016-07-05
Fix #1717, Explodapalooza / Turning Wheel. Fix advanceable traps.

### LINKED ISSUE #1717: Turning Wheel adds a counter when accessing Explode-a-palooza
Somehow the `:run-ends` event is triggering before `:agenda-stolen`, so The Turning Wheel is getting a counter added when hitting Explode-a-palooza. The steal message and placement of the agenda in the Runner's area doesn't happen until after the Corp is done answering the Yes/No prompt for gaining the 5 credits, and The Turning Wheel is adding the counter before any of that.

### LINKED ISSUE #1732: Damage problem
I was playing a game with Cibernetics, humanity upgraded against Adam, twice in the game he run into a server with an Overwriter with 3 counters and a Project junebug with 2 counters havin just 2 cards in hand that would make him be flatlined and after I clicked to use the triggered ability when he accessed both cards the ability didn't triggered it could be because of one of the directives

nealterrell:
Old run code needs to be cleaned up, but for now, Exploda shows a waiting prompt to the run doesn't end after the runner elects to steal. (Run can end when closing a prompt while `:run` is active in state.) 

Fixes advanceable traps (fix #1732).

### DIFF
--- a/src/clj/game/cards-agendas.clj
+++ b/src/clj/game/cards-agendas.clj
@@ -233,9 +233,15 @@
     :abilities [{:cost [:click 1] :counter-cost [:agenda 1] :msg "draw 5 cards" :effect (effect (draw 5))}]}
 
    "Explode-a-palooza"
-   {:access {:optional {:prompt "Gain 5 [Credits] with Explode-a-palooza ability?"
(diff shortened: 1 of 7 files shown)


mtgred:
👍

PR #1728, opened 2016-07-05
Don't show side/identity for matches with disabled spectators.

nealterrell:
I saw this request somewhere random, but can't remember where. Don't show the faction icon in the game list for a game with disabled spectators, and don't reveal the identity either. The icon was added to make it easier to find games to spectate, which doesn't apply to disabled spectator games, and addresses concerns about wanting to "hide" information about what decks a player is practicing for big tournaments.


mtgred:
👍

PR #1733, opened 2016-07-06
Several spoiled 23 Seconds cards and tests

JoelCFC25:
Implementations for 10 of the cards in the upcoming pack, along with several tests.


mtgred:
👍

PR #1734, opened 2016-07-06
fix Drive By, Project Beale, trashable executives

### LINKED ISSUE #1716: The Board does not go into runner score area when trashing while accessing in HR
While using The Turning Wheel, trashing the board by the Runner when accessed via The Turning Wheel, goes to the Archives and not to the Runner's score area.

JoelCFC25:
Getting some quick fixes in before a deployment! This includes a band-aid for the "add as an agenda" trashable executives (fixes #1716). A more elegant solution would be better in the long run.


mtgred:
👍

Issue #1729, opened 2016-07-05
Out of the Ashes losing heap ability when trashed from stack


FarCryFromHuman:
I just played a game where I had a copy of Out of the Ashes trashed from the top of my stack by Kala Ghoda Real TV. OotA didn't give me the normal prompt at the start of my turn. I drug it from my heap into my grip and trashed it manually and then it resumed giving me the prompt next turn.

JoelCFC25:
Probably needs to employ `:mill-effect` like Subliminal Messaging does.

FarCryFromHuman:
Closing this since it was repaired last restart.

PR #1737, opened 2016-07-06
Fix for #1729 (out of the ashes)

kevkcc:
As with subliminal messaging, use :mill-effect to register the event when milled. Perhaps a better solution in the future would be to rewrite the mill code to use move. Not entirely sure what the ramifications of that would be though.


mtgred:
👍

Issue #1118, opened 2016-01-16
[Request] Change order of Advance/Score menu on agendas


Merlin8000:
Assets have the option to advance on the bottom of the context menu.  Sometimes when trying to over-advance agendas, I've found myself reflexively clicking score.  This is normally not an issue, but Mushin No Shin doesn't allow the agenda to be scored until the start of the next turn.  

For consistency, I think it would be good to keep Advance as the last menu item for both card types.

JoelCFC25:
![screen shot 2016-06-23 at 3 16 27 pm](https://cloud.githubusercontent.com/assets/10407969/16318614/88988d6c-3955-11e6-823a-7a8688b14119.png)

Issue #1673, opened 2016-06-23
Left clicking an ICE outside of a run shouldn't automatically rez it


Kiv:
It's way too easy to accidentally click on an ICE outside of a run and have it reveal hidden info to the Runner. Maybe when outside of a run, the menu could come up with a single option (rez) so you can confirm with a second click.

queueseven:
I like this idea

PR #1747, opened 2016-07-08
Forces action panel on installed agendas and ice outside runs

Saintis:
Forces the rez panel on ice outside of runs, fixing #1673.

Also forces advance panel on installed agendas. This makes them require as many clicks to advance as advanceable traps, since those have options of rez and advance. Maybe I'm just paranoid but I always worry someone will notice I advance slower on traps and faster on agendas.

As an additional bonus this will midly slow down fast advance, forcing small clot windows 😉


JoelCFC25:
You could solve #1118 really easily in this too.

Saintis:
Now fixes #1118 as well.

Screenshot of everything:
panel improvement

mtgred:
👍

PR #1749, opened 2016-07-08
Never trust the client (client/server security issues)

nealterrell:
Ready to merge.

Two big changes to secure against cheating players. (Not that we've seen this necessarily, but it's still the proper way to do things.)

1. When a player sends a UI command targeting a card in the game, that command contains the data of the targeted card. Currently we take that card data and override the server copy of the card with the data sent by the client. That allows a JavaScript hacker to send a command like `play {:title "Hedge Fund" :cost 0}` to play a Hedge Fund that gains 9 credits and costs 0 to play. Likewise, when starting a game, a player's deck data is sent from the client to the server, and we make no attempt at verifying the deck data is actually valid. Someone can (easily) hack their client deck data so _only their_ cards have modified play costs, agenda values, etc. Most of this would be easy to spot in-game thanks to the extensive log messages, but there are subtle ways to cheat with these flaws.

    I fixed all UI action functions to discard the card data sent from the client, only using it to load the corresponding server-side data for the targeted card. A Hedge Fund with hacked play cost would simply be used to load the actual server-side card, which would have the correct play cost. You can see this in `core-actions.clj`, where any function that takes a `card` parameter from the client immediately re-binds that variable to the server version of the card by loading it from the state with `get-card`. (The client-side card has enough information -- the `:cid` and `:zone` -- to do this reliably.) Thanks to #1572, we can also validate user decks upon game start by replacing the card data sent by the client with the corresponding card data from the server's copy of all cards.

2. Currently the entire game state is sent to each player's UI, and the UI code is responsible for not revealing private information (facedown cards, opponent's hand, both players' decks, etc.). But anyone can read the JavaScript state and/or the network transmissions to read this information illegally, letting them know what's in their opponent's hand/deck, what's in their draw pile, what's in that server, etc. Bad.

    I fix this by sending three different states from the game server to the UI server after each action. The three states contain all the public information available to the Corp, Runner, and spectators, respectively. Public information consists of a player's hand, all their installed cards, and all faceup cards the opponent has installed or discarded. If a card is not public information, then all its data is stripped from the state sent to a player except for the card's `:cid`, `:zone`, and a few other misc keys; these keys are minimally necessary to maintain targeting and other UI functionality. 

#### TO-DO

1. ~~Restore "look at deck" functionality.~~
2. ~~Randomize decks before assigning `:cid`, so a hacker can't surmise the identity of an unknown card based on its `:cid` and the order it appears in their decklist.~~
3. Assign new `:cid` every time a card is moved to a private zone, so a hacker can't record the `:cid` of a card they played and then shuffled back into their deck, to know where that card now is in the deck.

### DIFF
--- a/server.coffee
+++ b/server.coffee
@@ -89,6 +89,26 @@ requester.on 'connect', (fd, ep) ->
 requester.monitor(500, 0)
 requester.connect("tcp://#{clojure_hostname}:1043")
 
+sendGameResponse = (game, response) ->
+  diffs = response.runnerdiff
+
+  for player in game.players
+    socket = io.sockets.connected[player.id]
+    if player.side is "Corp"
+      # The response will either have a diff or a state. we don't actually send both,
+      # whichever is null will not be sent over the socket.
+      lobby.to(player.id).emit("netrunner", {type: response.action,\
+                                                   diff: response.corpdiff, \
+                                                   state: response.corpstate})
+    else if player.side is "Runner"
+      lobby.to(player.id).emit("netrunner", {type: response.action, \
+                                                   diff: response.runnerdiff, \
+                                                   state: response.runnerstate})
+  for spect in game.spectators
+    lobby.to(spect.id).emit("netrunner", {type: response.action,\
+                                                diff: response.spectdiff, \
+                                                state: response.spectstate})
+
 requester.on 'message', (data) ->
   response = JSON.parse(data)
   if response.action is "remove"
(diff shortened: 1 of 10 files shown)


nealterrell:
OK, I am punting on issue 3 until I do a little more research into what it might possibly break. It's a very minor security issue anyway. This is ready to merge.

mtgred:
👍

zaroth:
Awesome work, Neal!

PR #1750, opened 2016-07-08
Swap agenda function

Saintis:
Extracts `swap-agendas` from Turntable and Exchange of Information.

Also removes the weird unused functions in core-misc.


mtgred:
👍

PR #1754, opened 2016-07-10
Invasion of Privacy rework

### LINKED ISSUE #1634: Invasion of Privacy should be reworked
Log all cards in the Grip, like Salem's. Then give the Corp a menu to select the cards to trash (only events and resources). 

Currently the entire Grip is moved to the temporary zone, which kinda sucks.

kevkcc:
Fixes #1634 

Also some minor edits to Snoop and Salem's Hospitality log messages so that all cards which reveal the grip have similar messages.


mtgred:
👍

PR #1758, opened 2016-07-13
NAPD MWL update, 2016 July -- Merge only when ready to enforce new MWL rules

nealterrell:
Remove: AstroScript Pilot Program

Add: D4v1d, Faust, Wyldside; Breaking News, Mumba Temple.


mtgred:
👍

PR #1759, opened 2016-07-13
Fix #1757, Bazaar triggering from non-grip installs

### LINKED ISSUE #1757: Bazaar triggers on installs other than from grip
Bazaar triggers on installs from Street Peddler. 

I have not tested to see if it triggers on install from Personal Workshop or The Supplier, but it should not on either, or any other effects that install from hosting card.

Similarly, with Eureka! or Artist Colony to install from stack.

I am not aware of install-hardware-from-heap abilities, but if they exist, Bazaar should ignore them also.

nealterrell:


mtgred:
👍

PR #1762, opened 2016-07-13
Visible hands for spectators

nealterrell:
Requires #1749.

Adds a "Make players' hands visible to spectators" option when creating a game. When checked, shows the Grip and HQ to all spectators of a game. The option can only be selected if spectators are enabled, and will show a warning message recommending the use of a password.


mtgred:
👍

PR #1763, opened 2016-07-14
Fix double-prompts for Patron. Fix #1744.

### LINKED ISSUE #1744: Prompt for choosing Patron server appears twice
I had Jak Sinclair and Patron installed, so when I start my turn the prompt to do actions before taking clicks appears. I click first on Patron, and then on Jak. The problem is that, after taking my clicks, the prompt for Patron is shown again. Apart from that, both cards work normally.

nealterrell:


mtgred:
👍

PR #1775, opened 2016-07-15
Fix Mumbad Virtual Tour and Turning Wheel message

JoelCFC25:
The forced trash effect of MVT is wrongly persisting when the Runner hits an installed, unrezzed copy and the forced trash occurs. 

Turning Wheel message now references the Runner if the Corp fires an ETR subroutine.


mtgred:
👍

PR #1776, opened 2016-07-15
Fixes: Rebirth, invisible Parasite/Pawn/programs hosted on ice, dragging from :play-area. Rework Indexing, Making an Entrance.

### LINKED ISSUE #1127: Successful Indexing run makes R&D non-operable
After a successful Indexing run, I get the option to resolve the Run Effect. I click it and see the top five cards of R&D. However, now I cannot drag them back into R&D. I can drag them into Archives. While that is happening, the Corp cannot drag them to R&D either. After I move all five cards to Archives and fully resolve Indexing, *sometimes* the Corp can drag the cards to R&D, sometimes he or she cannot.

I played a bunch of Indexing lately, sometimes this happens and sometimes not. I could not identify a clear pattern. I don't think it ever happens with the first Indexing of the game, sometimes it happens with second or third one (sometimes it does not).

Thank you!

### LINKED ISSUE #1413: When Indexing cards are sometimes not able to be dragged back to R&D from the temp zone.
Indexing normally works fine but on occasion I am unable to drag cards from the temp zone back to R&D and the cards are trapped in the temp zone. This appears to happen when the corp is Jinteki and the only other commonality I can remember is that there was a caprice in the top 5 cards every time. The browser used was Firefox 45.0.1 and the OS Windows 7.

### LINKED ISSUE #1767: Pawn is broken
I install Pawn on an ICE and I can't see it.  The opponent said they could see it, but i couldn't, so couldn't click on it to perform actions.    
Tried with both Chrome and Firefox.  Problem did not happen at 3:30pm Pacific DT.  But does now.

Also, maybe related, when opponent played Oversight AI, the OAI on the card appeard facedown to me (but not to my opponent)

### LINKED ISSUE #1769: Rebirth not working (tested on Andromeda, Val)
Rebirth is not doing anything after selecting the runner to rebirth to.

Tested rebirthing from: Andromeda, Val
Tested rebirthing into: Leela, Gabe, Noise

Using Chrome / Windows 10.

nealterrell:
Fix #1769, as a consequence of #1749. 

Fix #1767 by making your cards that are hosted on opponent's cards public to you. Also fixes Parasite.

Fix all outstanding issues with Indexing by reworking the card to use a recursive prompt where you select the cards to place on R&D. Includes a final "This is what R&D will look like, are you sure?" prompt. (Fix #1413, Fix #1127.)

image

Also applied this logic to Making an Entrance.

Also includes a 50-80% speedup in `lein test test.all` runtime by not needing to hit the database for each card's data in each test.

### DIFF
--- a/src/clj/game/cards-events.clj
+++ b/src/clj/game/cards-events.clj
@@ -749,7 +749,7 @@
                                              (not (= "Draft" (:setname c)))
                                              (not (= (:title c) (-> @state :runner :identity :title)))))
                         swappable-ids (filter is-swappable (vals @all-cards))]
-                        (cancellable swappable-ids :sorted)))
+                    (cancellable swappable-ids :sorted)))
 
      :effect (req
                (move state side (last (:discard runner)) :rfg)
--- a/src/clj/game/cards-events.clj
+++ b/src/clj/game/cards-events.clj
@@ -426,12 +426,37 @@
       :msg (msg "trash " (count targets) " card" (when (not= 1(count targets)) "s") " and draw " (cards-to-draw targets) " cards")})
 
    "Indexing"
-   {:effect (effect (run :rd {:replace-access
-                              {:msg "rearrange the top 5 cards of R&D"
-                               :effect (req (prompt! state side card
-                                                     (str "Drag cards from the Temporary Zone back onto R&D") ["OK"] {})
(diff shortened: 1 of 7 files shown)


mtgred:
👍

PR #1786, opened 2016-07-17
Band-aid fix for #1782.

### LINKED ISSUE #1782: Typing "null" into chat crashes the game
I'm not even sure if this is a real bug or a weird coincidence, but over multiple trials typing the word "null" into chat would shortly afterwards crash the game for everyone involved, including spectators.

nealterrell:
Fix #1782 by replacing any message that is just "null" (the only message that seems to be crashing things) with " null", which renders identically but does not seems to trip up the JavaScript UI. My notes in #1782 might be helpful for anyone who seeks to understand exactly why this happening, which I have not determined.


mtgred:
👍

Issue #1784, opened 2016-07-17
Canceled a self-modifying code


Montoli:
Please describe the circumstances of your error here.

Clicked "self modifying code" to trigger it.  Decided, upon reflection that I didn't really want to use it now.  Hit cancel.  Got this error.

Stack Trace:
```clojure
#error {
 :cause nil
 :via
 [{:type java.lang.NullPointerException
   :message nil
   :at nil}]
 :trace
 []}
```

nealterrell:
@mtgred are you still starting the game server with the special JVM option?

nealterrell:
Looks like the code to look for "Cancel" as a prompt option is broken, so SMC gets invoked with "Cancel" as the targeted card, which... yeah.

nealterrell:
I guess that would happen when I accidentally turn "Cancel" into "Can cel". :rage:

PR #1787, opened 2016-07-17
"Can cel" should be "Cancel".

### LINKED ISSUE #1785: Cancelling SMC causes a crash and will not let you cancel SMC.
This was when there was only one program left in the stack. Every time "Cancel" is hit, another 2c is deducted for using SMC.

Stack Trace:
```clojure
#error {
 :cause nil
 :via
 [{:type java.lang.NullPointerException
   :message nil
   :at nil}]
 :trace
 []}
```

nealterrell:
Looks like a typo introduced in #1776. Whoops. Fix #1785, #1784.


mtgred:
👍

PR #1789, opened 2016-07-17
Have Null's ability track the specific ice it is used on

### LINKED ISSUE #1788: Null strength penalty added to ice incorrectly when previous ice was destroyed
This seems to occur during a run when a previously Parasited ice is destroyed via being at 2 or less strength when Null's ability is triggered. The next ice on the server is also at -2 (in this case it was a NEXT Silver, which had its strength display as -1).

nealterrell:
Fix #1788.


mtgred:
👍

PR #1804, opened 2016-07-22
Implement Full Immersion RecStudio; improvements to corp hosting overall

### LINKED ISSUE #1644: Worlds Plaza does not place credits on Adonis
When using Worlds Plaza's click to install and rez ability, no credits are placed on Adonis. It just sits there with no credits on it.

### LINKED ISSUE #1752: Full Immersion Recstudio
the card allows you to host up to 2 agendas/ assets.  trash cost increases by 3 for each hosted card.  This ability is not active in Jinteki

thanks

### LINKED ISSUE #1799: Corp cards hosted on other cards appearing face down
So far I've seen this happen with both Casting Call and Oversight AI. I was spectating both times so this might be a spectator-only issue.

nealterrell:
Implements Full Immersion RecStudio. Gives FIR a "Click: Install an asset or agenda" ability to target a card in hand. Also allows FIR to be the target of an install triggered by another ability, like Interns or Lateral Growth, by expanding the "no server specified" path of `corp-install` to include rezzed cards with positive `:can-host` functions. (See FIR implementation.) When showing a list of servers the corp can install to, this menu will include Full Immersion RecStudio if there are not already 2 cards hosted. 

FIR does not have the equivalent of Dinosaurus' "move an installed card onto Dinosaurus" ability. That ability is for interactions with Clone Chip etc. We don't want that for Corp cards, because the equivalent operation (installing to a new remove server, then "moving" onto FIR) would trigger Turtlebacks and Astrolabe from the new server creation. 

In the future I would like to apply the same treatment to runner cards, so that if Dinosaurus is in play and you use Clone Chip, the target program won't go directly into your rig, instead you will get a menu to choose to install onto Dinosaurus or onto the rig. 

Fixes a few other issues involving hosted corp cards. Fix #1644, Fix #1799, Fix #1752.

NOTE: the UI isn't acceptable IMO, but I don't know how to fix it. Paging @mtgred and @domtancredi:

image

There are two cards hosted on this FIR, but the offset is maybe 10 pixels between the two?

### DIFF
--- a/src/clj/game/cards-assets.clj
+++ b/src/clj/game/cards-assets.clj
@@ -350,6 +350,20 @@
                       :msg "add it to their score area and gain 1 agenda point"
                       :effect (effect (as-agenda :corp card 1))}}}
 
+   "Full Immersion RecStudio"
+   {:can-host (req (and (or (is-type? target "Asset") (is-type? target "Agenda"))
+                        (> 2 (count (:hosted card)))))
+    :trash-cost-bonus (req (* 3 (count (:hosted card))))
+    :abilities [{:label "Install an asset or agenda on Full Immersion RecStudio"
+                 :req (req (< (count (:hosted card)) 2))
+                 :cost [:click 1]
+                 :prompt "Choose an asset or agenda to install"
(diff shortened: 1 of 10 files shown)


mtgred:
👍

PR #1807, opened 2016-07-22
Fixes numbers on Corp servers (the zeroes).

### LINKED ISSUE #1778: [COSMETIC] (0) displayed in upper left corner of Archives when empty; issue with HQ card count
When a new game starts the corp's archives shows the new face-up/face-down card counts under the Archives tag, but it also shows the old (0) amount behind the tag. The count goes away as soon as there are any cards in archives. I've yet to see if the count comes back if Archives is emptied out again by recursion effects.

In the same game the HQ tag was missing and (0) displayed in the upper left of the identity card even though the corp player had cards in hand.

Saintis:
Also adds license to bower and npm package files.
Also updates bower and coffee script version.

Fixes #1778.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,7 @@
 {
   "name": "netrunner",
   "version": "0.0.1",
+  "license": "MIT",
   "dependencies": {
     "jquery": "~2.0.3",
     "jqueryui": "~1.11.4",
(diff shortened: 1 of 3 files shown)


Saintis:
If updating bower / coffeescript seems unnecessary it is easy to revert that change.

mtgred:
👍

PR #1808, opened 2016-07-22
Updates packages.

Saintis:
Migrates from Express 3.x to 4.x (lots of work).

Currently complains that there are no specified option `extended` for `body-parser`, or `resave`, `saveUninitialized` for `express-session`.

Not tested that server will run in `"production"` mode.


zaroth:
I think we should use a proper public test environment for checking this out, will update later.

erbridge:
Oh sorry, I didn't realize you'd already done this @Saintis.

PR #1817, opened 2016-07-25
Refactor recursive reordering prompts + extension to other cards

kevkcc:
Refactors the new recursive prompt code that are used for Indexing, Making an Entrance, etc. I like them a lot over dragging from play area so I aimed to make it more consistent all around.

Changes:
1. Refactor the recursive prompt code into 2 functions in cards.clj
2. Extends the system to a lot of cards that still use drag from play area (Rolodex, Data Hound, Shiro, Spy Camera). Also reworked CBI Raid along the same lines but it is slightly different so doesn't use the refactored functions.
3. Fix bugs for the already reworked cards (Indexing, etc.) where rearranging 0 cards (empty R&D or stack) will freeze the game.
4. Tests for all of the above 

also...
5. Fix that long standing pending parasite/builder test
6. An addition to my previous PR for jack-out prevention, now prevents jacking out on 'An offer you can't refuse' runs

### DIFF
--- a/src/clj/game/cards.clj
+++ b/src/clj/game/cards.clj
@@ -27,6 +27,43 @@
   triggers"
   {:effect (req (toast state :corp "Reminder: You have unrezzed cards with \"when turn begins\" abilities." "info"))})
 
+(declare reorder-final) ;forward reference since reorder-choice and reorder-final are mutually recursive
+(defn reorder-choice
+  "Generates a recursive prompt structure for cards that do reordering (Indexing, Making an Entrance, etc.)
+
+  reorder_side is the side to be reordered, i.e. :corp for Indexing and Precognition.
+  wait_side is the side that has a wait prompt while ordering is in progress, i.e. :corp for Indexing and Spy Camera.
+
+  This is part 1 - the player keeps choosing cards until there are no more available choices. A wait prompt should
+  exist before calling this function. See Indexing and Making an Entrance for examples on how to call this function."
+
+  [reorder_side wait_side remaining chosen n original]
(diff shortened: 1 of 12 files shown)


src/clj/game/cards.clj
```diff
@@ -27,6 +27,43 @@
...
+  "Generates a recursive prompt structure for cards that do reordering (Indexing, Making an Entrance, etc.)
+  This is part 2 - the player is asked for final confirmation of the reorder and is provided an opportunity to start over."
+
+  [reorder_side wait_side chosen original]
+  {:prompt (str "The top cards of " (if (= reorder_side :corp) "R&D" "your Stack")
+                " will be " (clojure.string/join  ", " (map :title chosen)) ".")
+   :choices ["Done" "Start over"]
+   :delayed-completion true
+   :effect (req (if (= target "Done")
+                  (do (swap! state update-in [(if (= reorder_side :corp) :corp :runner) :deck]
```
nealterrell: Instead of `(if (= reorder_side :corp) :corp :runner)`, can you do `reorder_side`? You do this in a few places.

nealterrell:
Really nice job. Is CBI Raid different because it's not rearranging things that are already on R&D?

Saintis:
This is great!

kevkcc:
Thanks for the feedback, made the change. Don't know why I did it like that in the first place lol

And yes, CBI Raid moves cards from hand onto R&D instead of a reorder so while the recursive prompt works, the operation is different. Didn't seem to make sense to me to do a more generic refactor for this one case. Also I have no idea why circleci is failing now.

mtgred:
👍

PR #1820, opened 2016-07-25
FAQ 3.1 search changes; fix for facedown hosted cards (again)

nealterrell:
Updates all cards that search R&D/Stack so that they shuffle immediately after a card is chosen, before the effect is executed. I was thorough and changed cards that don't really need to be changed (anything that brings the target into the player's hand, like Project Atlas), just in case there is some future card that messes with the draw pile when a card is tutored. This has a functional effect on Hostage (into Street Peddler) and Heritage Committee (via Mumbad City Hall).

Along the way, I noticed that several runner cards with tutor effects were not actually shuffling afterwards. Whoops.

Also fixes (I hope) issues with runner cards being drawn face-down when hosted on other cards. For future reference, there are two reasons why the UI will draw a card face down:

1. The card is missing a `:code` field. `:code` identifies the card and allows its image to be drawn. We don't want this information to be known if the card is private/unknown to the player, so this field is stripped from private states. Any card without a code, regardless of anything else, will be drawn face down.
2. The function `face-down?` in `gameboard.cljs` returns true. This function will return true if:

    1. The card has a `:facedown` key, which is used by Street Peddler and Apex effects.
    2. The card is a corp card that is `(not rezzed)`.
    3. Unless that corp card is an Operation (Oversight AI, Patch). 

The same rules are then applied to any hosted cards. In the past, case 1 was broken, and the private states were stripping out codes for too many cards. This time case 2 was broken, thanks to the upgrades for Full Immersion RecStudio in #1804.

### DIFF
--- a/src/clj/game/cards-agendas.clj
+++ b/src/clj/game/cards-agendas.clj
@@ -5,15 +5,15 @@
 (def cards-agendas
 
   {"15 Minutes"
-     {:abilities [{:cost [:click 1] :msg "shuffle 15 Minutes into R&D"
-                   :label "Shuffle 15 Minutes into R&D"
-                   :effect (req (let [corp-agendas (get-in corp [:scored])
-                                      agenda-owner (if (some #(= (:cid %) (:cid card)) corp-agendas) :corp :runner)]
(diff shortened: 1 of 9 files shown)


mtgred:
:+1:

PR #1821, opened 2016-07-26
Fixes: Controlling the Message, Temujin Contract, C.I. Fund

### LINKED ISSUE #1796: NBN: Controlling the Message requires /close-prompt command
Intermittently (meaning: sorry, I can't give more details about how to reproduce), I've seen the runner need to use /close-prompt to get out of a run after Controlling the Message's trace fires.

### LINKED ISSUE #1819: Temüjin Contract locked to a server even after changing zone
I realise the card has only just been automated, but wanted to point this out!

I was playing a game and hastily clicked the wrong server for my Temujin on install; I clicked RnD when I was intending to put it on Archives. Moving it back to my hand didn't allow me to replay it, and this didn't work via my stack or heap either. The other copy of Temujin in my heap had the same issue; it was locked to archives, but I couldn't play it.

Thank you!

JoelCFC25:
* Fix #1819: Change to `dissoc-card` so it will drop `:server-target` from deactivated cards; update Diwan to use this instead of `:named-target`
* Fix #1796: Add an unsuccessful effect to CtM trace so Runner's wait prompt will be cleared. 
* Prevent a Corp turn start bug when C.I. Fund has no credits on it.

### DIFF
--- a/src/clj/game/cards-assets.clj
+++ b/src/clj/game/cards-assets.clj
@@ -153,7 +153,7 @@
                  :msg (msg "trash it and gain " (get-in card [:counter :credit] 0) " [Credits]")
                  :effect (effect (gain :credit (get-in card [:counter :credit] 0))
                                  (trash card {:cause :ability-cost}))}]
-    :events {:corp-turn-begins {:req (req (>= (get-in card [:counter :credit]) 6))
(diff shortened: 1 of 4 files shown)


mtgred:
👍

PR #1822, opened 2016-07-26
correct ICE removal in front of Caprice Nisei

JoelCFC25:
This PR removes the obsolete `trash-ice-in-run` function, which was only ever a workaround. As I was alerted on Stimhack Slack, the current behavior of ICE destruction when it occurs immediately in front of a rezzed Caprice Nisei is wrong. Both instant destruction via Parasite and self-trashing ICE (Special Offer, Data Mine, etc) are using `trash-ice-in-run` to update the run position, but this wrongly sends the Runner directly to the successful run prompt, skipping the place where Caprice should fire. 

Now we just call `no-action` followed by `continue` in the ICE (and simply `continue` in Parasite) to accomplish the same goal (automatically update the run position) without barreling past the Caprice trigger point.

### DIFF
--- a/src/clj/game/cards-ice.clj
+++ b/src/clj/game/cards-ice.clj
@@ -414,7 +414,8 @@
    {:abilities [{:msg "do 1 net damage"
                  :effect (req (damage state :runner eid :net 1 {:card card})
                               (when current-ice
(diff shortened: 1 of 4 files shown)


mtgred:
👍

PR #1826, opened 2016-07-28
Prevent corp from rezzing second copy of a unique card

### LINKED ISSUE #1814: [Feature Request] Rezzing ICE does not check for its uniqueness
That awkward moment when your opponent tells you that you already have one instance of that ICE rezzed and that ICE is **unique** 😅 

I was recently playing a game where I had an [Excalibur](https://netrunnerdb.com/en/card/06111) rezzed and tried to rez another one.

It would be a nice feature to have to either:
- Preventing from rezzing the ICE at all and displaying a warning
- Prompting a confirmation to rez the ICE knowing that with the other one rezzed will put the board into an invalid state

(I'm not quite sure if this issue is limited to ICE only, so if the same happens for assets/upgrades would be good to apply the same solution for all of it)

nealterrell:
Fix #1814.


mtgred:
👍

lostie:
@nealterrell great thanks! Awesome work guys 👏  🎩

PR #1828, opened 2016-07-29
Fixes for corp hosting

### LINKED ISSUE #1827: Full Immersion Rec Studio reveals hosted cards to runner in log
When hosting a card on Rec Studio it is being revealed to the runner in the logs. I.e. x spends click to use Full Immersion RecStudio to install and host **Corporate Sales Team**. This information shouldn't be getting revealed to the runner - the cards are installed face down and should behave like any other asset until accessed or rezzed.

nealterrell:
Notably with Full Immersion RecStudio. Fix #1827. Also has a minor fix for Sandburg (immediately update ice strength when rezzed), and more tests for corp hosting (Patch, Sandburg on FIR).


mtgred:
👍

PR #1829, opened 2016-07-30
Fix Fan Site and Full Immersion RecStudio

JoelCFC25:
Requires that Fan Site be installed in order to handle the `:agenda-scored` event, and stops revealing the names of cards hosted on Full Immersion RecStudio.


JoelCFC25:
Superseded by #1832.

PR #1833, opened 2016-07-30
Implements 3 Blood Money cards

Saintis:
Implements Prisec and Hatchet Job. Also soft-implements Builder of Nations (damage just as an ability).

Also makes `qty` (in the test ns) throw an Exception if the wanted card is missing from `@all-cards`.

Adds test for Prisec.

### DIFF
--- a/src/clj/game/cards-operations.clj
+++ b/src/clj/game/cards-operations.clj
@@ -305,19 +305,19 @@
    {:msg "gain 3 [Credits] and draw 1 card"
     :effect (effect (gain :credit 3) (draw))}
 
-   "Hatchet Job"
-   {:trace {:base 5
-            :choices {:req #(and (installed? %)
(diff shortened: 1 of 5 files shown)


src/clj/game/cards-operations.clj
```diff
@@ -312,6 +312,13 @@
             :label "Give the Runner 4 tags"
             :effect (effect (tag-runner :runner 4))}}
 
+   "Hatchet Job"
+   {:trace {:base 5
+            :choices {:req #(and (installed? %)
+                                 (not (has-subtype? % "Virtual")))}
+            :msg "add an installed non-virtual card to the Runner's grip"
+            :effect (effect (move :runner target :hand true))}}
```
Saintis: Not sure why there is `true` here at the end, but that's how Archangle does it...

mtgred:
👍

PR #1835, opened 2016-08-01
Fix the saving of game stats.

nealterrell:
My bad. Game server needs to send the string `remove` with the last game state back to the Node server for stats to be saved. We send the "old state" back from the game server, because applying the "remove" command removes the game state from the server (since the game was closed).


mtgred:
👍

PR #1836, opened 2016-08-02
More Blood Money cards

JoelCFC25:
Fixes the event listener on Jinteki: Potential Unleashed so no mill occurs if damage is prevented. 

Implements Algo Trading, Golden, Nfr, Special Report, Enforcing Loyalty, Aiki, and Fairchild 2.0 from the upcoming pack.


mtgred:
👍

Issue #1721, opened 2016-07-03
Sometime, the run ends before a card is trashed by the runner


mtgred:
It results in bad pubs being cleared before it can be used to pay for the trash cost.

nealterrell:
I'm trying to look into this, but basic tests are passing. Remember any other details? I know you were playing Val, but where were you trashing from, what were you trashing, etc?

artturipi:
I think it happens at least when running on a remote containing only a CVS and nothing else. I'm pretty sure it happens on centrals too if the last card accessed is CVS. Probably also other cards where the Corp can trigger something, like Snare! (+IG to make the BP matter)

nealterrell:
You nailed it, @artturipi.

PR #1846, opened 2016-08-05
New cards, some fixes

### LINKED ISSUE #1849: Doppelganger triggering after accessing upgrade in central servers
Happened in my game:
1. Corp had unrezzed upgrade in HQ.
2. I had Doppelganger installed and successfully ran HQ.
3. I chose to access the unrezzed upgrade first. Doppelganger then immediately prompts me to make another run.
4. As I have yet to access HQ (Corp's hand), this somehow breaks the run into 2.

Good luck!

nealterrell:
Impelement:

1. Georgia Emelyov
2. Another Day, Another Paycheck
3. Paperclip. When in the heap, shows a prompt upon starting a run, passing an ice, or rezzing an ice, provided the current-ice in the run is a barrier.
4. Rumor Mill. Requires several changes so that cards marked as `:disabled true` do not have most of their effects triggered.
5. Data Breach. Requires extending when-completed to the "run-ends" events.

Fixes:

1. Punitive Counterstrike FAQ change
2. Subcontract/AccDiag interaction with Terminal operations
3. Fix #1849, #1721. If a corp card that shows a prompt to the corp (CVS, Shi.Kyu) is accessed, then `handle-run-end` will end the run, because a run "ends" when the runner has no more prompts showing, and the prompt to trash the accessed card doesn't show until the corp has dealt with their option. This causes "when a run ends" effects to fire prematurely (Doppelganger in #1849), and BP credits to be lost before the runner gets to trash the accessed card (#1721).

    This requires a more thorough rework of the run-ends code, but for now, a very simple hack is to only end a run if BOTH players have no prompts showing. 

I need to take some time off to prepare for my upcoming semester, so hopefully this will cover us for a while.


mtgred:
👍

Issue #1845, opened 2016-08-03
Vulcan Coverup not triggerable by 24/7 News Cycle


TheRealC:
Hiya there.

I was playing a game some hours ago as Titan Transnational against a Valencia-Rebirthed-into-Quetzal. I had two Vulcan Coverups in my score area; one of them I had scored myself (it had the Titan Transnational agenda counter on it), one I had been sent by a Turntable. I went for the obvious play, which was, clearly:

Power Shutdown my R&D into Archives -> Jackson Howard three 24/7 News Cycles into R&D -> Accelerated Diagnostics to play my three 24/7s.

Now the expected behaviour would be that I could re-trigger Vulcan Coverup's "When scored, do 2 meat damage." However, I could not target neither of the Coverups ("my own" or the Turntabled one) with the 24/7 News Cycles; clicking on them, even after very explicitly choosing to use each 24/7, in turn and order, did nothing (I'm 99.9% sure I was using Accelerated Diagnostics right).

Possible causes I could think of: Turntable screwed up my agenda zone (still weird that the "real" Coverup didn't work), 24/7 got confused since its additional cost of forfeiting an agenda suddenly no longer applied, Rebirth somehow broke it, the Agenda Counter from Titan Transnational somehow broke it.

Since it's the only real win condition of that overly janky Titan deck, it would be nice to see this fixed.

I have not been able to confirm whether you are normally able to 24/7 a Vulcan; maybe that just doesn't work at all for some reason.

As an aside, we used console commands to resolve the damage, and I hit his only I've Had Worse on the second 24/7. It was worth the try.

Cheers, and thanks for all your hard work,

The Real C

Saintis:
Had a look at the code. This is due to an incorrect check for 24/7 (I think). 
The `:choices` req is currently:
```clojure
:choices {:req #(and (is-type? % "Agenda")
                                (= (first (:zone %)) :scored)
                                (:abilities %))}
```

But the relevant attribute for 24/7 is `:effect` and not `:abilities`. This should be an easy fix.

TheRealC:
Awesome, thanks!

Saintis:
Actually might not be that easy... related to #1555.

The `:abilities` bit seems to be intended to ensure only agendas on the corp side are targetable. Adding an explicit side check: that the card is in `[:corp :scored]` and not `[:runner :scored]` would replace the use of `:abilities`

kevkcc:
I'm working on this as part of my agenda PR

Issue #1840, opened 2016-08-02
Personality Profiles with Empty Grip


benhosp:
Personality Profiles doesn't let "search your stack" or "install from your heap" effects resolve if the Runner has no cards in their grip.

kevkcc:
I got this one as well

Issue #1687, opened 2016-06-27
Exchange of Information on Mandatory Ugrades does not automatically add a click every turn


promisedlandt:
If the corp scores Mandatory Upgrades, the click gets added automatically every turn.
Given a Mandatory Upgrades in the runner's score area, when the corp plays Exchange of Information on the Mandatory Upgrades, it gets added to their score area, but does not automatically add the click every turn.

Could be similar to #1555 / #1125, just with Exchange of Information instead of Turntable.

JoelCFC25:
Thanks for the report--unfortunately this one is expected. Those are indeed the relevant issues, since EoI just reused a large chunk of Turntable's code.

Issue #1555, opened 2016-05-26
Stolen/Turntabled Chronos Project Not Triggering with 24/7


se4n:
This is certainly an edge case but I'll describe what happened in a game today -- I, playing PE, had a naked Chronos Project on the table, with a Profiteering and a full House of Knives scored. The runner had Turntable installed, ran on the Chronos Project, and instead of stealing, swapped it with my full House of Knives. Now I had a Chronos Project and a Profiteering in my score area, and they had the full House of Knives in theirs. My turn, I play 24/7 News Cycle -- sacrificing the Profiteering, trying to trigger the Chronos Project, but it won't work.

My assumption is that since the Chronos Project was never actually *scored* (just stolen and then moved to my score area) that perhaps 24/7 needs fixing? That is, it seems to be trying to retrigger an existing thing that happens when a card is scored -- but in this case, because of Turntable (and soon with Exchange of Information), there will be cards with "when scored" abilities in the Corp's score area that were never actually scored per se.

JoelCFC25:
I noted this at #1125, which I closed at the time because I have no idea how to handle this case.

se4n:
Ah, thanks, sorry for the duplicate report. I didn't see #1125.

JoelCFC25:
No, it's no problem--it was closed after all!

PR #1847, opened 2016-08-05
Agenda overhaul

### LINKED ISSUE #1783: 15 Minutes + Exchange of Information
After I used Exchange of Information to swap a scored 15 Minutes for an agenda in the Runner's score area, I wasn't able to click and use the 15 Minutes ability.

### LINKED ISSUE #1837: Project Ares Bug
Scoring project ares with only 4 points still gives you bad pub, even if the runner trashes nothing.

kevkcc:
This PR fixes most of the lingering issues with swapping agendas, Mandatory Upgrades, 15 minutes, 24/7, etc. 

The way I saw it was that agendas (in general) had effects 'when scored', 'when stolen' and just passive effects (Mandatory Upgrades, Rebranding Team) that are always in play. The agenda code seemed alright so I didn't tinker much, just added a :swapped key. Passive effects were previously grouped into the same category as 'when scored' but they can now work properly if you call the effect in both the scored (:effect) and swapped (:swapped) scenarios. This results in some code duplication and might not be the perfect solution so could be looked at more in the future.

Also fixes issues with 24/7 news cycle and 15 minutes with swapped agendas.

Fixes #1783, #1687, #1555, #1845 
Also fixes #1837 (project ares), #1840 (personality profiles), along with a bunch of other bugs that I discovered while going through all the agendas in the game that no one plays.

Also updated old tests and added a bunch more!


mtgred:
👍

PR #1851, opened 2016-08-06
Display text or images in card browser

elmarco:
It is convenient to be able to copy & paste or search in the text of cards.

The second patch is just cleanup (unless it's useful for debugging?).

(I have no prior experience with om/clojure, so this might not be the best way to do it, I welcome help or ideas)

### DIFF
--- a/server.coffee
+++ b/server.coffee
@@ -90,10 +90,8 @@ requester.monitor(500, 0)
 requester.connect("tcp://#{clojure_hostname}:1043")
 
 sendGameResponse = (game, response) ->
-  diffs = response.runnerdiff
 
   for player in game.players
-    socket = io.sockets.connected[player.id]
     if player.side is "Corp"
       # The response will either have a diff or a state. we don't actually send both,
       # whichever is null will not be sent over the socket.
--- a/src/cljs/netrunner/cardbrowser.cljs
+++ b/src/cljs/netrunner/cardbrowser.cljs
@@ -34,40 +34,43 @@
       (make-span "\\[Link\\]" "link")
       (make-span "\\[Trash\\]" "trash")))
 
-(defn card-view [card owner]
-  (om/component
-   (sab/html
-    [:div.card-preview.blue-shade
-     [:h4 (:title card)]
-     (when-let [memory (:memoryunits card)]
-       (if (< memory 3)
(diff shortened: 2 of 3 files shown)


elmarco:
Screenshot:
text

nealterrell:
Did you test this change on a live game? `card-view` is used in `gameboard.cljs` for displaying the card zoom view, and is also used in `deckbuilder.cljs`. By adding a third parameter to the method I am concerned you may have broken those.

(Props for the use of `partial`, though.)

elmarco:
You are right, I broke `deckbuilder`, however, it seems `gameboard` uses its own `card-view`.
Fixing it, by using a partial. I am not particularly proud of using it there, I was looking for a way to give the `:appearance` state to the component. I don't think it's the "om" way what I did.

nealterrell:
Look for `cb/card-view` at the end of gameboard.

elmarco:
Ok, will fix it.

Btw, not sure where to ask this, but since you are quite familiar with the project, let me ask. clojurescript+react seems quite awesome, and I am glad to learn it. However, since I am considering writing a game engine (for conquest), would you say that clojure was/is a good choice? Reading the code, I am not quite convinced of using functional programming is really a win. Any opinions?

nealterrell:
Clojure(Script) wouldn't have been my first choice if I was founding a project like this; I probably would have thought to make a more traditional backend in ASP.NET (my only real web backend experience) with a JS user interface. I do like that the game engine and UI can be written in the same language with Clojure(Script), and have come to greatly enjoy working with Clojure, but there are other languages that can also do this (JavaScript via Node, Scala(.js)). I think functional programming is a style that every skilled programmer can and should incorporate into most software projects, but I don't always agree with pure-functional approaches to complicated software like games.

zaroth:
It may be a good idea if you want to attract to project bored programmers thirsty for some functional programming distraction, though ;-)

elmarco:
Hi Neal, unfortunately, I am quite busy with other projects these days. Feel free to pick up where I left off.

Saintis:
@jwarwick is this PR still relevant after #2727 has been merged?

jwarwick:
@Saintis probably safe to close it. I think #2727 added the same functionality in a slightly different way.

PR #1854, opened 2016-08-09
Removes action and ability panels from disabled cards.

### LINKED ISSUE #1715: Error when clicking on an agenda hosted on film critic
As the runner when you click on the agenda hosted by Film Critic an error report popup happens, but the game keeps running fine. It's really easy to accidentally click on the agenda instead of film critic itself.

Saintis:
Also disables agendas hosted on Film Critic in order to fix #1715. Also adds default advancement count to the `score` function to prevent error when trying to score an unadvanced agenda.

Might require rewrite of Rumor Mill depending on how that was implemented (could incorrectly prevent disabled cards form being rezzed, pagind @nealterrell).

### DIFF
--- a/src/clj/game/cards-resources.clj
+++ b/src/clj/game/cards-resources.clj
@@ -381,7 +381,9 @@
                                 (is-type? (:card (first (get-in @state [side :prompt]))) "Agenda")))
                  :label "Host an agenda being accessed"
                  :effect (req (when-let [agenda (:card (first (get-in @state [side :prompt])))]
-                                (host state side card (move state side agenda :play-area))
+                                (let [agenda (move state side agenda :play-area)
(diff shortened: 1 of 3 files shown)


Saintis:
Ah, seems there is a test that relies on illegal score attempts throwing errors. I can either revert the small fix, or change the test. Any preferences?

nealterrell:
Remove the test.

I'd also prefer to not break the Rumor Mill functionality with this extension. `:disabled` is used to indicate cards that are "blanked" by Rumor Mill, Cerebral Static, etc. I agree it's not a great adjective, and perhaps should be changed to `:textbox-blank` or similar, but I don't think it should be extended to represent another feature (disabled due to hosting).

Saintis:
Good points. Will rethink my approach.

Issue #1856, opened 2016-08-10
24/7 accelerated diagnostics interaction


PowerfulWizard:
24/7 doesn't resolve when playing it through accelerated diagnostics with just one agenda on the score area. If you have 2+ agendas scored it resolves normally (ingoring the aditional cost of forefeiting an agenda, as it should) .

Saintis:
24/7 has a `req` of more than one agenda scored. This should probably be removed or changed to at least one agenda scored. Relevant line [here](https://github.com/mtgred/netrunner/blob/master/src/clj/game/cards-operations.clj#L5).

PR #1858, opened 2016-08-10
24/7 accelerated diagnostics interaction fix

Hoclor:
Simple fix for #1856, allows 24/7 News Cycle to be played with only 1 agenda scored.

### DIFF
--- a/src/clj/game/cards-operations.clj
+++ b/src/clj/game/cards-operations.clj
@@ -2,7 +2,7 @@
 
 (def cards-operations
   {"24/7 News Cycle"
-   {:req (req (> (count (:scored corp)) 0))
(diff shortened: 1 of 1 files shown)


src/clj/game/cards-operations.clj
```diff
@@ -2,7 +2,7 @@
 
 (def cards-operations
   {"24/7 News Cycle"
-   {:req (req (> (count (:scored corp)) 1))
+   {:req (req (> (count (:scored corp)) 0))
```
Saintis: Can otherwise use `pos?` as well. As in `(pos? (count (:scored corp)))`

mtgred:
👍

PR #1859, opened 2016-08-10
Efficiency committee bandaid fix

### LINKED ISSUE #1857: Efficiency Commitee getting stuck without advanced forever
When i was playing with a friend of mine, he scored Efficiency Commitee and during a turn he used to get clicks, but then next turn he was still stuck without advancing, he tried spending all agenda counters, and the next turn he was still stuck without advancing.
Card was in a server with an upgrade, Efficiency Committee was the second Agenda scored.

kevkcc:
Fixes #1857 

Only a bandaid fix

@JoelCFC25 I couldn't easily transition from using the persistent flag system to the turn flag system. can-advance doesn't currently handle the turn-flag system and simply adding a check for a turn-flag in that function didn't seem to work.

Won't have time to develop for next couple weeks but I can have a further look in the future.

### DIFF
--- a/src/clj/game/cards-agendas.clj
+++ b/src/clj/game/cards-agendas.clj
@@ -227,8 +227,7 @@
                                      ((constantly true)
                                        (toast state :corp "Cannot advance cards this turn due to Efficiency Committee." "warning")))))
                  :msg "gain [Click][Click]"}]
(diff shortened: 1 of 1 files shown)


mtgred:
👍

PR #1867, opened 2016-08-14
update to card-list to be scrollable without position absolute

### LINKED ISSUE #1815: Card search box on Cards page/tab appears behind the cards
As the title suggests, the search function appears behind the cards on the card page. I made a temporary fix for myself by accessing **inspect element** on the page, and added a z-index element style to it so it can appear in front, but that is tedious and the box still partially blocks at least one card, so I was wondering if there was some way to fix the issue. This issue was not present until between a few days and a week ago.

Thanks in advance

### LINKED ISSUE #1866: Bug: Chrome: Cards: Filter overlay issue
In the Cards view, the Filter is being overlaid by the cards now in Chrome.

To fix this, we'll remove the "position: absolute" element of it, and instead replace with a max-height (based on current screen dimensions), with overlay: scroll for anything larger.

domtancredi:
Fixes #1866, fixes #1815.

If we want this to be a flexible scroll, we may want to set a height to the entire wrapper.

### DIFF
--- a/src/css/base.styl
+++ b/src/css/base.styl
@@ -735,9 +735,7 @@ nav ul
       border: 1px solid orange
 
 .card-list
-  position: absolute
-  top: 0
-  bottom: 0
+  max-height: 960px
   overflow: auto
   -webkit-overflow-scrolling: touch
   padding-top: 40px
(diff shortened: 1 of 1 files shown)


mtgred:
By fixing the max-heigh, the card list doesn't take the screen on large screens.

A proper fix is to set the left properties: https://github.com/mtgred/netrunner/commit/d98178a1f54a42eaf01daef8c49172f74a6d9d0d

PR #1871, opened 2016-08-16
implement Financial Collapse and Scarcity of Resources

JoelCFC25:


nealterrell:
For Financial Collapse, you'll want to use `continue-ability` instead of `resolve-ability`, and for each effect that calls continue-ability, mark that effect as `:delayed-completion true`. That will fix the interaction with AccDiag / Subcontract.

mtgred:
👍

PR #1872, opened 2016-08-16
fix Salsette Slums + Controlling the Message interaction

JoelCFC25:
This is a relatively [recent ruling](https://twitter.com/ANR_RulesWiki/status/760495404639068160?s=09). 

Salsette Slums fires a "fake" `:runner-trash` (with a target of `nil` to avoid triggering Hostile Infrastructure) in order to "switch off" NBN: Controlling the Message for the rest of the turn.


mtgred:
👍

PR #1875, opened 2016-08-17
Identity fixes: Null, Khan, Chronos Protocol

### LINKED ISSUE #1862: Bio-ethics does not trigger Chronos Protocol if CP was used during the runner's previous turn
`:runner-turn-begins` and `:corp-turn-begins` in Chronos Protocol should be `:runner-phase-12` and `:corp-phase-12` so they clear during the "Start of turn" window instead of the mandatory draw window.

### LINKED ISSUE #1873: Khan icebreaker install discount not applied
The 1-credit discount Khan provides when installing an icebreaker through her ability is not applied, and I can't see anything in the relevant source code that indicates that the discount is implemented.

Khan's card text for reference (emphasis mine):
> The first time you pass a piece of ice each turn, you may install an icebreaker from your hand, **lowering the install cost by 1**.

JoelCFC25:
* Fix #1873: Was missing icebreaker installation discount
* Fix #1862: Move timing of CP's effect activation
* Fix Null: Whistleblower to actually discard the targeted card.

### DIFF
--- a/src/clj/game/cards-identities.clj
+++ b/src/clj/game/cards-identities.clj
@@ -101,8 +101,8 @@
 
    "Chronos Protocol: Selective Mind-mapping"
    {:events
-    {:corp-turn-begins {:effect (effect (enable-corp-damage-choice))}
-     :runner-turn-begins {:effect (effect (enable-corp-damage-choice))}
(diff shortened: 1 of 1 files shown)


mtgred:
👍

PR #1877, opened 2016-08-17
Add a /move-bottom console command

### LINKED ISSUE #1834: Add ability to manually add card from HQ to bottom of R&D
Sensi Actors Union + Daily Business Show currently doesn't work right.

JoelCFC25:
Fixes #1834. 

This is a workaround for the current inability to put what should be eligible cards on the bottom of R&D when using Sensie Actors Union in multi-draw situations.


mtgred:
👍

PR #1883, opened 2016-08-19
fix Salsette Slums and trashable executives; fix Liquidation

### LINKED ISSUE #1886: Pawn + Scheherazade + Deep Red
http://imgur.com/a/NI4Wm

If you, in response to the "Do you want to use Deep Red's ability"-prompt, click on Scheherazade to host the Pawn, nothing will happen, but once you press OK on the prompt you will get Scheherazade's prompt, and then once you've clicked on Pawn for that, then when you click on a piece of ICE for Pawn's ability, you will get one on the ICE and one on Scheherazade.

The pawn created by Scheherazade disappears if it goes into the heap - The heap says that X cards are in it, but you can only see and select X - 1 cards, with the Pawn being hidden. In all other ways the Pawn seems to behave normally.

JoelCFC25:
* Salsette Slums has the same problem Film Critic used to have (fixed at #1277)--using it during a run would result in `:access true` lingering on in the game state, so a Corp player subsequently doing a voluntary trash from hand of one of the 2-point executives would see the card wrongly teleport to the Runner's area. 
* Liquidation is using the wrong calculation to figure credits paid--it's always doing 3x number of rezzed cards, regardless of how many the Corp actually selects.
* Fix #1886: Bumped up the priority of Scheherazade's "host an installed program" ability to slightly ease the pain of the tortuous Pawn+Deep Red+Scheherazade engine, avoiding duplicated Pawns.

### DIFF
--- a/src/clj/game/cards-operations.clj
+++ b/src/clj/game/cards-operations.clj
@@ -460,10 +460,10 @@
                    (continue-ability state side
                      {:prompt "Choose any number of rezzed cards to trash"
                       :choices {:max n :req #(and (rezzed? %) (not (is-type? % "Agenda")))}
-                      :msg (msg "trash " (join ", " (map :title targets)) " and gain " (* n 3) " [Credits]")
+                      :msg (msg "trash " (join ", " (map :title targets)) " and gain " (* (count targets) 3) " [Credits]")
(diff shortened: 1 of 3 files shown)


mtgred:
👍

PR #1885, opened 2016-08-19
Various fixes

### LINKED ISSUE #1865: Doppelganger run ending
I think the fix in #1849 broke Doppelganger. Nothing happens when you pick "yes" on the prompt, so I'm guessing the run is getting ended immediately or just not triggering.

### LINKED ISSUE #1880: "Successful run" prompt available when targeted marketing with Sol
When New Angeles Sol gets an agenda stolen, they can refire targeted marketing. This gives them the ability to choose a card to name. While the card is being named, the runner gets a "successful run" prompt, and if they click they get another access on the server. If runner doesn't click, then the prompt disappears when the corp finishes choosing. Note: if the runner has used something like the turning wheel or medium to access additional cards, then these apply to the second successful run too.

nealterrell:
Fix #1880 by causing Currents to trigger effect-completed events when appropriate.

Fix several run issues when accessing damage-causing cards like Prisec and Shi.Kyu. __Any__ effect that calls `damage` should be marked `:delayed-completion true`.

Fix #1865 by better following the game rules regarding when "successful run ends" event should be triggered. The event should be triggered after the run has closed (BP credits lost, icebreakers reset). This requires slight tweaks to some cards, as `:run` in state will no longer exist when the event is triggered. Instead we send a copy of the `:run` map to the event handlers.

### DIFF
--- a/src/clj/game/cards-assets.clj
+++ b/src/clj/game/cards-assets.clj
@@ -217,7 +217,9 @@
                  :choices {:req #(has-subtype? % "Connection")}
                  :msg (msg "to trash " (:title target)) :effect (effect (trash card) (trash target))}
                 {:cost [:click 1] :req (req (>= (:advance-counter card) 2))
-                 :msg "do 2 meat damage" :effect (effect (trash card) (damage eid :meat 2 {:card card}))}]}
+                 :delayed-completion true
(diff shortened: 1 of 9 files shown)


mtgred:
👍

PR #1891, opened 2016-08-21
More fixes

### LINKED ISSUE #1879: Sandburg appears to apply strength twice if Rumor Mill has been used on it and then lost.
Unsure of the exact steps that caused this, but I'm pretty sure there was a combination of trashing a sandburg, having it brought back with team sponsorship, rumor mill being played, and a hostile takeover being scored to cancel the rumor mill. When the sandburg was ressed again, the ice was given +6 strength despite the corp only having 15 credits.

### LINKED ISSUE #1888: Flare sub fails when no hardware to trash
When flare's sub trace succeeds, if the runner has no hardware, the meat damage and ETR don't fire.

### LINKED ISSUE #1889: Imp Virus counters and purge

### LINKED ISSUE #1890: Museum of history only allowed to rez one at a time, but acted as if two were rezzed
It's possible this was caused my a rule change I missed. If it was I may have found a rare bug
I was under the impression I could have multiple Museum of History cards rezzed at the same time.
After I played a second museum of history card, but before my next turn started I tried to click on the 2nd museum, but no effect or error message occurred. After a quick chat with my opponent and a quicker rule search, we agreed I could test it by de-rezzing the first MOH, getting a free credit, and then rezzing the 2nd. It did allow me to rez the 2nd. It would not allow me to rez the 1st again.
However once I decided to move on, assuming it had been changed to unique and I couldn't find that rule change. I then choose to activate one MOH before the mandatory draw step, and it forced me to shuffle two cards back into R&D, even though only one MOH was rezzed.
I have a screen shot of the table at the time if it will help.

nealterrell:
Fix #1888 by giving Flare a `:cancel-effect`.

Fix #1890 by having MoH only count rezzed copies.

Fix #1879 by only registering events when re-enabling a card if that card is active (rezzed).

Fix #1889 by having `purge` use `all-installed` to find nested hosted viruses.

### DIFF
--- a/src/clj/game/cards-assets.clj
+++ b/src/clj/game/cards-assets.clj
@@ -653,7 +653,7 @@
                                 (str "Choose " (if (< 1 mus) (str mus " cards") "a card")
                                      " in Archives to shuffle into R&D")))
                  :choices {:req #(and (card-is? % :side :corp) (= (:zone %) [:discard]))
-                           :max (req (count (filter #(= "10019" (:code %)) (all-installed state :corp))))}
+                           :max (req (count (filter #(and (= "10019" (:code %)) (rezzed? %)) (all-installed state :corp))))}
                  :show-discard true
                  :priority 1
                  :once :per-turn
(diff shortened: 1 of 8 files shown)


mtgred:
👍

PR #1892, opened 2016-08-21
Fix game password entry issue

### LINKED ISSUE #1818: Join a game with password error
When I join a game with password and enter the wrong password, if a new party is create or an other delete in the same time, I "jump" to an other party with the same password field.

I try to join party n°3 in the list, miss password, a party is create so my party go to n°4 but game make me join party n°3 (old n°2) with n°3's password field.

cryogen:
Added temp state for the game currently awaiting password entry so that new
games don't overwrite the current state causing users to be joined to the
wrong game.

This is my first dabble into clojure so go easy on me, but do point out if i'm barking up the wrong tree and I'll get it right!

Fixes #1818

### DIFF
--- a/src/cljs/netrunner/gamelobby.cljs
+++ b/src/cljs/netrunner/gamelobby.cljs
@@ -108,11 +108,13 @@
   (send {:action "leave-lobby" :gameid (:gameid @app-state)})
   (om/update! cursor :gameid nil)
   (om/update! cursor :message []))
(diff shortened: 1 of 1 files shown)


Saintis:
Looks good! Could probably use `(swap! app-state dissoc :password-game)` instead of `(swap! app-state assoc :password-game nil)`.

Don't worry about the test, it is just having memory issues.

cryogen:
Made the change suggested, thanks :)

PR #1897, opened 2016-08-22
Address Circle-CI OOM error.

nealterrell:
Hopefully.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -11,9 +11,11 @@ dependencies:
   post:
     - cd data; coffee fetch.coffee
 
+
 test:
   override:
     - lein compile game.main
+    - lein compile test.core
     - lein test test.core
     - lein test test.cards.agendas
     - lein test test.cards.assets


Saintis:
All hail Neil, the fixer of CircleCI!

nealterrell:
Ne**a**l :stuck_out_tongue:.

`lein test` will compile any files that have not been compiled yet. Each test adds another function to be compiled, and that seems to be why we're running out of memory on Circle. If we compile individual files, then fewer things need to be compiled in one individual process, so it's less likely to run OOM. Right now we compile all the core game files with `lein compile game.main` (which depends on `game.core` [which has card definitions], `game.utils`, and `game.macros`), then we let `lein test` compile all the `test.*` files at once. That's what was breaking, so I forced `test.core` to be compiled separately first. 

In the future, if this happens again, you can break up the compilation even more finely depending on which phase is breaking. Either `game.core` can be compiled separately if the first phase breaks, or individual `test.cards.*` files can be compiled.

PR #1903, opened 2016-08-24
Card fixes

### LINKED ISSUE #1902: Algo trading Paid Ability cost incorrect
Firing the ability on Algo trading currently costs 2 creds and no clicks, when it should be 1 click and no creds to trash. Easy enough to fix while playing, but I assume (!) a relatively straight forward fix in code too!

Thank you :)

### LINKED ISSUE #1908: Space Camp and Medium interaction
Hello,
When doing an R&D run using Medium, if Space Camp triggers, there's an access error.  Here's the interaction:

Runner runs R&D with multiple accesses because of medium and encounters Space Camp.
Corp gets the option to use Space Camp's advancement tokens AND the runner gets access to the Successful Run button
If the Runner clicks "Successful Run" before the corp uses the advancement tokens, the run starts over with additional accesses. 
For instance, if the runner begins the run with 4 virus counters on medium, the first access is for 5 cards. If this bug occurs, the runner starts over with 10 cards, then 16, and so on.  Here's the log from the corp view:

"lunalovegood sets virus counters to 4 on Medium.
lunalovegood spends  to make a run on R&D.
wombat929 has no further action.
lunalovegood accesses 5 cards.
lunalovegood accesses 10 cards.
lunalovegood steals Global Food Initiative and gains 2 agenda points.
lunalovegood accesses 16 cards.
lunalovegood left the game."

Thanks for all your hard work on this project!
wombat929
briley@curragh-labs.org

JoelCFC25:
* Fix #1902. Sorry, just failed to correct Algo Trading's cost after the copy/paste from C.I. Fund.
* Fixes Jamie Micken, whose ability is wrongly firing both on an install done after the first one of the turn, and also when the card being installed is the one that results in more installed Shaper cards than other factions. 
* Fix #1908 by adding the `:delayed-completion` framework and wait prompts to Space Camp
* Cards trashed from Runner's hand by Salem's should not be able to be prevented (e.g., Fall Guy)

### DIFF
--- a/src/clj/game/cards-assets.clj
+++ b/src/clj/game/cards-assets.clj
@@ -981,9 +981,15 @@
                                card nil))}}
 
    "Space Camp"
-   {:access {:msg (msg "place 1 advancement token on " (card-str state target))
-             :choices {:req can-be-advanced?}
-             :effect (effect (add-prop target :advance-counter 1 {:placed true}))}}
+   {:access {:delayed-completion true
+             :effect (effect (show-wait-prompt :runner "Corp to place an advancement with Space Camp")
+                             (continue-ability
+                               {:msg (msg "place 1 advancement token on " (card-str state target))
+                                :choices {:req can-be-advanced?}
+                                :cancel-effect (effect (clear-wait-prompt :runner))
+                                :effect (effect (add-prop target :advance-counter 1 {:placed true})
+                                                (clear-wait-prompt :runner))}
+                              card nil))}}
 
    "Sundew"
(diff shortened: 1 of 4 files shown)


JoelCFC25:
I'm not sure why all upgrade tests are bombing, since I didn't change anything there.

kevkcc:
If you have time, Shi-Kyu could also use a wait prompt. I tried to do the fix myself but the weird character prevents the card definition from loading in my browser..

JoelCFC25:
Done!

mtgred:
👍

Issue #1852, opened 2016-08-08
Functionality: Update API for netrunnerdb


domtancredi:
netrunnerdb is retiring its current API, so we may need to update our fetch routines. 

They are also renaming faction "neutral" into "neutral-runner" and "neutral-corp"; need to investigate if that affects the API

Saintis:
Whoever looks into this might also look at NRDB API for loading DotW and player's own decks.

Saintis:
Should be done now.

PR #1905, opened 2016-08-24
Fetch NRDB API 2.0

nealterrell:
Update fetch URL to use new NRDB API 2.0.

Map new API keys to old keys so we don't have to change any card logic. For example, we don't need to update cards that target ice to look for "ice" type instead of "ICE".

Update Rebirth logic to look for "00" in a card's code to look for draft IDs, since we no longer have "setname" in the card data.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -11,9 +11,11 @@ dependencies:
   post:
     - cd data; coffee fetch.coffee
 
+
 test:
   override:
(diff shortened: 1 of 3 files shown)


JoelCFC25:
Fixes #1852.

mtgred:
👍

PR #1906, opened 2016-08-26
Update to installed macro

### LINKED ISSUE #1839: The Future Perfect triggers psi game on Full Immersion RecStudio
Just played a game in which The Future Perfect was installed and hosted on Full Immersion RecStudio.

When accessing The Future Perfect triggered a psi game even though it was installed and the psi game should no longer be triggered.

JoelCFC25:
Fixes #1839.

The `installed` macro needs to check not the `card` alone, but rather `(get-nested-host card)` in order to return an accurate result.


mtgred:
👍

PR #1907, opened 2016-08-26
Announcements

nealterrell:
Allows an admin to push announcement toasts to all active games. 

I think this is a "good enough" solution. Upon submitting a message, a toast is shown in all active games. It is not shown in main chat, and it is not shown to any game that starts after the message is submitted. The purpose of this feature is to warn users when @mtgred has to bring the server down for upgrades. A message in main chat is good to warn people who are chatting, but insufficient for those who are playing.

This feature is enabled at the URL `./announce`. Only administrators can see this page. Administrators must be marked in the users database as `isadmin: true`; Minh will know how to do this. I copied the layout of this page from the Reset Password page, so it's a little basic, but it gets the job done:

image

The message goes to the Node server, which sends it to the game server. The game server inserts a Warning toast into each active game state. This took some minor surgery on `main`.

### DIFF
--- a/data/fetch.coffee
+++ b/data/fetch.coffee
@@ -15,61 +15,91 @@ appName = process.env.OPENSHIFT_APP_NAME || 'netrunner'
 
 db = mongoskin.db("mongodb://#{login}#{mongoHost}:#{mongoPort}/#{appName}")
 
-setFields = [
-  "name",
-  "available"
-]
+same = (key, t) ->
+  return [key, t]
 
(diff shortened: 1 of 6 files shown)


Saintis:
Looks good. How much work would it be to also make it send some **ANNOUNCEMENT - RESTART SOON** in main chat (custom server message rather than just Minh posting)?

mtgred:
Amazing!

PR #1911, opened 2016-08-28
Implementation Notes

Saintis:
Now more ready to merge...

Need some magic CSS help again to (ideally) make the log panel shrink to fit the implementation note (see below.

---

Adds notes to cards that are unimplemented or have partial implementations to make users aware of there situations (should cut down on issues posted about Enhanced Login Protocols atleast).

Notes are generates as following:
- If the card has no card-def then `:implementation nil` is associated to the card signifying that it is unimplemented
- If the card has a card-def but no `:implementation` key in the card-def then it is assumed to be fully implemented and `:implementation :full` is associated to the card
- If the card has a card-def and an `:implementation` key in it, the value of that key is passed along (the implementation note)
- If the card has a `:recurring` key then a notice that recurring credit usage is not restricted is added to the implementation note

Adds implementation notes to most of the cards listed as "partial" implementation in the card implementation document. Have ignored subroutine count and adding related implementations since most of that is done manually anyway.

### Examples:
screen shot 2016-08-28 at 12 12 29
screen shot 2016-08-28 at 12 12 31
screen shot 2016-08-28 at 13 54 55

### CSS issues (due to static log panel height):
Empty space for fully implemented cards
screen shot 2016-08-28 at 14 00 00

Overflow / clash on multiline implementation notes
screen shot 2016-08-28 at 13 59 22
screen shot 2016-08-28 at 14 06 04


zaroth:
Absolutely splendid idea and work done here. Unfortunately can't help with the CSS issues, as it's not my forte, but I hope it gets through ASAP.

Some ideas:
* if the card view with implementation note is somehow modularized, it could be shown in the card browser/deckbuilder too, if desired (however the card data probably can't be stored in :implementation instead card logic, but somewhere external)
* as per above, maybe consider storing the implementation notes card mapping in a separate clojure file that may be included wherever relevant (i.e. also outside of the game view)? Downside of such approach is that it isn't stored alongside game logic, so it may be forgotten. Upside, outside of mentioned out-of-game use flexibility, is that we could possible auto-generate this file from our google document, so it doesn't have to be changed in two places.

Saintis:
Alternate message location. Needs some CSS help as well to make the container fit to the text. 
screen shot 2016-09-10 at 13 07 27
screen shot 2016-09-10 at 13 07 42

Saintis:
Sorry, made a mess rebasing this and the alternative styling branches I have locally...

I think the altarnative message location shown above is the way to go (probably easiest as well) so that has been merged into this PR.

Can even leave the message box at fixed width - in that case this is ready to merge.

Let me know what you prefer @mtgred.

src/css/base.styl
```diff
@@ -1674,10 +1674,41 @@ nav ul
         .text
           height: 195px
 
+    // Card implementation div
+    .implementation
+      width: 150px
```
erbridge: To deal with the width issues, couldn't you just change this to `max-width` to make it automatically fit to the contents?

Saintis:
Ready to merge again.

zaroth:
You didn't change the branch to be merged into to `dev`, though... Unless you meant for it to go straight to `master` ("Edit" at the top)

Saintis:
Hmm, that was unintended

Saintis:
I guess you get all the stuff in `master` for free 😛

zaroth:
Saved me a merge ;-)

src/css/base.styl
```diff
@@ -1700,10 +1700,41 @@ nav ul
...
+      // stuff from .panel
+      radius: 4px
+      padding: 5px 10px
+      border: 1px solid rgba(39, 56, 76, 0.9)
+      margin: 3px
+
+      // stuff from .blue-shade
+      color: white
+      background-color: transparent-blue
+      box-shadow: 0 0 12px rgba(0, 0, 0, 0.4), inset 0 0 0 1px rgba(255, 255, 255, 0.1)
```
erbridge: Why copy from `.panel` and `.blue-shade` rather than just adding the classes?

PR #1913, opened 2016-08-28
Change mill definition to use move

kevkcc:
Small PR. :mill-effect was something I introduced for Subliminal Messaging but it's redundant because :move-zone with a check to discard fulfills the same purpose. So if we just change the mill definition to use the move function, we can remove :mill-effect and reduce code duplication.


mtgred:
👍

PR #1917, opened 2016-08-30
Adds cards trashed from damage message.

Saintis:
Explicitly prints which cards were trashed due to damage.

screen shot 2016-08-30 at 19 06 39


mtgred:
👍

PR #1923, opened 2016-09-02
Mr. Li rework (Genetics Pavilion/Mr. Li interaction)

### LINKED ISSUE #1594: Mr. Li and Genetics Pavilion interation
Genetics Pavilion reads: The Runner cannot draw more than 2 cards during his or her turn.
Mr. Li reads: Draw 2 cards. Add 1 of these cards to the bottom of your stack.

Issue: The runner is not prevented from drawing more cards after using Mr. Li or after previously drawing cards during their turn.
Fix: The runner should be prevented from drawing more than 2 cards on their turn even when using Mr. Li.

kevkcc:
Fixes #1594 

Mr. Li now uses the draw function so that Genetics Pavilion can track the draws.
Looks like Daily Business Show could use a similar update but that one looks a lot harder. I will have a further look sometime.

### DIFF
--- a/src/clj/game/cards-resources.clj
+++ b/src/clj/game/cards-resources.clj
@@ -622,13 +622,17 @@
     :abilities [ability]})
 
    "Mr. Li"
-   {:abilities [{:cost [:click 1] :prompt "Card to keep?"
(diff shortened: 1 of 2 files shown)


mtgred:
👍

PR #1928, opened 2016-09-02
Implement Equivocation. Improvements to simultaneous effect resolution.

nealterrell:
Implementation and test for Equivocation. To do this, it was necessary to extend the simultaneous effect resolution prompts to the `:successful-run` event, in order to account for the interaction with Laramy Fisk. A few cards handling this event had to be cleaned up, following these general guidelines:

1. An event handler should be `:silent (req true)` if it never needs to show up in the simultaneous effect resolution prompts. This is a judgment call, but I think it's reasonably clear when a handling effect has no possible interaction with other handlers of the same event. Example: Medium gains a token on successful run; there are no interactions with other successful-run effects that would make it necessary to manually control when Medium gets its token in the effect-resolution window; therefore it can be silent.
2. An event handler should be `:interactive (req true)` if it has complicated interactions with other handling effects. If at least one handler is interactive, then the simultaneous resolution prompt is shown with all non-silent handlers. In this case we mark both Fisk and Equivocation as interactive.

I also improved several agendas and their interactions with the simultaneous effect resolution prompts. It is currently unimportant whether AstroScript gains its token before or after other "when scored" effects, so it (and similar agendas) is marked as `:silent`. Other agendas are now marked `:interactive` if there are existing interactions or conceivable future interactions that would make the order of resolution important. For example, Corporate War currently has no interactions, but we can conceive of a future current or asset that says "When you score an agenda, gain some number of credits" which would have an interaction with Corp War.


mtgred:
👍

Issue #1931, opened 2016-09-02
Straighten out NRDB API migration issues


zaroth:
Amongst others:
- deckbuilder stores the whole card identity data along with the deck. That's a recipe for problems if the card ever gets errata'd or their fields change (as it happened now, checking validity is now via `pack_code`, not `setname`) 
- card fields are currently renamed by the `fetch.coffee` script to the old, depreciated ones. To reduce confusion and debugging times, the new terminology from API 2.0 should be used.
- consider referring to cards by their IDs instead of their names, as it is way more future-proof. Also, what if FFG prints another version of the card with the same name? It may mess up a bit current alt-arts implementation, though.

zaroth:
Some fixes to this were pushed by @minh in f0c07ff 919b048 , but I don't think we should store whole identity data of decks, just their ID - so this remains to be fixed.

PR #1933, opened 2016-09-02
A stopgap measure for all cards suddenly being unreleased

zaroth:
Currently fixed it by mapping the codes to names as @nealterrell suggested. However, it is not very sustainable for the longterm - see #1931


Saintis:
If you want it less hack-y you could use `/api/2.0/public/pack/{pack_code}` to get the names perhaps.

mtgred:
I just pushed a proper fix: https://github.com/mtgred/netrunner/commit/f0c07ff8a1e143c56a962a5a48500b368ae36c16

PR #1934, opened 2016-09-03
Fix Space Camp and other run-ending-too-soon issues

### LINKED ISSUE #1929: Space camp in archives doesn't end run properly
1. Put space camp in Archives
2. Runner runs on it
3. After the advancement token is placed, the run should end since that was the last thing to access but the Runner gets to click successful run again

Other notes: If space camp isn't the last effect in Archives, for instance, if you access Space Camp first and then a News Team, run will end properly. Also the :cancel-effect on Space Camp is busted, it just kicks you out of the run completely.

Shi-kyu might have a similar issue since it's an Archives trap but I can't test it due to the weird character not rendering in my browser

nealterrell:
Fix #1929. It's not actually a problem with Space Camp... I overlooked archives access when fixing some of the run routines to not end runs too early. While doing this, I discovered potential issues when accessing multiple unrezzed upgrades in a server, so I fixed that too. Tests for each scenario. Thanks to @kevkcc for the reproduction steps.

### DIFF
--- a/src/clj/game/core-runs.clj
+++ b/src/clj/game/core-runs.clj
@@ -226,6 +226,7 @@
 (defn access-helper-remote [cards]
   {:prompt "Click a card to access it. You must access all cards in this server."
    :choices {:req #(some (fn [c] (= (:cid %) (:cid c))) cards)}
+   :delayed-completion true
    :effect (req (when-completed (handle-access state side [target])
                                 (if (< 1 (count cards))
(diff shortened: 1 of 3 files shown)


mtgred:
👍

wombat929:
Hello! Thanks for all your work on this! I think it's still not working correctly when there are more than one Space Camp in the Archives. Thanks!

PR #1936, opened 2016-09-03
Lets Adam Pull Directives from `@all-cards`

### LINKED ISSUE #1239: Make Adam pull directives from DB instead of making him include them in his deck
A bandaid on the Adam situation was put in place in #1232. However, it's not ideal. As per @nealterrell:

> the game engine doesn't have the ability to load cards from thin air; the web server sends the entire decklist to the game server when creating a game, and those are the only cards the server knows about.
> This will need to change for the spoiled card Rebirth

Saintis:
Rewrites the Adam implementation to let the directives be fetched from `@all-cards` if no directives are specified in the deck.

~~The old way of adding 3 directives to the deck still works.~~

Removes the Adam specific stuff in the deckbuilder since this is no longer relevant. Updates the help page.

Adds a test for Adam with no directives.

Fixes #1239

### DIFF
--- a/src/clj/game/cards-identities.clj
+++ b/src/clj/game/cards-identities.clj
@@ -10,43 +10,21 @@
    {:events {:pre-start-game
              {:req (req (= side :runner))
               :effect (req (let [is-directive? #(has-subtype? % "Directive")
-                                 one-of-each (fn [cards] (->> cards (group-by :title) (map second) (map first)))
-                                 directives (filter is-directive? (concat (:deck runner) (:hand runner)))
-                                 directives (one-of-each directives)]
-                             (case (count directives)
-                               ;; Have directives in the deck
-                               3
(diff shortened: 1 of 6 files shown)


zaroth:
Make sure to update `help.cljs` too!

zaroth:
Also, I'm not sure about that part:

> The old way of adding 3 directives to the deck still works.

Do you mean that to have one spare of each directive inside the deck (beside the pre-installed ones) one would have to build 51 card deck (45 base + 6 directives)? That's fairly unintuitive. I think I'd rather have it consistent, even if it means that all existing decks will have 1 spare directive each for now :)

Saintis:
Can remove the old way of directives. Means some kind of announcement would be useful for when this is deployed so any Adam players out there can update their decks.

mtgred:
Awesome 👍

PR #1939, opened 2016-09-06
Fixes: Shi.Kyu in R&D, remove halo from stolen agendas

### LINKED ISSUE #1935: Remove fresh card highlight from stolen agendas
The highlight indicating that the card was freshly installed doesn't ever go away if the Runner steals the agenda very next turn. It should be removed right then.

### LINKED ISSUE #1937: Shi-Kyu fires from R&D
There seems to be some bug with Shi-Kyu half-firing on the runner's side from R&D. If it is accessed from R&D then they have to use the /close-prompt action to be able to continue playing the game.

JoelCFC25:
* Fix #1937: The `:req` is at the wrong spot, leaving the Runner blocked by a wait prompt even when hitting the card in R&D
* Fix #1935: Dissoc `:new` from agendas when they are stolen so an orange halo around it won't persist.

### DIFF
--- a/src/clj/game/cards-assets.clj
+++ b/src/clj/game/cards-assets.clj
@@ -938,11 +938,11 @@
    "Shi.Kyū"
    {:access
     {:delayed-completion true
+     :req (req (not= (first (:zone card)) :deck))
      :effect (effect (show-wait-prompt :runner "Corp to use Shi.Kyū")
                      (continue-ability
                        {:optional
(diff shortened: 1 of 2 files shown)


zaroth:
Are there any more cases of lingering halo after this fix?

JoelCFC25:
Yeah, the case of the Corp voluntarily trashing cards before the halo is removed on its own still exists. I can add one more commit that will take care of that part.

mtgred:
👍

PR #1945, opened 2016-09-07
Deckbuilder improvements

### LINKED ISSUE #1926: Unify deckbuilder symbols with netrunnerDB
★ - MWL
● - influence taken by a card
​​○​​ - influence that would be taken by an alliance card but isn't due to meeting alliance conditions

Saintis:
Refactors some stuff:
- Alliance helpers
- Dots construction for costs
- Influence mapping function
- Draft identity check
- Deck parsing from the database and text
- Checking for MWL penalty (see below for more details)

Significant Changes:
- Looks up the identity like any other card. Significant change is that now only save the `:title :side :code` keys of the identity back to the database. First change should fix any weird "unreleased cards" issues from the `:setname` of the id. Second change will reduce data stored in the database.
- Now loads in the MWL lists from NRDB. Uses the currently active MWL list to check for penalty influence.
- Removes the OCTGN Export button (and the functionality).

Ideally the id should just be the card code (and all the other cards as well), but in the current implementation cards are looked up by `:title` and the search is filtered by the `:side` of the identity so both those keys are also needed.

MWL changes: fetches MWL lists from NRDB, defaults to using currently active list to check for influence penalty but has functionality to specify the code of the list to use for checking. This could allow people to specify what penalty list they are building for (e.g. MWL 1.0, 1.1 or ANRPC GLC 1.0).

Adds empty dots for free Alliance influence -- fixes #1926 

Makes functions more pure by extracting `sets` variable up to the `deck-builder` function.

Improves Draft identity support. Draft identities are no longer restricted by maximum card counts and can freely add more cards.

Screenshots:
screen shot 2016-09-07 at 20 53 58
screen shot 2016-09-07 at 22 25 03

### DIFF
--- a/data/fetch.coffee
+++ b/data/fetch.coffee
@@ -30,6 +30,13 @@ setFields = {
   "cycle_code" : (k, t) -> ["cycle", capitalize(t.replace(/-/g, " "))]
 }
 
+mwlFields = {
+  "name" : same
+  "code" : same
+  "active" : same
+  "cards" : same
+}
+
 mapFactions = {
   "haas-bioroid" : "Haas-Bioroid",
   "jinteki" : "Jinteki",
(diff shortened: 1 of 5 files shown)


Saintis:
Will keep on working on this a bit to extract / refactor some more stuff.

Saintis:
Just pushed some more significant changes.

Saintis:
Ok, this should be ready to merge now.

mtgred:

Saintis:
Ready for merge again.

mtgred:
👍

PR #1950, opened 2016-09-09
Minor identity handling refactor

### LINKED ISSUE #1948: Make runner identities load link from card data
Currently the link value of ids is hard coded. A better solution would be to update the link value when the id is initialized from the `base_link` card data property.

Would also make Rebirth's implementation a bit cleaner.

Saintis:
Creates new function `identity-init` that initialises the runner starting link. Removes link gaining from identity carddefs. Updates Rebirth to use new function.

Adds two tests to ensure link is gained/lost properly.

Fixes #1948


mtgred:
👍

PR #1960, opened 2016-09-13
prevent Archangel from targeting Corp cards

JoelCFC25:
Discovered in a report on Stimhack Slack. Also added a gross hack to stop Mumbad Virtual Tour's forced auto trash if Salsette Slums is installed.


mtgred:
👍

PR #1963, opened 2016-09-15
Patent-pending grab bag of card fixes

### LINKED ISSUE #1947: Edward Kim not trashing on access after Retrieval Run
After using Retrieval Run and accessing no other cards in the turn, if Edward Kim accesses an operation, it won't be trashed.

### LINKED ISSUE #1962: Controlling the Message fails to trigger when a runner card has been trashed earlier that turn
For example, if the runner uses a Street Peddler before trashing the first corp card, CTM will not trigger.

I'm not 100% sure this is the issue, but it seems to fit every time CTM fails to trigger in a game I'm playing.

JoelCFC25:
* Fix #1962: For NBN: Controlling the Message, scan all the instances of `:runner-trash` for the turn and make sure none are Corp cards. 
* Fix #1947: Solves Edward Kim + Retrieval Run by only switching off his ability for the turn if no access replacement was used on the Archives run.
* Fixes Award Bait issue where the stolen agenda is never placed in the Runner's score area if the Corp has no advanceable cards in play.
* Fixes Obelus interaction with access replacement run events (where `:cards-accessed` would be nil). 
* Runner needs to lose/gain 1 click if Victoria Jenkins is rezzed or trashed during the Runner's turn

### DIFF
--- a/src/clj/game/cards-agendas.clj
+++ b/src/clj/game/cards-agendas.clj
@@ -73,6 +73,7 @@
 
    "Award Bait"
    {:access {:delayed-completion true
+             :req (req (not-empty (filter #(can-be-advanced? %) (all-installed state :corp))))
              :effect (effect (show-wait-prompt :runner "Corp to place advancement tokens with Award Bait")
                              (continue-ability
                                {:delayed-completion true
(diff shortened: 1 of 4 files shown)


mtgred:
👍

PR #1966, opened 2016-09-16
Space camp fix

kevkcc:
Small PR but noticed that space camp is (still) not 100% since cancelling the effect kicks you out of an archives run before accessing anything else prematurely. This should be good now.


nealterrell:
OK, I see why the previous fix didn't do anything: `:cancel-effect` is being called directly instead of run through `resolve-ability`, so `effect-completed` isn't being called to indicate the ability has finished.

mtgred:
👍

PR #1953, opened 2016-09-10
Fix several successful-run issues

nealterrell:
Fix #1949, fix #1951, fix #1952.

Cleans up `trigger-event-simult`. Ignores handlers that are suppressed (Crisium Grid) or whose `:req` is false. (Not sure why I didn't do this before. Maybe it'll come to me in a dream.)

PR #1967, opened 2016-09-16
Fix for Manhunt and other successful-run cards

nealterrell:
I thought I fixed this in #1953, but did not write a test to verify the correctness. Bad Neal.


mtgred:
👍

PR #1971, opened 2016-09-18
Damage statement fix + Argus message improvement

### LINKED ISSUE #1969: Cards trashed by damage message displaying when no cards trashed
This is a small cosmetic thing:

image

When all the points of damage from a given source are prevented, the new cards trashed due to damage message still displays, but with no listed cards.

As a side note, it might be nice to have a message indicating which choice the runner makes when Argus triggers. It's implied by the damage prevention, but I think the log would look cleaner and more intuitive if it said "RunnerName chooses to take 2 meat damage" after resolving the dialogue.

Saintis:
Only perform card trash / hand size modification / flatline check / "trashes ... due to damage" if the damage count is positive.

Elaborates system message when runner chooses an option from Argus -- adds "chooses to". Also moves this message to be printed directly below the argus message, before any messages due to damage or tag taking.

Fixes #1969.

Screenshots:
screen shot 2016-09-18 at 20 49 13
screen shot 2016-09-18 at 20 50 52
(Yes, the empty "trashes due to damage" statement is no longer printed if all damage is prevented, it is not just cropped away :wink:)

### DIFF
--- a/src/clj/game/cards-identities.clj
+++ b/src/clj/game/cards-identities.clj
@@ -48,11 +48,11 @@
               :choices ["1 tag" "2 meat damage"] :player :runner
               :msg "make the Runner take 1 tag or suffer 2 meat damage"
               :effect (req (if (= target "1 tag")
-                             (do (tag-runner state :runner 1)
-                                 (system-msg state side "takes 1 tag")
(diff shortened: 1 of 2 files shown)


FarCryFromHuman:
Looks great!

mtgred:
:+1:

PR #1975, opened 2016-09-19
Refactors how flags work.

Saintis:
Also refactors the genetics cards a bit.

Adds a test and a pending test for Councilman.

I'm looking into fixing the Councilman + zone movement issue (the pending test) but it is likely to need some non-trivial code changes so I'll tack it on to this as a second PR if I get anywhere with it.


nealterrell:
👍

PR #1980, opened 2016-09-20
Give Chum ability to give +2 strength to ICE

JoelCFC25:
Minor usability improvement for Chum. The targeted ICE must be rezzed, so the Corp should hold off on using this ability until it's clear that the Runner isn't bypassing (e.g., Security Nexus or some other shenanigans). 

Also fixes the Rigged Results typo, Khan's misplaced install cost bonus, and stops Hatchet Job from targeting Corp cards.


nealterrell:
👍

PR #1982, opened 2016-09-20
implement Net Mercur

JoelCFC25:
This card is bonkers. We simply check for the Stealth subtype on a card when recurring credits are taken and then trigger a new event. Ghost Runner and Net Mercur have to manually trigger this event themselves since they don't use recurring credits.


nealterrell:
👍

PR #1984, opened 2016-09-22
implement Ark Lockdown

JoelCFC25:
We briefly discussed how the auto-complete text entry boxes on Targeted Marketing and Salem's Hospitality are not the best fit here, since the Corp would never name a card which isn't present in the Heap (would they??).


nealterrell:
👍

PR #1988, opened 2016-09-23
Provide a base colour for input fields

erbridge:
In Firefox, themes are able to set their own default colours. On a dark theme, this can result in unreadable text inputs.


erbridge:
screenshot from 2016-09-23 20-36-04

vs

screenshot from 2016-09-23 20-34-53

Issue #1388, opened 2016-03-25
Visual indicators of main ICE subtypes: B/C/S


zaroth:
Using the framework put down by Saintis for Femme Fatale counter, we could consider using similar (maybe a bit less visually prominent) counters for ICE subtypes. This would make it easy to tell what subtype a piece of ICE is at a glance, even for beginners: 

- **B**(arrier) 
- **C**(ode gate) 
- **S**(entry)

Some kind of color coding of the counters could be also nice to make it even more clear when you're encountering a code gate / sentry / barrier. To take it further, breakers could also have similar symbol(s) with matching color to signify which types they can break.

Cards that would gain the most from this kind of display:

- Tinkering
- Paintbrush
- Chimera
- Changeling
- Lycan
- Wendigo
- Mother Goddess
- Rielle "Kit" Peddler: Transhuman
- Sub Boost

Saintis:
Some of these will be improved by #1990.

JoelCFC25:
Are we able to close this now? Seems like it was mostly addressed.

zaroth:
Definitely :)

PR #1990, opened 2016-09-23
Add subtype labels

erbridge:
Add labels like the server designation of Patron, but for the chosen subtype for Chameleon and Chimera.

### DIFF
--- a/src/clj/game/cards-agendas.clj
+++ b/src/clj/game/cards-agendas.clj
@@ -611,36 +611,27 @@
              :msg "add it to their score area and gain 1 agenda point"}}
 
    "Rebranding Team"
-   {:interactive (req true)
-    :msg "make all assets gain Advertisement"
-    :effect (req (doseq [c (filter #(is-type? % "Asset")
-                                   (concat (all-installed state :corp)
-                                           (:deck corp)
-                                           (:hand corp)
-                                           (:discard corp)))]
-                   (update! state side (assoc c :subtype
-                                              (->> (vec (.split (or (:subtype c) "") " - "))
-                                                   (cons "Advertisement") ;append ad even if it is already an ad
-                                                   (join " - "))))))
-    :swapped {:msg "make all assets gain Advertisement"
-              :effect (req (doseq [c (filter #(is-type? % "Asset")
-                                             (concat (all-installed state :corp)
-                                                     (:deck corp)
-                                                     (:hand corp)
-                                                     (:discard corp)))]
-                             (update! state side (assoc c :subtype
-                                                          (->> (vec (.split (or (:subtype c) "") " - "))
-                                                               (cons "Advertisement")
-                                                               (join " - "))))))}
-    :leave-play (req (doseq [c (filter #(is-type? % "Asset")
-                                       (concat (all-installed state :corp)
-                                               (:deck corp)
-                                               (:hand corp)
-                                               (:discard corp)))]
-                       (update! state side (assoc c :subtype
-                                                    (->> (vec (.split (or (:subtype c) "") " - "))
-                                                         (drop 1) ;so that all actual ads remain ads if agenda leaves play
-                                                         (join " - "))))))}
+   (let [get-assets (fn [state corp]
+                      (filter #(is-type? % "Asset")
+                              (concat (all-installed state :corp)
+                                      (:deck corp)
+                                      (:hand corp)
+                                      (:discard corp))))
+         add-ad (fn [state side c]
+                  (update! state side
+                           (assoc c :subtype (combine-subtypes false ;append ad even if it is already an ad
+                                                               (or (:subtype c) "")
+                                                               "Advertisement"))))]
+     {:interactive (req true)
+      :msg "make all assets gain Advertisement"
+      :effect (req (doseq [c (get-assets state corp)] (add-ad state side c)))
+      :swapped {:msg "make all assets gain Advertisement"
+                :effect (req (doseq [c (get-assets state corp)] (add-ad state side c)))}
+      :leave-play (req (doseq [c (get-assets state corp)]
+                         (update! state side (assoc c :subtype
+                                                      (->> (split (or (:subtype c) "") #" - ")
+                                                           (drop 1) ;so that all actual ads remain ads if agenda leaves play
+                                                           (join " - "))))))})
 
    "Remote Data Farm"
    {:silent (req true)
--- a/src/clj/game/cards-agendas.clj
+++ b/src/clj/game/cards-agendas.clj
@@ -611,17 +611,15 @@
              :msg "add it to their score area and gain 1 agenda point"}}
 
    "Rebranding Team"
-   (let [get-assets (fn [state corp]
-                      (filter #(is-type? % "Asset")
-                              (concat (all-installed state :corp)
-                                      (:deck corp)
-                                      (:hand corp)
-                                      (:discard corp))))
-         add-ad (fn [state side c]
-                  (update! state side
-                           (assoc c :subtype (combine-subtypes false ;append ad even if it is already an ad
-                                                               (or (:subtype c) "")
-                                                               "Advertisement"))))]
+   (letfn [(get-assets [state corp]
+             (filter #(is-type? % "Asset") (concat (all-installed state :corp)
+                                                   (:deck corp)
+                                                   (:hand corp)
+                                                   (:discard corp))))
+           (add-ad [state side c]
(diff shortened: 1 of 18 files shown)


src/css/base.styl
```diff
@@ -1528,6 +1528,10 @@ nav ul
             .strength
               transform(rotate(90deg))
 
+            .subtype-target
+              transform(rotate(90deg), translateX(100%))
```
Saintis: This rotates it the wrong way for the runner... missing the corresponding code under `.corp-board.opponent` to rotate / translate it back. `transform(rotate(-90deg), translateY(-100%))` seems to do it for me.

Comparison:
Old
screen shot 2016-09-23 at 19 47 19
New
screen shot 2016-09-23 at 19 54 03

erbridge: Aha. New to the codebase, so missed that. I'll fix that.

Saintis:
Some pics to show the awsomeness:
Corp-side:
screen shot 2016-09-23 at 19 45 59


Chameleon:
screen shot 2016-09-23 at 19 47 57

Saintis:
I do have another discussion point: colours?

Would love to have the following colour scheme:
- Barrier ~ Anarch orange
- Code Gate ~ Shaper green
- Sentry ~ Criminal blue

Would make the icons a bit more fun than the current "sienna" colour 😄

Thoughts?

src/css/base.styl
erbridge: Should be fixed now.

Saintis:
Seems like you have to update the tests to look for `:subtype-target` rather than `:ice-type` for the morph ice.

erbridge:
Setting the colours based on the type looks like it needs a better grasp of Clojure than I have (this is the first time I've even looked at the language :)). Doesn't feel like a necessary feature to me, though, but feel free to correct me :).

Looking at the failing test, now. It doesn't refer to `:ice-type` anywhere in the code, and that change was related to Chimera, not the moprh ice anyway. It looks like `update-ice-strength` isn't being fired, but I don't understand why that would be...

Saintis:
I'll have a look.

src/clj/game/cards-ice.clj
```diff
@@ -336,9 +337,10 @@
                                       (->> (vec (.split (:subtype card) " - "))
                                            (concat [target])
                                            (join " - "))))
+                      (update! (assoc card :subtype-target target))
```
Saintis: Found the problem. The `update!` changes the card as stored in `state` (signified by the `!`).
Now above this line the card is also `update!`ed to have the new `:subtype`. The `:subtype-target` `update!` overwrites this change (as you are `assoc`ing with the old `card`), which is why the tests fail (the card never actually morphs).

This works for me:
```clojure
(defn morph [state side card new old]
  (update! state side (assoc card
                        :subtype-target new
                        :subtype (->> (remove #(= old %) (.split (:subtype card) " - "))
                                      vec (concat [new]) distinct (join " - "))))
  (update-ice-strength state side card))
```

erbridge: Aha! This makes sense. Is this why I see `(get-card state card)` all over the place?

Saintis: So `(get-card state card)` refreshes the card as defined in the state (so that could have been an alternative solution for the problem above). It's been added in a lot of places recently to prevent client side hacking of trash / rez / play costs or other devious things by reseting the card definition to what the server has. Except for client-facing functions there is normally not much need to call `(get-card state card)` since its often clearer to use a `let` or similar to define the "updated" card before you send it to `update!` or other similar function.

@nealterrell anything to add to this?

erbridge: That's what I thought. That's useful. Just testing the fix now.

erbridge:
So I have noticed that if I `/counter 0` a Wendigo with an odd number of advancement tokens, it retains the Barrier type. Is there an event I need to be listening to for this, or is this an issue with `/counter`?

Saintis:
Appears to be an issues with `/counter 0`. I don't think it triggers `:advancement-placed`. Feel free to make an issue 😄

erbridge:
Alright :).

I think all the issues are fixed (other than the colours). I thought about giving Mother Goddess the same treatment, but she can have a lot of subtypes, so it wouldn't really work in this way. Also Panchatantra, but that's not even implemented.

src/cljs/netrunner/gameboard.cljs
```diff
@@ -512,7 +512,12 @@
           current-strength [:div.darkbg.strength current-strength])
         (when-let [{:keys [char color]} icon] [:div.darkbg.icon {:class color} char])
         (when server-target [:div.darkbg.server-target server-target])
-        (when subtype-target [:div.darkbg.subtype-target subtype-target])
+        (when subtype-target
+          (let [colour-type (case subtype-target
+                              ("Barrier" "Sentry") (lower-case subtype-target)
+                              "Code Gate" "code-gate"
+                              nil)]
```
erbridge: This was the bit I struggled with :).

Saintis:
Mother Goddess + Panchatantra would both require some way to have more than one subtype label, which would be cool but probably run into space issues.

Also Tinkering which adds Barrier, Code Gate and Sentry.

Some quick colours:
screen shot 2016-09-23 at 22 13 57
screen shot 2016-09-23 at 22 14 03

src/cljs/netrunner/gameboard.cljs
Saintis: Feel free to get rid of it if you want. I wanted to see if a PR to your branch worked and this happened instead 😕 😄

Saintis:
@mtgred this is ready for merge now. Feel free to skip / revert the colours (or change them) depending on your personal preference.

erbridge:
That white on green is pretty low contrast... Might be better to mkae the colours a bit more muted, or change the text colour.

Saintis:
Alternative darker colours:
screen shot 2016-09-23 at 22 43 29

```stylus
&.barrier
  background-color: #cc5200

&.code-gate
  background-color: #29a329

&.sentry
  background-color: #007acc
```

src/clj/game/cards-ice.clj
nealterrell: Not much to add. Yes, be careful with multiple `update!` to the same card. Values are immutable so the first does not change your local copy of the card data.

src/cljs/netrunner/gameboard.cljs
erbridge: Pretty cool, though. Maybe update it with the new colours, though?

Saintis:
Pushed the change to the darker colours.

Saintis:
If we want to add more labels there is a fairly comprehensive list of related cards in #1388.

erbridge:
Added Mother Goddess and refactored a bunch of stuff. She gets the first character of subtypes if she has more than one (to prevent wrapping issues).

Also refactored a lot of repeated subtype modification code.

The cards that modify a card other than themselves are a little problematic, due to potential clashes. Probably subtype targets should be given a similar treatment as Rebranding Team gives to the Avertisement subtype.

JoelCFC25:
Really just fantastic work--it looks SO much cleaner.

src/clj/game/cards-agendas.clj
```diff
@@ -612,36 +612,27 @@
...
+   (let [get-assets (fn [state corp]
+                      (filter #(is-type? % "Asset")
+                              (concat (all-installed state :corp)
+                                      (:deck corp)
+                                      (:hand corp)
+                                      (:discard corp))))
+         add-ad (fn [state side c]
+                  (update! state side
+                           (assoc c :subtype (combine-subtypes false ;append ad even if it is already an ad
+                                                               (or (:subtype c) "")
```
Saintis: This could be `(:subtype c "")`

src/clj/game/cards-agendas.clj
```diff
@@ -612,36 +612,27 @@
...
-    :leave-play (req (doseq [c (filter #(is-type? % "Asset")
-                                       (concat (all-installed state :corp)
-                                               (:deck corp)
-                                               (:hand corp)
-                                               (:discard corp)))]
-                       (update! state side (assoc c :subtype
-                                                    (->> (vec (.split (or (:subtype c) "") " - "))
-                                                         (drop 1) ;so that all actual ads remain ads if agenda leaves play
-                                                         (join " - "))))))}
+   (let [get-assets (fn [state corp]
```
Saintis: Could alternatively use `letfn`

mtgred:
Look good. :)

PR #1991, opened 2016-09-23
Cleans up file names

Saintis:
Moves the `cards-*` and `core-*` files into `cards` and `core` folders respectively and drops their prefixes.

Cleans up some `:refer`s and `declare`s.

Moved `game-states` and `old-states` atoms to where they were being used, `core/turns` and `main` respectively.

This should either be merged first, and all other PR's rebased / fixed (probably a lot of work), or merged last (and I'm the only one that has to do work). Hopefully git can manage renamed files to make it less painful.

Also looked at spliting `game.core` namespace into `game.core` and `game.cards` but it's not possible due to the circular dependency:
- Card definitions needs `game.core` functions
- `game.core` defines and heavily uses `card-def` which depends on the card definitions


erbridge:
Merge last means never :).

Saintis:
Now back and ready to merge again! 😁

PR #1992, opened 2016-09-23
Make /counter and /adv-counter trigger :advancement-placed

erbridge:
This fixes some issues with advanceable ice (I noticed it while working on #1990 for the morph ice).

### DIFF
--- a/src/clj/game/core-io.clj
+++ b/src/clj/game/core-io.clj
@@ -80,7 +80,7 @@
          :else nil))
 
 
-; In-game chat commands
+;;; In-game chat commands
 (defn set-adv-counter [state side target value]
   (set-prop state side target :advance-counter value)
   (trigger-event state side :advancement-placed target)
@@ -103,7 +103,7 @@
                                      (and (is-type? target "Agenda") (is-scored? target)) :agenda
                                      (and (card-is? target :side :runner) (has-subtype? target "Virus")) :virus)
                         advance (= :advance-counter c-type)]
-                    (cond advance (do (set-adv-counter state side target value))
+                    (cond advance (set-adv-counter state side target value)
                           (not c-type) (toast state side "You need to specify a counter type for that card." "error"
                                               {:time-out 0 :close-button true})
                           :else (do (set-prop state side target :counter (merge (:counter target) {c-type value}))
--- a/src/clj/game/core-io.clj
+++ b/src/clj/game/core-io.clj
@@ -81,10 +81,15 @@
 
 
 ; In-game chat commands
+(defn set-adv-counter [state side target value]
+  (set-prop state side target :advance-counter value)
+  (trigger-event state side :advancement-placed target)
+  (system-msg state side (str "sets advancement counters to " value " on "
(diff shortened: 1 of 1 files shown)


src/clj/game/core-io.clj
```diff
@@ -98,9 +103,7 @@
                                      (and (is-type? target "Agenda") (is-scored? target)) :agenda
                                      (and (card-is? target :side :runner) (has-subtype? target "Virus")) :virus)
                         advance (= :advance-counter c-type)]
-                    (cond advance (do (set-prop state side target :advance-counter value)
-                                      (system-msg state side (str "sets advancement counters to " value " on "
-                                                                  (card-str state target))))
+                    (cond advance (do (set-adv-counter state side target value))
```
Saintis: The `do` statement is not needed here since you are only doing a single thing.

src/clj/game/core-io.clj
```diff
@@ -81,10 +81,15 @@
 
 
 ; In-game chat commands
```
Saintis: While you are editing this file, please throw a two more `;` here (so it starts with `;;; `). Makes it a nice header-like comment.

src/clj/game/core-io.clj
```diff
@@ -81,10 +81,15 @@
...
+(defn set-adv-counter [state side target value]
+  (set-prop state side target :advance-counter value)
+  (trigger-event state side :advancement-placed target)
+  (system-msg state side (str "sets advancement counters to " value " on "
```
Saintis: I would prefer that the `system-msg` comes before `trigger-event`, otherwise your messages might end up out of order (for example if we add a message to morph ice).

src/clj/game/core-io.clj
erbridge: Sure thing :).

src/clj/game/core-io.clj
erbridge: Good point. So events trigger immediately when triggered, then? There's no concept of an update loop, or something? Good to know.

src/clj/game/core-io.clj
erbridge: Excuse Clojure newness :).

erbridge:
Should be resolved now.

src/clj/game/core-io.clj
nealterrell: Triggering an event notifies all listening cards one at a time. Each card's ability is resolved before moving to the next. "Resolving" an ability either causes an immediate change to the logical game state (what the player considers to be "the state of the game": credits, cards, points, etc.), or inserts a prompt into a queue with instructions on what to do when the prompt is handled by the player. 

The "update loop" equivalent is: player performs some action on client -> message sent to Node server -> message forwarded to game server -> message parsed, game actions invoked, new game state calculated -> state returned to Node server -> state returned to clients, which render the new state.

erbridge: Thanks. That's clear now.

nealterrell:
👍

PR #1995, opened 2016-09-24
Make Snitch confirm when the runner has decided not to jack out

erbridge:
Adds the following message when "No" is selected at the "Jack Out?" prompt:

> username uses Snitch to continue.

### DIFF
--- a/src/clj/game/cards-programs.clj
+++ b/src/clj/game/cards-programs.clj
@@ -705,7 +705,7 @@
                                                 {:optional {:prompt "Jack out?"
                                                             :yes-ability {:msg "jack out"
                                                                           :effect (effect (jack-out nil))}
-                                                            :no-ability {:msg "continue"}}}
+                                                            :no-ability {:msg "continue the run"}}}
                                                 card nil)))}]}
 
    "Surfer"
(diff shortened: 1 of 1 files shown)


src/clj/game/cards-programs.clj
```diff
@@ -704,7 +704,8 @@
                                                 state side
                                                 {:optional {:prompt "Jack out?"
                                                             :yes-ability {:msg "jack out"
-                                                                          :effect (effect (jack-out nil))}}}
+                                                                          :effect (effect (jack-out nil))}
+                                                            :no-ability {:msg "continue"}}}
```
nealterrell: Let's make it "continue the run" and call it a day.

erbridge: Done.

kevkcc:
While we're at it, can you give Yagura a similar treatment for the no case?

erbridge:
Not sure Yagura needs it, since the runner doesn't see anything (unless I'm mistaken). The issue I wanted to resolve with Snitch is that the corp needs to be told that the runner has decided what to do. With Yagura, the fact that they can now continue is sufficient, I think

Issue #1695, opened 2016-06-28
Political Dealings + Merger (+ Palana Agroplex?)


atreios:
When you try to trigger Political Dealings after drawing Merger, this agenda is not listed as an option for the install, whether you drew it alone or with other agendas. If you draw it with other agendas, those others will be shown in the list, but Merger will not. 

In my present case, I was using 2x Palana Agroplex to draw 3 cards at the start of every turn, and only saw the bug on the agroplex draws, though other means of drawing may also be affected.

JoelCFC25:
Fascinating--I've confirmed this one and I'm totally stumped by it! Any drawn agendas that don't include Merger and Political Dealings will show them just fine. Thanks for the report!

JoelCFC25:
OK, I see the issue now. Political Dealings listens for the `:corp-draw` event, and then scans the `n` rightmost cards in the Corp's hand (where `n` is the number of cards just drawn) to see if any agendas are in there. In the case of an Agroplex where you allow it to fire automatically instead of triggering it manually before your mandatory draw, the system is doing `(draw 1)` twice instead of `(draw 2)`. So if Merger is the top card, the opportunity for Political Dealings to see it is lost when the second `(draw 1)` occurs. 

The workaround here is to always fire your Agroplexes manually prior to your mandatory draw if you have Political Dealings in play.

queueseven:
`Whenever you draw` trigger are quite tricky. A lot of cards don't work correctly (DBS and SAU come to my mind). I don't have a good solution to this at the moment.

PR #2001, opened 2016-09-25
Daily Business Show + Political Dealings upgrades

### LINKED ISSUE #1450: Daily Business + Political dealings
When drawing extra cards with Daily Business, Political Dealings doesn't recognize anything after the first card as valid to reveal and install. If its the first on daily i was ok. any other card it doesn't register.

kevkcc:
This is part 2 of my updates to draw cards. See #1923 for part 1.
Fixes #1450, #1695 

Thanks to @nealterrell for some help he gave.

This PR upgrades Daily Business Show to use the draw function instead of take x from top of deck. Political Dealings was also reworked to fix some lingering issues and to more closely follow the rules since apparently you can install two agendas off of the DBS draw, triggering turtlebacks, and then put one on bottom of R&D (see http://ancur.wikia.com/wiki/Democracy_and_Dogma_UFAQ)

Additions to the engine:

- draw-bonus function that registers a bonus of n draws to the next draw attempt.

- :pre-corp-draw and :post-corp-draw event hooks (pre-corp-draw signals an attempt to draw, the draw happens, corp-draw is called, post-corp-draw happens after all listeners on corp-draw are finished)

- [side :register :most-recent-drawn] contains the cards in the most recent draw attempt (will be 1 card most of the time but can be 3 from Sensie or 0 if all draws prevented). This seems better than take n from the rightmost part of your hand as done previously.

Lingering issues:

- Really complex scenarios such as using Sensie, DBS and Political Dealings simultaneously make the prompts go a bit awry (you can see below that Political Dealings doesn't fully resolve before the Sensie prompt comes up again - I was unable to fix this):

test uses Sensie Actors Union to draw 3 cards.
test uses Daily Business Show to draw additional cards.
test uses Political Dealings to reveal Chronos Project.
*test uses Sensie Actors Union to add 1 card from HQ to the bottom of R&D.
test installs Chronos Project in Server 4 (new remote).
test uses Daily Business Show to add 1 card to the bottom of R&D.

- Not even entirely sure this is the best approach but thoughts would be appreciated!

### DIFF
--- a/src/clj/game/cards-assets.clj
+++ b/src/clj/game/cards-assets.clj
@@ -246,18 +246,13 @@
               :effect (req (let [dbs (count (filter #(and (rezzed? %) (= (:title %) "Daily Business Show"))
                                                     (all-installed state :corp)))]
                              (draw-bonus state side dbs)))}
-             :corp-draw
-             {:once :per-turn
-              :once-key :daily-business-show-register-draws
-              :effect (req (let [drawn (take-last (get-in @state [:corp :register :drawn-this-turn]) (:hand corp))]
-                             (swap! state assoc-in [:corp :register :dbs-drawn] drawn)))}
              :post-corp-draw
              {:once :per-turn
               :once-key :daily-business-show-put-bottom
               :delayed-completion true
               :effect (req (let [dbs (count (filter #(and (rezzed? %) (= (:title %) "Daily Business Show"))
                                                     (all-installed state :corp)))
-                                 drawn (get-in @state [:corp :register :dbs-drawn])]
+                                 drawn (get-in @state [:corp :register :most-recent-drawn])]
                              (continue-ability
                                state side
                                {:prompt (str "Choose " dbs " card" (if (> dbs 1) "s" "") " to add to the bottom of R&D")
@@ -766,10 +761,10 @@
      {:events
       {:corp-draw
        {:delayed-completion true
-        :req (req (let [drawn (take-last target (:hand corp))
+        :req (req (let [drawn (get-in @state [:corp :register :most-recent-drawn])
                         agendas (filter #(is-type? % "Agenda") drawn)]
                     (seq agendas)))
-        :effect (req (let [drawn (take-last target (:hand corp))
+        :effect (req (let [drawn (get-in @state [:corp :register :most-recent-drawn])
                            agendas (filter #(is-type? % "Agenda") drawn)]
                        (continue-ability state side (pdhelper agendas 0) card nil)))}}})
 
--- a/src/clj/game/cards-assets.clj
+++ b/src/clj/game/cards-assets.clj
@@ -238,28 +238,27 @@
    {:in-play [:hand-size-modification 4]}
 
    "Daily Business Show"
-   {:events {:corp-draw
+   {:events {:pre-corp-draw
              {:msg "draw additional cards"
               :once :per-turn
               :once-key :daily-business-show
-              :req (req (first-event state side :corp-draw))
-              :effect (req
-                        (let [dbs (count (filter #(and (rezzed? %) (= (:title %) "Daily Business Show"))
-                                                  (all-installed state :corp)))
-                              newcards (take dbs (:deck corp))
-                              drawn (concat newcards (take-last target (:hand corp)))]
-                          (if (not= (count newcards) dbs)
-                            ; couldn't draw from R&D, so corp is decked
-                            (win-decked state)
-                            (do (doseq [c newcards] (move state side c :hand))
-                                (resolve-ability
-                                  state side
-                                  {:prompt (str "Choose " dbs " card" (if (> dbs 1) "s" "") " to add to the bottom of R&D")
-                                   :choices {:max dbs
-                                             :req #(and (in-hand? %)
-                                                        (some (fn [c] (= (:cid c) (:cid %))) drawn))}
-                                   :msg (msg "add " dbs " card" (if (> dbs 1) "s" "") " to bottom of R&D")
-                                   :effect (req (doseq [c targets] (move state side c :deck)))} card targets)))))}}}
+              :req (req (first-event state side :pre-corp-draw))
+              :effect (req (let [dbs (count (filter #(and (rezzed? %) (= (:title %) "Daily Business Show"))
+                                                    (all-installed state :corp)))]
+                             (draw-bonus state side dbs)))}
+             :corp-draw
+             {:delayed-completion true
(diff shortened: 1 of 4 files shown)


src/clj/game/cards-assets.clj
```diff
@@ -741,22 +741,32 @@
...
+   (let [pdhelper (fn pd [agendas n]
+                    {:optional
+                     {:prompt (msg "Reveal and install " (:title (nth agendas n)) "?")
+                      :yes-ability {:delayed-completion true
+                                    :msg (msg "reveal " (:title (nth agendas n)))
+                                    :effect (req (when-completed
+                                                   (corp-install state side (nth agendas n) nil
+                                                                 {:install-state
+                                                                  (or (:install-state (card-def (nth agendas n)))
+                                                                      :rezzed-no-cost)})
```
nealterrell: Instead of the `or` here, just pass `:rezzed-no-cost` as the second argument to `:install-state` as `(:install-state (card-def ...) :rezzed-no-cost)`. When you use a keyword as a lookup function, an optional second argument will be used as the value if the result would otherwise be `nil`.

src/clj/game/cards-assets.clj
```diff
@@ -238,28 +238,28 @@
...
-                                  state side
-                                  {:prompt (str "Choose " dbs " card" (if (> dbs 1) "s" "") " to add to the bottom of R&D")
-                                   :choices {:max dbs
-                                             :req #(and (in-hand? %)
-                                                        (some (fn [c] (= (:cid c) (:cid %))) drawn))}
-                                   :msg (msg "add " dbs " card" (if (> dbs 1) "s" "") " to bottom of R&D")
-                                   :effect (req (doseq [c targets] (move state side c :deck)))} card targets)))))}}}
+              :once-key :daily-business-show-draw-bonus
+              :req (req (first-event state side :pre-corp-draw))
+              :effect (req (let [dbs (count (filter #(and (rezzed? %) (= (:title %) "Daily Business Show"))
```
nealterrell: I know this comparison with the card's title is legacy, but perhaps we should consider using the `:code` of the card instead of the `:title`, which will be a faster string comparison. The `:code` for DBS is `"06086"`, from the URL to the card on NetrunnerDB: https://netrunnerdb.com/en/card/06086

src/clj/game/core-rules.clj
```diff
@@ -65,30 +65,42 @@
...
+  "Registers a bonus of n draws to the next draw (Daily Business Show)"
+  [state side n]
+  (swap! state update-in [:bonus :draw] (fnil #(+ % n) 0)))
+
 (defn draw
   "Draw n cards from :deck to :hand."
   ([state side] (draw state side 1 nil))
   ([state side n] (draw state side n nil))
   ([state side n {:keys [suppress-event] :as args}]
+   (swap! state assoc-in [side :register :most-recent-drawn] nil) ;clear the most recent draw in case draw prevented
```
nealterrell: FYI (not necessary to implement), although there is no `dissoc-in` in Clojure, you can mimic it with `(swap! state assoc-in [side :register] dissoc :most-recent-drawn)`.

src/clj/game/core-rules.clj
```diff
@@ -65,30 +65,42 @@
...
+         (swap! state assoc-in [side :register :most-recent-drawn] drawn)
+         (swap! state update-in [side :register :drawn-this-turn] (fnil #(+ % draws-after-prevent) 0))
+         (swap! state update-in [:bonus] dissoc :draw)
          (when (and (not suppress-event) (pos? deck-count))
-           (trigger-event state side (if (= side :corp) :corp-draw :runner-draw) n))
+           (when-completed
+             (trigger-event-sync state side (if (= side :corp) :corp-draw :runner-draw) draws-after-prevent)
+             (trigger-event state side (if (= side :corp) :post-corp-draw :post-runner-draw) draws-after-prevent)))
          (when (= 0 (remaining-draws state side))
            (prevent-draw state side))))
```
nealterrell: I don't know if this is relevant, but be careful here. This `when` and anything that follows will not wait for the `:corp-draw` or the `:post-corp-draw` events to finish resolving before executing. I don't know what it's for or if it matters, but it's something to keep in mind when using `when-completed` (which you otherwise nailed).

src/clj/game/cards-assets.clj
```diff
@@ -238,28 +238,28 @@
...
+             {:once :per-turn
+              :once-key :daily-business-show-put-bottom
+              :delayed-completion true
+              :effect (req (let [dbs (count (filter #(and (rezzed? %) (= (:title %) "Daily Business Show"))
+                                                    (all-installed state :corp)))
+                                 drawn (get-in @state [:corp :register :most-recent-drawn])]
+                             (continue-ability
+                               state side
+                               {:prompt (str "Choose " dbs " card" (if (> dbs 1) "s" "") " to add to the bottom of R&D")
+                                :msg (msg "add " dbs " card" (if (> dbs 1) "s" "") " to the bottom of R&D")
```
Saintis: These `(if ? "s" "")` could be `(when ? "s")` instead.

src/clj/test/cards/assets.clj
```diff
@@ -157,6 +157,77 @@
...
+      (prompt-select :corp (find-card "Hedge Fund" (:hand (get-corp)))) ; Sensie target
+      (is (= 3 (count (:hand (get-corp)))))
+      (is (= "Hedge Fund" (:title (last (:deck (get-corp))))) "Hedge Fund last card in deck")
+      (is (= "Resistor" (:title (last (butlast (:deck (get-corp))))))
+          "Resistor second last card in deck")
+      ; Try to use first Sensie again
+      (card-ability state :corp sensie1 0)
+      (is (empty? (get-in @state [:corp :prompt])) "Sensie didn't activate")
+      (is (= 3 (count (:hand (get-corp)))))
+      ; Use second Sensie
```
Saintis: Pet peeve alert! :wink: - comments should be `;;` if single-line comments, `;` is for end of line comments.

src/clj/test/cards/assets.clj
```diff
@@ -671,31 +742,65 @@
     (is (find-card "Braintrust" (:scored (get-corp))) "Braintrust is scored")))
 
 (deftest political-dealings
+  "Political Dealings - Full test"
```
Saintis: I agree with Neal that we should make these test comments use `;;` instead of strings as `deftest` doesn't actually support `doc-string`s (see comment [here](https://github.com/mtgred/netrunner/pull/2002#discussion_r80395634)).

Saintis:
Good stuff! Made some additional, style related, comments.

nealterrell:
👍

PR #2003, opened 2016-09-26
Minor fixes

### LINKED ISSUE #1257: Add a warning when you're about to install a program that you don't have memory for
At the moment one can see a warning next to the memory units when there is not enough MU after the install, however, undoing such missclicks can be a pain. It would be nice if a confirmation prompt showed before installing a program that exceeds your available memory units.

### LINKED ISSUE #1774: Mumbad Virtual Tour + NBN: Controlling the Message
If the runner accesses an installed, unrezzed Mumbad Virtual Tour and is able to trash it, they don't see what the card they accessed was or that they were forced to trash it before NBN: Controlling the Message triggers and initiates a trace.

### LINKED ISSUE #1927: Uniqueness rez/install should prompt to trash, not prevent the action
If the runner already has a unique non-console installed, Jinteki.net currently prevents them from installing another copy of that card entirely. This isn't exactly correct -- players are allowed to install new copies of unique cards, but upon doing so, all existing copies of that card that are installed are trashed.

Similarly, if the corp has a unique card rezzed, other copies of that card are entirely prevented from being rezzed unless the unique card is trashed. In reality, players are allowed to rez multiple copies of unique cards, but upon doing so, existing copies are trashed.

This is relevant especially now that Temujin Contracts exists, as many players who aren't familiar with the rules of unique cards assume that once you install a Temujin, it's stuck there until it's empty, not realizing that they are, in fact, allowed to install another one and trash the previous one. Similarly on the Corp side, new players might not realize that you can rez another copy of Caprice Nisei, trashing an existing one, if needed.

My suggestion is to provide the player with a confirmation prompt when they attempt to install another copy of a unique card, or rez another copy of a unique card, rather than prevent the install/rez altogether and assume the player knows they are allowed to trash the existing ones.

### LINKED ISSUE #1977: Heartbeat and trashed due to damage message
When Heartbeat is used to prevent damage, and a face down card is trashed, the card's name is not displayed in the following trashed due to damage message (or this is the same issue as #1969 and Heartbeat needs a message that lists what cards it used to pay for its ability, including face down cards).

image

Saintis:
A collection of fixes:

- Specifies facedown card trashed with Heartbeat, fixes #1977 
- Toast when you install a programming making your MU negative, fixes #1257
- Improves Mumbad Virtual Tour, now also does an info toast on forced trash, fixes #1774
- Toasts on install and rez restrictions, fixes #1927

Screenshots:
Heartbeat
screen shot 2016-09-26 at 18 09 47

MU Toast
screen shot 2016-09-26 at 18 23 18

Mumbad Virtual Tour
screen shot 2016-09-26 at 19 50 04
screen shot 2016-09-26 at 19 50 37

Rez / install toasts
screen shot 2016-09-26 at 21 02 44
screen shot 2016-09-26 at 21 02 59
screen shot 2016-09-26 at 21 03 10
screen shot 2016-09-26 at 21 03 22
screen shot 2016-09-26 at 21 18 31

More coming...

### DIFF
--- a/src/clj/test/cards/upgrades.clj
+++ b/src/clj/test/cards/upgrades.clj
@@ -397,7 +397,7 @@
     (is (= "Mumbad Virtual Tour" (:title (first (:discard (get-corp))))) "MVT trashed")))
 
 (deftest mumbad-virtual-tour-slow-trash
-  ;; Tests that Mumbad Virtual Tour does not force trash with no :slow-trash
+  ;; Tests that Mumbad Virtual Tour does not force trash with :slow-trash
   (do-game
     (new-game (default-corp [(qty "Mumbad Virtual Tour" 2)])
               (default-runner [(qty "Imp" 1) (qty "Salsette Slums" 1)]))
--- a/src/clj/test/cards/upgrades.clj
+++ b/src/clj/test/cards/upgrades.clj
@@ -379,6 +379,57 @@
       (is (prompt-is-card? :corp sn) "Corp prompt is on Security Nexus")
       (is (prompt-is-type? :runner :waiting) "Runner prompt is waiting for Corp"))))
 
+(deftest mumbad-virtual-tour-force-trash
+  ;; Tests that Mumbad Virtual Tour forces trash when no :slow-trash
+  (do-game
+    (new-game (default-corp [(qty "Mumbad Virtual Tour" 2)])
+              (default-runner))
+    (play-from-hand state :corp "Mumbad Virtual Tour" "New remote")
+    (take-credits state :corp)
+    (run-empty-server state "HQ")
+    ;; MVT does not force trash when not installed
+    (prompt-choice :runner "No")
+    (is (= 5 (:credit (get-runner))) "Runner not forced to trash MVT in HQ")
+    (is (empty? (:discard (get-corp))) "MVT in HQ is not trashed")
+    (run-empty-server state "Server 1")
+    ;; Toast should show at this point to notify runner they were forced to trash MVT
+    (is (= 0 (:credit (get-runner))) "Runner forced to trash MVT")
+    (is (= "Mumbad Virtual Tour" (:title (first (:discard (get-corp))))) "MVT trashed")))
+
+(deftest mumbad-virtual-tour-slow-trash
+  ;; Tests that Mumbad Virtual Tour does not force trash with no :slow-trash
+  (do-game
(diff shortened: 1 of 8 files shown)


nealterrell:
Feel like applying your MU toast to #1927? :)

erbridge:
I think the "Please trash  X first" is possibly misleading. It implies I can just trash Jackson Howard or Temujin, which I can't.

src/clj/test/cards/upgrades.clj
```diff
@@ -379,6 +379,57 @@
...
+    (prompt-choice :runner "No")
+    (is (= 5 (:credit (get-runner))) "Runner not forced to trash MVT in HQ")
+    (is (empty? (:discard (get-corp))) "MVT in HQ is not trashed")
+    (run-empty-server state "Server 1")
+    ;; Toast should show at this point to notify runner they were forced to trash MVT
+    (is (= 0 (:credit (get-runner))) "Runner forced to trash MVT")
+    (is (= "Mumbad Virtual Tour" (:title (first (:discard (get-corp))))) "MVT trashed")))
+
+(deftest mumbad-virtual-tour-slow-trash
+  ;; Tests that Mumbad Virtual Tour does not force trash with no :slow-trash
```
erbridge: I think you mean `with :slow-trash` not `with no :slow trash`.

nealterrell:
Not quite, @erbridge. From the Core rule book: "Unique Cards
Some cards have a unique symbol (◆) in front of their title. There can be only one unique card of the same title active at a time. If a card with a unique title becomes active, any other card that shares its title is immediately trashed. This trashing cannot be prevented." Here, "becomes active" means "installed" (Runner) or "rezzed" (Corp).

erbridge:
You've already installed both Jacksons. You try to rez Jackson 2, see the message, and think, ok, well I'll trash that already rezzed Jackson, then, so I can rez this better protected one (or whatever). But you can't actually do that without installing something new in that server.

nealterrell:
That actually is a valid play. The core game rules say that if a second copy of a unique card becomes active, the first copy must be trashed. "Active" for Corp means rezzed, for Runner means installed.

erbridge:
Oh wow. I didn't know that. Puts Temujin lock in a very different light. Nevermind then :).

src/clj/test/cards/upgrades.clj
Saintis: Ah, yes

PR #2004, opened 2016-09-26
Added test for Noise ID (tentative toe in the water)

kevinpowe:
Hi! This is a very tentative toe in the water to start adding test coverage to wrap my head around the codebase. I've added a test for Noise that appears to do the right thing locally.

### DIFF
--- a/src/clj/test/cards/identities.clj
+++ b/src/clj/test/cards/identities.clj
@@ -599,6 +599,21 @@
       (prompt-choice :runner "1 [Credits]")
       (is (= 5 (:credit (get-corp))) "Gained 1 credit from psi game"))))
 
+(deftest noise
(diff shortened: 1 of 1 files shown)


JoelCFC25:
It's not a *huge* thing, but it could also be extended to show that Noise will still mill the Corp during the Corp's turn (e.g., installing a virus with a Clone Chip).

kevinpowe:
@JoelCFC25 Great suggestion! I'll put that on my to-do list once I've skilled up a little more. ;-)

nealterrell:
Welcome aboard 👍

kevinpowe:
Understood re: alphabetizing, @Saintis. I'll look at the suggested enhancements and add that to the list for a second pass.

Thank you for the feedback all - makes ramping up much easier!

PR #2009, opened 2016-09-27
Convert test comments to use ;; instead of strings

kevkcc:
Padding my commit history. jk
Decided to just do this in one go so newer developers won't get confused by notation


nealterrell:
It's dirty work but somebody's gotta do it. 👍

Issue #677, opened 2015-09-06
Security update available for jws library


astrostl:
Howdy,

`npm install` yields "npm WARN deprecated jws@0.2.6: Security update: Versions below 3.0.0 are deprecated."

Just saying :)

meatcompute:
Is this warning still an issue? I did not run into this setting up on my machine, os x 10.10.5

erbridge:
I didn't notice the warning, but we're still using the insecure version:

```
$  npm list jws
netrunner@0.0.1 /home/felix/dev/netrunner
└─┬ jsonwebtoken@0.4.1
  └── jws@0.2.6
```

erbridge:
Fixed by #2011.

PR #2011, opened 2016-09-28
Update most of the JS dependencies

### LINKED ISSUE #676: Critical vulnerability in jsonwebtoken library
Howdy,

`npm install` yields "npm WARN deprecated jsonwebtoken@0.4.1: Critical vulnerability fix in v5.0.0. See https://auth0.com/blog/2015/03/31/critical-vulnerabilities-in-json-web-token-libraries/"

Just saying :)

erbridge:
This should fix a number of potential security issues. The last major outdated dependency is `socket.io`, and their docs suck, so I'll probably look at that for a different PR. The upgrade issue is related to `socket.request.user` being reset between socket events (and so not sending the user to the cljs).

I've tested these changes by playing a few local games, and clicking most of the bits of UI, but it's possible that I've missed something, so if other people could test it, that would be good.

Fixes #676 and #677.


erbridge:
Oh. Looks like CircleCI is using an ancient version of Node... It should to be upgraded to 4.x or later. Also, it needs to have the `node_modules` cleared, so it will reinstall the correct version of `mongodb`.

Saintis:
Seems that we just need to specify the Node version in circle.yml as suggested [here](https://circleci.com/docs/build-image-precise/#nodejs)

Saintis:
Well that seems to have solved the Node issue, but the `mongodb` issue still persists. Also seems like `node-trello` has some deep dependency on `async ~0.9.0`...

Saintis:
Yay! That works!

Saintis:
Maybe do #2012 while you (we) are at it ;)

Saintis:
Did a minimal test locally. Seems to work well.

erbridge:
Given that this has the potential to majorly break the app, my inclination is to keep the changeset to the minimum. If you feel that #2012 is a requirement (give that pug hasn't actually had any releases since the last jade release), then I can look at doing that, too.

Saintis:
Nah, just general _update everything!_ feels. Did it locally, seemed straight-forward. But probably safest to leave it.

erbridge:
The argument the other way, is that if we're going to need to test it thoroughly for this, then maybe it's best to lump them all in together to minimize testing burden...

PR #2014, opened 2016-09-28
Be more precise in messages about cards that become agendas

erbridge:
Possibly overkill, but I'm a pedant :).

### DIFF
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -46,6 +46,13 @@
      :events {:corp-turn-begins ability}
      :abilities [ability]}))
 
+(defn as-trashed-agenda
+  "Adds the given card to the given side's :scored area as an agenda worth n points after resolving the trash prompt."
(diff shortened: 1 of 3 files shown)


src/clj/game/cards/assets.clj
```diff
@@ -797,7 +805,7 @@
     :events {:corp-turn-begins
              {:effect (req (add-counter state side card :power -1)
                            (when (<= (get-in card [:counter :power]) 1)
-                             (system-msg state :corp "adds Public Support to his scored area and gains 1 agenda point")
+                             (system-msg state :corp "uses Public Support to add it to their score area as an agenda worth 1 agenda point")
```
zaroth: regarding his->their, how (in)consistent is it currently in code? Maybe it would be worth it doing a search & replace sweep?

erbridge: I think that was the only instance.

mtgred:
👍

PR #2015, opened 2016-09-29
stop Targeted Marketing from paying out if a named Runner Current is played

JoelCFC25:
Fixes issue raised in Stimslack:

Targeted Marketing is wrongly paying out 10cr if it names a Runner Current and that Current gets played (see official FAQ page 11). Targeted Marketing is supposed to be trashed immediately before it has a chance to trigger. This just postpones `:play-operation` or `:play-event` until after the Current trash is done.


mtgred:
👍

PR #2019, opened 2016-09-29
Indicate whether trashed cards were inside a central or protecting it

erbridge:
The result is "pays 4 [Credits] to trash PAD Campaign from HQ" vs "pays 1 [Credits] to trash Cyberdex Virus Suite from HQ Server".

### DIFF
--- a/src/clj/game/core/runs.clj
+++ b/src/clj/game/core/runs.clj
@@ -125,9 +125,11 @@
              {:prompt (str "Pay " trash-cost "[Credits] to trash " name "?")
               :yes-ability {:cost [:credit trash-cost]
                             :delayed-completion true
-                            :effect (effect (trash eid card nil)
-                                            (system-msg (str "pays " trash-cost
-                                                             " [Credits] to trash " name)))}}}
+                            :effect (req (trash state side eid card nil)
+                                         (let [src (name-zone :corp (:zone card))]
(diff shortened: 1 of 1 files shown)


src/clj/game/core/runs.clj
```diff
@@ -125,9 +125,11 @@
...
               :yes-ability {:cost [:credit trash-cost]
                             :delayed-completion true
-                            :effect (effect (trash eid card nil)
-                                            (system-msg (str "pays " trash-cost
-                                                             " [Credits] to trash " name)))}}}
+                            :effect (req (trash state side eid card nil)
+                                         (let [src (name-zone :corp (:zone card))]
+                                           (system-msg state side (str "pays " trash-cost
+                                                                       " [Credits] to trash " name
+                                                                       (when-not (nil? src) (str " from " src))))))}}}
```
Saintis: I think `"in the root of" src` would be more correct for the centrals. But this way makes it compatible with remote server right?

erbridge: Currently it only works for centrals. `name-zone` returns `HQ` for cards in HQ and `HQ Server` for cards install in the root of HQ. Remotes appear to return nil. I don't know if it's right to alter `name-zone`, or whether to do something more custom.

erbridge:
Closing in favour of #2064.

PR #2020, opened 2016-09-29
ELP implementation

mattchainsaw:
I was not completely sure on a more elegant approach to writing 

`:req (req (not (has-subtype? (get (get (get @state :run) :run-effect) :card) "Run")))`

but was unsure as to how to tell is a run is started from an Event or not. It works though.
Also, my test suite for the card was a little funky. I originally (at the very end of the test) had
```
(deftest
...
(is (= 1 (:click (get-runner))))
(is (not (core/can-run-server? state "HQ")) "Runner can't run - needs two clicks")))
```
but I was getting failures even though it worked in practice. The failure said it was `(not (not true))`, which I wanted it to be. But it didn't evaluate all the way to true. I'm still new to clojure, so if anyone knows why this happened, please let me know. My replacement simply runs and jacks out and asserts that the clicks left for the runner is still at 1, meaning the run was impossible to do and didn't happen.

### DIFF
--- a/src/clj/game/cards/operations.clj
+++ b/src/clj/game/cards/operations.clj
@@ -294,13 +294,13 @@
    "Enhanced Login Protocol"
    {:events {:run {:once :per-turn
                    ;; TODO: Find more elegant way of seeing if its a run event
-                   :req (req (not (has-subtype? (get (get (get @state :run) :run-effect) :card) "Run")))
+                   :req (req (nil? (:card (:run-effect (:run @state)))))
                    :effect (effect (update! (assoc card :elp-fire true))
-                             (lose :runner :click 1))
+                                   (lose :runner :click 1))
                    :msg "force the Runner to spend an additional [Click]"}
              :runner-spent-click {:req (req (and (not (:elp-fire card)) (>= 2 (:click runner))))
                                   :effect (effect (prevent-run))}
-             :runner-phase-12 {:effect (effect (update! (assoc card :elp-fire (not true))))}}}
+             :runner-phase-12 {:effect (effect (update! (assoc card :elp-fire false)))}}}
 
    "Exchange of Information"
    {:req (req (and tagged
--- a/src/clj/game/cards/operations.clj
+++ b/src/clj/game/cards/operations.clj
@@ -291,6 +291,17 @@
                               :effect (effect (trash target))}
                             card nil)))}}
 
+   "Enhanced Login Protocol"
+   {:events {:run {:once :per-turn
+                   ;; TODO: Find more elegant way of seeing if its a run event
+                   :req (req (not (has-subtype? (get (get (get @state :run) :run-effect) :card) "Run")))
+                   :effect (effect (update! (assoc card :elp-fire true))
+                             (lose :runner :click 1))
+                   :msg "force the Runner to spend an additional [Click]"}
+             :runner-spent-click {:req (req (and (not (:elp-fire card)) (>= 2 (:click runner))))
+                                  :effect (effect (prevent-run))}
+             :runner-phase-12 {:effect (effect (update! (assoc card :elp-fire (not true))))}}}
+
    "Exchange of Information"
    {:req (req (and tagged
                    (seq (:scored runner))
--- a/src/clj/test/cards/operations.clj
+++ b/src/clj/test/cards/operations.clj
@@ -315,8 +315,8 @@
     (take-credits state :corp)
     (play-from-hand state :runner "Keyhole")
     (is (= 3 (:click (get-runner))) "Runner has 3 clicks before run")
-    (let [sneak (get-in @state [:runner :rig :program 0])]
-      (card-ability state :runner sneak 0)
+    (let [keyhole (get-in @state [:runner :rig :program 0])]
+      (card-ability state :runner keyhole 0)
       (is (= 2 (:click (get-runner))) "Runner has 2 clicks after using sneakdoor"))))
 
 (deftest exchange-of-information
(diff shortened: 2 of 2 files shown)


erbridge:
A result of `(not (not true))` is telling you that `(not (core/can-run-server? state "HQ"))` is evaluating to `(not true)`. As a simpler example, `(is (= 1 2))` gives the following test output:

```
expected: (= 1 2)
  actual: (not (= 1 2))
```

As for your question of simplifying, I'm pretty sure you can convert `(get (get (get @state :run) :run-effect) :card)` into `(:card (:run-effect (:run state)))` (but I'm also a Clojure newbie).

src/clj/game/cards/operations.clj
```diff
@@ -291,6 +291,17 @@
...
+   "Enhanced Login Protocol"
+   {:events {:run {:once :per-turn
+                   ;; TODO: Find more elegant way of seeing if its a run event
+                   :req (req (not (has-subtype? (get (get (get @state :run) :run-effect) :card) "Run")))
+                   :effect (effect (update! (assoc card :elp-fire true))
+                             (lose :runner :click 1))
+                   :msg "force the Runner to spend an additional [Click]"}
+             :runner-spent-click {:req (req (and (not (:elp-fire card)) (>= 2 (:click runner))))
+                                  :effect (effect (prevent-run))}
+             :runner-phase-12 {:effect (effect (update! (assoc card :elp-fire (not true))))}}}
```
erbridge: `(not true) = false`

src/clj/test/cards/operations.clj
```diff
@@ -273,6 +273,40 @@
...
+    (run-jack-out state)
+    (is (= 2 (:click (get-runner))) "first run takes 2 clicks")
+    (run-on state :hq)
+    (run-jack-out state)
+    (is (= 1 (:click (get-runner))) "second run takes 1 click")
+    (take-credits state :runner)
+    (take-credits state :corp)
+    (take-credits state :runner 2)
+    (is (= 2 (:click (get-runner))))
+    (is (boolean (core/can-run-server? state "HQ")) "Runner can run - two clicks")
```
erbridge: `boolean` is unnecessary here. `is` looks for truthiness, so implicitly casts to a `boolean`.

src/clj/game/cards/operations.clj
```diff
@@ -291,6 +291,17 @@
                               :effect (effect (trash target))}
                             card nil)))}}
 
+   "Enhanced Login Protocol"
+   {:events {:run {:once :per-turn
+                   ;; TODO: Find more elegant way of seeing if its a run event
+                   :req (req (not (has-subtype? (get (get (get @state :run) :run-effect) :card) "Run")))
+                   :effect (effect (update! (assoc card :elp-fire true))
+                             (lose :runner :click 1))
```
erbridge: The indentation style is to align arguments after a line break, so this is missing 6 spaces.

src/clj/test/cards/operations.clj
```diff
@@ -273,6 +273,40 @@
...
+    (play-from-hand state :corp "Enhanced Login Protocol")
+    (take-credits state :corp)
+    (is (= 4 (:click (get-runner))) "Runner starts turn with 4 clicks")
+    (play-from-hand state :runner "Dirty Laundry")
+    (prompt-choice :runner "Archives")
+    (run-successful state)
+    (is (= 3 (:click (get-runner))) "Run event only takes 1 click")
+    (run-on state :hq)
+    (run-jack-out state)
+    (is (= 1 (:click (get-runner))) "first non-event run takes 2 clicks")
```
erbridge: Better to check this happens immediately after the run begins, not after it ends, in case ice allows spending clicks.

kevkcc:
What happens when you run with keyhole or sneakdoor beta?

src/clj/test/cards/operations.clj
```diff
@@ -273,6 +273,40 @@
...
+    (is (= 1 (:click (get-runner))) "second run takes 1 click")
+    (take-credits state :runner)
+    (take-credits state :corp)
+    (take-credits state :runner 2)
+    (is (= 2 (:click (get-runner))))
+    (is (boolean (core/can-run-server? state "HQ")) "Runner can run - two clicks")
+    (take-credits state :runner 1)
+    (is (= 1 (:click (get-runner))))
+    (run-on state :hq)
+    (run-jack-out state)
```
erbridge: Would be good to check that the run was actually prevented, by e.g. doing the `core/can-run-server?` check again before making the run.

mattchainsaw:
@kevkcc Didn't even think about programs making runs! I'll fix it up and write another test.

src/clj/test/cards/operations.clj
mattchainsaw: Since ELP doesn't prevent a run on a specific server, that function does not correctly tell if the run is viable. I looked into how Excaliber was tested and will change to that as it makes more since.

zaroth:
It would be neat if the "make a run" button was greyed out or disappeared when you don't have enough clicks for it, similar to what happens to "purge viruses" button for Corp having less than 3 clicks.

mattchainsaw:
@zaroth The current implementation greys out the run ability if you don't have enough clicks.

src/clj/game/cards/operations.clj
```diff
@@ -291,6 +291,17 @@
                               :effect (effect (trash target))}
                             card nil)))}}
 
+   "Enhanced Login Protocol"
+   {:events {:run {:once :per-turn
+                   ;; TODO: Find more elegant way of seeing if its a run event
+                   :req (req (nil? (:card (:run-effect (:run @state)))))
```
Saintis: can also use `(get-in @state [:run :run-effect :card])`

PR #2021, opened 2016-09-30
Improve some of the styling

erbridge:
- Make the game count orange to distinguish it from links in the navbar (works as grey, too).
- Wrap an `<il>` in a `<ul>` as it should be, and simplify the navbar styles (no visual changes).
- Tidy up indentation and spacing in the stylus files.

### DIFF
--- a/src/cljs/netrunner/auth.cljs
+++ b/src/cljs/netrunner/auth.cljs
@@ -21,13 +21,14 @@
 (defn logged-menu [user owner]
   (om/component
    (sab/html
-    [:li.dropdown.usermenu
-     [:a.dropdown-toggle {:href "" :data-toggle "dropdown"}
-      (om/build avatar user {:key :username :opts {:size 22}})
-      (:username user)
-      [:b.caret]]
-     [:div.dropdown-menu.blue-shade.float-right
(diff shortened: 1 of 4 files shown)


nealterrell:
Can you post before & after screenshots?

erbridge:
Sure, but the only change is "0 Games" goes from white to orange.

zaroth:
when changing UI, quick visual appraisal is everything ;-)

erbridge:
Sure. Not objecting :).

erbridge:
Before:
screenshot from 2016-09-30 18-17-20

After:
screenshot from 2016-09-30 18-15-23

zaroth:
I like the styling cleanup, but to be honest I'm not too convinced by the orange games count indicator... I think it's more confusing this way (by looking like either a hovered real link or a fake link (which also does something on hover).

erbridge:
Yeah. We don't have a huge amount of variety in colours, though. Could go with grey?

erbridge:
screenshot from 2016-10-12 18-17-24

Saintis:
I like how sad it looks when there is only 1 game.

PR #2022, opened 2016-09-30
Tricksy fixes

### LINKED ISSUE #1916: Khan and Caprice bug
This is a niche one given the lack of devoted Khan players but here goes:

When Khan's ability is triggered on a server with one piece of ICE and a rezzed Caprice (ie so both Kahn and Caprice trigger at the same time) it can cause a bug. Caprice's psi game will trigger but the runner will be given a successful run even if the corp wins the psi game. 

I forgot to get a screenshot of this but will do so if I run into it again.

### LINKED ISSUE #2008: Leela Patel and upgrades
On multiple occasions now, I have stolen an agenda during an access on a central server and bounced an unrezzed upgrade on the server that I am currently accessing, but when I finish the central accesses, it still wants me to access the now bounced upgrade. Unless I /close-prompt I cannot exit the run windows.

nealterrell:
Fix #1916 by making the `:pass-ice` event synchronous, so Khan's install prompt has to resolve before Caprice is triggered.

Fix #2008 by reworking central-server access so that upgrades which are moved in the middle of the access phase via Leela are do not show up in the remaining "Choose a card to access" prompts.

### DIFF
--- a/src/clj/game/cards/identities.clj
+++ b/src/clj/game/cards/identities.clj
@@ -223,19 +223,22 @@
    "Haas-Bioroid: Architects of Tomorrow"
    {:events {:pass-ice
              {:once :per-turn
+              :delayed-completion true
               :req (req (and (rezzed? target)
                              (has-subtype? target "Bioroid")))
-              :effect (req (when (some #(and (has-subtype? % "Bioroid") (not (rezzed? %))) (all-installed state :corp))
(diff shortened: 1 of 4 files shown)


FarCryFromHuman:
Seems like this is a piece of the puzzle you'd need to get cards like Mongoose and Brahman a little more automated as well?

mtgred:
👍

PR #2023, opened 2016-09-30
Duplicate ELP implementation

mattchainsaw:
I took the advice of @Saintis in #2020 and added a bonus cost to the `click-run` action. I used the same tests. The only noticable difference in the UI is that in #2020 the action of clicking Run in the game is greyed out if you can not pay for it. In this implementation, if you click on run and can not pay for it, a notice appears alerting you that you do not have the ability to pay for a run.

I was a little unsure as to what the msg for the card should be in this case, as when the Runner clicks to run, the prompt says that they spend 2 clicks.


JoelCFC25:
The newly spoiled Service Outage current in the Escalation pack imposes a 1 credit additional cost on the first run of any kind each turn--using a click or a card ability.

zaroth:
So what happens when Adam with Always Be Running has 0 credits at the beginning on his turn with Service Outage current active? Checkmate, bioroids!

Saintis:
Additional cost so can decline to make a run maybe?

erbridge:
That's how additional scosts work in general, if I have it right. Same as Gagarin remote access. You cannot be forced to spend the additional cost, meaning you can avoid the directive.

src/clj/game/core/actions.clj
```diff
@@ -343,14 +343,17 @@
...
+  (trigger-event state side :pre-click-run)
+  (let [bonus (get-in @state [:bonus :click-run])]
+    (when (and (not (get-in @state [:runner :register :cannot-run]))
+               (can-run-server? state server)
+               (can-pay? state :runner "a run" :click 1 bonus))
+      (run state side server)
+      (let [cost (pay state :runner nil :click 1 bonus)
+            spent (build-spend-msg cost "make a run on")
+            message (str spent server)]
+        (clear-click-run-bonus state side)
```
erbridge: Shouldn't cards be responsible for removing the bonus themselves?

src/clj/test/cards/operations.clj
```diff
@@ -273,6 +273,52 @@
...
+           (run-jack-out state)
+           (take-credits state :runner)
+           (take-credits state :corp)
+           (take-credits state :runner 2)
+           (is (= 2 (:click (get-runner))))
+           (is (core/can-run-server? state "HQ") "Runner can run - two clicks")
+           (take-credits state :runner 1)
+           (is (= 1 (:click (get-runner))))
+           (run-on state :hq)
+           (is (not (:run @state)) "No run initiated - not enough clicks")))
```
erbridge: Does the runner still have a click after this?

erbridge:
This has merge conflicts after #2071.

src/clj/game/cards/operations.clj
```diff
@@ -291,6 +291,11 @@
                               :effect (effect (trash target))}
                             card nil)))}}
 
+   "Enhanced Login Protocol"
+   {:events {:pre-click-run {:once :per-turn
+                             :msg "increase the cost of running"
+                             :effect (effect (click-run-bonus [:click 1]))}}}
```
erbridge: It might be worth taking a look at how #2071 implemented Service Outage (though obviously only add additional costs to click runs). I made Service Outage responsible for removing the bonus it added.

src/clj/game/core/costs.clj
```diff
@@ -123,3 +123,9 @@
      (swap! state update-in [:bonus] dissoc :install-cost)
      (swap! state update-in [:bonus] dissoc :ignore-install-cost)
      cost)))
+
+(defn click-run-bonus [state side & n]
+  (swap! state update-in [:bonus :click-run] #(merge-costs (concat % n))))
```
erbridge: Probably best to call the key `:click-run-cost` for clarity.

src/clj/test/cards/operations.clj
```diff
@@ -273,6 +273,52 @@
     (play-from-hand state :corp "Diversified Portfolio")
     (is (= 7 (:credit (get-corp))) "Ignored remote with ICE but no server contents")))
 
+(deftest enhanced-login-protocol
+         ;; Enhanced Login Protocol - tests interaction with run events, interaction with runs, and inability to run last click
+         (do-game
```
erbridge: Indentation for both tests is too deep. Should be 2 spaces.

src/clj/test/cards/operations.clj
```diff
@@ -273,6 +273,52 @@
...
+           (new-game (default-corp [(qty "Enhanced Login Protocol" 1)])
+                     (default-runner [(qty "Keyhole" 1)]))
+           (play-from-hand state :corp "Enhanced Login Protocol")
+           (take-credits state :corp)
+           (play-from-hand state :runner "Keyhole")
+           (is (= 3 (:click (get-runner))) "Runner has 3 clicks before run")
+           (let [keyhole (get-in @state [:runner :rig :program 0])]
+                (card-ability state :runner keyhole 0)
+                (is (= 2 (:click (get-runner))) "Runner has 2 clicks after using sneakdoor"))))
```
erbridge: You used Keyhole, not Sneakdoor Beta.

src/clj/game/core/costs.clj
```diff
@@ -123,3 +123,9 @@
...
+
+(defn clear-click-run-bonus [state side]
+  (swap! state update-in [:bonus] dissoc :click-run))
```
erbridge: I don't think this function is necessary (or safe).

src/clj/game/core/actions.clj
erbridge: Probably best to make cards responsible for removing their own click run cost.

src/clj/test/cards/operations.clj
erbridge: Maybe check if the runner still has their last click.

src/clj/game/core/actions.clj
```diff
@@ -343,14 +343,17 @@
...
   [state side {:keys [server] :as args}]
-  (when (and (not (get-in @state [:runner :register :cannot-run]))
-             (can-run-server? state server)
-             (can-pay? state :runner "a run" :click 1))
-    (run state side server)
-    (let [cost (pay state :runner nil :click 1)
-          spent (build-spend-msg cost "make a run on")
-          message (str spent server)]
-      (system-msg state :runner message))))
+  (trigger-event state side :pre-click-run)
```
erbridge: Is it possible to avoid adding a new event by using turn events instead? Look at Service Outage for an example.

mattchainsaw:
I cleaned up my changes a lot and cleaned up the merge conflicts. I took out the pre-click-run triggered event and the dangerous clearing function. It passes the same tests. In the tests I added a check to see if the runner had their last click when ELP denies them the abillity to run

mattchainsaw:
Moved to #2079

PR #2025, opened 2016-10-01
Implement Omar Keung

erbridge:
Implements Omar Keung using a new `:pre-successful-run` event. If there are any other novel interactions to worry about, let me know, as this has been problematic in different ways for every test...


mtgred:
👍

PR #2027, opened 2016-10-01
Use npm scripts rather than global modules

erbridge:
This removes the requirement to globally install Coffeescript and Bower (and make sure they're at the correct versions). It also simplifies the installation process.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,6 @@ The implementation of the game rules is in progress. About 95% of the cards are 
 * Node.js, Node Package Manager
 * Leiningen (version 2+)
 * MongoDB
-* Coffeescript
 * Zero MQ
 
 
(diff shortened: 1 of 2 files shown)


zaroth:
Does npm still install coffeescript this way even if there was one installed globally from the distro's repository?

erbridge:
Yup. NPM will always install all local dependencies regardless of what's installed globally. Their executables get put in `node_modules/.bin`, but NPM scripts add it to the path, so local dependencies are preferred over global ones. Obviously, manually running `coffee` from a terminal will still use the global version (unless they do some clever stuff to actually run the local version instead, which some CLIs do).

erbridge:
Oh, and it's probably worth mentioning that some of the scripts are special, like `postinstall` which always gets executed immediately after `npm install`.

See: https://docs.npmjs.com/misc/scripts.

mtgred:
👍

PR #2030, opened 2016-10-02
call play-subroutine in Archangel and Chrysalis

### LINKED ISSUE #2024: NPE when Runner lets Archangel/Chrysalis trace fire
Please describe the circumstances of your error here.
Accessed a chrysalis in R&D, clicked "yes" to "allow all subs to fire" text > npe.

Presumably because it was not installed yet still acted as encountered ice...

Stack Trace:
```clojure
Error do choice Yes #error {
 :cause nil
 :via
 [{:type java.lang.NullPointerException
   :message nil
   :at nil}]
 :trace
 []}
```

JoelCFC25:
Fixes #2024.

When the Runner allows the effects of these cards to proceed (can't or won't break when accessing them), we now need to use `play-subroutine`

Also adds a `:no-ability` to Yagura sub so it isn't completely silent when the Corp chooses **not** to bottom the top card of R&D.

### DIFF
--- a/src/clj/game/cards/ice.clj
+++ b/src/clj/game/cards/ice.clj
@@ -1273,7 +1273,8 @@
                   {:msg "look at the top card of R&D"
                    :optional {:prompt (msg "Move " (:title (first (:deck corp))) " to the bottom of R&D?")
                               :yes-ability {:effect (effect (move (first (:deck corp)) :deck)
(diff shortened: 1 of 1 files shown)


zaroth:
it seems like circleci had problems with grabbing data from NRDB?

JoelCFC25:
Their SSL cert was expired for a period of time over the weekend, could that be why?

mtgred:
👍

Issue #2032, opened 2016-10-02
Compilation of clojurescript for production causes thousands of warnings about externs


zaroth:
Probably due to issues with minified JS names, `lein cljsbuild once prod` prints out 100MB of similar warnings:

    Oct 02, 2016 10:51:21 AM com.google.javascript.jscomp.LoggerErrorManager println WARNING:
    /home/netrunner/resources/public/lib/react/react.min.js:16: WARNING - accessing name window
    in externs has no effect. Perhaps you forgot to add a var keyword? c=m+o(s,g),v+=u(s,c,n,r);
    else{var y=p(e);if(y){var C,b=y.call(e);if(y!==e.entries)for(var=0;! (... more minified code...)

I put a bandaid on the issue via suppressing the extern validation warnings in caf5a3e, but would welcome someone able to fix the cause, not the effect of the warnings. Resources I found on the matter:

* https://github.com/clojure/clojurescript/wiki/Compiler-Options
* http://lukevanderhart.com/2011/09/30/using-javascript-and-clojurescript.html
* https://groups.google.com/forum/#!topic/clojure/VN6YzyJIg2s
* https://stackoverflow.com/questions/6612369/google-closure-error-with-externs

zaroth:
Fixed by @erbridge in #2033.

PR #2033, opened 2016-10-02
Create a cljs externs file

erbridge:
Related to #2032.

I believe this adds all the needed externs for compilation, but I can't get the production code running locally, so haven't been able to confirm. I have checked by eye that some functions called from the cljs don't get renamed, so the mechanism is working, but there might be missing definitions.


zaroth:
Looks awesome! How was the extern.js file generated? Could you maybe document this proecess on the wiki so we're able to update it easily when more minified libraries are added?

Saintis:
I like the move of `app-state` to a separate file.

erbridge:
It was generated using the generator at the top of the file. Ideally we would write a script to generate it ourselves since it's a pretty manual process to do it using that site. If it works I might take a look at automating it.

PR #2044, opened 2016-10-04
more Escalation cards

JoelCFC25:
Adds Black Orchestra, Peregrine, Houdini, Fairchild 3.0, plus a partial for First Responders (for now it only checks for damage this turn, not that the source was a Corp card). This will fail tests only because of the Adam test that is getting fixed in #2042.


mtgred:
👍

PR #2057, opened 2016-10-06
implement Find the Truth

JoelCFC25:
Adam's new directive. I've gone with a toast for now since it doesn't impose additional clicks on the Runner for the first successful run event. However, another approach would be the prompt that you have to dismiss like in Motivation--the advantage there is the Runner would be able to hover over the card title and see the card preview if they can't remember the details of the card. Open to opinions there.

### DIFF
--- a/src/clj/game/cards/resources.clj
+++ b/src/clj/game/cards/resources.clj
@@ -429,6 +429,18 @@
                              (str "add " (:title c) " to their score area and gain " (get-agenda-points state :runner c)
                                   " agenda point" (when (> (get-agenda-points state :runner c) 1) "s"))))}]}
 
+   "Find the Truth"
+   {:events {:post-runner-draw {:msg (msg "reveal that they drew: "
+                                          (join ", " (map :title (get-in @state [:runner :register :most-recent-drawn]))))}
+             :successful-run {:optional
+                              {:delayed-completion true
+                               :req (req (first-event state side :successful-run))
+                               :prompt "Use Find the Truth to look at the top card of R&D?"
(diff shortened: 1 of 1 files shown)


src/clj/game/cards/resources.clj
```diff
@@ -429,6 +429,18 @@
                              (str "add " (:title c) " to their score area and gain " (get-agenda-points state :runner c)
                                   " agenda point" (when (> (get-agenda-points state :runner c) 1) "s"))))}]}
 
+   "Find the Truth"
+   {:events {:post-runner-draw {:msg (msg "reveal that they drew: "
```
FarCryFromHuman: Should be "reveals," correct? Using "they" as a singular pronoun is fine in modern English, but the runner's name and his action still need subject-verb agreement.

FarCryFromHuman:
For what my opinion is worth, I think a runner using FtT has the intent to do some analysis around the top card of R&D, so I prefer the prompt with card hover.

src/clj/game/cards/resources.clj
JoelCFC25: When using `:msg`, the string is automatically prefixed with:  "<Runner name> uses <card name> to "

JoelCFC25: screen shot 2016-10-06 at 12 29 22 pm

FarCryFromHuman: Ah, clever. That certainly fixes it!

JoelCFC25:
The dismissable prompt version: 
screen shot 2016-10-06 at 12 53 36 pm

mtgred:
👍

PR #2062, opened 2016-10-08
fix Lib Chela agenda forfeit

### LINKED ISSUE #2055: Liberated Chela causes runner win before agenda forfeit
If the runner is on 5 points and uses Liberated Chela to forfeit a 1 point agenda, Jinteki will output a game win for the runner before subtracting the 1 point from the forfeited agenda

kleer1:
Fixes #2055 :Removed :forfeit from cost and manually checked that :forfeit is able to be paid. Allowed for forfeiting of agenda prior to Lib Chela being added to score area.

Test failure from space camp.

### DIFF
--- a/src/clj/game/cards/resources.clj
+++ b/src/clj/game/cards/resources.clj
@@ -609,29 +609,42 @@
                                 (trash state :runner card {:unpreventable true})))}]}
 
    "Liberated Chela"
-   {:abilities [{:cost [:click 5 :forfeit]
-                 :msg "add it to their score area"
-                 :effect (req (if (not (empty? (:scored corp)))
-                                (do (show-wait-prompt state :runner "Corp to decide whether or not to prevent Liberated Chela")
+   {:abilities [{:cost [:click 5]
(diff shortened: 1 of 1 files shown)


kevkcc:
I wonder if there is way that the forfeit function could check for this situation. That would seem to be the more proper way to do it. I would prefer not to see forfeit removed from cost since that is technically correct.

kleer1:
I agree. It's not the ideal solution. The problem I'm running into is that when I trace through the definitions and function calls, it seems like the forfeit of the agenda should occur before the do-effect in the do-ability definition. If that was the case, then I think a flag to say whether the forfeit is preventable could work and then handle the prompts in the forfeit or pay definitions. Unfortunately, the forfeit is happening after the effect. I doubled check it against artist colony and it seems to work the same (ie the msg that you forfeited an agenda comes first in the log, but the actual forfeiting of the agenda happens after the effect). With my less than a week of clojure experience, I haven't been able to pinpoint what is causing this (or if it's intentional for other reasons). I'm definitely new at this, so any input/pointers are appreciated.

PR #2063, opened 2016-10-08
Let the corp choose the number of copies revealed by Reclamation Order

erbridge:
Before:

1. Corp selects a card from Archives
2. All copies are moved to HQ
3. `name uses Reclamation Order to add copies of Adonis Campaign to HQ.` is printed to the log (note no mention of quantity)

After:

1. Corp selects a card from Archives
2.  `name uses Reclamation Order to name Adonis Campaign.` is printed to the log
3. Corp selects the number of cards to reveal
4. That many copies are moved to HQ, starting with `:seen` copies
5. `name uses Reclamation Order to reveal 2 copies of Adonis Campaign from Archives and add them to HQ.` is printed to the log

### DIFF
--- a/src/clj/game/cards/operations.clj
+++ b/src/clj/game/cards/operations.clj
@@ -689,14 +689,33 @@
                             (system-msg (str "does " (or (:stole-agenda runner-reg) 0) " meat damage")))}}
 
    "Reclamation Order"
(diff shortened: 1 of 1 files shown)


mtgred:
👍

PR #2065, opened 2016-10-08
Implement ELP and Service Outage

erbridge:
Includes a variation on #2023 (replacing it).


erbridge:
Service Outage is wrong, so reimplementing.

PR #2066, opened 2016-10-08
Implement Wetwork Refit

erbridge:


erbridge:
That would add a fair amount of work, since you'd need to track which ice it was attached to and make sure to only remove the version of the brain damage subroutine added by it.

erbridge:
I made it remove the first subroutine on `:leave-play`, assuming that is the one it added.

mtgred:
👍

PR #2071, opened 2016-10-09
Implement Service Outage

erbridge:
This adds a new key to abilities, `:makes-run`, to indicate if the ability makes a run (and should therefore be subject to additional costs).


mtgred:
👍

PR #2076, opened 2016-10-11
Miscellaneous fixes

JoelCFC25:
* Fixes failing Adam test (non-exiled Find the Truth is prompting to look at the top of R&D)
* Give Runner-activated pay credits abilities to Pup and Swarm
* Deactivate target of the `/rfg` command so an exiled Adam directive will not see its effects persist

### DIFF
--- a/src/clj/game/core/io.clj
+++ b/src/clj/game/core/io.clj
@@ -187,7 +187,8 @@
                                          {:title "/rez command"} nil))
         "/rez-all"    #(when (= %2 :corp) (command-rezall %1 %2 value))
         "/rfg"        #(resolve-ability %1 %2 {:prompt "Select a card to remove from the game"
-                                               :effect (effect (move target :rfg))
+                                               :effect (req (let [c (deactivate %1 %2 target)]
+                                                              (move %1 %2 c :rfg)))
                                                :choices {:req (fn [t] (card-is? t :side %2))}}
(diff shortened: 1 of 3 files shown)


src/clj/game/core/io.clj
```diff
@@ -187,7 +187,8 @@
                                          {:title "/rez command"} nil))
         "/rez-all"    #(when (= %2 :corp) (command-rezall %1 %2 value))
         "/rfg"        #(resolve-ability %1 %2 {:prompt "Select a card to remove from the game"
-                                               :effect (effect (move target :rfg))
+                                               :effect (req (let [c (deactivate %1 %2 target)]
+                                                              (move %1 %2 c :rfg)))
```
zaroth: we don't have a general-purpose `rfg` function?

src/clj/test/cards/identities.clj
```diff
@@ -38,6 +38,7 @@
     (advance state (get-content state :remote1 0) 2)
     (take-credits state :corp)
     (run-empty-server state :remote1)
+    (prompt-choice :runner "No") ; Dismiss prompt from non-exiled Find the Truth directive
```
zaroth: Yay! Let's make the tests green again!

src/clj/game/core/io.clj
erbridge: Maybe we should make one?

Issue #968, opened 2015-12-09
Enhancement request.  Sounds


AlvazHall:
Sounds for major events such as:
Playing as Corp
   opponent end turn
   run initiated
   agenda stolen
   ice breaker installed
   current installed
Playing as runner
   opponent end turn
   new remote created
   agenda scored
   ice installed
   current installed

nealterrell:
We'd like it too. It was discussed in #342 but we never heard back from the sound designer. @bozmond... any updates? :D

bozmond:
Hi guys. Yes, it's been playing on my mind I haven't done anything yet. Been pretty busy setting up a new studio here in Melbourne, along with being pretty busy with other projects. I have some time coming up over the holidays so I'll endeavor to get something happening. The new studio I'm setting up is worth a look if you're interested in old synths. http://mess.foundation

xanshin:
good to see this feature is being looked at, it's probably the main thing i miss from OCTGN when using jinteki.

I'd like to add a request for one event:
**-new card is installed in a sever**

how did things go over the holidays @bozmond ?

bozmond:
Hey @xanshin and all,

I'm pretty flat out for the next three weeks. I'm hoping to look at this after that! 

Just a comment overall about the sounds. Personally, I don't think OCTGN sounds are that great, but my impression is people end up liking them because they are synonymous with playing AN online, and after a few hundred plays a little familiarity is bound to set in. 

My thinking about the sounds was that I'd make something distinct, but probably more abstract. Unlike the literal door opening and closing sounds on OCTGN - something clean and clear, more synth based but with a bit of mechanical flavour thrown in.

I'd be keen if anyone has some references to other interface sounds in other games, films or whatever that they think might be a good match for jinteki. It'll help me to figure out what might be a good fit for the community overall.

xanshin:
thanks @bozmond 
If its helpful at all I had a dig around in freesound.org and found some nice tactile mechanical sounds and some abstract type sounds that might suit certain events.

opponent end turn
https://www.freesound.org/people/suzenako/sounds/48670/
https://www.freesound.org/people/juskiddink/sounds/122683/

current installed
https://www.freesound.org/people/E330/sounds/99444/
https://www.freesound.org/people/BMacZero/sounds/94133/

new card in sever
https://www.freesound.org/people/Omar%20Alvarado/sounds/251544/

new remote created
https://www.freesound.org/people/starinatwalls/sounds/331155/

ice installed
[...]

PR #2078, opened 2016-10-12
Sound framework

zaroth:
A step on the way for finishing #968


views/layout.jade
```diff
@@ -18,6 +18,7 @@ html
     script(src='/lib/moment/min/moment.min.js')
     script(src='/lib/marked/marked.min.js')
     script(src='/lib/toastr/toastr.min.js')
+    script(src='/lib/howler/howler.min.js')
```
erbridge: Need to update the externs, too.

PR #2079, opened 2016-10-12
ELP implementation

mattchainsaw:
Moved over from #2023 to avoid all the rebase nonsense


src/clj/game/core/actions.clj
```diff
@@ -342,13 +342,14 @@
...
   [state side {:keys [server] :as args}]
   (let [cost-bonus (get-in @state [:bonus :run-cost])]
-    (when (and (not (get-in @state [:runner :register :cannot-run]))
-               (can-run-server? state server)
-               (can-pay? state :runner "a run" :click 1 cost-bonus))
-      (run state side server)
-      (when-let [cost-str (pay state :runner nil :click 1 cost-bonus)]
-        (system-msg state :runner
-                    (str (build-spend-msg cost-str "make a run on") server))))))
+    (let [click-cost-bonus (get-in @state [:bonus :click-run-cost])]
```
erbridge: You can combine this let with the previous one.

src/clj/game/cards/operations.clj
```diff
@@ -291,6 +291,23 @@
...
+   (let [remove-effect (req (when (:elp-click-run card)
+                              (click-run-cost-bonus state side [:click -1])
+                              (update! state side (dissoc card :elp-click-run))))
+         remove-ability {:req (req (and
+                                     (:elp-click-run card)
+                                     (:elp-fired card)))
+                         :effect remove-effect}]
+     {:events {:runner-turn-begins {:msg "add an additional cost of 1 [Click] to make the first run not through a card ability this turn"
+                                    :effect (effect (update! (assoc card :elp-click-run true))
+                                              (click-run-cost-bonus [:click 1]))}
```
erbridge: What happens if the card is played on the Runner turn e.g. with Sol? This is a problem for Service Outage, too. Probably ok to require a manual fix for now, though.

src/clj/game/cards/operations.clj
```diff
@@ -291,6 +291,23 @@
...
+               :runner-turn-ends {:effect remove-effect}
+               :run {:req (req (when (nil? (:card (:run-effect (:run @state))))
+                                 (update! state side (assoc card :elp-fired true))))}
```
erbridge: Do we need this event? What does it do?

erbridge: Nevermind, I see, but when does it get cleaned up?

src/clj/game/cards/operations.clj
JoelCFC25: Add a top level `:effect` with the same ability block as the `:runner-turn-begins` event and add a `:req` so it only happens if the active player is the Runner?

src/clj/game/cards/operations.clj
```diff
@@ -292,19 +292,24 @@
...
-                                     (:elp-fired card)))
+                                     (:elp-trigger-effect card)
+                                     (:clicked-to-run card)))
                          :effect remove-effect}]
-     {:events {:runner-turn-begins {:msg "add an additional cost of 1 [Click] to make the first run not through a card ability this turn"
-                                    :effect (effect (update! (assoc card :elp-click-run true))
-                                              (click-run-cost-bonus [:click 1]))}
+     {:msg "add an additional cost of 1 [Click] to make the first run not through a card ability this turn"
+      :effect (effect (update! (assoc card :elp-trigger-effect true))
+                      (click-run-cost-bonus [:click 1]))
```
erbridge: Would be good to refactor this like you do for the removal effects.

src/clj/game/cards/operations.clj
```diff
@@ -291,6 +291,28 @@
...
+   (let [remove-effect (req (when (:elp-trigger-effect card)
+                              (click-run-cost-bonus state side [:click -1])
+                              (update! state side (dissoc card :elp-trigger-effect))))
+         remove-ability {:req (req (and
+                                     (:elp-trigger-effect card)
+                                     (:clicked-to-run card)))
+                         :effect remove-effect}]
+     {:msg "add an additional cost of 1 [Click] to make the first run not through a card ability this turn"
+      :effect (effect (update! (assoc card :elp-trigger-effect true))
+                      (click-run-cost-bonus [:click 1]))
```
erbridge: Maybe give Service Outage the same treatment while you're at it?

src/clj/game/cards/operations.clj
erbridge: Oh wait. This is actually more subtle than that, because it's activated by the first click run each turn, which could have happened before playing it.

mtgred:
👍

PR #2096, opened 2016-10-17
Added test for Deja Vu, recurring non-virus and virus cards

kevinpowe:
Added basic testing for Deja Vu, recurring a non-virus and a virus card and checking behaviour of the UI for both.

Would have added a check for only viruses being presented as options in testing the virus recursion, but still cutting my teeth on how to interrogate which options are returned by the UI for a prompt.

Have ensured that I've pushed all files from local repo this time! ;-)

### DIFF
--- a/src/clj/test/cards/events.clj
+++ b/src/clj/test/cards/events.clj
@@ -265,6 +265,55 @@
     (run-jack-out state)
     (is (empty? (:prompt (get-runner))) "No option to run again on unsuccessful run")))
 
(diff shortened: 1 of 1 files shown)


kevinpowe:
Got it. Thanks @nealterrell. I'd used that form in the test for Noise, but wasn't 100% sure it was doing what I wanted here. Glad to have that behaviour confirmed!

mtgred:
👍

PR #2105, opened 2016-10-18
Choose Adam directives before the start of the game

erbridge:
The runner is prompted to choose 3 directives from the temporary zone before either side draws their hand. Any unused directives disappear into the aether.

Note: Due to the limitations for the choices interface, the runner is able to choose fewer than 3 directives, and currently has no way back from that.


mtgred:
👍

PR #2107, opened 2016-10-18
Display the number of games that are pending on the lobby tabs

### LINKED ISSUE #1918: An indicator of number of public open games per lobby room
As per title. Would certainly aid noticing when a game opens in a not very populated room (such as Competitive outside of peak hours).

erbridge:
Resolves #1918.


Saintis:
This can be closed now that #2746 has been merged.

Issue #1244, opened 2016-02-17
Blackguard + add'l costs


TalionKris:
Currently, using Blackguard forces the rez of cards like Archer, which it shouldn't, due to the ruling about additional costs (i.e. the Runner can't force the Corp to pay them).

Here's the current list of cards it interacts with:

Archer
Bribery
Cortez Chip
Enforcer 1.0
Running Interference

zaroth:
The runner's cards you mentioned look pretty situational and rarely played, especially in this situation. I'd rather worry about:

Corporate Town
Ibrahim Salem

JoelCFC25:
Archer, Enforcer 1.0, Corporate Town, and Ibrahim Salem look like they can be handled pretty easily--`(when (:additional-cost (card-def target)))` blah blah, throw an optional prompt at the Corp player asking if they want to rez or decline. 

Bribery and Running Interference apply `rez-cost-bonus` to their targets and are contingent on mid-run trickery like Raymond Flint or Snitch. I don't remember how these bonuses are represented in state so I don't know offhand if Blackguard could pick them out.

zaroth:
Also, a big one - **Hacktivist Meeting**.

PR #2111, opened 2016-10-20
marginal Blackguard improvement against "additional cost to rez" cards

JoelCFC25:
Related to #1244. Not a total fix, but this will keep the Corp from being forced to forfeit an agenda to rez Archer, Enforcer 1.0, Corporate Town, or Ibrahim Salem when they are exposed with Blackguard in play. As they have additional costs to rez, the Corp is supposed to be able to decline to pay them and thus avoid rezzing the card entirely.


mtgred:
👍

PR #2114, opened 2016-10-21
Omni-drive fix

poi2000:
Omni-drive is missing an effect to remove the hosted card's memory from the Runner's memory when you host a card that is already in play. I've tried to make a quick fix but the code test keeps denying my changes. The most recent error is saying that the map literal contains an even number of forms. What exactly does this mean and how can I fix it.(Fixed).


kevkcc:
Your changes don't currently compile. Basically, you don't need to call effect twice. Also, if you have time, consider helping us expand our unit tests by adding a test for Omni-drive!

mtgred:
👍

PR #2116, opened 2016-10-22
Standardize some strings and refactor bits

erbridge:


src/clj/game/cards/events.clj
```diff
@@ -50,22 +64,23 @@
...
+   (run-event
+    {:req (req has-bad-pub)
+     :msg "prevent ICE from being rezzed during this run"}
+    nil
+    (effect (register-run-flag!
+             card
+             :can-rez
+             (fn [state side card]
+               (if (ice? card)
+                 ((constantly false) (system-msg state side (str "is prevented from rezzing ICE on this run by Blackmail")))
```
JoelCFC25: This was the first run-flag card, and still writes to the game log instead of using a toast like everything else now does.

erbridge:
I think CircleCI is confused by having started on `dev`, which has a different `circle.yml`, with a different version of Node, which is incompatible with `master`'s `package.json`.

mtgred:

erbridge:
Done.

Saintis:
Looks like this needs another rebase :)

Saintis:
Merged in #2793. Thanks for the ground work!

PR #2130, opened 2016-10-30
stop Tri-maf/Rolodex effects from triggering when lost to damage or manually discarded

### LINKED ISSUE #2128: Tri-Maf Contact
Does 3 damage when trashed from hand due to damage or due to hand size - card text should not be active when not installed.

JoelCFC25:
Fixes #2128. 

Maybe _(pretty please?)_ this will kill off Tri-maf problems, but then again I've said that before. Tri-maf's 3 meat damage and Rolodex's "mill 3" leave play effects were changed to `:trash-effect` to solve a different kind of wrong triggering recently, but now they've been firing if you trash these cards from hand (over hand limit) or they get hit by damage. 

This adds a couple more conditions to `resolve-trash-end` so that a Runner card must be _installed_ for its `:trash-effect` to fire (solves Tri-maf and Rolodex), but also allows a card to use `:when-inactive true` in its trash effect to let it work from a player's hand (this preserves I've Had Worse functionality and the 4 Corp trashable Executives are changed to use this as well).

### DIFF
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -193,7 +193,7 @@
    "Chairman Hiro"
    {:effect (effect (lose :runner :hand-size-modification 2))
     :leave-play (effect (gain :runner :hand-size-modification 2))
-    :trash-effect {:when-unrezzed true
+    :trash-effect {:when-inactive true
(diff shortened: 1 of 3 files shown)


mtgred:
👍

PR #2131, opened 2016-10-30
Many fixes

### LINKED ISSUE #2046: Close-prompt during Celebrity Gift or Mushin No Shin
After unintentionally clicking the wrong card and closing the prompt I noticed that the game starts behaving very strangely past that point. This lasts seemingly forever as I made a deck containing nothing but Mushin No Shin, Celebrity Gift and Pad Campaigns to put down on the board and went through my entire deck.

When the prompt is closed on either of the two cards with or without the other in hand the following will happen.

The next card, not the one you closed the prompt on, will behave as the other one. 

In a sequence played as such. Where ok means it works as intented and not ok is the "opposite"

> Close-prompt mushin
>    play mushin - ok
>       play gift - not ok
>          play mushin - not ok
>             play mushin - ok

You close the Mushin which makes the next Celebrity Gift behave as a Mushin, which in turn makes the next Mushin after that behave as a Celebrity Gift. 
However, any number of consecutive number of Mushins or Celebrity Gifts will behave normally past the first. They behave normally until the other card is played.

The behaviour is only reversed for the for the first card played after the other. Alternating Mushin and Celebrity Gift will guarantee the reversed behavior every card. They are always "the first card of the other kind."

Expecting this to be related to doubles or perhaps targeted operations. I have tried both with Corporate Shuffle and Reuse.

Corporate Shuffle does not alter its behavior as far as I have been able to tell. Perhaps because it is not a targeted card.

Reuse however also shares the same problem as Celebrity Gift and Mushin. With three targeted cards in the deck the order seems to work out to this.

The same things apply as before. Consecutives operations work past the first etc.

The operation you are playing right now will become what the card of the last one you played says it is.

> Close-prompt gift
> play reuse - becomes gift
> play mushin - becomes reuse
> play gift - becomes mushin

I also tried some non double targeted operations.

Rework and Lateral Growth but also combined with reuse to see if the behavior is somehow not consistent between double or normal operations.
The behavior is still consistent from what I can tell. It still becomes what the last card played says it is. 

In all cases the text above the hand looks proper. Rework asks you to pick a card, Celebrity Gift asks you to pick five. But if the card that is actually "in effect" allows more or less it allows this despite the text. Rework can target five cards to trash as Reuse and gift is forced to shuffle in the first card clicked.

I have not tried this with an opponent who also uses targeted events to see if it breaks across players nor have I tried it as a runner. But I expect it to be the same for the runner.

### LINKED ISSUE #2070: Encountering Archangel from a central prompts the runner for successful run after the encounter ends
Running on a central server (HQ or R&D especially) and hitting an Archangel that is not installed but is in hand or in R&D will prompt the runner to have a successful run after the encounter ends. This has the effect of repeating the access step of the run. If they hit Archangel _again_ they could conceivably be stuck in an infinite loop. A /close-prompt is needed to put an end to this.

### LINKED ISSUE #2113: Gang sign lets/forces runner to access installed upgrades in root of HQ.
Runner accessed two cards from hand, then accessed installed Prisec in HQ. Unsure whether this was part of the first or second gang sign access and whether or not a popup came up for the runner to access installed upgrade.

### LINKED ISSUE #2127: Psychic Field not trashed after being exposed with Drive By
Using Drive By on a Psychic Field correctly results in a psi game first, and the damage is applied correctly if the Runner loses, but after the psi game finishes the Psychic Field remains installed instead of getting trashed.

nealterrell:
Fix #2127 by correctly setting the async result after exposing a card with an on-expose effect.

Fix #2113 by tricking the access helper into thinking all the cards in HQ root were already accessed.

Fix #2046 by making `close-prompt` trigger the closed prompt as "effect completed".

Fix #2070 by reworking Archangel's `:access` effect to use delayed completion.

Fix an issue reported in chat involving installing Shards as Laramy Fisk leading to an infinite loop.

### DIFF
--- a/src/clj/game/cards/ice.clj
+++ b/src/clj/game/cards/ice.clj
@@ -190,22 +190,30 @@
 
    "Archangel"
    {:access
-    {:req (req (not= (first (:zone card)) :discard))
+    {:delayed-completion true
+     :req (req (not= (first (:zone card)) :discard))
      :effect (effect (show-wait-prompt :runner "Corp to decide to trigger Archangel")
-                     (resolve-ability
+                     (continue-ability
                        {:optional
                         {:prompt "Pay 3 [Credits] to force Runner to encounter Archangel?"
                          :yes-ability {:cost [:credit 3]
-                                       :effect (req (system-msg state :corp "pays 3 [Credits] to force the Runner to encounter Archangel")
-                                                    (clear-wait-prompt state :runner)
-                                                    (resolve-ability state :runner
-                                                                     {:optional
-                                                                      {:player :runner
-                                                                       :prompt "Allow Archangel trace to fire?" :priority 1
-                                                                       :yes-ability {:effect (req (play-subroutine state side {:card card :subroutine 0}))}}}
-                                                                     card nil))}
-                         :no-ability {:effect (req (system-msg state :corp "declines to force the Runner to encounter Archangel")
-                                                   (clear-wait-prompt state :runner))}}} card nil))}
+                                       :delayed-completion true
+                                       :effect (effect (system-msg :corp "pays 3 [Credits] to force the Runner to encounter Archangel")
+                                                       (clear-wait-prompt :runner)
+                                                       (continue-ability
+                                                         :runner {:optional
+                                                                  {:delayed-completion true
+                                                                   :player :runner
+                                                                   :prompt "Allow Archangel trace to fire?"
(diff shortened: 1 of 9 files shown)


nealterrell:
@JoelCFC25 shouldn't you be watching the last Cubs World Series game for another 100 years?

mtgred:
👍

PR #2138, opened 2016-11-04
Four Intervention cards

nealterrell:
Bulwark (nothing notable).

Government Investigation. (Improve `all-active` to return currents. Add a card flag to restrict specific bets on psi games.)

Top Hat. (Was pleasantly surprised that this worked first try. Shows optional prompt on successful R&D run. If yes, replaces access with a mandatory "pick a card, any card" routine. Run does not end until a card is chosen.)

The Gauntlet. (We don't have a way to track "how many ice you broke all subroutines on". It's more complicated than counting the ice protecting HQ, thanks to the Bird breakers, Cutlery, etc.  So we just ask the runner. Unfortunately the `:number` UI for prompts requires an upper limit, so I fixed that at 10. I'm open to better suggestions, noting that the higher the number, the longer and more unwieldy the UI list becomes.)

### DIFF
--- a/src/clj/game/cards/events.clj
+++ b/src/clj/game/cards/events.clj
@@ -398,6 +398,9 @@
    {:msg (msg "draw " (- (hand-size state :runner) (count (:hand runner))) " cards")
     :effect (effect (draw (- (hand-size state :runner) (count (:hand runner)))))}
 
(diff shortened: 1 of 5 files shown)


mtgred:
👍

PR #2141, opened 2016-11-04
more Intervention cards

JoelCFC25:


src/clj/game/core/abilities.clj
```diff
@@ -422,7 +422,7 @@
           (system-msg state side (str "spends " bet " [Credits]"))
           (trigger-event state side (keyword (str "psi-bet-" (name side))) bet)
           (trigger-event state side (keyword (str "psi-bet-" (name opponent))) opponent-bet)
-          (trigger-event state side :psi-game nil)
+          (trigger-event state side :psi-game bet opponent-bet)
```
nealterrell: Slight problem here... delayed-completion only works with trigger-event-sync or trigger-event-simult, so Fumiko's not going to have delayed completion here. The only problem I can foresee is a Runner who has both meat and net damage prevention hitting a Psychic Field with Fumiko rezzed. I imagine we'll get overlapping prevention prompts because Psychic Field will trigger (from `:access`) before the prevention for Fumiko's meat damage has finished.

JoelCFC25: So...the solution is?

JoelCFC25: Will try that out with your test case...

mtgred:
👍

PR #2142, opened 2016-11-04
Remove a flex setting to fix layout on Chrome

erbridge:
Not sure why I put the `flex(0)` in there, but it seems to be fine without it.


nealterrell:
👍

PR #2143, opened 2016-11-04
Fix the show discard for e.g. Jackson Howard

erbridge:
Add .me to things as well as .opponent to fix the discard auto-popup.


nealterrell:
👍

PR #2145, opened 2016-11-05
Frantic coding

kevkcc:
So I guess there's the possibility that if the runner had recurring credits and didn't take them before playing frantic coding, something they could possibly install wouldn't be on the list since i populate the list of installable targets at the time the card is played but it's not a huge deal and I'm not sure how best to deal with it. This situation is gonna come up a lot more with Net Mercur.

Also updated Eureka which is a similar card.


mtgred:
👍

Issue #2154, opened 2016-11-10
Playing currents in sol consistently triggers erros


tolaasin:
Please describe the circumstances of your error here.

Stack Trace:
```clojure
Error do choice Hyperdriver #error {
 :cause nil
 :via
 [{:type java.lang.NullPointerException
   :message nil
   :at [clojure.string$lower_case invokeStatic "string.clj" 217]}]
 :trace
 [[clojure.string$lower_case invokeStatic "string.clj" 217]
  [clojure.string$lower_case invoke "string.clj" 213]
  [game.core$resolve_prompt$fn__4447 invoke "actions.clj" 129]
  [clojure.core$some invokeStatic "core.clj" 2592]
  [clojure.core$some invoke "core.clj" 2583]
  [game.core$resolve_prompt invokeStatic "actions.clj" 129]
  [game.core$resolve_prompt invoke "actions.clj" 113]
  [game.main$handle_do invokeStatic "main.clj" 77]
  [game.main$handle_do invoke "main.clj" 74]
  [game.main$handle_command invokeStatic "main.clj" 150]
  [game.main$handle_command invoke "main.clj" 139]
  [game.main$run$fn__25520 invoke "main.clj" 185]
  [game.main$run invokeStatic "main.clj" 175]
  [game.main$run invoke "main.clj" 169]
  [game.main$_main$fn__25555 invoke "main.clj" 239]
  [clojure.lang.AFn run "AFn.java" 22]
  [java.lang.Thread run "Thread.java" 745]]}
```

tolaasin:
Targeted marketing consistently gives game errors at the moment.

JoelCFC25:
Thanks--duplicate of #2152, #2148, #2146.

nealterrell:
Is this just with Targeted Marketing? (Implied otherwise in title.)

tolaasin:
Yes, it's just with tarmar. Scarcity worked fine.

PR #2156, opened 2016-11-11
Targeted Marketing fix

nealterrell:
I still believe this is an issue with the database on live, as the only possible NPE on the line reported in #2154 etc. is if the `:title` of one of the entries in `@all-cards` is `nil`. This check will coerce a nil title to the empty string, so `lower-case` won't throw.


mtgred:
👍

PR #2158, opened 2016-11-11
Improvements to Medium, Nerve Agent via @JoelCFC25

### LINKED ISSUE #386: Medium/Nerve Agent extra accesses should apply outside of runs
Nerve Agent currently modifies `access-bonus`, which is a key confined only to runs. When 2 or more virus counters are on Nerve Agent it should act like a mini HQ Interface and grant additional accesses even when they occur outside a run (Raymond Flint, Gang Sign, Kitsune).

### LINKED ISSUE #2150: Gauntlet not allowing access of more than 1 card
Gauntlet prompts to ask how many ICE on HQ all subs were broken on but runner does not access greater than 1 card.
Reproduced multiple times, selecting 1 or 2 as number of ICE with all subs broken.

nealterrell:
This is @JoelCFC25's work on making the `:pre-access` event synchronized with `trigger-event-sync`. Enables cards handling that event to pause the access phase until their effect is resolved, which allows Medium and Nerve Agent to prompt to select the number of additional accesses once the Runner hits the access step. Beautiful.

Fixes #2150 and moves the Gauntlet handler to `:pre-access` so it doesn't fire on Account Siphon (etc) runs.

Fixes #386 by triggering `:pre-access` in Gang Sign etc. so Nerve Agent and Medium can set the access amount.

Adds a `:default` field to prompts that use `:number`, allowing you to set the default-selected option. In this case, Medium sets the default to the largest number of accesses, instead of 0.


mtgred:
👍

PR #2159, opened 2016-11-12
fix #561: dont show card text when img-url present

### LINKED ISSUE #561: Jinteki biotech showing the hidden identity
http://puu.sh/jHusa/c9915be62d.png <-upper right corner, just under where it shows their identity-card. One line of text.

DeepSpawn:
fixes #561 

screen shot 2016-11-11 at 11 58 28 pm

screen shot 2016-11-12 at 3 55 50 pm

Previously we were rendering the card text and then putting the img over the top of this.
When the card text was long this would result in card text being visible at the bottom.
We now only show the card text if we are unable to create the img url for the card.

### DIFF
--- a/src/cljs/netrunner/cardbrowser.cljs
+++ b/src/cljs/netrunner/cardbrowser.cljs
@@ -38,39 +38,45 @@
       (make-span "\\[Trash\\]" "trash")))
 
 (defn card-view [card owner]
-  (om/component
-   (sab/html
-    [:div.card-preview.blue-shade
-     [:h4 (:title card)]
-     (when-let [memory (:memoryunits card)]
-       (if (< memory 3)
-         [:div.anr-icon {:class (str "mu" memory)} ""]
-         [:div.heading (str "Memory: " memory) [:span.anr-icon.mu]]))
-     (when-let [cost (:cost card)]
-       [:div.heading (str "Cost: " cost)])
-     (when-let [trash-cost (:trash card)]
-       [:div.heading (str "Trash cost: " trash-cost)])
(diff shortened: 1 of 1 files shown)


DeepSpawn:
Hey I have updated the PR based upon your feedback. 

I modified the card-view component to contain a state to determine if the it should show the text view or not, which defaults to false. The onError event handle updates the state of the component to toggle this, which will trigger a re-render of the component showing the text view.

(I tried to work out how to extract the text view code to a separate function but am not experienced enough with clojure to get it working so I have just left it as it is)

DeepSpawn:
screen shot 2016-11-15 at 10 09 39 pm
When testing I noticed that the text view work correctly for identities, but not for other cards. I am not sure if this is expected behavior or not. I was able to reproduce it on master when the img url is hardcoded to something that will fail (attached screenshot is from master).

DeepSpawn:
Hey @nealterrell how do I keep this PR moving? I changed my approach after you first looked at it and I think the current approach is a reasonable way of solving the problem.

nealterrell:
Sorry about that, got caught up in dev server merges and then the holidays. Let me review and then authorize.

nealterrell:
Could you clarify the screenshot you posted in your comment? I'm not sure what the issue is.

DeepSpawn:
So that last screen shot is showing that we do not currently get the text for the card body when viewing non identity cards, we can see its name and type etc but not what the card actually does.

nealterrell:
OK, I don't know why that is, but it's a very minor issue.

mtgred:
👍

PR #2160, opened 2016-11-12
Make secondary-pane z index 1

DeepSpawn:
I noticed that when you have a logos trigger with a diverse deck, the prompt to select a card is rendered underneath your hand making some cards not selectable.

This PR is just a CSS change to move the secondary pane to +1 z level so that it will get rendered over top of the players hand. 

Before:
screen shot 2016-11-12 at 6 03 10 pm

After:
screen shot 2016-11-12 at 6 51 01 pm

### DIFF
--- a/src/css/base.styl
+++ b/src/css/base.styl
@@ -1436,6 +1436,7 @@ nav ul
 
   .secondary-pane
     width: 186px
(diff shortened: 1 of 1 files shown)


mtgred:
👍

PR #2164, opened 2016-11-15
implement Hasty Relocation

JoelCFC25:
Stole from CBI Raid here. Could use some eyeballs--@nealterrell, @Saintis, @zaroth, @erbridge etc. 

This **will not** work correctly in conjunction with Accelerated Diagnostics (which, if correct, would stop the mill). But people are asking about this card a lot and a starting point is better than nothing I guess.


mtgred:
👍

PR #2167, opened 2016-11-18
apply Checkpoint damage at correct timing

### LINKED ISSUE #2129: minor log issue with ELP
TheBigBoy spends 2  and  to play Enhanced Login Protocol.
TheBigBoy uses Enhanced Login Protocol to .
TheBigBoy uses Jeeves Model Bioroids to gain .

second line is borked

### LINKED ISSUE #2165: Checkpoint deals damage after a run is successful rather than when it is successful
Like the title says the damage is applied at the wrong time. The runner can access cards before the damage.

JoelCFC25:
* Fixes #2165: Checkpoint should use event registration instead of swapping things into `[:run]`. This will also solve the broken interaction of a successful Checkpoint trace on a Dirty Laundry run. 
* Fixes #2129: Use `system-msg` to prevent a truncated message when ELP is played.
* Removes subtypes (if necessary) from Mother Goddess when other ICE gets trashed

### DIFF
--- a/src/clj/game/cards/ice.clj
+++ b/src/clj/game/cards/ice.clj
@@ -349,13 +349,14 @@
    {:effect take-bad-pub
     :subroutines [(trace-ability 5 {:label "Do 3 meat damage when this run is successful"
                                     :msg "do 3 meat damage when this run is successful"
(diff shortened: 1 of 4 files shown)


mtgred:
👍

PR #2172, opened 2016-11-23
partial implementation of Raman Rai

JoelCFC25:
This is a "most of the way" treatment of this really bizarre card, stealing techniques from DBS and Midori. 

I think in theory during an instance of multi-draw you'd have to decide after each card whether or not to use Raman Rai, but for this I just took the easy route and did what we do with DBS--allow you to choose any of the cards in the `:most-recent-drawn` stored in the register. I also thought it would be really annoying if this card prompted the user after every single draw, so it's just silent and we leave it up to players to decide when to fire it and not abuse it (hopefully). 

screen shot 2016-11-23 at 9 43 57 am


mtgred:
👍

Issue #2184, opened 2016-11-28
Implement Reaver


shazzner:
Please implement Apex's latest card Reaver. I implemented the code myself but as I don't have contributor rights, I'm unable to create a pull request.

`"Reaver"
   {:events {:runner-trash {:req (req (and (first-event state :runner :runner-trash) (:installed target)))
                     :effect (effect (draw :runner 1))
                     :msg "to draw 1 card"}}}`

JoelCFC25:
You would need to clone this repository, create a new branch in your own repo with the code, and then you could submit the PR from that.

JoelCFC25:
On a related note, was discussing this card in Stimhack Slack this morning, and most people think it's meant to trigger on "first time this turn" trash of a Corp card as well, the FFG news article notwithstanding. That would square with the text as written, too. I guess it will remain to be seen--but the 4 influence rating would make more sense that way. Historically the FFG news articles have been shot through with rules errors and other flat-out wrong references.

nealterrell:
Either way, we encourage you to follow the "typical" GitHub workflow and take official credit in the git logs for your submission. Fork this project. Clone it to your desktop. Implement the code in your repo, sync to GItHub upstream, and then submit a Pull Request :).

shazzner:
Created PQ, apologies for getting back to this so late!

PR #2185, opened 2016-11-29
Added in partial implementation of Reaver

shazzner:
As per: https://github.com/mtgred/netrunner/issues/2184

Partial implementation of Reaver program. Partial as JoelCFC25 pointed out, should probably fire on corp trash.

### DIFF
--- a/src/clj/game/cards/programs.clj
+++ b/src/clj/game/cards/programs.clj
@@ -588,6 +588,11 @@
                           :effect (effect (update! (assoc card :hosted-programs (remove #(= (:cid target) %) (:hosted-programs card))))
                                           (lose :memory (:memoryunits target)))}}}
 
+   "Reaver"
+   {:events {:runner-trash {:req (req (and (first-event state :runner :runner-trash) (:installed target)))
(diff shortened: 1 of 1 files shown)


src/clj/game/cards/programs.clj
```diff
@@ -588,6 +588,11 @@
                           :effect (effect (update! (assoc card :hosted-programs (remove #(= (:cid target) %) (:hosted-programs card))))
                                           (lose :memory (:memoryunits target)))}}}
 
+   "Reaver"
+   {:events {:runner-trash {:req (req (and (first-event state :runner :runner-trash) (:installed target)))
```
JoelCFC25: If in fact it turns out this is meant to work on Corp cards, use `(installed? target)`, because only Runner cards ever have `:installed true`.

JoelCFC25:
This one would be all set with the change I noted plus justifying the `:effect` and `:msg` lines so they line up with `:req`.

shazzner:
Apologies for the lateness, added in the effect (didn't test, hope it's correct) and space-aligned the :msg text.

nealterrell:
👍

PR #2186, opened 2016-11-30
implement Credit Crash and Hellion Beta Test

### LINKED ISSUE #1089: Agenda point ambushes asking to trash after firing
News team (and possibly shikyu) firing from not-Archives still ask you whether to trash them even if you choose to take them to your score area. It works correctly, but can get pretty confusing.

JoelCFC25:
Bear with me on this one...it's the result of several crucial clarifications from rules guru @jakodrako on Stimslack, since there still are no released UFAQs for anything after the 23 Seconds pack. 

He says Credit Crash was ruled to act like a "when you access" trigger, despite not being worded that way _(!!)_. As such, it needs to resolve prior to a Corp card's "when the Runner accesses" effects--and can even make them fizzle entirely, as we'll get to in a moment. The other quirk of this card is that if the Corp opts to pay the credits to prevent "this trash", that only applies to the no-cost trash--the Runner _still_ subsequently gets a chance to pay credits to trash conventionally (assuming the card has a trash cost, of course). 

How this might play out in an actual scenario...the Runner plays Credit Crash and runs R&D. They hit an Archangel, which has a "when accessed" effect. With the `when-completed` framework applied to the `:pre-access-card` event, Credit Crash can fully resolve first. The Corp is asked if they want to pay Archangel's 4 credit rez cost to prevent it being trashed by the Runner. They decline. Archangel is then instantly trashed at no cost and its "when accessed" effect never happens.

This PR also fixes #1089. Just as we have to check "is this card still in the same place?" after resolving any `:pre-access` effect like Credit Crash, we have to check again after any "when accessed" effect resolves, so that a News Team or Shi.Kyu the Runner elects to take as -1 point will no longer ask the Runner to trash them.

EDIT: Also added in Hellion Beta Test.

### DIFF
--- a/src/clj/game/cards/events.clj
+++ b/src/clj/game/cards/events.clj
@@ -127,6 +127,38 @@
     :effect (req (swap! state update-in [:corp :has-bad-pub] inc))
     :leave-play (req (swap! state update-in [:corp :has-bad-pub] dec))}
 
+   "Credit Crash"
+   {:prompt "Choose a server" :choices (req runnable-servers)
+    :effect (effect (run target nil card)
+                    (register-events (:events (card-def card))
+                                     (assoc card :zone '(:discard))))
+    :events {:pre-access-card
+             {:once :per-run
+              :delayed-completion true
+              :req (req (not= (:type target) "Agenda"))
+              :effect (req (let [c target
+                                 cost (:cost c)
+                                 title (:title c)]
+                             (if (can-pay? state :corp nil :credit cost)
+                               (do (show-wait-prompt state :runner "Corp to decide whether or not to prevent the trash")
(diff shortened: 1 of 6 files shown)


kevkcc:
I'm wondering if :trashed-card might be better named as :trashed-accessed-card so it's more accurate. Also does Hellion Beta Test work with Gang Sign? Otherwise this is great stuff!!

JoelCFC25:
> Also does Hellion Beta Test work with Gang Sign?

It should, since if you follow the chain...Gang Sign calls `access-helper-hq` which in turn calls `msg-handle-access` which in turn calls `handle-access`, which is where all the magic happens.

mtgred:
👍

PR #2187, opened 2016-11-30
Batch of fixes

### LINKED ISSUE #2168: Misusing Blockade Runner
After using Blockade Runner, I clicked Sneakdoor Betta in my hand instead of in the list. This installed Sneakdoor Beta, then I clicked Sneakdoor Beta in the list in hopes of fixing it. At that point the game seemed to crash. I was no longer able to interact with the game or type in the chat, and when I left the game and rejoined as a spectator it did not appear. Thanks for the awesome work you guys!

### LINKED ISSUE #2171: Jackson shuffle issue when targeting 0 cards (presumably SMC is similar)
When using Jackson to shuffle for 0 (on a run following Indexing, though presumably it doesn't matter), I tried to Jackson, selecting 0 cards.  Now AFAIK, that should still shuffle, unless I'm just flat out wrong about the rules.  It didn't.  Presumably this happens with other targeted shuffle effects when maybe it shouldn't (SMC only one that comes to mind at the moment for me).

By all means, tell me I'm an idiot if I'm wrong.

### LINKED ISSUE #2175: Bug with Plan B after Medical Breakthrough & Mushin No Shin
From my friend who doesn't have or want a GitHub account:

> Using Firefox 49.0.2 on fully patched Ubuntu Linux 14.04.
> 
> Playing as Corporation, I had already scored one Medical Breakthrough agenda (https://netrunnerdb.com/en/card/05005) which states:
> "Lower the advancement requirement of each Medical Breakthrough by 1." I have found that this effect is implemented correctly in jinteki.net when subsequent Medical Breakthrough agendas are scored independently.
> 
> However, with another Medical Breakthrough sitting in my hand, I used Mushin No Shin (https://netrunnerdb.com/en/card/05015) to install a Plan B asset (https://netrunnerdb.com/en/card/05023) with 3 agenda counters on it. The runner accessed the Plan B asset on their turn and I was prompted by Plan B to select the agenda in my hand to score. It would not allow me to score the Medical Breakthrough agenda in my hand (the second one) with only 3 agenda counters on Plan B.
> 
> I did try manually adding another agenda counter to Plan B and having the runner access it again and this time it did score the second Medical Breakthrough - but with 4 agenda counters. So it appears that Plan B does not recognize that previous Medical Breakthroughs have been scored and to adjust the number of required agenda counters accordingly. Other "regular" agendas appear to be scored correctly by Plan B.

### LINKED ISSUE #2177: "Showing Off" event not working with Medium (Or multiple Mediums)
Currently, when you use the access ability of Showing Off, instead of accessing cards from the bottom of RND upwards, it just accesses from the top anyway.

### LINKED ISSUE #2178: Magnet ICE not making hosted Parasite (when directly installed onto it) blank
Parasite was still fully interacting with Magnet when directly installed onto. Gained a counter, and was also trashing Magnet when Null's ID ability was used to lower the ICE's STR to zero.

### LINKED ISSUE #2183: Find the Truth + Top Hat interaction
If the first successful run of the turn is on R/D, the prompt for Top Hat triggers first, whereas  you would want to trigger Find the Truth to know whether to use Top Hat.

JoelCFC25:
* Fix #2168: Switch to use card targeting in Blockade Runner
* Fix #2171: Add a `:cancel-effect` to `rfg-and-shuffle-rd-effect` so you can Jackson back 0 cards and still get a reshuffle of R&D
* Fix #2175: Make Plan B check for altered agenda advancement requirements
* Fix #2177: Overhaul Showing Off to work with current access logic
* Fix #2178: Have Magnet blank all cards that get hosted directly onto it
* Fix #2183: Make Find the Truth and Top Hat `:interactive true` so the Runner can choose the order of their resolution
* Stop Palana Foods from firing if 0 cards are drawn
* Stop Scarcity of Resources from charging additional credits when Apex installs a facedown resource
* Fix incredibly rare bug--a Runner with Gang Sign installed runs and accesses an advanced Plan B that the Corp uses to score an agenda from hand, resulting in the "Card from hand" access prompt generated by Gang Sign getting shown to the Corp

### DIFF
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -786,15 +786,17 @@
     0
     {:req (req (pos? (:advance-counter (get-card state card) 0)))
      :effect
-     (effect (resolve-ability
-              {:prompt "Choose an Agenda in HQ to score"
-               :choices {:req #(and (is-type? % "Agenda")
-                                    (<= (:advancementcost %) (:advance-counter (get-card state card) 0))
-                                    (in-hand? %))}
-               :msg (msg "score " (:title target))
-               :effect (effect (score (assoc target :advance-counter
-                                             (:advancementcost target))))}
-              card nil))}
+     (req (doseq [ag (filter #(is-type? % "Agenda") (get-in @state [:corp :hand]))]
+            (update-advancement-cost state side ag))
+          (resolve-ability state side
+            {:prompt "Choose an Agenda in HQ to score"
+             :choices {:req #(and (is-type? % "Agenda")
+                                  (<= (:current-cost %) (:advance-counter (get-card state card) 0))
+                                  (in-hand? %))}
+             :msg (msg "score " (:title target))
+             :effect (effect (score (assoc target :advance-counter
+                                           (:current-cost target))))}
+           card nil))}
     "Score an Agenda from HQ?")
 
    "Political Dealings"
--- a/src/clj/game/cards/events.clj
+++ b/src/clj/game/cards/events.clj
@@ -993,11 +993,17 @@
                       {:replace-access
                        {:msg "access cards from the bottom of R&D"
                         :delayed-completion true
-                        :effect (req (swap! state assoc-in [:corp :deck]
+                        :effect (req (when-completed (resolve-ability state side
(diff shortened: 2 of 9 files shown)


mtgred:
👍

PR #2188, opened 2016-12-01
Fix #2180, Obelus + Hades Shard interaction

### LINKED ISSUE #2180: using Hades shard  on R+D run doesn't count for Obelus-Draw
My opponent used Hades Shard on a R+D run (Medium with 0 counters installed too) and only draw one card automaticly (ten cards in archives at this moment, only 6 in stack left).

nealterrell:
With test.


mtgred:
👍

PR #2192, opened 2016-12-06
add Martial Law cards

JoelCFC25:
Adds several Martial Law cards, plus a simplification of Chrysalis (the "when accessed" encounter just happens, it's not optional for the Corp).


mtgred:
👍

PR #2206, opened 2016-12-09
Admin paths for version string and for refreshing all-cards

nealterrell:
Moves app version information into the database, instead of an environment variable. 

Adds a URL path `/admin/version` for updating this version string. The string is used when referencing `app.js` in the HTML; a new string forces browsers to retrieve a new copy of the file. Updating the string while the site is live will cause all future connections to pull the new game script without having to reboot Node.

Adds a URL path `/admin/init`, which resends the "initialize" data sent to the game server upon initial web server connection. This data includes the list of all cards in the game, used for Adam's directives, Targeted Marketing, Rebirth, etc. If the initial connection fails, or if the card data sent to the server is incomplete, this URL will force the latest data from the database to be sent to the game server, without having to reboot Node.

Moves `/announce` to `/admin/announce` for consistency.

### DIFF
--- a/server.coffee
+++ b/server.coffee
@@ -330,7 +330,9 @@ app.use passport.session()
 app.use stylus.middleware({src: __dirname + '/src', dest: __dirname + '/resources/public'})
 app.use express.static(__dirname + '/resources/public')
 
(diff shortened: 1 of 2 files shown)


mtgred:
👍

PR #2207, opened 2016-12-09
implement Nihongai Grid, minor upgrade enhancements

JoelCFC25:
Implements Nihongai Grid. The tricksy bit was taking the deck (a sequence) and vectorizing it to reuse all the swapping code elsewhere, then turning it back into a sequence. 

Fully automates Will-o'-the-Wisp and adds interactivity checks to it and Bernice Mai, so you can do this properly like you ought to be able to: 

screen shot 2016-12-09 at 2 59 03 pm


mtgred:
👍

PR #2208, opened 2016-12-10
implement Interdiction

JoelCFC25:
Thanks to @Saintis great work on refactoring the flags system awhile back, this card is much easier now. Has a `:leave-play` effect in the _exceedingly_ unlikely event that the Corp manages a Plan B score on the Runner's turn. :P


mtgred:
👍

PR #2215, opened 2016-12-12
implement Manta Grid

JoelCFC25:
There is an extra layer of care in here for what I think would be an _exceedingly rare_ situation in which Manta Grid can trigger but the Runner still manages to trash it before the end of their turn. I think you could in theory have this happen with either a Data Breach or a Doppelganger "free" run that happens after you're out of clicks and you can resurrect an Imp mid-run, allowing you to trash Manta Grid after its effect has occurred.


JoelCFC25:
Back to the drawing board. This card is so dumb

PR #2217, opened 2016-12-13
implement On the Lam

JoelCFC25:
The first installable Runner event. It's a little weird in that the card will go to the Heap immediately but then come out when you select a host for it.


mtgred:
👍

PR #2225, opened 2016-12-14
A few card fixes

### LINKED ISSUE #2212: After accessing Sapper, the Runner accesses an additional card
I've noticed in a couple of games that if the Runner accesses Sapper from HQ (not sure about R&D) the Runner gets to access an additional card. This happens whether or not the runner trashes Sapper.

### LINKED ISSUE #2220: MaXXX and Motivation
Dear Developers,

Today I found a bug which is the following:
If I play MaXXX - Maximum Punk Rock and play Motivation, then at the beginning of the turn Motivation always triggers first. I'd like to trigger MaX's ability first, then motivation, but the game doesn't let me do it.

I would be very grateful if you could fix this bug.

Thank you: Meronim

### LINKED ISSUE #2222: Feature Suggestion: Friends in High Places to print an install message in log
It would be neat if Friends in High Places printed a message in the log telling you which face-up cards were installed in which servers like Team Sponsorship, Crick et al do.

### LINKED ISSUE #2244: London Library + On the Lam funny interaction
At the moment, it seems that the _condition_ of London Library to trash any **programs** at the end of the Runners turn does not take into account the **type** of card **hosted**, at least I've ran into a funny interaction where I've hosted On the Lam on London Library and at the end of my turn, On the Lam was trashed :D

JoelCFC25:
* Fix #2244: Only trash programs on London Library when turn ends (sparing On the Lam)
* Fix #2222: Have Friends in High Places identify cards installed from Archives that were face up.
* Fix #2220: Motivation needs to be able to be triggered after MaxX if the Runner wishes
* Fix #2212: Prevent a lingering prompt issue by preventing the Runner from seeing "Trash Sapper?"  until after the subroutine has fully resolved
* Use `runner-phase-12` in Hyperdriver so it can be fired before copies of Data Folding trigger.

### DIFF
--- a/src/clj/game/cards.clj
+++ b/src/clj/game/cards.clj
@@ -5,7 +5,8 @@
                     :msg (msg "trash " (:title target))
                     :choices {:req #(and (installed? %)
                                          (is-type? % "Program"))}
-                    :effect (effect (trash target {:cause :subroutine}))})
+                    :effect (effect (trash target {:cause :subroutine})
+                                    (clear-wait-prompt :runner))})
 
 (def trash-hardware {:prompt "Choose a piece of hardware to trash"
                      :label "Trash a piece of hardware"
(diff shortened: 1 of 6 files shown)


nealterrell:
👍

PR #2226, opened 2016-12-14
Fix The Foundry interaction with Accelerated Beta Test

### LINKED ISSUE #1672: When Accelerated Beta Test triggers The Foundry, cards in the temporary zone should be shuffled into R&D

nealterrell:
Not that anyone was holding their breath.

Make the `rez` function work with `when-completed`. Make ABT watch for shuffle effects during ice installs; abort the next prompt if a shuffle is detected. ABT now auto-trashes once all ice are installed, or if there are no ice in top 3. A prompt is shown in the latter case. The wording of the prompt is a little silly; if it's too unprofessional, we can change it. See line 44 in agendas.clj.

Fix #1672.

### DIFF
--- a/src/clj/game/cards/agendas.clj
+++ b/src/clj/game/cards/agendas.clj
@@ -17,26 +17,47 @@
 
    "Accelerated Beta Test"
    (letfn [(abt [n i]
-             {:req (req (pos? i))
-              :prompt "Select a piece of ICE from the Temporary Zone to install"
-              :choices {:req #(and (= (:side %) "Corp")
-                                   (ice? %)
-                                   (= (:zone %) [:play-area]))}
-              :effect (req (corp-install state side target nil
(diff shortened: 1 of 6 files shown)


src/clj/game/cards/agendas.clj
```diff
@@ -17,26 +17,47 @@
...
+                                                       (effect-completed state side eid)))
+                                                 (do (doseq [c (get-in @state [:corp :play-area])]
+                                                       (move state side c :deck))
+                                                     (shuffle! state side :deck)
+                                                     (effect-completed state side eid))))))
+                :cancel-effect (req (doseq [c (get-in @state [:corp :play-area])]
+                                      (system-msg state side "trashes a card")
+                                      (trash state side c {:unpreventable true})))}
+               {:prompt "None of the cards are ice. Say goodbye!"
+                :choices ["I have no regrets"]
```
erbridge: 👍

PR #2229, opened 2016-12-14
fix Blacklist interaction with Deja Vu, Trope, Ark Lockdown

### LINKED ISSUE #1946: Rezzed Blacklist not preventing Trope from moving cards out of Heap
This one surprises me, since `move` has a check to see that the origin zone isn't locked.

JoelCFC25:
Fixes #1946. 

The locked zone check in `move` doesn't seem to be working at all, so I stole the one that does work from `runner-install`. Then we add a "force" option to `move` so that you can still perform drag and drop manually.


nealterrell:
👍

PR #2231, opened 2016-12-14
Allow volume control for sounds.

nealterrell:
Only real complaint regarding the sounds so far has been volume, so I added a volume slider to My Account. It's not SUPER pretty, but it's functional. I also cleaned up the alt-art editor so it won't display a list of blank entries if the alt-art GET request completes before the all-cards GET request.

image

Works mid-game too, once you hit "Update Profile".

This also renames "My Account" to "Settings", and places it  before "About" (the traditional "last menu") in the carousel.


mtgred:
👍

PR #2233, opened 2016-12-14
Extend when-completed to tagging runner

### LINKED ISSUE #2228: Account Siphon incorrectly gives money before tags can be avoided.
Scenario: a runner makes an Account Siphon run on an undefended HQ (surprise Hayley Siphon!) with one credit in the bank and New Angeles City Hall on the field. After using the run ability, they have 11 credits in the bank and the interface asks them if they want to avoid tags, at which point they can use NACH's tag avoidance to pay 4 and go down to 7 credits post-run.

This behavior is incorrect, as the runner should not be able to use their ill-gotten credits until after the opportunity to avoid tags has passed, as per the FAQ: http://ancur.wikia.com/wiki/FAQ_1.2.18

nealterrell:
Fixes #2228 and a few other issues, by enabling `tag-runner` to be awaited with `when-completed`.

Also fixes an interaction with Accelerated Beta Test and Team Sponsorship... I think.

I don't know why there are all these other commits here, oh well.

### DIFF
--- a/src/clj/game/cards/agendas.clj
+++ b/src/clj/game/cards/agendas.clj
@@ -128,7 +128,8 @@
                             :effect (req (rez-cost-bonus state side (- (get-in card [:counter :agenda] 0))))}}}
 
    "Breaking News"
-   {:effect (effect (tag-runner :runner 2))
+   {:delayed-completion true
+    :effect (effect (tag-runner :runner eid 2))
     :silent (req true)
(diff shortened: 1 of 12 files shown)


erbridge:
Looks like the branch was started on top of some other changes. Maybe just do an interactive rebase onto `master` and drop the unwanted commits?

nealterrell:
OK, I think I did it.

PR #2235, opened 2016-12-15
Old Hollywood Grid refresh

### LINKED ISSUE #2169: Old Hollywood Grid vs Gang Sign
I had Old Hollywood Grid installed and rezzed on HQ, scored an agenda, runner fired Gang Sign and was apparently forced to steal an agenda from HQ with it. Later during a regular run on HQ, Old Hollywood Grid did function as expected, denying the steal on an accessed agenda.

kevkcc:
I think this is the way to go but tbh, this one was sort of tricky and might warrant a review. Copies ideas from Red Herrings. Fixes #2169 

Instead of registering a run-flag as before, we now register a persistent-flag when the :pre-steal-cost event triggers (even though there is no actual additional cost, we avoid unnecessary triggers over using :pre-access-card). The persistent flag is immediately cleared after the access, upon a new event hook, :post-access-card.

### DIFF
--- a/src/clj/game/cards/upgrades.clj
+++ b/src/clj/game/cards/upgrades.clj
@@ -355,14 +355,22 @@
     :leave-play (req (enable-run-on-server state card (second (:zone card))))}
 
    "Old Hollywood Grid"
-   (let [ohg {:req (req this-server)
-              :effect (effect (register-run-flag!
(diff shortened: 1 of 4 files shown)


nealterrell:
👍

Issue #2204, opened 2016-12-09
SMC menu priority issue.


poi2000:
I'm assuming this was an issue with priorities in layered menus.As runner, when I hit the Run button for my action, I had trouble selecting the server to run on because the Self-modifying code was right underneath the server menu and when I tried clicking on most of the button to select a server, it brought up the SMC menu instead. The far edge of the menu was clickable but it served as somewhat of a nuisance.

JoelCFC25:
Thanks for the report. I've noticed this recently too, especially when programs occupy the 2nd row of the rig. Notice in the picture below that the R&D "button" of the Run menu is not lit up in orange like it should be on hover (oops, guess my pointer didn't appear in the screen grab), and also that the virus counters on Cache are above everything. 

![screen shot 2016-12-08 at 8 45 32 pm](https://cloud.githubusercontent.com/assets/10407969/21035844/8806585a-bd87-11e6-9ddd-c21653478e3e.png)

JoelCFC25:
This probably tracks back to #2160.

lpoulter:
I did a quick test and this look to be a z-index issue. The z-index of the `.gameboard .counters`is set to 10 when you disable this the issue is fixed. I'm not sure why however as the run menu has z-index 40.

DeepSpawn:
My bad, I will look into this bug and see if I can get a fix in this weekend.

lpoulter:
i've found the issue. #2189 set the z-index of `right-inner-leftpane` to 1. This is inherited by the run sub menu. As counters is z-index 10 it appears higher than the sub-menu. @erbridge you set the index to 1 in #2804

DeepSpawn:
Yep, that is exactly what is happening. I just finished going down the rabbit hole of reading up on the stacking context rules.

Setting right-inner-leftpanel to Z level > 10 will fix this issue, but I am unsure if this will have any other consequences. Are we ok to go ahead and make that change?

erbridge:
Well discovered. I wish we weren't using `z-index` for layout at all...

lpoulter:
@erbridge out of intrest what are alternatives to z-index?

erbridge:
Natural DOM ordering. Things higher in the DOM appear beneath things below them.

PR #2237, opened 2016-12-16
issue #2204: increase run menu z level

DeepSpawn:
This fixes a regression where adding a z-level to the right-inner-leftpane caused a stacking context to be formed, which resulted in the run menu appearing below counters on cards (with z-level 10).

This turns z-level up to 11 to stop this from happening.

### DIFF
--- a/src/css/base.styl
+++ b/src/css/base.styl
@@ -1459,7 +1459,7 @@ nav ul
 
       .right-inner-leftpane
         width: 186px
-        z-index: 1
+        z-index: 11
         justify-content(space-between)
         display-flex()
         flex-direction(column)


DeepSpawn:
cc @lpoulter

lpoulter:
Looks good to me and certainly fixes the issue. However I don't know the code base well enough to understand what elements might be affected. I would also question why the z-index was added in the first place.

DeepSpawn:
So the z-index was added so that the list of choices gets rendered on top of your hand when you are presented with a large list of cards to choose from, eg when picking a card from deck from a logos trigger.

I am happy to investigate alternative solutions the problem that caused me to raise #2160 rather than just fixing the z level if you think that is the better approach.

JoelCFC25:
Awesome--thanks to both of you for the investigation and the cure!

erbridge:
A better fix (IMO) is in #2239. It fixed the natural DOM order to avoid needing to hack the `z-index`.

nealterrell:
https://www.youtube.com/watch?v=4xgx4k83zzc

Thanks for your efforts to fix this. I'm going to agree that reordering the DOM elements is a preferred solution.

PR #2238, opened 2016-12-16
band-aid fix for Deep Red

JoelCFC25:
Deep Red is completely broken (again) right now. This at least makes it functional, and you can sneak in a usage of Scheherazade to host a Caissa installed from the Heap by a fully advanced Pawn before responding to Deep Red's prompt. 

Still working on restoring automated Pawn advancement...


nealterrell:
👍

PR #2239, opened 2016-12-16
Order gameboard panes naturally

### LINKED ISSUE #2204: SMC menu priority issue.

erbridge:
Pull the button pane into its own component and reverse the pane order, to fix z-indexing.

Fixes #2204.

### DIFF
--- a/src/cljs/netrunner/gameboard.cljs
+++ b/src/cljs/netrunner/gameboard.cljs
@@ -95,7 +95,7 @@
         (reset! lock false)
         (case (:type msg)
           ("do" "notification" "quit") (do (swap! game-state (if (:diff msg) #(differ/patch @last-state (:diff msg))
-                                                                             #(assoc (:state msg) :side (:side @game-state))))
+                                                                 #(assoc (:state msg) :side (:side @game-state))))
                                            (swap! last-state #(identity @game-state)))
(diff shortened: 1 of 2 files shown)


src/cljs/netrunner/gameboard.cljs
```diff
@@ -327,18 +326,15 @@
             height (.-scrollHeight div)]
         (when (or (zero? scrolltop)
                   (< (- height scrolltop (.height (js/$ ".gameboard .log"))) 500))
-          (aset div "scrollTop" height))))
-
-
-    om/IDidMount
+          (aset div "scrollTop" height)))) om/IDidMount
```
DeepSpawn: Looks like this got pulled into the previous line during the reformatting

erbridge: That's still valid code, since this is just a list of arguments, but I'll fix it.

mtgred:
👍

PR #2240, opened 2016-12-17
Make the run arrow split into segments instead of overlaying

erbridge:
This should fix the layout and allow menus and cards to be interacted with, without being blocked by the arrow, while still allowing it to be visible. It looks like this:

screenshot from 2016-12-17 07-17-21

The downside is that it pushes the cards apart during a run. Ideally that would be smoothed out by a height transition, but that would require a fair amount of extra work.

### DIFF
--- a/src/cljs/netrunner/gameboard.cljs
+++ b/src/cljs/netrunner/gameboard.cljs
@@ -842,6 +842,14 @@
        [:div (str (+ hand-size-base hand-size-modification) " Max hand size")
         (when me? (controls :hand-size-modification))]]))))
 
+(defn run-arrow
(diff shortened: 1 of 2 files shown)


domtancredi:
That is pretty amazing. It fixes the top // etc.

jbrz:
Perhaps a solution to the server height-growth would be to stack/compress passed ice?

image

erbridge:
That could be problematic with cards hosted on ice. Also the layering is such that ice closer to the server is higher, so their menus appear on top.

jbrz:
Good point, didn't think about that. Just spit-balling ideas.

PR #2247, opened 2016-12-19
Fix Rumor Mill to affect all cards, not just installed ones

### LINKED ISSUE #2241: Rumor Mill + trashable "as 2 agenda points" executives
Rumor Mill is only applying to installed Corp cards right now, when it should apply equally to cards in R&D or HQ. Thus, the Runner is wrongly getting the trashable execs as 2 agenda points if Rumor Mill is active and the Runner trashes one of these from R&D or HQ.

kyzen:
Fixes #2241

Rumor Mill effect previously only checked against installed cards, now checks against installed cards as well as cards in HQ, R&D, and archives.  Also updated unit test to check the situation mentioned in the issue.


nealterrell:
Nice job.

nealterrell:
👍

PR #2264, opened 2016-12-23
Fix icebreaker abilities when encountering ice.

nealterrell:
Was not calculating "dynamic" abilities correctly, leading to off-by-one errors whenever a "Match strength with X" ability was added during a run.


mtgred:
👍

PR #2265, opened 2016-12-23
Fix server response when saving a new deck.

nealterrell:
Don't know why, but the database data returned from insert is now in `data.ops`, not just `data`.


mtgred:
👍

PR #2266, opened 2016-12-23
Fix On the Lam.

nealterrell:
This is the first Event card that can be hosted on others, so `host` had to be changed to account for the ability.


mtgred:
👍

PR #2271, opened 2016-12-24
Make the central pane overflow

### LINKED ISSUE #2261: No Horizontal Scrolling on Firefox
At some point in the last three days I lost the ability to scroll horizontally. Normally this isn't that problematic, but when there is a long row of servers, resources, hardware, or programs, my screen aligns with the outermost one and I lose the ability to interact with any cards or command buttons that are not in the frame of view.

erbridge:
Causes the central pane to scroll independently of the main window, so the left and right panes stay visible and in position.

I looked at adding shadows to indicate the scrollable area's edges using http://lea.verou.me/2012/04/background-attachment-local/, but that solution requires a solid background colour.

Fixes #2261.

### DIFF
--- a/src/css/base.styl
+++ b/src/css/base.styl
@@ -1510,6 +1510,7 @@ nav ul
     flex(1)
     display-flex()
     flex-direction(column)
+    overflow: auto
 
     .card-wrapper
       margin: 3px 6px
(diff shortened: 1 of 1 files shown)


nealterrell:
So this makes the chat box stay put when the game area goes wide?

erbridge:
Yup.

nealterrell:
🥇

Issue #2276, opened 2016-12-25
Archives Interface Infinite Run - Cannot Jack OUt


Chemscribbler:
When playing Whizzard I ran an empty archives with Archives Interface out. I clicked the option to remove a card from the game. This led to a Run is successful pop-up, which then reprompted me to pick the AI choice again. My Temujin emptied out additional times as well. /close-prompt and /end-run were not successful, but did log a run ended and prompt closed.

JoelCFC25:
Thanks for the report. Archives Interface just needs a `not-empty` check of Archives before prompting.

PR #2279, opened 2016-12-29
Archives Interface is not triggered when there's no cards to trash

lostie:
Fixes https://github.com/mtgred/netrunner/issues/2276

Adds a simple check not to open **Archives Interface** prompt if Archives is empty

### DIFF
--- a/src/clj/game/cards/hardware.clj
+++ b/src/clj/game/cards/hardware.clj
@@ -9,7 +9,7 @@
     {:successful-run
      {:silent (req true)
       :delayed-completion true
-      :req (req (= target :archives))
+      :req (req (and (= target :archives) (not-empty (:discard corp))))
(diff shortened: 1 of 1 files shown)


src/clj/game/cards/hardware.clj
```diff
@@ -9,7 +9,7 @@
     {:successful-run
      {:silent (req true)
       :delayed-completion true
-      :req (req (= target :archives))
+      :req (req (and (= target :archives) (not-empty (:discard corp))))
```
nealterrell: Looks good. Clojure people would prefer using the `seq` function to test for not-emptiness, as using `seq` on something that is empty will return `nil`. I think that's less readable, and prefer `not-empty`.

domtancredi:
Looks solid to me. Good work!

PR #2286, opened 2017-01-03
Refactored to add Mongoose, and model Post for native News

domtancredi:
- Added Mongoose (ORM!), 
- created a model (Post, instead of *News* because the singular version of News is no fun), 
- setup the logic to retrieve news,
- setup the logic for an admin to post to create _new_ news items


domtancredi:
Should resolve #2051

models/post.coffee
```diff
@@ -0,0 +1,14 @@
+mongoose = require('mongoose')
```
erbridge: Thank you for not inlining this into `server.coffee` :).

package.json
```diff
@@ -33,6 +33,7 @@
     "mkdirp": "^0.5.1",
     "moment": "^2.15.1",
     "mongodb": "^2.2.10",
+    "mongoose": "^4.7.5",
     "mongoskin": "^2.1.0",
     "node-trello": "github:adunkman/node-trello#eb3b4fc53e5d3c2bd4f913a806a88dfe8fa43218",
```
erbridge: Let's remove `node-trello` if we're no longer using it, too.

mtgred:
Indeed. What's cool about the Trello connection is that it acts as a minimalist CMS.

I'm also not fan of Mongoose or any ORM. It adds another dependency, layer of complexity, stuffs to learn and maintain. It also makes the data harder to change. What's great about MongoDB compared to an SQL database or Datomic is precisely that's it has no schema. The simplicity of having JSON data means you can change the structure quickly without having to write schema migration scripts. It's kind of similar to data structures in C, Javascript or Clojure: just arrays and key-value dictionaries. Instead of C++, Java where you have to define classes and interfaces that made your data structures harder to understand, inspect or change. There are certainly advantages to schemas or object oriented programming, but i prefer the simplicity bare data structures.

erbridge:
I actually disagree. I find an ORM makes it easier to reason about the structure of the data. And even without it, you still have an implicit schema. If you want to change the structure of the data, you would still need to migrate it, but as the schema isn't explicit, it becomes harder to work out what exactly has changed.

Having said that, I don't think the data needs of this project are heavy enough that it matters either way.

models/post.coffee
domtancredi: What do you mean not inlining it? It is in *server.coffee* because it needs to connect to the database.

erbridge: I'm sorry, that wasn't sarcastic at all. I was referring to the fact that this is a separate file from `server.coffee`, which I keep meaning to break down into smaller components. All good and positive :).

domtancredi:
I can extend this pull request to include the UI CRUD for News. 

An ORM is as essential an architectural element as an API for products of a sufficient size and scope (IMO). They abstract data, set a structure to it, and make it easier for team members to add new features and reduce mistakes. Data can change but should data structure really change that much? We aren't talking about adding migrations or seeders or anything like that and can play a little fast and loose with data structure

Issue #2245, opened 2016-12-19
An offer you can't refuse scoring issue


lostie:
I came across a minor issue (unfortunately, I forgot to grab a screenshot of it), but when playing **An Offer You can't refuse**, the runner chose to add **An Offer You can't refuse** to the agenda score as 1 agenda point, but I somehow ended up with **Celebrity Gift** (which was in Archives) in the score agenda instead. The agenda point was correct, just the actual card that should be be in the agenda score that wasn't.

PR #2290, opened 2017-01-05
Add correct Offer You Can't Refuse card to Corp score area

lostie:
Solves https://github.com/mtgred/netrunner/issues/2245

There's some instances the Runner takes a little while to decide. In the
meanwhile, if some other cards go into Archives, the last card discarded
would be the one added to the Corp score area.

This change ensures the actual card that is played to trigger the Runner
decision is the one added to the Corp score area if he/she decides not
to run.

### DIFF
--- a/src/clj/game/cards/operations.clj
+++ b/src/clj/game/cards/operations.clj
@@ -77,7 +77,7 @@
                                      :effect (effect (clear-wait-prompt :corp)
                                                      (game.core/run eid serv nil card))}
                        :no-ability {:effect (effect (clear-wait-prompt :corp)
-                                                    (as-agenda :corp (last (:discard corp)) 1))
+                                                    (as-agenda :corp (some #(when (= (:cid card) (:cid %)) %) (:discard corp)) 1))
                                     :msg "add it to their score area as an agenda worth 1 agenda point"}}}
                     card nil)))}
 
--- a/src/clj/test/cards/operations.clj
+++ b/src/clj/test/cards/operations.clj
@@ -107,6 +107,22 @@
       (prompt-select :corp (refresh co))
       (is (= 15 (:credit (get-corp))) "Corp gained 6 credits for Back Channels"))))
 
+(deftest an-offer-you-cant-refuse
+  ;; An Offer You Can't Refuse - exact card added to score area, not the last discarded one
+  (do-game
+    (new-game (default-corp [(qty "Celebrity Gift" 1) (qty "An Offer You Can't Refuse" 1)])
+              (default-runner))
+    (play-from-hand state :corp "An Offer You Can't Refuse")
+    (prompt-choice :corp "R&D")
+    (core/move state :corp (find-card "Celebrity Gift" (:hand (get-corp))) :discard)
+    (is (= 2 (count (:discard (get-corp)))))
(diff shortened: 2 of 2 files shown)


lostie:
@Saintis @nealterrell could you please review?

src/clj/test/cards/operations.clj
```diff
@@ -1358,3 +1358,19 @@
...
+    (new-game (default-corp [(qty "Celebrity Gift" 1) (qty "An Offer You Can't Refuse" 1)])
+              (default-runner))
+    (play-from-hand state :corp "An Offer You Can't Refuse")
+    (prompt-choice :corp "R&D")
+    (core/move state :corp (find-card "Celebrity Gift" (:hand (get-corp))) :discard)
+    (is (= 2 (count (:discard (get-corp)))))
+    (prompt-choice :runner "No")
+    (is (= 1 (:agenda-point (get-corp))) "An Offer the Runner refused")
+    (is (= 1 (count (:scored (get-corp)))))
+    (is (= "An Offer You Can't Refuse" (:title (first (:scored (get-corp))))))
```
nealterrell: A nice utility in the test framework is `find-card`, which takes a title and collection and returns the card with that title if it is in the colletion, `nil` otherwise. You can use this with `is` pretty nicely:

`(is (find-card "An Offer You Can't Refuse" (:scored (get-corp))) ...)`

which will fail if the card can't be found.

src/clj/game/cards/operations.clj
```diff
@@ -77,7 +77,7 @@
                                      :effect (effect (clear-wait-prompt :corp)
                                                      (game.core/run eid serv nil card))}
                        :no-ability {:effect (effect (clear-wait-prompt :corp)
-                                                    (as-agenda :corp (last (:discard corp)) 1))
+                                                    (as-agenda :corp (first (filter #(= (:cid card) (:cid %)) (:discard corp))) 1))
```
nealterrell: A slightly better way to do this pattern (finding the first instance of something matching a filter predicate):

`(some #(when (= (:cid card) (:cid %)) %) (:discard corp))`

`some` takes a function and a collection, and returns the first first truthy value found when applying the function to each element of the collection in turn. Here, the `when` will test your condition and return a card if the card fits the condition, otherwise nil; the `some` will continue iterating cards until it finds a card that the function returns non-nil with, i.e., a card that matches the filter.

src/clj/test/cards/operations.clj
```diff
@@ -1358,3 +1358,19 @@
       (prompt-select :corp (refresh vanilla))
       (is (not= "Wetwork Refit" (:title (first (:hosted (refresh vanilla)))))
           "Wetwork Refit is not hosted on Vanilla"))))
+
+(deftest offer-you-cant-refuse
```
erbridge: We normally alphabetize the tests by the main card they refer to.

src/clj/test/cards/operations.clj
lostie: I was originally trying to do something like that, but didn't know you could use 

`(is (find-card ..))` 

directly. I can have it changed to follow that approach instead.

lostie:
♻️ (ed) with the suggested improvements 🎩

Issue #2016, opened 2016-09-29
[Run timing structure] Hades Shard vs Jackson Howard


Repulze:
A few patches ago, a prompt was coded in so that it allowed the Corp to take "further action" (step 4.3) just before a successful run. This is how it should work:
1. Runner decides to continue the run (step 4.2)
2. Paid ability window/"Corp takes further action" (step 4.3) - Jackson Howard.
3. Successful run (step 4.4) - Hades Shard.

However, this is how it currently works:
1. Runner decides to continue the run (step 4.2)
**2. Paid ability window (step 4.3) - Hades Shard first _(this is wrong)_, then Jackson Howard.**
3. Successful run (step 4.4) - Runner is forced to access and has no chance to install Hades Shard.

This forces the Runner to install Hades Shard during the ability window, before they actually make a successful run. Many players started waiting to see if I would install Hades first before deciding to pop Jackson. I would have probably done the same since I've always (incorrectly) installed Hades Shard in the ability window.

I believe this is a problem because there is no successful run trigger for Hades. Maybe a code for Hades that says: "if Hades Shard is in grip, prevent (Archives) access and ask if the Runner would like to install Hades instead. If no, access Archives."

Once again, thanks for the work. o7

zaroth:
The way it should work is to make the shards in Grip register `:replace-access` event for runs on appropriate servers, with relevant labels. Current implementation is actually instant paid ability, activable only on position zero of these servers and leads to the issue you described.

However, I'm not aware of any precedents of cards that register triggers merely by being in somebody's hand - so this framework would need to be built first. Quite a bit of work for three underplayed cards, but maybe somebody will tackle it ;-)

Saintis:
Could maybe use the `:move-zone` trigger found in Subliminal Messaging to register an event when a Shard moves to the Grip, and then deregister it when they move away.

zaroth:
True, if Subliminal Messaging is finally integrated and fully working, these triggers should help :)

erbridge:
Problem with using `:move-zone` (at least in its current form) is that neither starting hands, nor drawing, trigger the event. Using `:runner-draw` is also problematic, since starting hands suppress those events, too.

PR #2303, opened 2017-01-09
Shard(s) replacement access

lostie:
At the moment if user performs a run on any central server and the run is successful, there's some occasions where the user has the opportunity to access cards and while doing it install the related Shard at no cost.

These changes ensure that if the user choose to access, the Shard can no longer be installed as a replacement access condition.

Related to https://github.com/mtgred/netrunner/issues/2016

**NOTE:** These changes do NOT solve the other main issue, which is the Shard(s) effect is a replacement access effect that occur at 4.4 step of a run. Currently it's handled as a trigger ability.


JoelCFC25:
Superseded by #2792.

PR #2305, opened 2017-01-11
Ravana 1.0, Card tests - Akamatsu Mem, Dorm Computer, Snare!, Desperado, CyberSolutions Mem Chip, Freedom

danhut:
Hi guys - I am a netrunner player and been looking to learn a functional language so some first contributions here.  Lots to learn on code base and Clojure still.  Wanted to submit some small changes first. 

Not sure if you'd rather bank them up for a larger pull - let me know.

Edit: Had a go at Ravana 1.0 - it will make the UI clear to show that Ravana fired, and let the corp target only rezzed bioroids when they fire the sub.  Not sure if you had a plan to do something more automated but seems adequate to me.

### DIFF
--- a/src/clj/game/cards/ice.clj
+++ b/src/clj/game/cards/ice.clj
@@ -1061,9 +1061,9 @@
    {:subroutines [end-the-run]}
 
    "Ravana 1.0"
-   {:subroutines [{:label "Resolve a subroutine on another piece of rezzed bioroid ice."
+   {:subroutines [{:label "Resolve a subroutine on another piece of rezzed bioroid ICE"
                    :choices {:req #(and (rezzed? %) (ice? %) (has-subtype? % "Bioroid"))}
-                   :msg (msg "Resolve subroutine on " (:title target))}]
+                   :msg (msg "resolve a subroutine on " (:title target))}]
     :runner-abilities [(runner-break [:click 1] 1)]}
 
    "Red Tape"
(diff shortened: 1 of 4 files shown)


src/clj/game/cards/ice.clj
```diff
@@ -1060,6 +1060,12 @@
    "Rainbow"
    {:subroutines [end-the-run]}
 
+   "Ravana 1.0"
+   {:subroutines [{:label "Resolve a subroutine on another piece of rezzed bioroid ice."
```
JoelCFC25: For consistency almost all messages use "ICE" (although in fairness, to match the printed rules and card texts, it probably should be all lowercase). You can drop the period at the end, too.

src/clj/game/cards/ice.clj
```diff
@@ -1060,6 +1060,12 @@
...
+                   :choices {:req #(and (rezzed? %) (ice? %) (has-subtype? % "Bioroid"))}
+                   :msg (msg "Resolve subroutine on " (:title target))}]
```
JoelCFC25: Make "Resolve" all lowercase and put "a" in between it and "subroutine". Currently your message will come out in the log as:  "Playername uses Ravana 1.0 to Resolve subroutine on ICE-name-here"

src/clj/game/cards/ice.clj
danhut: Thanks - pushed a new commit.

src/clj/game/cards/ice.clj
danhut: Thanks - will take more care on that.  Pushed a new commit.

PR #2310, opened 2017-01-12
Fix #2308 - Baba Yaga

### LINKED ISSUE #2308: Baba Yaga hosting already installed programs double charges you for installing
Hosting a program on Baba Yaga not directly from your grip causes the program to be treated as a fresh install on Baba Yaga. For example, when the following occurs:

1. Have both Baba Yaga and Self Modifying Code installed;
2. Fetch an icebreaker with Self Modifying Code, installing it;
3. Use the "host installed icebreaker..." function on Baba Yaga to host the newly installed icebreaker

The icebreaker is moved to being hosted by Baba Yaga, but is treated as a fresh install, while not forgetting about the previous install (that is, it charges you again for the install cost, and subtracts the associated MU, without freeing up the MU from the first install).

danhut:
This is a fix for ability 1 on Baba Yaga when a breaker is hosted from the in-play rig.  Memory is correct, and no install charge for the breaker.  Also updated test to include this ability.

Fix #2308

### DIFF
--- a/src/clj/game/cards/icebreakers.clj
+++ b/src/clj/game/cards/icebreakers.clj
@@ -230,7 +230,7 @@
                                          (not (has-subtype? % "AI"))
                                          (installed? %))}
                     :effect (req (when (host state side card target)
-                            (gain :memory (:memoryunits target))))}
+                                   (gain :memory (:memoryunits target))))}
          gain-abis (req (let [new-abis (mapcat (fn [c] (map-indexed #(assoc %2 :dynamic :copy, :source (:title c)
                                                                                :index %1, :label (make-label %2))
                                                                     (filter #(not= :manual-state (:ability-type %))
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/icebreakers.clj
```diff
@@ -229,7 +229,8 @@
                     :choices {:req #(and (has-subtype? % "Icebreaker")
                                          (not (has-subtype? % "AI"))
                                          (installed? %))}
-                    :effect (effect (runner-install target {:host-card card}))}
+                    :effect (req (when (host state side card target)
+                            (gain :memory (:memoryunits target))))}
```
nealterrell: Please fix the alignment of the `gain` function call.

danhut:
Done.  I was originally using Atom - moved to IntelliJ which handles this a lot better.

PR #2311, opened 2017-01-12
more Quorum cards

JoelCFC25:
Implements: Aaron Marrón, Tapwrm, Tracker, Fawkes, Sūnya, Peace in Our Time, and Şifr.

### DIFF
--- a/src/clj/game/cards/hardware.clj
+++ b/src/clj/game/cards/hardware.clj
@@ -680,6 +680,26 @@
                          :effect (effect (tag-runner :runner eid 1) (end-run))
                          :unsuccessful {:msg "bypass the current ICE"}}}]}
 
+   "Şifr"
+   {:in-play [:memory 2]
+    :abilities [{:once :per-turn
+                 :msg (msg "lower their maximum hand size by 1 and lower the strength of " (:title current-ice) " to 0")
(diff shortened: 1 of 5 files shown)


src/clj/game/cards/hardware.clj
```diff
@@ -680,6 +680,26 @@
...
+                                    :runner-turn-begins {:effect (effect (gain :runner :hand-size-modification 1)
+                                                                         (unregister-events card))}
+                                    :runner-turn-ends {:effect (effect (update! (dissoc card :sifr-target)))}
+                                    :pass-ice {:effect (effect (update! (dissoc card :sifr-target)))}} card)
+                                 (update-ice-strength current-ice))}]
+    :events {:runner-turn-begins nil
+             :pre-ice-strength nil
+             :runner-turn-ends nil
+             :pass-ice nil}}
+
```
nealterrell: Let's migrate the strength-modifying events into the permanent `:events` map, rather than "temporarily" registering them -- like how Datasucker works. Also, did you mean `:run-ends` instead of `:runner-turn-ends`?

Also, the `:runner-turn-begins` code for restoring hand size is another example of a need for a "floating" ability resolution framework... if Sifr somehow gets trashed before the next turn begins, the Runner won't get their +1 size.

Issue #2259, opened 2016-12-21
Easy to misclick Builder of Nations


richoward:
It would be better if the option to deal a meat damage from builder of nations was behind a submenu to prevent misclicks dealing damage to runners

danhut:
Submitted an enhancement for this - so you have to confirm the damage, or can cancel the ability.  Also made it once per turn max.

JoelCFC25:
In at #2316.

PR #2316, opened 2017-01-14
Builder of Nations enhancement Issue #2259 & made once per turn

danhut:
Submitted an enhancement for this - so you have to confirm the damage, or can cancel the ability.  Also made it once per turn max.  Also added test cases.

### DIFF
--- a/src/clj/game/cards/identities.clj
+++ b/src/clj/game/cards/identities.clj
@@ -739,9 +739,13 @@
    "Weyland Consortium: Builder of Nations"
    {:implementation "Damage triggered manually"
     :abilities [{:label "Do 1 meat damage"
+                 :once :per-turn
+                 :prompt "Do a meat damage from identity ability?"
+                 :choices (cancellable ["Yes"])
                  :delayed-completion true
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/identities.clj
```diff
@@ -739,9 +739,13 @@
...
    {:implementation "Damage triggered manually"
     :abilities [{:label "Do 1 meat damage"
+                 :once :per-turn
+                 :prompt "Do a meat damage from identity ability?"
+                 :choices (cancellable ["Yes"])
                  :delayed-completion true
-                 :msg "do 1 meat damage"
-                 :effect (effect (damage eid :meat 1 {:card card}))}]}
+                 :effect (req (when (= target "Yes")
+                                (do (damage state :runner eid :meat 1 {:card card})
```
nealterrell: Minor style thing: you don't need a `do` in the body of a `when`; the `do` is implicit, unlike `if`.

danhut: Thanks - yes I just realised that when sitting in the movies.  Fixed it up

PR #2317, opened 2017-01-14
fix Peace in Our Time

### LINKED ISSUE #2314: Peace in Our Time
Peace in our Time prints a message to the logs stating how many credits each player should get, but it doesn't actually give anyone any credits.

JoelCFC25:
* Fixes #2314: _Giving_ the credits instead of just talking about it would be a nice touch.


nealterrell:
:P

Issue #2320, opened 2017-01-17
Retarget Security Testing on turn begin


triorph-emacs:
Sometimes on turn startup you misclick your sectest, which is a pain in the ass. The only way to really fix this is to trash sectest and reinstall, and that doesn't work so well because it forces the access. 

It would be nice, if on clicking Security Testing, there is an option to select a server to retarget it to.

danhut:
If this were to exist it would need to only be allowed before you first click - or other action - such as Jak Sinclair right?

triorph-emacs:
Not necessarily. There are a lot of actions that jinteki allows the player to make that are technically illegal but allow you to fix mistakes as you bluster past, or just other actions that are illegal outside of start of turn but are still allowed (Sensie Actor's Union can be used immediately after install if you want to cheat). I see no reason why we can't trust the player not to abuse security testing in the same manner.

JoelCFC25:
FWIW, there are plenty of "when your turn begins" cards for both Runner and Corp that could be tightened up to restrict the timing of usage with `:req (req (:corp-phase-12 @state))`  (and the Runner analogue).

danhut:
I tried out a couple of possible implementations for this.  

The first one allows the runner to re-target security testing at any point in their turn - but still only lets it fire once per turn.  This just removes the existing check - commented out below which blocks re-targeting once a target server has been set.
```
   "Security Testing"
   (let [ability {:prompt "Choose a server for Security Testing" :choices (req servers)
                  :msg (msg "target " target)
                  ;:req (req (not (:server-target card)))
                  :effect (effect (update! (assoc card :server-target target)))}]
```
The second version allows re-targetting any time up until your first click has been taken - this does not include lost clicks such as from Wyldside

```
   "Security Testing"
   (let [ability {:prompt "Choose a server for Security Testing" :choices (req servers)
                  :msg (msg "target " target)
                  :req (req (not (contains? (into {} (get @state :turn-events)) :runner-spent-click)))
                  :effect (effect (update! (assoc card :server-target target)))}]
```

Personally I prefer the latter for more automation.

JoelCFC25:
Went in at #2323.

PR #2322, opened 2017-01-20
Allow Sec Test/Patron to be re-targetted Issue #2320

danhut:
This updates Sec Testing & Patron as follows:

Allow their target to be changed (in case of misclick), unless they have been used this turn, or you've already spent a click.  Also works for #1744 - except we assume the only reason someone would set Patron/SecTest manually in Step 1.2 would be if they are actually using it.

Added a couple of utility functions which I could not see existing code for to enable this.


danhut:
Did not commit fully.  Fixing and will resubmit

PR #2335, opened 2017-01-26
Fix #2331 - news team with blacklist issue

### LINKED ISSUE #2331: News Team not leaving archives after runner adds it to score area.
As the title states, and [as can be seen here](http://prnt.sc/dzhkdk), News team remains in archives after the runner chooses to add it to their score area.  
Issue might be related to #932

danhut:
Added a flag to News Team to force move to occur even if Blacklist is rezzed.  This seemed the best way to do this - the move function checks the current side for locked zone which means blacklist stops news team moving from corp discard.

Fix #2331

### DIFF
--- a/src/clj/test/cards/assets.clj
+++ b/src/clj/test/cards/assets.clj
@@ -735,8 +735,7 @@
     (core/rez state :corp (get-content state :remote1 0))
     (run-empty-server state :archives)
     (prompt-choice :runner "Add News Team to score area")
-    (is (= 2 (count (:scored (get-runner)))) "News Team added to Runner score area with Blacklist rez")
-    ))
+    (is (= 2 (count (:scored (get-runner)))) "News Team added to Runner score area with Blacklist rez")))
 
 (deftest net-police
   ;; Net Police - Recurring credits equal to Runner's link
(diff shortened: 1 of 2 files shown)


danhut:
Thanks Neal - updated based on this and re-committed

src/clj/test/cards/assets.clj
```diff
@@ -717,6 +717,27 @@
...
+    (is (= 2 (:tag (get-runner))) "Runner has 2 tags")
+    (run-empty-server state :archives)
+    (prompt-choice :runner "Add News Team to score area")
+    (is (= 1 (count (:scored (get-runner)))) "News Team added to Runner score area")
+    (trash-from-hand state :corp "News Team")
+    (core/rez state :corp (get-content state :remote1 0))
+    (run-empty-server state :archives)
+    (prompt-choice :runner "Add News Team to score area")
+    (is (= 2 (count (:scored (get-runner)))) "News Team added to Runner score area with Blacklist rez")
+    ))
```
erbridge: This should be on the previous line.

danhut: fixed those brackets.

PR #2342, opened 2017-01-29
Jeeves automation - Actual Pull Request now 

danhut:
Hi team,
I added automation for Jeeves to cover everything except Click abilities on cards.  This needed a bit of work.  

Can someone please confirm they are ok with the approach within the game engine as I am clearly still learning it (and Clojure).

1. Added event emitters for several click abilities that did not exist. New events were:

- corp-click-install 
- pay-click-operation
- pay-double-operation
- corp-click-purge
- corp-click-trash 

2. Made Jeeves listen for them
3. Jeeves has its own key in the @state atom which gets cleared each turn.  Here we track what events it has seen each turn
4. Added unit test

What remains if this is not a terrible way to do this...

- Clickable card abilities - via another event emitter
- More test cases

(Edit - added a :leave-play key to clear :jeeves from @state if it is trashed)

### DIFF
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -584,20 +584,18 @@
           ability {:label "Gain [Click]"
                    :msg "gain [Click]"
                    :once :per-turn
-                   :effect jeeves}]
+                   :effect jeeves}
+          cleanup (effect (update! (dissoc card :seen-this-turn)))]
 
     {:abilities  [ability]
-     :leave-play {:effect (effect (update! (dissoc card :seen-this-turn)))}
-     :trash-effect {:effect (effect (update! (dissoc card :seen-this-turn)))}
+     :leave-play {:effect cleanup}
+     :trash-effect {:effect cleanup}
      :events {
              :corp-spent-click
               {:effect (req (update! state side (update-in card [:seen-this-turn target] (fnil + 0) (second targets)))
                             (when (>= (get-in (get-card state card) [:seen-this-turn target]) 3)
-                                 (resolve-ability state side
-                                                  {:once :per-turn
-                                                   :effect jeeves
-                                                   :msg (msg "gain a [Click]")} card nil)))}
-              :corp-turn-ends {:effect (effect (update! (dissoc card :seen-this-turn)))}}})
+                                 (resolve-ability state side ability card nil)))}
+              :corp-turn-ends {:effect cleanup}}})
 
    "Kala Ghoda Real TV"
    {:flags {:corp-phase-12 (req true)}
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -584,22 +584,20 @@
           ability {:label "Gain [Click]"
                    :msg "gain [Click]"
                    :once :per-turn
-                   :effect jeeves}
-          matrix  (atom {})]
+                   :effect jeeves}]
 
     {:abilities  [ability]
-     :leave-play (req (reset! matrix {}))
-     :trash-effect (req (reset! matrix {}))
+     :leave-play {:effect (effect (update! (dissoc card :seen-this-turn)))}
+     :trash-effect {:effect (effect (update! (dissoc card :seen-this-turn)))}
      :events {
-             :corp-spend-click
-              {:effect (req (swap! matrix update-in [target] (fnil + 0) (second targets))
-                            (let [qty (get-in @matrix [target])]
-                              (resolve-ability state side
-                                             {:req (req (>= qty 3))
-                                              :once :per-turn
-                                              :effect jeeves
-                                              :msg (msg "gain a [Click]")} card nil)))}
-             :corp-turn-begins {:effect (req (reset! matrix {}))}}})
+             :corp-spent-click
+              {:effect (req (update! state side (update-in card [:seen-this-turn target] (fnil + 0) (second targets)))
+                            (when (>= (get-in (get-card state card) [:seen-this-turn target]) 3)
+                                 (resolve-ability state side
+                                                  {:once :per-turn
+                                                   :effect jeeves
+                                                   :msg (msg "gain a [Click]")} card nil)))}
+              :corp-turn-ends {:effect (effect (update! (dissoc card :seen-this-turn)))}}})
 
    "Kala Ghoda Real TV"
    {:flags {:corp-phase-12 (req true)}
@@ -697,8 +695,7 @@
     :events {:corp-turn-begins ability}})
 
    "Melange Mining Corp."
-   {:abilities [{:cost [:click 3] :effect (effect (do (gain :credit 7)
-                                                      (trigger-event state side :corp-spend-click (:cid card) 3)))
+   {:abilities [{:cost [:click 3] :effect (effect (gain :credit 7))
                  :msg "gain 7 [Credits]"}]}
 
    "Mental Health Clinic"
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -695,8 +695,7 @@
     :events {:corp-turn-begins ability}})
 
    "Melange Mining Corp."
-   {:abilities [{:cost [:click 3] :effect (effect (gain :credit 7))
-                 :msg "gain 7 [Credits]"}]}
+   {:abilities [{:cost [:click 3] :effect (effect (gain :credit 7)) :msg "gain 7 [Credits]"}]}
 
    "Mental Health Clinic"
    (let [ability {:msg "gain 1 [Credits]"
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -54,17 +54,6 @@
       (move state :runner (assoc (deactivate state side card) :agendapoints n :zone [:discard]) :scored options)) ; allow force option in case of Blacklist/News Team
   (gain-agenda-point state side n)))
 
-
-(defn jeeves-event-map
-  "Returns a map of the events seen by Jeeves & their count this turn"
-  [state]
-  (frequencies (get @state :jeeves)))
-
-(defn jeeves-event-count
-  "Returns summed occurences of an event vector this turn or zero if nil"
-  [state eventkeys]
-  (reduce + (map #(or (% (jeeves-event-map state)) 0) eventkeys)))
-
 ;;; Card definitions
 (declare in-server?)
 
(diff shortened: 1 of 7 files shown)


erbridge:
This looks like an interesting approach, but I wonder if there's a way to track the clicks spent another way, without creating a load of new events. It would also be nice to count the spent clicks for each category, not the events.

danhut:
The reason I went this way - is that Jeeves has to only see the events if he is rezzed.  Originally I had a general purpose solution where I could count all events of a type per turn then Jeeves would check that when an event occurred - but the issue was Jeeves could fire on stuff from before it was rezzed.

It might be possible to have fewer events and then log how many clicks each took - but I had concerns about existing functionality.  Additionally I was worried about things that affect how many clicks something could take like Accelerated Diagnostics etc.

But right now a Double operation will create two events - one for the payment of a click, and one for the 2nd click.  I felt that was safer that aggregating but very open to correction.  I am new here ;)

Of course this is all for one card!  But i think the event triggers are general enough to be used elsewhere.

erbridge:
Could you maybe do the click counting as described with a single event for `:spent-click` that contains the number of clicks and type of event in the state, and bind to that with Jeeves?

I'm reluctant to add a load of new events unless they are clearly needed for a class of card interactions (though someone else feel free to contradict me :)).

danhut:
That is a better way around it.  If no one says otherwise I will change it to that way around.

nealterrell:
If Jeeves needs his own atom in the full state (and not tucked into the card itself, which is what I'd prefer), then you should put it in `[:corp :register]`, which gets auto-cleared for you at the start of a turn.

danhut:
@nealterrell Do you see a better way to do this?   Very open to it.  

I think Jeeves needs his own state outside the card as I need to track specifically which Click actions he has seen since rez.  The other way would be to track within :turn-events when he was rezzed, and only count events since that time.  That seems more complex to me.

You'll also see I went this way with manta grid as the effect still stand even if the card is later nuked by something.  Given my lack of experience with the game engine there could be lots of better ways that my thinking.

danhut:
Question on this.
We want a single event :spend-click that Jeeves registers as a handler for.  
I would need the event emitter to include additional data - click reason and click cost.  Looking at the trigger event code below:
```
(defn trigger-event
  "Resolves all abilities registered as handlers for the given event key, passing them
  the targets given."
  [state side event & targets]
```
I think i could handle this by turning this function into a multiple arity function with a 5th binding like so:
```
(defn trigger-event
  "Resolves all abilities registered as handlers for the given event key, passing them
  the targets given."
  ([state side event & targets] (trigger-event [state side event nil & targets]))
  ([state side event eventdata & targets]
         rest of fn
  )
```

eventdata would be a hashmap:
```
{:eventtype type_of_event
 :clickcost:  click_cost
 }
```

Thus nothing else in the engine should care, and we could use this for other stuff in future ... if needed such as :creditcost

And a question on the event engine - when Jeeves gets the event is there a way to pass this eventdata to it directly?  Or it needs to go grab it from :turn-events?

nealterrell:
If Jeeves can only see actions that occur while he's rezzed, then doesn't it make sense that he should track it in the card state, since it's the card's memory? A second Jeeves being rezzed (after first is trashed) shouldn't share memory, right?

Re: varargs in functions. Clojure does not support overloaded functions if the function has a form that takes a varargs list (the `&` syntax). Suppose on overload takes 4 parameters and then a varargs list, and a second form takes 5 parameters and a varargs list, then we call the function with 5 arguments. How is the compiler for our poor dynamically typed language supposed to know if we mean to call the 4 argument function with a vararg list of 1, or the 5 argument function with an empty vararg list? It can't, so it's not allowed.

But all that is irrelevant, because `trigger-event already does accept your "event data" as the `targets` list.  When you call `trigger-event`, you are allowed to put whatever you want after the event key, and all the values you give will be passed to all event handlers as a variable called `targets`. 

Example: https://github.com/mtgred/netrunner/blob/master/src/clj/game/core/rules.clj#L182 -- we trigger `:pre-resolve-damage` and pass three items as "targets": the type of damage being resolved, the card dealing the damage, and the amount of damage after prevention. Any card that listens to this event will have a variable `targets` in the `:effect` of the handler; `targets` is a list that contains the arguments passed to the trigger, in this case the type, card, and amount. You can access those parameters using `(first targets)`, `(second targets)`, `(nth targets 2)`, etc. (__NOTE__: the symbol `target` is always equal to `(first targets)` inside of ability handlers.)

You can see how `targets` for this event is used here: https://github.com/mtgred/netrunner/blob/master/src/clj/game/cards/upgrades.clj#L547 -- Tori Hanzo checks to see if `target` (the first parameter, which is the damage type) is Net damage, and that the `(last targets)` (the last parameter, the amount of damage done) is positive.

danhut:
Thanks Neal. This is very helpful.

Will use it to clean up recon drone I just implemented too.

It would be great to cut and paste this into the event section of the wiki.

Sent from my iPhone

On 8 Feb 2017, at 6:01 am, Neal Terrell <notifications@github.com<mailto:notifications@github.com>> wrote:


If Jeeves can only see actions that occur while he's rezzed, then doesn't it make sense that he should track it in the card state, since it's the card's memory? A second Jeeves being rezzed (after first is trashed) shouldn't share memory, right?

Re: varargs in functions. Clojure does not support overloaded functions if the function has a form that takes a varargs list (the & syntax). Suppose on overload takes 4 parameters and then a varargs list, and a second form takes 5 parameters and a varargs list, then we call the function with 5 arguments. How is the compiler for our poor dynamically typed language supposed to know if we mean to call the 4 argument function with a vararg list of 1, or the 5 argument function with an empty vararg list? It can't, so it's not allowed.

But all that is irrelevant, because trigger-event already does accept your "event data" as thetargetslist. When you calltrigger-event, you are allowed to put whatever you want after the event key, and all the values you give will be passed to all event handlers as a variable calledtargets`.

Example: https://github.com/mtgred/netrunner/blob/master/src/clj/game/core/rules.clj#L182 -- we trigger :pre-resolve-damage and pass three items as "targets": the type of damage being resolved, the card dealing the damage, and the amount of damage after prevention. Any card that listens to this event will have a variable targets in the :effect of the handler; targets is a list that contains the arguments passed to the trigger, in this case the type, card, and amount. You can access those parameters using (first targets), (second targets), (nth targets 2), etc. (NOTE: the symbol target is always equal to (first targets) inside of ability handlers.)

You can see how targets for this event is used here: https://github.com/mtgred/netrunner/blob/master/src/clj/game/cards/upgrades.clj#L547 -- Tori Hanzo checks to see if target (the first parameter, which is the damage type) is Net damage, and that the (last targets) (the last parameter, the amount of damage done) is positive.

-

JoelCFC25:
PTSD flashbacks from Tori Hanzo

danhut:
I had another go at this.  Can the team please take a look?  It is all done except I need to go through the "clickable" cards and add a "corp-spend-click" event to them.  I did Melange already.

Changes:

1. One event for all click spend - :corp-spend-click
2. Actions/rules pass the event name & number of clicks in targets
3. Melange passed :cid and number of clicks in targets
4. State kept on card via an atom

src/clj/game/cards/assets.clj
```diff
@@ -580,10 +580,26 @@
...
-   {:implementation "Trigger is manual"
-    :abilities [{:label "Gain [Click]"
-                 :msg "gain [Click]" :once :per-turn
-                 :effect (effect (gain :click 1))}]}
+    (let [jeeves (effect (gain :click 1))
+          ability {:label "Gain [Click]"
+                   :msg "gain [Click]"
+                   :once :per-turn
+                   :effect jeeves}
+          matrix  (atom {})]
```
nealterrell: So this atom will have a fun side effect: all Jeeves across all games currently running will share the same state ;). The atom is declared inside the card definitions map, which is a single data structure used by the server for all games.

The state needs to be saved inside the card. See Security Testing for an example of saving information on a card instance in response to an event, then using that information later. (Two parts: using `assoc` on a card to save new information into its map, and passing that to `update!` which updates the game state; and later retrieving the same key from the game state using `get-card`.)

src/clj/game/cards/assets.clj
```diff
@@ -580,10 +580,26 @@
...
+
+    {:abilities  [ability]
+     :leave-play (req (reset! matrix {}))
+     :trash-effect (req (reset! matrix {}))
+     :events {
+             :corp-spend-click
+              {:effect (req (swap! matrix update-in [target] (fnil + 0) (second targets))
+                            (let [qty (get-in @matrix [target])]
+                              (resolve-ability state side
+                                             {:req (req (>= qty 3))
```
nealterrell: I'd rather have you only resolve the ability is `qty` is >= 3, instead of always resolving an ability with this `:req`.

src/clj/game/cards/assets.clj
```diff
@@ -681,7 +697,9 @@
     :events {:corp-turn-begins ability}})
 
    "Melange Mining Corp."
-   {:abilities [{:cost [:click 3] :effect (effect (gain :credit 7)) :msg "gain 7 [Credits]"}]}
+   {:abilities [{:cost [:click 3] :effect (effect (do (gain :credit 7)
+                                                      (trigger-event state side :corp-spend-click (:cid card) 3)))
```
nealterrell: Why is this necessary? Why don't other cards need this trigger? Also, did you test this... the `do` is making me nervous and is unnecessary. (It's implicit inside `effect` and `req`.)

src/clj/game/cards/assets.clj
danhut: Thanks... need to read more on this - I expected everything inside the (lets) to be local.  Might have been fun on the live server!

src/clj/game/cards/assets.clj
danhut: I tested it and it was needed.  Did not trace why but think I have seen this elsewhere too.

erbridge: Are you aware that `effect` injects `state side` as the first two arguments? So this would look like:

```
(effect (gain :credit 7)
        (trigger-event :corp-spend-click (:cid card) 3))
```

But I agree with @nealterrell that this seems unnecessary. Can we not check to see if a cost contains clicks when resolving an ability and trigger the event there, in general, like you do with the other events?

danhut: Will check if I can make that work.  Now understand the issue with Melange.  thanks.

danhut:
Found where to do this. Soon Jeeves!

Sent from my iPhone

On 11 Feb 2017, at 11:04 pm, Felix Laurie von Massenbach <notifications@github.com<mailto:notifications@github.com>> wrote:


@erbridge requested changes on this pull request.

________________________________

In src/clj/game/cards/assets.clj<https://github.com/mtgred/netrunner/pull/2342#pullrequestreview-21384720>:

> @@ -681,7 +697,9 @@
     :events {:corp-turn-begins ability}})

    "Melange Mining Corp."
-   {:abilities [{:cost [:click 3] :effect (effect (gain :credit 7)) :msg "gain 7 [Credits]"}]}
+   {:abilities [{:cost [:click 3] :effect (effect (do (gain :credit 7)
+                                                      (trigger-event state side :corp-spend-click (:cid card) 3)))


Are you aware that effect injects state side as the first two arguments? So this would look like:

(effect (gain :credit 7)
        (trigger-event :corp-spend-click (:cid card) 3))


But I agree with @nealterrell<https://github.com/nealterrell> that this seems unnecessary. Can we not check to see if a cost contains clicks when resolving an ability and trigger the event there, in general, like you do with the other events?

-

danhut:
Thanks for the review team.  I have implemented what I think is better now!
- Card state is keep on the card
- All corp and common click event-triggers add a map containing {:action :what-you-did}
- All events that jeeves sees are send via the pay function - which means no changes to cards with click abilities - so no ongoing maintenance! Hooray,  Much better.

src/clj/game/cards/assets.clj
```diff
@@ -580,10 +580,24 @@
...
    "Jeeves Model Bioroids"
-   {:implementation "Trigger is manual"
-    :abilities [{:label "Gain [Click]"
-                 :msg "gain [Click]" :once :per-turn
-                 :effect (effect (gain :click 1))}]}
+    (let [jeeves (effect (gain :click 1))
+          ability {:label "Gain [Click]"
+                   :msg "gain [Click]"
+                   :once :per-turn
+                   :effect jeeves}]
```
erbridge: Can we use this same ability in the automatic effect, too?

src/clj/game/core/actions.clj
```diff
@@ -15,7 +15,7 @@
     (case (:type card)
       ("Event" "Operation") (play-instant state side card {:extra-cost [:click 1]})
       ("Hardware" "Resource" "Program") (runner-install state side (make-eid state) card {:extra-cost [:click 1]})
-      ("ICE" "Upgrade" "Asset" "Agenda") (corp-install state side card server {:extra-cost [:click 1]}))
+      ("ICE" "Upgrade" "Asset" "Agenda") (do (corp-install state side card server {:extra-cost [:click 1]})))
```
erbridge: This `do` seems unnecessary to me.

src/clj/game/cards/assets.clj
```diff
@@ -580,10 +580,24 @@
...
+
+    {:abilities  [ability]
+     :leave-play {:effect (effect (update! (dissoc card :seen-this-turn)))}
```
erbridge: It would be nice to refactor and reuse this effect.

src/clj/game/core/costs.clj
```diff
@@ -43,6 +44,8 @@
 
 (defn pay
   "Deducts each cost from the player."
+  ; args format  ([:click 1 :credit 0] [:forfeit] {:action :corp-click-credit})
```
erbridge: You could add this directly to the docstring.

danhut:
thanks Felix - getting there!  Will take a look.  Appreciate it.

danhut:
All done.  Cleaner again.

src/clj/game/core/actions.clj
```diff
@@ -15,8 +15,8 @@
     (case (:type card)
       ("Event" "Operation") (play-instant state side card {:extra-cost [:click 1]})
       ("Hardware" "Resource" "Program") (runner-install state side (make-eid state) card {:extra-cost [:click 1]})
-      ("ICE" "Upgrade" "Asset" "Agenda") (corp-install state side card server {:extra-cost [:click 1]}))
-    (trigger-event state side :play card)))
+      ("ICE" "Upgrade" "Asset" "Agenda") (corp-install state side card server {:extra-cost [:click 1]})))
+  (trigger-event state side :play card))
```
erbridge: What was the intention behind this change? Looks like you've moved `trigger-event` from being the default clause of the `case` to just running afterwards.

danhut: No intent.  Failed cleanup of my earlier issue with "do" on this function. ie, It was not needed.

danhut:
Someone please merge this before it gets all conflicted?

Issue #2282, opened 2016-12-31
Rumor Mill + trashed Cybernetics Court


JoelCFC25:
It looks like this is the only asset or upgrade affected by this...

If Rumor Mill is in play and suppressing the effect of a rezzed Cybernetics Court--reducing the Corp max hand size from 9 back to the standard 5--trashing the Court by any method (Runner paying to trash in a run, voluntary overwrite by Corp, or card effect like Product Recall) will result in the max hand size dropping by 4 again, even though the card's effect is nullified by Rumor Mill.

kyzen:
Created pull request #2343  as potential fix

PR #2343, opened 2017-01-29
Address issue #2282

kyzen:
Fix for Rumor Mill interaction with Cybernetics Court (also Chairman Hiro) as per issue: #2282

### DIFF
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -205,6 +205,7 @@
    "Chairman Hiro"
    {:effect (effect (lose :runner :hand-size-modification 2))
     :leave-play (effect (gain :runner :hand-size-modification 2))
+    :flags {:constant-effect true}
     :trash-effect {:when-inactive true
                    :req (req (:access @state))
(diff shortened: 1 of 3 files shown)


nealterrell:
@kyzen I think this is overkill.. can you try just doing the replacement of `:in-play` with `:effect` and `:leave-play`? `:in-play` does not check to see if the card is disabled or not. (You could also look at `deactivate`, where `:in-play` is checked, to add a condition that the card is not disabled before applying the `lose` function.)

kyzen:
Just doing the replacement of :in-play with the other two isn't sufficient, because in the (admittedly edge) case where Rumor Mill is active and the corp rezzes Cybernetics or Hiro, the effect will initially be inactive but will remain inactive even once Rumor Mill leaves play.  This is because, prior to this change, the **enable-card** function always calls **card-init** with a "resolve" argument of false.  I believe the flag, or something like it, is necessary to check if we need to resolve a rezzed card's ability when re-enabling that card after Rumor Mill.  Of course, I may be wrong too.

JoelCFC25:
This is fixed in a much tidier way in #2708.

Issue #2340, opened 2017-01-28
Street Peddler and Currents


jbrz:
If a current ends up on a Street Peddler, game thinks it's active.

**Rumor Mill** stopped Jackson from working.
**Hacktivist** caused trashes on asset-rez.

However, **Employee Strike** did _not_ appear to fire on asset-trash against CtM.

Things like Clot, The Source, etc. appear to be inactive, as expected.  So might just be isolated cases with some currents/events - I didn't do exhaustive research outside of the aforementioned cases.

jbrz:
Not sure what the "proper" fix is, still not familiar enough with the syntax or the structure, but maybe just adding a `can-trigger?` to events(or all cards?) to check for `facedown`?

That's just my gut reaction, but again, that doesn't necessarily seem to be the problem here as it's not all events or even all currents, and might not be that simple/elegant of a solution.

JoelCFC25:
The correct fix is to look at the changes made in #2266 and add a check for `(not facedown)` in the case of the hosted card being an Event.

PR #2346, opened 2017-01-31
Fix for #2340 Peddler + Current issue

danhut:
Fixed this issue by checking the currents are in the :current area when the effect occurs.


danhut:
Fixed the rest of impacted currents too... (i think!)

danhut:
I'll fix up the host code instead, want to keep these unit tests

PR #2351, opened 2017-02-03
Clear :access flag on run-end - fix #2319

### LINKED ISSUE #2319: Trashing Chairman Hiro as Corporation adds him to the runner's score area.
Playing as the Corporation, attempting to trash Chairman Hiro from hand at the end of the turn caused him to be added to the runner's score area instead. Dragging him back to hand and trashing him from hand did not address the issue.

Installing Chairman Hiro into a new server and then trashing him by writing over him also resulted in this issue. The only way I discovered to avoid this was to use '/discard #n' commands to manually trash him from hand.

This appears to be an issue that has been closed in the past [#1181 and #1312]. The runner had not previously accessed Chairman Hiro in that game. Playing as the Corporation on the previous day, I had trashed Chairman Hiro from hand before without causing the issue. There was no use of Film Critic, but it's possible the Runner may have saw Chairman Hiro in R&D through use of Spy Camera [I don't remember.]

I'm attaching a full transcript of the game in question in case it helps address the issue.

[Transcript.rtfd.zip](https://github.com/mtgred/netrunner/files/706183/Transcript.rtfd.zip)

danhut:
fix #2319

### DIFF
--- a/src/clj/game/core/runs.clj
+++ b/src/clj/game/core/runs.clj
@@ -666,6 +666,7 @@
     (swap! state assoc-in [:runner :run-credit] 0)
     (swap! state assoc :run nil)
     (update-all-ice state side)
+    (swap! state dissoc :access)
     (clear-run-register! state)
     (trigger-run-end-events state side eid run)))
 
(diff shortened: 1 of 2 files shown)


danhut:
Added that test.  I wrote the shock test for that reason and did not add it for some reason.

nealterrell:
Looks good. `run-cleanup` is definitely a better place for this.

PR #2355, opened 2017-02-04
test for clone chip and dont install programs you cant afford

mattchainsaw:
Currently if you try to install a program you can not afford, the prompt says that you did install it and trashed clone chip, but the program is still in the heap. The new behavior is that clone chip is not trashed. I was not able to change the prompt from firing.

### DIFF
--- a/src/clj/game/cards/hardware.clj
+++ b/src/clj/game/cards/hardware.clj
@@ -141,7 +141,7 @@
                                                    (lose state side :tag 1)))} card nil))}]}
 
    "Clone Chip"
-   {:abilities [{:prompt "Choose a program to install from your Heap" :msg (msg "install " (:title target))
+   {:abilities [{:prompt "Choose a program to install from your Heap"
                  :priority true :show-discard true
                  :req (req (not (seq (get-in @state [:runner :locked :discard]))))
                  :choices {:req #(and (is-type? % "Program")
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/hardware.clj
```diff
@@ -146,7 +146,10 @@
                  :req (req (not (seq (get-in @state [:runner :locked :discard]))))
                  :choices {:req #(and (is-type? % "Program")
                                       (= (:zone %) [:discard]))}
-                 :effect (effect (trash card {:cause :ability-cost}) (runner-install target))}]}
+                 :effect (req (when (>= (:credit runner) (:cost target))
+                                    (do
+                                      (trash state side card {:cause :ability-cost})
```
erbridge: Call `system-msg` here and remove the `:msg` entry to make the message fire later.

mattchainsaw: Thanks!

src/clj/game/cards/hardware.clj
```diff
@@ -141,12 +141,16 @@
...
    "Clone Chip"
-   {:abilities [{:prompt "Choose a program to install from your Heap" :msg (msg "install " (:title target))
+   {:abilities [{:prompt "Choose a program to install from your Heap"
                  :priority true :show-discard true
                  :req (req (not (seq (get-in @state [:runner :locked :discard]))))
                  :choices {:req #(and (is-type? % "Program")
                                       (= (:zone %) [:discard]))}
-                 :effect (effect (trash card {:cause :ability-cost}) (runner-install target))}]}
+                 :effect (req (when (>= (:credit runner) (:cost target))
+                                    (do
```
nealterrell: You don't need `do` inside a `when`, unlike with `if`.

PR #2359, opened 2017-02-06
Implement recon drone

danhut:


danhut:
And of course I only block net damage. New commit shortly.

danhut:
Added meat & brain fully now

PR #2362, opened 2017-02-10
Prevent multiple install prompts for conspiracy breakers when user denies first prompt

### LINKED ISSUE #2360: Multiple Conspiracy Breaker (Paperclip, Black Orchestra, MKUltra) prompts
Not a big issue, but if you have multiple of the same conspiracy breaker in the archives and encounter the right type of ice, the install prompt will also be shown multiple times. I think it could sometimes be relevant to install multiple or to install one that you already have installed, but if you answer "No" on the first prompt, it seems like the single prompt would be enough. It's at least slightly confusing for the user (when they decline an install, but the prompt is still there).

nealterrell:
Conspiracy breakers already won't ask to install if another copy is already installed. Now they won't ask to install if you just said "no" to another copy for the same piece of ice.

Fix #2360, not 2630 (whoops).

### DIFF
--- a/src/clj/game/cards/icebreakers.clj
+++ b/src/clj/game/cards/icebreakers.clj
@@ -145,11 +145,15 @@
   (let [install-prompt {:req (req (and (= (:zone card) [:discard])
                                        (rezzed? current-ice)
                                        (has-subtype? current-ice type)
-                                       (not (some #(= title (:title %)) (all-installed state :runner)))))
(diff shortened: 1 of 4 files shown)


danhut:
Neal - does this work if the Ice has multiple types (eg Barrier and Code Gate) And you have say 2x  Paperclip & Black Orchestra in the bin?
 
I read it as no. But did not try it.

nealterrell:
You are correct, that's an oversight on my part. Will fix.

PR #2379, opened 2017-02-20
Implements Self-Adapting Code Wall 

Antigrapist:
I was able to reuse Lotus Field's test rather than trying to write my own.

### DIFF
--- a/src/clj/game/cards/ice.clj
+++ b/src/clj/game/cards/ice.clj
@@ -1174,7 +1174,7 @@
                            
     "Self-Adapting Code Wall"
     {:subroutines [end-the-run]
-        :flags {:cannot-lower-strength true}}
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/ice.clj
```diff
@@ -1171,6 +1171,10 @@
                            :delayed-completion true
                            :effect (effect (tag-runner :runner eid 1))
                            :msg "give the Runner 1 tag"}}]}
+                           
+    "Self-Adapting Code Wall"
+    {:subroutines [end-the-run]
+        :flags {:cannot-lower-strength true}}
```
JoelCFC25: Justify this so that the colons preceding `subroutines` and `flags` are vertically aligned.

PR #2391, opened 2017-02-24
Fix: tests, Archives Interface, TS, Stinson; implement: Mad Dash, Jemison, Los, Clan Vengeance

### LINKED ISSUE #2386: Team sponsorship allows installing runner cards from the heap
Found this out accidentally while checking to make sure I hadn't blasted through a clot trigger while the TS trigger was active. Clicking the runner's heap selected the top card of the heap for my TS shenanigans, letting me install a runner card (cloak) in a new remote. Later confirmed that this is incredibly repeatable, just whenever you have a team sponsorship trigger, click on any of the runner's cards in their heap and you can install it in any server.

At this moment I'm unsure what happens when they run it, but have confirmation that Leela cannot bounce them back into hand.

### LINKED ISSUE #2389: Eater and Archives Interface
I know the software can't enforce every rule, I was just wondering if that was the case here, or it was an interaction that hadn't come up before.

My understanding is that if you Eater on an Archives run you can't use Archives Interface because you would have to access the card to use AI's replacement effect, and obviously Eater prevents that.

In my last game I wasn't sure of the rule so said okay, and it allowed him to remove my card.

JoelCFC25:
Gets tests passing again (failing due to new Shaper identity), fixes #2386, fixes #2389. 

Moves Bryan Stinson to upgrades, adds missing restriction that the operation must have the Transaction subtype.

Also implemented: Mad Dash and Jemison Astronautics (with tests) from the current pack; Clan Vengeance and Los: Data Hijacker from Station One.

### DIFF
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -1149,6 +1149,7 @@
                              :interactive (req true)
                              :delayed-completion true
                              :choices {:req #(and (not (is-type? % "Operation"))
+                                                  (= (:side %) "Corp")
                                                   (#{[:hand] [:discard]} (:zone %)))}
                              :msg (msg (corp-install-msg target))
                              :effect (effect (corp-install eid target nil {:no-install-cost true}))}}}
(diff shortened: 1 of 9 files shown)


danhut:
Joel - can you add my test case?  Might as well have it?

[Gist](https://gist.github.com/danhut/a7daa0393c21c05dbe0c16015238930f)

JoelCFC25:
Just realized Clan Vengeance is wrong, it's milling from R&D instead of taking from HQ. Fixing shortly.

PR #2393, opened 2017-02-26
implement Mad Dash and Signal Jamming

danhut:
Test fail due to Ayla + Rebirth.  Fixed in #2391 already


danhut:
Will close this one - and back out Mad Dash as Joel did it before I committed.

PR #2394, opened 2017-02-26
implement signal jam

danhut:
Test fail due to Ayla + Rebirth. Fixed in #2391 already


JoelCFC25:
Wow, sweet. I actually thought of another (never seen) card that could do mid-run installs--Savoir-faire!

danhut:
For cards I did not "augment" there is a toast and they are not allowed to install. I can check those but it is quite safe. Main thing was to stop wasted SMC and Peddlers.

danhut:
Because they self trash before install is fired

danhut:
Added some extra checks for Artist Colony, DaVinci, Savoir-Faire to stop them getting waste.  I don't think it is needed for any others.

PR #2395, opened 2017-02-27
implement network exchange & Synth DNA Modification

danhut:


danhut:
Test fail due to Ayla + Rebirth. Fixed in #2391 already

PR #2407, opened 2017-03-05
Fail if unable to connect to mongodb

ouroboros8:
Currently, if unable to connect to mongo, the script will run
"successfully" but silently fail to write to the database. Silent
failures like this are bad, and make it hard to debug often trivial
issues.

If for some reason there's a need to enable the fetch without database-connectivity, I could add a flag to enable that behaviour (i.e. ignore db connection exceptions), but as far as I'm concerned that's not necessary.


ouroboros8:
lmk if you want me to retroactively raise an issue for this change.

PR #2408, opened 2017-03-05
Fix docker config for development

ouroboros8:
The old docker config in `docker-compose.yml` no longer worked, so I've rewritten it pretty much entirely. The new config is probably not suitable for production, but this application probably isn't that suitable as a whole for containerisation; I just want a quick way to run a dev instance.


mtgred:
👍

PR #2409, opened 2017-03-06
implement quarantine system, SYNC BRE

danhut:


src/clj/game/cards/assets.clj
```diff
@@ -913,6 +913,20 @@
                              (system-msg state :corp "uses Public Support to add it to their score area as an agenda worth 1 agenda point")
                              (as-agenda state :corp (dissoc card :counter) 1)))} }}
 
+   "Quarantine System"
+   (letfn [(rez-ice [cnt ap] {:prompt "Choose an ICE to rez"
+                              :delayed-completion true
+                              :choices {:req #(and (is-type? % "ICE") (complement rezzed?))}
```
JoelCFC25: Quibbling, but `(ice? %)` will work.

src/clj/game/cards/assets.clj
```diff
@@ -913,6 +913,20 @@
...
+                              :msg (msg "rez " (:title target))
+                              :effect (req (rez-cost-bonus state side (* ap -2))
+                                           (rez state side target {:no-warning true})
+                                           (if (< cnt 3) (continue-ability state side (rez-ice (inc cnt) ap) card nil)
+                                                         (effect-completed state side eid)))})]
+     {:abilities [{:label "Forfeit agenda to rez upto 3 ICE with a 2 [Credit] discount per agenda point"
+                   :req (req (pos? (count (:scored corp))))
+                   :cost [:forfeit]
+                   :effect (req (continue-ability state side (rez-ice 1 (:agendapoints (last (:rfg corp)))) card nil))}]})
```
JoelCFC25: We also need to check that the item forfeited was of the Agenda type, because the card refers to "printed agenda points" and things like Franchise City technically have 0 printed agenda points, though they'll have `:agendapoints 1` in state due to their card logic. Same would apply to things like Fan Site and whatnot that could in theory get swapped to the Corp.

src/clj/game/cards/assets.clj
```diff
@@ -913,6 +913,20 @@
...
+   "Quarantine System"
+   (letfn [(rez-ice [cnt ap] {:prompt "Choose an ICE to rez"
+                              :delayed-completion true
+                              :choices {:req #(and (is-type? % "ICE") (complement rezzed?))}
+                              :msg (msg "rez " (:title target))
+                              :effect (req (rez-cost-bonus state side (* ap -2))
+                                           (rez state side target {:no-warning true})
+                                           (if (< cnt 3) (continue-ability state side (rez-ice (inc cnt) ap) card nil)
+                                                         (effect-completed state side eid)))})]
+     {:abilities [{:label "Forfeit agenda to rez upto 3 ICE with a 2 [Credit] discount per agenda point"
```
JoelCFC25: "up to"

src/clj/game/cards/assets.clj
danhut: I think you could forfeit anything that is currently an agenda even if zero points to fire something like Jemison.  But you are right it should be the printed points.  I believe we need to change fn as-agenda to give Franchise City a :agendapoint-gained attribute so we can ignore it from this card?  Thoughts?

danhut: Decided to do it on the card.  Should be safe unless they ever publish a non-agenda with printed agenda points...

PR #2445, opened 2017-03-19
Adding 4 new and upcoming events

PesiTheWizard:
Added Build Script, Deep Data Mining, Mars for Martians, Spear Phishing.

### DIFF
--- a/src/clj/game/cards/events.clj
+++ b/src/clj/game/cards/events.clj
@@ -69,6 +69,10 @@
     :effect (effect (resolve-ability {:prompt "Choose a server" :choices (req runnable-servers)
                                       :effect (effect (run target nil card))} card nil))}
 
+   "Build Script"
+   {:msg (msg "gain 1 [Credits] and draw 2 cards")
+    :effect (effect (gain :credit 1) (draw 2))}
+   
    "Calling in Favors"
    {:msg (msg "gain " (count (filter #(has-subtype? % "Connection") (all-installed state :runner)))
               " [Credits]")
@@ -226,6 +230,13 @@
                                                      (filter #(has-subtype? % "Virus") (:discard runner)) :sorted))
                                      :effect (effect (move target :hand))} card nil)))}
 
+   "Deep Data Mining"
+   {:effect (effect (run :rd nil card)
+                    (register-events (:events (card-def card)) (assoc card :zone '(:discard))))
+    :events {:successful-run {:silent (req true)
+                              :effect (effect (access-bonus (min 4 (:memory runner))))}
+             :run-ends {:effect (effect (unregister-events card))}}}
+   
    "Demolition Run"
    {:prompt "Choose a server" :choices ["HQ" "R&D"]
     :abilities [{:msg (msg "trash " (:title (:card (first (get-in @state [side :prompt])))) " at no cost")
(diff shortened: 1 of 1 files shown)


src/clj/game/cards/events.clj
```diff
@@ -809,6 +820,10 @@
       :effect (req (show-wait-prompt state :corp "Runner to rearrange the top cards of their stack")
                    (let [from (take 6 (:deck runner))]
                      (continue-ability state side (entrance-trash from) card nil)))})
+   
+   "Mars for Martians"
+   {:msg (msg "draw " (count (filter #(and (has-subtype? % "Clan") (is-type? % "Resource")) (all-installed state :runner))) " card(s) and gain " (:tag runner) " [Credits]")
```
JoelCFC25: Can streamline by starting with: 

`(let [n (count (filter #(and (has-subtype? % "Clan") (is-type? % "Resource")) (all-installed state :runner)))]`

src/clj/game/cards/events.clj
```diff
@@ -1199,6 +1214,12 @@
     :events {:pre-rez-cost nil}
     :end-turn {:effect (effect (unregister-events card))}}
 
+   "Spear Phishing"
+   {:implementation "Bypass is manual"
+    :prompt "Choose a server"
+    :choices (req runnable-servers)
+    :effect (effect (run target nil card))}
+
```
JoelCFC25: Fine as a starting point--for real flair, since this card references ICE at a defined position, you could have it register an `:encounter-ice` event where if the `ice-index` of the event target is 0, you print a message to the game log.

PesiTheWizard: I decided against it due to the ways bypassing can be prevented, and I wasn't sure if it would break if new ice is added to the server during the run at a particular point.

src/clj/game/cards/events.clj
```diff
@@ -69,6 +69,10 @@
     :effect (effect (resolve-ability {:prompt "Choose a server" :choices (req runnable-servers)
                                       :effect (effect (run target nil card))} card nil))}
 
+   "Build Script"
+   {:msg (msg "gain 1 [Credits] and draw 2 cards")
```
JoelCFC25: `(msg .... )` is only required if you are using any functions to determine the contents of the message.

src/clj/game/cards/events.clj
```diff
@@ -809,6 +820,12 @@
       :effect (req (show-wait-prompt state :corp "Runner to rearrange the top cards of their stack")
                    (let [from (take 6 (:deck runner))]
                      (continue-ability state side (entrance-trash from) card nil)))})
+   
+   "Mars for Martians"
+   {:msg (msg "draw " (count (filter #(and (has-subtype? % "Clan")
+                                           (is-type? % "Resource")) (all-installed state :runner))) " card(s) and gain " (:tag runner) " [Credits]")
+    :effect (effect (draw (count (filter #(and (has-subtype? % "Clan")
+                                               (is-type? % "Resource")) (all-installed state :runner)))) (gain :credit (:tag runner)))}
```
JoelCFC25: Should look like this:  https://pastebin.com/qMmK490T

JoelCFC25: Should look like this:  https://pastebin.com/qMmK490T

And either just say "cards" in the message or look at a card like Daily Business Show for a `when` statement in the message to determine if plural is necessary.

JoelCFC25:
Looking good! There's no NRDB data yet for Mars for Martians, but it'll work fine once that appears. 👍

JoelCFC25:
@PesiTheWizard the Mars for Martians code can be exhumed and resubmitted at any time--I had reverted it quite awhile back for making tests fail when there was no card data. Now there is.

PR #2447, opened 2017-03-20
Changing Anson Rose to automatically add advancement counters.

poi2000:
I used code from Mumbad Construction Co. to automatically put
advancement tokens onto Anson Rose. This should cause Anson to
automatically get advancement tokens at the start of the Corporations
turn.


kevkcc:
You accidentally committed your intelliJ files

JoelCFC25:
Are you just trying to eliminate the "order my when-turn-begins" pause from `:corp-phase-12`? In theory there could be some future card combo where it becomes necessary to do something else ahead of Anson Rose's effect, and you'd have to undo this change. Maybe the thing to do is just alter the `:flags` so it only uses phase 1.2 if another rezzed card also does, otherwise it just fires on turn start without giving you the extra UI step.

poi2000:
Thanks for the input, I hadn't thought about interactions like EBC where you would want to choose the order of your start-of-turn events.
So if I'm understanding this correctly, giving Anson Rose the :corp-phase-12 flag if another rezzed card has the :corp-phase-12 flag, otherwise letting Anson fire on its own?

poi2000:
Might get back to this when I have time but not really a pressing issue.

PR #2448, opened 2017-03-23
fix #2291

### LINKED ISSUE #2291: Reaver/Wasteland requirements are slightly off
"Wasteland"
   {:events {:runner-trash {:req (req (and (first-event state :runner :runner-trash) (:installed target)))
                     :effect (effect (gain :credit 1))
                     :msg "to gain 1[Credit]"}}}

Steps to reproduce:
Have Faust
Have Reaver/Wasteland
Have Spy Cam
Boost Faust Str
Trash Spy Cam
-> No trigger

This seems to be because runner trashing a non-installed card (via Faust, other options exist) is the first runner-trash event.

danhut:
fix #2291

### DIFF
--- a/src/clj/game/core/rules.clj
+++ b/src/clj/game/core/rules.clj
@@ -1,7 +1,7 @@
 (in-ns 'game.core)
 
-(declare card-init card-str close-access-prompt enforce-msg gain-agenda-point get-agenda-points is-type? in-corp-scored?
-         prevent-draw resolve-steal-events make-result say show-prompt system-msg trash-cards untrashable-while-rezzed?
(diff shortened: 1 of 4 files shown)


src/clj/game/core/rules.clj
```diff
@@ -311,6 +311,7 @@
 (defn- resolve-trash
   [state side eid {:keys [zone type] :as card}
    {:keys [unpreventable cause keep-server-alive suppress-event] :as args} & targets]
+  (when (:installed card) (swap! state update-in [side :register :trashed-installed] (fnil inc 0)))
```
JoelCFC25: Note that Corp cards never have `(:installed true)`. So if you want this to work equally well for both sides, it should be `(installed? card)` and you'll have to declare `installed?` up at the top of this file.

PR #2451, opened 2017-03-24
fix The Supplier, Personal Workshop

### LINKED ISSUE #2358: Brain Chip hosted on Supplier issue
Install The Supplier, host a Brain Chip on it. Brain Chip should be inactive by definition. Give yourself tags, then as Corp, trash The Supplier. Brain Chip's `:leave-play` effect is firing for some reason, despite the card not being installed, and the Runner's MU and maximum hand size get reduced.

JoelCFC25:
Fixes #2358 (yes, finally). This will prevent `:leave-play` effects from firing for hosted, inactive, faceup Runner cards if their host is trashed. Currently this applies to The Supplier and Personal Workshop, but will also include Customized Secretary pretty soon.

### DIFF
--- a/src/clj/game/core/installing.clj
+++ b/src/clj/game/core/installing.clj
@@ -20,6 +20,7 @@
   [state side {:keys [disabled installed rezzed facedown zone host] :as card}]
   (when-let [leave-effect (:leave-play (card-def card))]
     (when (and (not disabled)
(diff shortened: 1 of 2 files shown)


danhut:
Nice 1 Joel!! :1st_place_medal:

PR #2457, opened 2017-03-28
Add MCA Informant

GassedUpOldMan:
I don't actually know Clojure so take it as you will

### DIFF
--- a/src/clj/game/cards/operations.clj
+++ b/src/clj/game/cards/operations.clj
@@ -615,7 +615,18 @@
                               :trace {:base 2 :msg "give the Runner 1 tag"
                                       :delayed-completion true
                                       :effect (effect (tag-runner :runner eid 1))}}}}
-
+   
+   "MCA Informant"
+   {:req (req (some #(has-subtype? % "Connection") (all-installed state %))) 
+    :prompt "Choose a connection to host MCA Informant" 
+    :choices {:req #(and (has-subtype? % "Connection") (installed? %))} 
+    :effect (effect (host target (assoc card :zone [:discard]))
+            (system-msg (str "hosts MCA Informant on " (:title target))) 
+            (swap! state update-in [:runner :tag] inc)) 
+    :abilities [{:label "[Click], 2 [Credits]: Trash this card."
+                 :effect (effect (trash card {:cause :ability cost}))}] 
+    :leave-play (req (swap! state-update-in [:runner :tag] dec))}
+   
    "Medical Research Fundraiser"
    {:msg "gain 8 [Credits]. The Runner gains 3 [Credits]"
     :effect (effect (gain :credit 8) (gain :runner :credit 3))}
--- a/src/clj/game/cards/operations.clj
+++ b/src/clj/game/cards/operations.clj
@@ -617,15 +617,18 @@
                                       :effect (effect (tag-runner :runner eid 1))}}}}
    
    "MCA Informant"
-   {:req (req (some #(has-subtype? % "Connection") (all-installed state %))) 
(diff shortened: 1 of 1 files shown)


src/clj/game/cards/operations.clj
```diff
@@ -615,7 +615,18 @@
                               :trace {:base 2 :msg "give the Runner 1 tag"
                                       :delayed-completion true
                                       :effect (effect (tag-runner :runner eid 1))}}}}
-
+   
+   "MCA Informant"
+   {:req (req (some #(has-subtype? % "Connection") (all-installed state %)))
```
JoelCFC25: `(all-installed state %)` doesn't work here. Would need to be `(concat (all-installed state :runner) (all-installed state :corp))`

src/clj/game/cards/operations.clj
```diff
@@ -615,7 +615,18 @@
...
+   
+   "MCA Informant"
+   {:req (req (some #(has-subtype? % "Connection") (all-installed state %))) 
+    :prompt "Choose a connection to host MCA Informant" 
+    :choices {:req #(and (has-subtype? % "Connection") (installed? %))} 
+    :effect (effect (host target (assoc card :zone [:discard]))
+            (system-msg (str "hosts MCA Informant on " (:title target))) 
+            (swap! state update-in [:runner :tag] inc)) 
+    :abilities [{:label "[Click], 2 [Credits]: Trash this card."
+                 :effect (effect (trash card {:cause :ability cost}))}]
```
JoelCFC25: This ability is not actually costing anything. Needs a costs vector like `:cost [:click 1 :credit 2]`. Also, this would add the ability to MCA Informant itself. The ability needs to go on the host. Wetwork Refit is a template here (and is the card I started from in my failing-so-far efforts to get this card working).

src/clj/game/cards/operations.clj
```diff
@@ -615,7 +615,18 @@
...
+   "MCA Informant"
+   {:req (req (some #(has-subtype? % "Connection") (all-installed state %))) 
+    :prompt "Choose a connection to host MCA Informant" 
+    :choices {:req #(and (has-subtype? % "Connection") (installed? %))} 
+    :effect (effect (host target (assoc card :zone [:discard]))
+            (system-msg (str "hosts MCA Informant on " (:title target))) 
+            (swap! state update-in [:runner :tag] inc)) 
+    :abilities [{:label "[Click], 2 [Credits]: Trash this card."
+                 :effect (effect (trash card {:cause :ability cost}))}] 
+    :leave-play (req (swap! state-update-in [:runner :tag] dec))}
```
JoelCFC25: This (and the corresponding line 625) is close--look at Paparazzi to see how the state of being tagged without a numeric quantity of tags is handled. Just change to `[:runner :tagged]` here.

JoelCFC25:
You're getting closer to where I got stuck and left off. I couldn't get the abilities of the host to match up to the labels. Also, the ability cost doesn't need to be included in the `:label`--it'll be put there automatically by virtue of having a `:cost` vector now.

nealterrell:
This probably needs to be done with a dynamic ability, a la Baba Yaga... the engine doesn't like a card having entries in its `:abilities` vector that don't line up with definitions in the card-def. (Wetwork Refit works because subroutines are treated differently than abilities.) 

You can try putting the "trash" ability on MCA Informant itself, then adding a `:dynamic :copy` ability to the hosted card, similar to how Baba Yaga adds a dynamic-copy ability to itself in order to duplicate its hosted's ability.

nealterrell:
Also, this isn't the same effect as Paparazzi -- a runner with 1 tag and MCA installed can be BOOM!'d, but not a runner with 1 tag and Paparazzi.

JoelCFC25:
We have a partial implementation for now and will iterate from there.

PR #2475, opened 2017-04-06
3 Low hanging fruits

PesiTheWizard:
Adding Process Automation, Battlement and IPO.


PesiTheWizard:
As Battlement isn't on Netrunnerdb yet, I'll have to redo this one.

PR #2500, opened 2017-04-16
Future card : Breached Dome

PesiTheWizard:
Adds Breached Dome
Probably needs plenty of revision

_Not to be merged until card exists_


PesiTheWizard:
Breached Dome isn't on NetrunnerDB yet. I posted it so that improvements could be made so I could see how to better make an effect like this. I hope this won't turn into a reputation.

PR #2505, opened 2017-04-17
partial implementation of MCA Informant

JoelCFC25:
By far the most requested card right now, so maybe "most of the way" is good enough for now. Hosts the card on a Connection and gives the Runner a tag, then the tag goes away when MCA leaves play. The Runner will have to manually remove 1 click and 2 credits and then manually trash the host, as I don't understand how to copy new abilities onto cards.

### DIFF
--- a/src/clj/game/cards/operations.clj
+++ b/src/clj/game/cards/operations.clj
@@ -650,6 +650,19 @@
                                       :delayed-completion true
                                       :effect (effect (tag-runner :runner eid 1))}}}}
 
+   "MCA Informant"
+   {:implementation "Runner must deduct 1 click and 2 credits, then trash host manually"
+    :req (req (not-empty (filter #(and (has-subtype? % "Connection")
+                                         (installed? %)) (concat (all-installed state :runner)
+                                                                 (all-installed state :corp)))))
+    :prompt "Choose a connection to host MCA Informant on it"
+    :choices {:req #(and (has-subtype? % "Connection") (installed? %))}
+    :msg (msg "host it on " (card-str state target) ". The Runner has an additional tag")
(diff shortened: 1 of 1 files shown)


JoelCFC25:
Actually this needs more work, ugh. Forgot that "considered to have" is not actually having in numeric terms, like with Corporate Scandal.

nealterrell:
What does your comment mean, Joel? Is there a specific card interaction you think you've gotten wrong with this?

JoelCFC25:
Similar to Satellite Grid and Corporate Scandal, "considered to have" means this additional tag should be included as part of the answer to "how many tags do you have?", but *not* included in the count of tokens (pretend we're talking physical components) that you can remove, pay with, etc. 

So when the Corp sticks this stupid card on their Corporate Troubleshooter, this implementation would not stop the Runner from using the Remove Tag button and spending a click and 2c to remove the tag. So just like Corp Scandal split bad publicity into `:bad-publicity` and `:has-bad-pub`, apparently now the Runner needs `:tag`, `:tagged`, and `:additional-tag`. Super.

nealterrell:
OK, that's what I figured you meant :+1:. 

You'll either have to settle for this and let the players learn they're not supposed to remove the MCA tag, or do your split and replace all cards that read `(:tag runner)` with a helper function that adds `:tag` and `:additional-tag`. That's really how it should've been designed in the first place... I've gotten pretty turned off on Clojure's overreliance on keywords lately; having a custom function whose implementation can change as the game mutates would've been a better way to go from the start. Oh well.

danhut:
Great! This will work nicely with Observe & Destroy - bleurgh

PR #2511, opened 2017-04-23
Rework Extra Costs, Choice of payment order on steal, implement Obakata Protocol, Donut Taganes, others

danhut:
Thanks for @JoelCFC25 for the Donut code.
The pay-forfeit function current only takes an amount of 1 - will need work if we ever see 2 or more. 

(Future) Other stuff we might do to move all costs into one place
- Move "pricing" for Doubles and Triples out of play-instance
- Move extra costs out of play 
- Move some of these cards to the "new way" 
https://netrunnerdb.com/find/?q=x%3A%22As+an+additional+cost+to+play%22+s%21Double

### DIFF
--- a/src/clj/game/cards/agendas.clj
+++ b/src/clj/game/cards/agendas.clj
@@ -560,6 +560,9 @@
                        :effect (req (gain state side :credit
                                           (if (>= (:advance-counter (get-card state card)) 5) 3 2)))}}}
 
+   "Obokata Protocol"
+   {:steal-cost-bonus (req [:net-damage 4 :credit 1])}
+
    "Personality Profiles"
    (let [pp {:req (req (pos? (count (:hand runner))))
              :effect (effect (trash (first (shuffle (:hand runner)))))
--- a/src/clj/game/cards/agendas.clj
+++ b/src/clj/game/cards/agendas.clj
@@ -561,7 +561,7 @@
                                           (if (>= (:advance-counter (get-card state card)) 5) 3 2)))}}}
 
    "Obokata Protocol"
-   {:steal-cost-bonus (req [:net-damage 4 :credit 1])}
+   {:steal-cost-bonus (req [:net-damage 4])}
 
    "Personality Profiles"
    (let [pp {:req (req (pos? (count (:hand runner))))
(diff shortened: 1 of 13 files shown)


PesiTheWizard:
Can we have Student Loans while we're at it?

danhut:
It is at least possible to implement now.  Something like:
   "Student Loans"
   {:msg "BLAH"
    :events {:pre-play-instant
             {:effect (effect (play-cost-bonus [:credit (* 2 NUMBER OF COPIES IN DISCARD)]))}}}

Need to build a real deck now... not played in weeks and not MWL compliant

JoelCFC25:
Stupendous. Would `pay` have to become delayed completion to allow for "trash an installed card" additional costs on Trade-In, The Price of Freedom, Scavenge, and Mutate? Something like: `:additional [:trash-installed {:type "Hardware"}]` so you can pass a parameter that the targeted card has to match? One other one that isn't handled is `[:tag 1]` for Observe and Destroy. 

I admit I don't fully understand it all yet, but man--looks awesome so far!

JoelCFC25:
This PR also makes Ben Musashi possible.

danhut:
A few things

- any new type of cost is expected to be passed as a vector [:cost-type cost-value].  I do some processing on :forfeit to make it fit this pattern.  I don't think we should have a cost that gets to break the rule.  
- for the example with trash type hardware.... it would be to be `:additional [:trash-installed cost value {:type "Hardware"}] ` and then merge costs would need a filter to ignore the map.  Trivial to implement i think.
- not sure about delayed completion.  Given pay-forfeit works fine if you have to choose which agenda to forfeit maybe not.
- as mentioned in the PR pay-forfeit only handles 1 agenda right now.  I was having issues with the function return value (cost string) which should pass back to pay when I tried to make it recursive
- for paying a tag I can do this easily

I'll have a look at these enhancements but i think this PR can go in when MT gets time

danhut:
Added Student Loans

danhut:
Added more to enable runner to choose order of paying costs when stealing.  This handles stacking Obokata Protocol and Ben Musashi  - but also others.

Issue #2058, opened 2016-10-06
1.1.1.1 format implementation


Imerxion:
Hello, guys! I had originally posted this on reddit, but I was told to post it here as well. 

I love Jinteki.net, and from my perspective it's seemed to be one of the greatest things that could've happened to the community. The ease of logging on, keeping track of decks, finding players, chatting, and choosing between casual and competitive has probably brought in and kept more players than anyone else thinks (but that's pure speculation on my part).
I also love MtG, but not as much as A:NR. But one of the greatest things about MtG is the sheer amount of ways to play the game, from all the different formats, adding players, Commander, etc. Depsite the game being balanced for two people, it still retains all these different ways to play.
I really think a great way to push new ways to play- a way to show more people that there can be more to this wonderful game, is to implement those new ways to play on Jinteki.net.
Next to the "Competitive" and "Casual" buttons, a third button could be added for the 1x1x1x1 format I was seeing pop-up a bunch on here. And (even more important in my book) a "Draft" tab for a 1v1 grid draft. Find a player, and a randomized 9x9 grid of cards from the latest Stimhack cube would appear on the screen for both players, and then they draft from those and play right then and there.
I am nowhere near a good enough programmer to implement these myself, but for the people out there who are and are reading this, please consider this. It'd be excellent for the casual players, new players, and even veterans who are still looking for more ways to play. When rotation comes around, a "Legacy" mode would also be great, and maybe in the far far future, we could even implement some 1v2 and 2v2 rules (though I know this would require a great more deal of designing and testing). Still, more ways to play the game we love is great, and I hope it can be as accessible as possible to everyone across the globe!

Saintis:
The 1.1.1.1 format is something that could happen in the future. Drafting would likely require a lot of effort to implement so it is a lot less likely this will be created unless one of the contributors decides to take on the project personally.

Generally we are against adding more tabs at it could cause longer waiting times for players. I suggest "Casual" be used as an umbrella tab for all games, including 1.1.1.1, personal drafts, "Legacy" games and others.

I am currently working on a rewrite of the deck-checker logic, and once that is done it should (hopefully) be easy to extend it to allow for 1.1.1.1 specific decks.

zaroth:
Split grid drafting onto separate issue, let this be about 1.1.1.1 format.

dStonecypher:
Can we merge into this a request for Cache Refresh Format as well?
[...]

PR #2542, opened 2017-05-15
#2058 - Cache Refresh and 1.1.1.1 format support

JarekWarsaw:
I've pushed my implementation of this feature. It includes both Cache Refresh and 1.1.1.1 format.
_Note: it's my first contact with Clojure, so I'm fairly sure that code is not optimized nor pretty._
Detection of Big Boxes is iffy - currently it's defined as packs with more than 20 cards.

Only in deckbuilder there's an additional tooltip naming example cards from deck violating given format. One card per problem in Cache Refresh, up to two cards per problem in 1.1.1.1. It should be enough for user to narrow down why deck is non-compliant with given format.

jarek_20170515_cache_refresh_jinteki

### DIFF
--- a/data/fetch.coffee
+++ b/data/fetch.coffee
@@ -28,6 +28,7 @@ setFields = {
   "name" : same
   "date_release" : (k, t) -> ["available", if t is null then "4096-01-01" else t]
   "cycle_code" : (k, t) -> ["cycle", capitalize(t.replace(/-/g, " "))]
+  "size" : (k, t) -> ["bigbox", t > 20]
 }
 
 mwlFields = {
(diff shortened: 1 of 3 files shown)


danhut:
Looks awesome to me.  Did not check it on the actual browser but code looks good.

danhut:
@JarekWarsaw Hi Jarek - i think this change has broked the "npm run fetch" task somehow.  If i have a new instance of Mongo it imports all cards fine - but if I re-run it I lose a bunch of cards- and more if I do it again.  This is causing the Rebirth issues (and other) on the live server.

Any ideas?  I will take a look too

JarekWarsaw:
@danhut Please check #2566 - it might be the cause of your problems

PR #2557, opened 2017-05-25
faq 4.0 changes

### LINKED ISSUE #2555: Scavenge needs an update based on FAQ 4.0
Scavenge no longer has a requirement for a program to be in the heap or grip. The only requirement needs to be that a program is installed. Also trashing the program is no longer an additional cost, but part of the card's effect.

### LINKED ISSUE #2556: FAQ 4.0 Changes
Summary of changes from Beyoken's video to look at:

**Power Shutdown**
X cannot exceed highest install cost among all programs/hardware

**Scavenge**
No longer has trashing as additional cost
Effectively reads "Trash a program, if you do, install a program"
Can now be played with Shadow Net
Valid target no longer required at time of playing scavenge

**MCA informant**
Cannot be hosted on corp card

**Sec Testing**
When your turn begins you MAY choose a server

**Fumiko Yamamori**
Triggers after psychic field so can kill if hand empty

**Timing changes of note - need to see if these need game engine changes.**
Explicit step for passing ICE
Explicity step in turn for refreshing

danhut:
resolve #2555 
resolve #2556

### DIFF
--- a/src/clj/game/cards/ice.clj
+++ b/src/clj/game/cards/ice.clj
@@ -278,7 +278,7 @@
     :runner-abilities [{:label "Take 1 tag"
                         :delayed-completion true
                         :effect (req (system-msg state :runner "takes 1 tag on encountering Authenticator to Bypass it")
-                                     (tag-runner state :runner eid 1))}]
+                                     (tag-runner state :runner eid 1 {:unpreventable true}))}]
     :subroutines [(gain-credits 2)
                   end-the-run]}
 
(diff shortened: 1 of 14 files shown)


danhut:
Think that's it all done.

src/clj/game/cards/ice.clj
```diff
@@ -272,6 +272,16 @@
    "Asteroid Belt"
    (space-ice end-the-run)
 
+   "Authenticator"
+   {:implementation "Encounter effect is manual"
+    :abilities [give-tag]
+    :runner-abilities [{:label "Take 1 tag"
+                        :delayed-completion true
+                        :effect (req (system-msg state :runner "takes 1 tag on encountering Authenticator to Bypass it")
+                                     (tag-runner state :runner eid 1))}]
```
JoelCFC25: Adding `{:unpreventable true}` to this tag is advised--Runner can't bypass if they don't actually take the tag.

PR #2573, opened 2017-06-01
Fixed race condition in fetch.coffee - fix #2566

### LINKED ISSUE #2566: Race condition in fetch.coffee - leads to missing cards
It seems that fetch.coffee has a race condition. It simultaneously removes records from collection and inserts new records. Depending on order, this might cause some newly-added cards to be removed from DB.
In my case fetch flip-flops between 1190 cards and 0 cards. I presume that when DB is empty, remove runs quickly and finishes first. When all cards are in DB, remove takes a bit longer and finishes second.

This applies to all 3 collections. Fix would require indenting block with insert, so it runs sequentially.

Generated JS with race condition
```
        db.collection("cards").remove(function() {});
        return db.collection("cards").insert(cards, function(err, result) {
          fs.writeFile("andb-cards.json", JSON.stringify(cards), function() {
            return console.log(cards.length + " cards fetched");
          });
          return callback(null, cards.length);
        });
```
Generated JS after fix
```
        return db.collection("cards").remove(function() {
          return db.collection("cards").insert(cards, function(err, result) {
            fs.writeFile("andb-cards.json", JSON.stringify(cards), function() {
              return console.log(cards.length + " cards fetched");
            });
            return callback(null, cards.length);
          });
        });
```

JarekWarsaw:


danhut:
Nice one!  Thanks

JoelCFC25:
So the upshot here is that I'd no longer have to wipe my local testing MongoDB every time I want to run a fetch to get new data from NRDB? That'd be so good.

mtgred:
Awesome 👍

PR #2582, opened 2017-06-03
implement Earth's Scion stuff

### LINKED ISSUE #2562: Careful Planning does not log which card was targeted
Rules are still followed, and the corp will be given the toast that it cannot rez the targeted card due to Careful Planning if she tries to do so, but the text log makes no mention of which card was targeted (or that a card was even targeted at all), which can make decision-making based around that difficult.

### LINKED ISSUE #2580: Steve Cambridge vs events that make runs on HQ.
Observed the following:

- My opponent, playing as Steve Cambridge, had two cards in the heap. Neither was an Inside Job.
- They used Inside Job to make a run on HQ. The run was successful.
- They used Steve's ability, choosing Inside Job and one of the other cards.

Since Inside Job hadn't finished resolving, they shouldn't have been able to choose it.

Presumably related to #322.

### LINKED ISSUE #2584: Chiyashi does not trash cards when AI is installed
I believe this is currently not automated? Easy to do manually, of course, but should be noted in one of those orange text boxes.

danhut:
fix #2562 
fix #2584
fix #2580

### DIFF
--- a/src/clj/game/cards/events.clj
+++ b/src/clj/game/cards/events.clj
@@ -94,15 +94,18 @@
     :effect (effect (install-cost-bonus [:credit -3]) (runner-install target))}
 
    "Careful Planning"
-   {:prompt "Choose a card in or protecting a remote server"
+   {:prompt  "Choose a card in or protecting a remote server"
     :choices {:req #(is-remote? (second (:zone %)))}
-    :effect (effect (register-turn-flag!
-                      card :can-rez
-                      (fn [state side card]
-                        (if (= (:cid card) (:cid target))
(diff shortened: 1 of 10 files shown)


JarekWarsaw:
Nice rework of Maw  :) Although I have a question - why is there a prompt for Runner if they want to use Maw? It's not an optional ability. And on the other hand, maybe add back ability to trigger Maw manually, in case someone clicks No or game state has to be reverted?

danhut:
Thanks - will adjust it.  I just did Aeneas Informant which needed the new "no-trash" event and did not notice it was non-optional.

PR #2634, opened 2017-07-11
Reeducation and Success implementations + small refactor

kevkcc:
Engine changes: 
Can now advance a card at no cost
Removed redundant parameter from reorder-choice function

EDIT: Reverted the refactor because I thought of a case where the 'redundant' parameter may be useful in the future (no cards use it right now but maybe later)


kevkcc:
Actually don't merge yet I thought of an issue

kevkcc:
Should be fine now :)

mtgred:
👍

PR #2647, opened 2017-07-15
All programs from Free Mars, Helheim servers, Mars for Martians

kevkcc:
Refactors the tagged ETR subroutine into its own function (IP block, pachinko, etc) in preparation for Data Loop.

I think Data Loop's on encounter effect is actually easier to do manually than through prompts so I'm leaving it manual until I hear otherwise

Also did god of war

### DIFF
--- a/src/clj/game/cards/events.clj
+++ b/src/clj/game/cards/events.clj
@@ -82,9 +82,9 @@
     :effect (effect (gain :credit 1) (draw 2))}
 
    "Calling in Favors"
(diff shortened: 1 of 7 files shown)


kevkcc:
Yea the current way is the 100% correct rules implementation but we could definitely go with the unpreventable tag since that's realistically what everyone would use. Maybe one edge case would be if you wanted to vamp yourself so they can't use bryan stinson. What do you think?

Issue #2654, opened 2017-07-19
Personality Profiles lies about the discard


PesiTheWizard:
Just had a game against Personality Profiles. It caused me to discard Order of Sol. Order of Sol was the top card of the heap and was visible from the main board. Log said I discarded Crash Space. Crash Space had been discarded earlier in the game and was at the bottom of the heap.

JoelCFC25:
Line 654 should be `(:title (last (:discard runner)))` instead of `(:title (first (:discard runner)))`.

PR #2664, opened 2017-07-26
Fixing issue #2654 - Personality Profiles

stoppableforce:
Per [#2654](https://github.com/mtgred/netrunner/issues/2654) Personality Profiles was always naming the lowest card in the Runner's discard pile instead of the one it actually discarded; this fixes that. Big ups to Joel for doing all the hard work of identifying the problem and exactly where to fix it. ;)

### DIFF
--- a/src/clj/game/cards/agendas.clj
+++ b/src/clj/game/cards/agendas.clj
@@ -665,7 +665,7 @@
    "Personality Profiles"
    (let [pp {:req (req (pos? (count (:hand runner))))
              :effect (effect (trash (first (shuffle (:hand runner)))))
-             :msg (msg "force the Runner to trash " (:title (first (:discard runner))) " from their Grip at random")}]
+             :msg (msg "force the Runner to trash " (:title (last (:discard runner))) " from their Grip at random")}]
      {:events {:searched-stack pp
                :runner-install (assoc pp :req (req (and (some #{:discard} (:previous-zone target))
                                                         (pos? (count (:hand runner))))))}})


danhut:
Looks good!

danhut:
Think this was not a good fix after all.  Causing the agenda test to fail.  Suggest you do a "lein test test.all" before future PRs if possible, or even better (as I learned) link a free CircleCI account to your own github so it runs the tests for you if you forgot.

This is the test that is now broken - but is still the right test:

```
(deftest personality-profiles
  ;; Personality Profiles - Full test
  (do-game
    (new-game (default-corp [(qty "Personality Profiles" 1)])
              (default-runner [(qty "Self-modifying Code" 1) (qty "Clone Chip" 1)
                               (qty "Corroder" 1) (qty "Patron" 2)]))
    (starting-hand state :runner ["Self-modifying Code" "Clone Chip" "Patron" "Patron"])
    (score-agenda state :corp (find-card "Personality Profiles" (:hand (get-corp))))
    (take-credits state :corp)
    (play-from-hand state :runner "Self-modifying Code")
    (play-from-hand state :runner "Clone Chip")
    (let [smc (get-in @state [:runner :rig :program 0])]
      (card-ability state :runner smc 0)
      (prompt-choice :runner (find-card "Corroder" (:deck (get-runner))))
      (is (= 2 (count (:discard (get-runner))))))
    (let [chip (get-in @state [:runner :rig :hardware 0])]
      (card-ability state :runner chip 0)
      (prompt-select :runner (find-card "Self-modifying Code" (:discard (get-runner))))
      (is (last-log-contains? state "Patron")
          "Personality Profiles trashed card name is in log")
      (is (= 3 (count (:discard (get-runner))))))))
```

danhut:
And the failure:
` :text " uses Personality Profiles to force the Runner to trash Clone Chip from their Grip at random."}], :trash {:trash-prevent nil}, :effect-completed {}}}] "Patron"))`

So we are claiming Clone Chip got trashed by personality profiles - which is the card we are firing - and Patron is what should appear.

JoelCFC25:
I'm willing to bet that the test would pass if Clone Chip was doing `runner-install` first and `trash` itself second.

danhut:
Probably good for now.  There is probably some problem longer term as you need strict sequencing in other cards with recursion to not have this agenda stuff up again.  I guess safer would be to set the reason for trash in the agenda to :card-effect :pp  - and then look for the newest card with that key before printing.

danhut:
So close Joel... The log is now right... just last log contains still sees the clone chip install:

 {:user "__system__", :text " uses Personality Profiles to force the Runner to trash Patron from their Grip at random."} {:user "__system__", :text " uses Clone Chip to install Self-modifying Code."}], 

Just need to add and "2nd Last Log Contains" function

danhut:
Solved here:
https://github.com/mtgred/netrunner/pull/2683

Issue #1530, opened 2016-05-16
when showing the hand, sort the cards


queueseven:
Salem's Hospitaliy / Celebrety Gift

PR #2665, opened 2017-07-26
Fix for #1530 (sorting card names when hand reveal prints them to log)

stoppableforce:
At least, assuming it means alphabetically. It doesn't really say.

Additionally, in a few cases where multiple card names are printed to the log by an effect other than showing a player's hand, those cards are now sorted alphabetically - namely Hellion Beta Test, Freelancer, Invasion of Privacy, and Biased Reporting.


danhut:
Looks good to me.

PR #2673, opened 2017-07-29
fixes + implement Open forum / Tithonium

### LINKED ISSUE #2655: Ayla with Rebirth - NVRAM cards go into Heap
They should go.... somewhere else ??? RFG?  Some other RAM?   Nowhere?

### LINKED ISSUE #2671: Reaver draws multiple cards when using Freelance Coding Contract.
Playing Freelance Coding Construct with an installed Reaver causes it to draw cards.

danhut:
fix #2671 
fix #2655


danhut:
Setting :condition true on stuff like Patch is for future when Rezzing Tithonium will trash anything without that flag on rez... When i get the energy

mtgred:
👍

PR #2700, opened 2017-08-15
Maw update

### LINKED ISSUE #2695: Maw - handle accessed card being face up
this is not handled from UFAQ:

If the Runner does not steal or trash the first card they access in HQ, and that card is then randomly selected for Maw's trigger, is the card trashed to Archives faceup or facedown?

    The Corp trashes that card to Archives faceup as it is currently visible to the Runner.

Saintis:
Adds :no-trash trigger for non-agenda cards with no trash cost. Makes Maw card discard face-up if it is the card being accessed (triggering the :no-trash event hook).

Adds test for this.

Currently auto-maw's once a non-agenda card with no trash cost is accessed. Would like some `when-complete` magic to only trash after the "OK" button is selected.

Fixes #2695

### DIFF
--- a/src/clj/game/cards/hardware.clj
+++ b/src/clj/game/cards/hardware.clj
@@ -446,7 +446,11 @@
                                  (not= (first (:zone target)) :discard)))
                   :once :per-turn
                   :msg "force the Corp to trash a random card from HQ"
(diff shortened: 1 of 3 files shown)


danhut:
Looks good - nice one!

mtgred:
👍

Issue #2701, opened 2017-08-16
Unnecessary/distracting start-of-turn prompts


TheRealC:
Certain cards like Apex and Out of the Ashes offer a start-of-turn pop-up asking you whether you want to use them. While thoughtful, this leads to a **lot** of turns opening with you having to click "No" several times - even if you _do_ intend to use these start-of-turn abilities, you might wish to order them differently than the order you are prompted in (e.g. first use Chopbot 3000, then activate Apex's ability, then decide whether you want to use Out of the Ashes - while prompt order is Apex first, then Out of the Ashes, then you can manually activate Chopbot).

I do not have an exhaustive list of such cards, but I'd suggest all such cards found get the same treatment as cards like Chopbot, Clone Suffrage and Hades Shard - usable on-demand, with a friendly reminder in the top-right corner that you have start-of-turn abilities you might consider using, as opposed to a potentially intrusive prompt.

Note that this is separate from the ordering of mandatory start-of-turn cards like Wyldside and PAD Campaign, but I guess that's covered already.

JoelCFC25:
For anyone looking to work on this, mimic other cards that employ `:flags {:runner-phase-12 (req true)}`. This gives you the blue notification toast and a pause before you take your clicks to manually fire "when your turn begins" cards in the order you choose. If left unfired, any remaining cards will fire themselves in an order you can't control when you take your clicks.

JoelCFC25:
I took care of Apex but I have no idea how to make this with for Out of the Ashes.

nealterrell:
The problem with Out of the Ashes is you cannot click cards in the Heap to activate them. The Ashes prompt has to be passive (automatically shown when triggered by the engine), making certain interactions impossible, like doing an Ashes run then following that with Chop Bot.

TheRealC:
Yeah, I've only been able to resolve stuff like that before by doing /rfg on the Ashes and adding a click. Not the most elegant solution, but it at least works. I guess implementing interactions with cards in Archives might be kinda rough, so it's workable for now. Honestly I'd be happier without the Ashes pop-up at all, but I realize there probably needs to exist a non-command way to trigger it, and that's seemingly it.

Thanks for your time, regardless ^^

NoahTheDuke:
Start of Turn effects are much better handled now. There are some occasional "unnecessary" prompts, but especially with simultaneous events, we should be down to specific, limited issues. Please open new issues for specific cards/events.

PR #2708, opened 2017-08-22
Fixes: Titanium Ribs, Security Nexus, The Price of Freedom, Space Camp

JoelCFC25:
* Fixes the much-plagued Titanium Ribs card (interactions with Stimhack and Salvaged Vanadis Armory) which is seeing a bewildering amount of use lately. 
* Reworks Apex to play nicer with other "when your turn begins" cards (per #2701) and update tests. 
* More thorough attempt at hard-to-reproduce Space Camp issue
* Fix hideously broken code for The Price of Freedom
* Fix once-per-turn on Security Nexus
* Prevent Bloo Moose if a Blacklist is rezzed
* Check for card being disabled when doing `deactivate` to take care of Rumor Mill + Cybernetics Court trashing and using Dr. Lovegood on an Independent Thinking that subsequently gets trashed.

### DIFF
--- a/src/clj/game/cards/identities.clj
+++ b/src/clj/game/cards/identities.clj
@@ -93,9 +93,11 @@
                   :choices {:max 1
                             :req #(and (= (:side %) "Runner")
                                        (in-hand? %))}
-                  :req (req (> (count (:hand runner)) 0))
+                  :req (req (and (pos? (count (:hand runner)))
+                                 (:runner-phase-12 @state)))
                   :effect (effect (runner-install target {:facedown true}))}]
      {:events {:runner-turn-begins ability}
+      :flags {:runner-phase-12 (req (pos? (count (:hand runner))))}
       :abilities [ability]})
 
    "Argus Security: Protection Guaranteed"
--- a/src/clj/game/cards/resources.clj
+++ b/src/clj/game/cards/resources.clj
@@ -236,7 +236,8 @@
 
    "Bloo Moose"
    {:flags {:runner-phase-12 (req true)}
-    :abilities [{:req (req (:runner-phase-12 @state))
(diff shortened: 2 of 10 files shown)


danhut:
Nice one Joel!!

src/clj/game/cards/identities.clj
```diff
@@ -93,9 +93,11 @@
                   :choices {:max 1
                             :req #(and (= (:side %) "Runner")
                                        (in-hand? %))}
-                  :req (req (> (count (:hand runner)) 0))
+                  :req (req (and (pos? (count (:hand runner)))
+                                 (:runner-phase-12 @state)))
```
Saintis: This might make Apex players sad if they don't remember to click Apex during the 1.2 window. Restricting when this can be used just makes correcting mistakes harder I think.

src/clj/game/cards/resources.clj
```diff
@@ -236,7 +236,8 @@
 
    "Bloo Moose"
    {:flags {:runner-phase-12 (req true)}
-    :abilities [{:req (req (:runner-phase-12 @state))
+    :abilities [{:req (req (and (:runner-phase-12 @state)
```
Saintis: Same as Apex comment.

src/clj/game/cards/identities.clj
JoelCFC25: No one will miss an opportunity to fire Apex. It will give you the blue notification and the pause before taking clicks to manually trigger it if it's important to order it relative to something else. If left unclicked during that pause, it will fire when you hit "Take Clicks" (via the `:runner-turn-begins` event).

Saintis: Good points.

mtgred:
👍

Issue #2661, opened 2017-07-24
Eater and Counter surveillance interaction.


vkl125:
Eater does not allow access yet counter surveillance will allow the runner to access x number of cards for the declared server on a counter surveillance run.

stoppableforce:
Counter Surveillance declares a max-access again, overriding the one from Eater; it shouldn't do so. The implementation appears to be based on a misunderstanding of how the card works, and also does not work correctly with other sources of bonus accesses like RDI, Medium, HQI, and Nerve Agent.

JoelCFC25:
> Counter Surveillance declares a `max-access` again

Yeah, this part is definitely wrong. Dropping that alone makes it combo with an HQ Interface correctly. I haven't tried all the other stuff yet.

stoppableforce:
>I haven't tried all the other stuff yet.

Medium/Nerve Agent are going to be the tricky one. They appear after you pay for Counter Surveillance, and the Medium/Nerve Agent access seems to completely replace Counter Surveillance's access. I'm guessing this is again due to CS being a weird access replacement that's currently implemented as a bonus access; Medium/Nerve Agent just replaces it a second time.

Saintis:
This is probably fixed in my PR, mostly since the `:max-access` is removed.

PR #2713, opened 2017-08-24
Counter Surveillance and Co. fixes

Saintis:
Should fix the Obelus + Counter Surveillance issue. Also fixes a few more minor things.


Saintis:
This should also allow Medium + RDI etc to increase number of Counter Surveillance accesses as per the UFAQ.

Saintis:
Probably fixes #2661.

mtgred:
👍

PR #2725, opened 2017-08-27
Displaying alt art cards in card browser and deck builder

jwarwick:
Displaying alternate/promo art cards in the card browser and deck builder if the user is authorized to see them (same logic as used in the game board).

Noticed that there doesn't seem to be an Easy Mark alt art png file, though it is listed in the promo.json file. I removed it from there, but someone will need to re-run `npm run promo` on the server to update the database.

I didn't see any automation around handling of alt art cards, not sure what process is in place today to get them added.

### DIFF
--- a/data/promo.json
+++ b/data/promo.json
@@ -212,9 +212,6 @@
     "code": "01111",
     "versions": ["wc2015"]
 }, {
-    "code": "01019",
-    "versions": ["alt"]
-}, {
     "code": "01109",
     "versions": ["alt"]
 }
--- a/src/cljs/netrunner/cardbrowser.cljs
+++ b/src/cljs/netrunner/cardbrowser.cljs
@@ -20,13 +20,27 @@
 (defn make-span [text symbol class]
   (.replace text (js/RegExp. symbol "g") (str "<span class='anr-icon " class "'></span>")))
 
-(defn image-url [{code :code}]
+(defn- image-url-string
+  [code, version-string]
+  (str "/img/cards/" code (when version-string (str "-" version-string)) ".png"))
+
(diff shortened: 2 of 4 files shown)


mtgred:
The card browser initially showed alt-art cards and had a "Alternate art" row in the set filter.

Alt art cards were actually listed in a different set altogether. This data implementation was changed for various reasons which made it incompatible with the card browser initial implementation. It resulted in the alt art cards not being showed in the card browser anymore.

It would be nice to restore the former behaviour of the card browser. That is show both regular and alt art version, with the ability to only show alt art cards via the set filter. Instead of only showing the regular or alternate version as implemented in this pull request.

jwarwick:
Okay, that was actually going to be my next step, but I'll roll it into this pull request.

I was thinking of showing the regular and alt cards at the same time and leaving the hiding of alt cards up to the settings and `{:user :special}` flag. Or would you prefer to let all users see the alt cards (still respecting their settings)?

jwarwick:
Updated the card browser to show all alt art cards. If the alt art preference is turned off, they won't show up. Otherwise, all cards show up in the card browser and there is an 'Alternative Art' cycle which can be used to filter them. In the deck builder, only users with a `:special` flag will see the alt art cards in their decks.

mtgred:
Thanks for the updated PR.

There are merges conflicts after the recent merges. Can you have a look at it?

jwarwick:
I think I'm going to retract this pull request. I've been working with @nealterrell on rethinking how we handle art in the deckbuilder and it will conflict in many ways with this. Probably better to roll this feature into that branch.

Issue #2728, opened 2017-08-31
PAD Factory doesn't let you advance your opponents' cards


MrCowbird:
Apologies for how narrow this issue is, but PAD Factory does not allow you to advance your opponent's cards (which is newly sorta relevant for Mass Commercialization purposes, and I suppose always interesting for Trick of Light).

Can resolve manually with "/counter ad 1" command, however.

Saintis:
Should be easy fix to allow for runner cards to be targeted as well.

PR #2729, opened 2017-08-31
Allow Pad Factory to put advancement tokens on Runner cards

jwarwick:
Fix for Issue #2728

### DIFF
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -988,8 +988,7 @@
    "PAD Factory"
    {:abilities [{:cost [:click 1]
                  :label "Place 1 advancement token on a card"
-                 :choices {:req #(and (= (:side %) "Corp")
-                                      (installed? %))}
+                 :choices {:req #(installed? %)}
                  :msg (msg "place 1 advancement token on " (card-str state target))
                  :effect (req (add-prop state :corp target :advance-counter 1 {:placed true})
                               (let [tgtcid (:cid target)]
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/assets.clj
```diff
@@ -988,8 +988,7 @@
    "PAD Factory"
    {:abilities [{:cost [:click 1]
                  :label "Place 1 advancement token on a card"
-                 :choices {:req #(and (= (:side %) "Corp")
-                                      (installed? %))}
+                 :choices {:req #(installed? %)}
```
JoelCFC25: This works fine--it can also be as simple as `{:req installed?}`

JoelCFC25:
Another idea to work in conjunction with this--Red Planet Couriers needs to scan all Runner cards for advancements in addition to Corp cards!

src/clj/game/cards/assets.clj
jwarwick: Cool, I assumed it has to be the whole anon function. I've simplified it in the latest commit.

jwarwick:
Looking at the code, it seems like there are a few cards that might need to be addressed: Shipment from Tennin, Dedication Ceremony.

Mass Commercialization appears to consider both sides already.

JoelCFC25:
Tennin can target Runner cards already. Dedication Ceremony would need the tweak:  Runner and installed OR Corp and rezzed

jwarwick:
I updated Dedication Ceremony and Red Planet Couriers.

I'm new to Netrunner, so I don't understand if I'm reading Shipment from Tennin incorrectly or reading the code wrong. It doesn't look like I can place advancement tokens on Runner cards currently.

mtgred:
👍

PR #2735, opened 2017-09-03
fixes

### LINKED ISSUE #2721: Aumakua not firing on archives
Hi! Aumakua is not adding a counter when accessing cards in Archives (and not trashing / stealing anything).

### LINKED ISSUE #2722: Aumakua adds a counter on a Medium dig where a card is trashed or stolen
Hi! When making a Medium dig with Aumakua installed, a counter is added to Aumakua if the first card is not trashed/stolen even if a card further down R&D is trashed/stolen on the same run.

### LINKED ISSUE #2734: Oversight AI won't work with Tithonium
Playing Oversight AI (rez ICE ignoring all costs) targeting Tithonium brings up the prompt "Do you want to pay the alternative Rez cost?" (forfeit an agenda). If the corp selects no, the ICE remains unrezzed. 

I did not have the 9 credits to pay the non-alternative rez cost, so I don't know if the game would take deduct the 9 at that point.

danhut:
fix #2734
fix #2722
fix #2721


butzopower:
Regarding fix for 2734, will that only cover cases of "ignoring all costs"?  I've noticed something similar with Tithonium and Muertos Gang Member, which will only ignore rez cost (not additional costs).

mtgred:
👍

Issue #297, opened 2015-06-29
cancel in the deck editor doesnt work


Balkmeister:
When you press cancel in the deck editor the deck is still saved as if you had pressed save.

mtgred:
It's not actually saved on the server. But the local "draft" is stil kept in the memory of your browser. If you refresh your browser, it will fetch the version on the server.

But I see how that can be confusing and will try add reset of the local copy when clicking on the cancel button.

PR #2740, opened 2017-09-03
Refresh deck list when cancelling an edit

jwarwick:
Fixes Issue #297

When the users chooses Cancel after editing a deck, this refreshes the deck view so the user does not see a stale list of cards. Retains currently selected deck.


JoelCFC25:
Holy smokes! Great job. I don't remember the last time an issue under 400 was solved. 🥇

jwarwick:
Thanks. I figured I'd work my way up from the bottom of the list... :)

mtgred:
👍

PR #2744, opened 2017-09-04
Base trace in prompt

Saintis:
Adds indication of base trace to Corp's trace prompt, and base link to Runner's prompt. Also some minor things.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -64,4 +64,7 @@
 
   :figwheel {:http-server-root "public"
              :server-port 3449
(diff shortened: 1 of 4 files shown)


Saintis:
Example (for Sunny).
Corp's side is the same, except it may have "2 + 1 + [choice]" if the corp has a bonus, so same how it is displayed in the chat window.
screen shot 2017-09-04 at 17 37 17

mtgred:
👍

PR #2745, opened 2017-09-05
Option to allow spectators to view hidden information.

### LINKED ISSUE #2743: Card visibility for spectators when "Make players' hands visible to spectators" is enabled
It would be useful if when the "Make players' hands visible to spectators" option is enabled on a game, then spectators could also mouseover facedown installed cards and facedown cards in Archives, so they have the same information as each of the players they are spectating. 

Games can be difficult to follow without this, as you have to notice what cards got installed as they're being installed and remember what cards are in what positions.

Saintis:
Changes specator option to allow view of _all_ hidden cards, not just hand.

Fixes #2743.

### DIFF
--- a/src/clj/game/main.clj
+++ b/src/clj/game/main.clj
@@ -114,6 +114,8 @@
     (private-card-vector state side deck)))
 
 (defn- private-states [state]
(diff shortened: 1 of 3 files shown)


mtgred:
👍

Issue #2531, opened 2017-05-06
[FEATURE] Links to Cards in Main Chatroom


zelmor:
Please make it possible for cards in the main chat rooms to be auto-linked when named, like it happens within the in-game chat during matches.

PR #2747, opened 2017-09-06
implement #395 & Fix #2724

danhut:
BOOM!  Bagged a sub-400 before they all went...
Fix #2724
Fix #395

PR #2750, opened 2017-09-07
Displaying card previews in chat

jwarwick:
Fixes issue #2531 

Moved the chat state into the main app-state. Using text parsing and preview display code from the gameboard.


danhut:
This will have a merge conflict with this
https://github.com/mtgred/netrunner/pull/2747

I think I need to change that one though as I merely renamed the chat atom. ???

jwarwick:
Yes, there are a few things in the pull request queue that are going to have conflicts. I'm not sure what the project policy is on how to resolve those in general.

For #2747, since I've already moved everything into the main `app-state`, we should be able to just add the wrapper around the input div.

I can roll that change into my branch easily if you want.

Saintis:
Generally what happens is some things get merged. Things with conflicts don't and then whoever has PRs with conflicts rebases and solves them 😉.

jwarwick:
Okay, I can work with that 😃

mtgred:
👍

Issue #2753, opened 2017-09-07
Clicking on an installed Earthrise Hotel causes a power counter to be spent and cards to be drawn


x3N1GM4x:
Perhaps there is some deliberate reason why the ability of Earthrise Hotel (to draw 2 cards for a cost of one power counter) can be triggered manually by clicking the card, but I thought it should only trigger on the start of the runner's turn?

Maybe there's some strange scenario where it might not trigger, so it's been written to allow this manual trigger, but even then it's not particularly difficult for the players to fix the game state themselves (removing a counter and drawing two cards for free), so I wonder if this might be a bug?

JoelCFC25:
> Perhaps there is some deliberate reason why the ability of Earthrise Hotel (to draw 2 cards for a cost of one power counter) can be triggered manually by clicking the card

Yes. The Runner can order "when your turn begins" effects in the order they see fit (or let them fire automatically). 

The ability needs `:req (req (:runner-phase-12 @state))` added.

PR #2755, opened 2017-09-08
Earthrise Hotel can only be used at start of Runner's turn

jwarwick:
Fixes Issue #2753 

Requires Runner to be in Phase 1.2 for usage.


mtgred:
👍

PR #2758, opened 2017-09-09
Fixed broken defn docstring/arg orderings

jwarwick:
Eastwood isn't able to figure out most of the namespace games, but it did catch a number of function definitions with the arguments and docstrings out of order.


mtgred:
Nice catch.

I wonder how the code actually compile ?!

jwarwick:
It just treats it as a string that gets discarded. Seems like it would be an easy thing for the compiler to catch

PR #2759, opened 2017-09-11
Fixed invalid argument in call to add-prop from add-counter

x3N1GM4x:
Fixed invalid `:advancement` argument in call to `add-prop` in function definition of `add-counter`.


x3N1GM4x:
Came across this when starting out writing some new unit tests and found that within the function definition of `add-counter`, when `add-prop` is called, the 4th argument is given as `:advancement`. However, looking at the source for `add-prop`, this is not valid and needs to be `advance-counter` to work properly.

mtgred:
👍

PR #2760, opened 2017-09-11
Handling mongo error when listing donors

jwarwick:
Showing up as a `TypeError: Cannot read property 'length' of undefined` in the server log. Can force it to happen by stopping the db.


mtgred:
👍

PR #2761, opened 2017-09-12
Handling next() middleware in register user

jwarwick:
The user registration handler was calling an undefined function (`next`) in certain error cases and attempting to send multiple responses in others. This was generating errors in the server log file.

Added the `next` parameter to the function args and straightened up the error flow.


mtgred:
👍

PR #2762, opened 2017-09-12
implement System Seizure

JoelCFC25:
The funky checks for `:per-turn` and manually setting it will allow icebreakers that can't auto-pump to strength match (e.g., Faust) to still get the all-run boost.


mtgred:
👍

PR #2763, opened 2017-09-12
Handling deck rotation information

jwarwick:
Updated the `fetch` script to pull the Cycle information from NetrunnerDB (and store it in Mongo), so we can access the `rotated` property of the cycles. Updated data packs (`sets`) with a `rotated` flag to determine if they are allowed in competitive play.

Updated the Deckbuilder `released?` function to check that a card does not belong to a rotated set.

You should run `npm run fetch` to update your local database to capture the new information.


JoelCFC25:
Awesome!

mtgred:
This is awesome indeed!

There's a merge conflict with after the batch of merges though. Can you have a look at it?

jwarwick:
Added a script to find duplicated cards in the db and add `replaced` and `replaced_by` fields. Need to either add an additional field to the db to capture `active` (eg. the cycle has been rotated, but the card is `replaced_by` something legal) or do that computation in the front end.

mtgred:
👍

PR #2771, opened 2017-09-16
Fix Leave button going missing if opponent leaves first

### LINKED ISSUE #1997: Problem leaving a game
If a game host leaves the game, the other player loses the "Leave Game" button, and needs to refresh the page to leave the game. There is also no message in the log.

danhut:
fix  #1997


mtgred:
👍

PR #2777, opened 2017-09-19
Implement Deck Stats

### LINKED ISSUE #2609: [Feature] Track win rates of decks
Hey guys, 

First I want to say thanks for the amazing work.

Idea: 

Would it be possible to keep track of the win rate of decks that we have saved on our accounts?
For example, together with the actual decks, keep track of win/loss on the database, and you can show the user, on the Deck Builder tab, the number of games played with that deck (counting from the last time that they saved any changes to the deck). You could show the values on the deck tiles on the left column in the Deck Builder.

wins: (int) — losses: (int) — win-rate: (%)

### LINKED ISSUE #2773: Opponent is typing indicator - can add vertical scrollbar at some resolutions
From Dev Slack:

I think the typing indicator looks fine , although for me (running Chrome 61.0.3163.91 at a res of 2048x1152 with an in-browser zoom of 125%), whenever the typing indicator appears, it causes a second vertical scrollbar to appear to the right of the chat window, which in turn pushes the chat window itself to the left slightly...

danhut:
This implements a games/wins/loses counter per deck.  

[Updated pic to show clear stats button]
screen shot 2017-09-22 at 9 14 17 am

Fix #2609
Fix #2773


danhut:
Reworked this to update deck-stats at the same time as game-stats via the node server.  Slight implication is the update is triggered when the second player leaves the game - which means their deck-stats won't be updated immediately on them going into the deck builder.

I don't think much can be done about it - unless we make the ZeroMQ comms a 2-way thing as we rely on a request from the node server to push game state right now.

Added a stats refresh button to allow user to pull stats without refresh to make this better.

mtgred:
👍

PR #2782, opened 2017-09-20
Titanium Ribs, The Cleaners fix

Saintis:
Fixes the Titanium Ribs extra damage issue and removes The Cleaners messages from other runner sources of Meat Damage.

Adding `:unboostable` should no longer be necessary as The Cleaners will only trigger from Corp source of damage now.


mtgred:
👍

PR #2785, opened 2017-09-23
Updates Node packages

Saintis:
Updates some node packages and adds package-lock file.

Does not update socket.io since that caused problems.

Things work locally, so node-trello and other production specific packages may cause issues.


mtgred:
👍

PR #2787, opened 2017-09-23
Add ability to block users

jwarwick:
There are a few changes here to allow you to filter out unwanted attention from other users.

1. A UI update to the Account page which allows you to add a username to your Block List and remove names from the list. It could use some UI help, but I'm at the limit of my CSS knowledge.

2. Client side filtering of messages from blocked users in the main Chat window

3. Server side check if any of the players in a game have a user on the blocked list on a `join` or `watch` request.

4. Client side filtering of the games lobby to both hide games from users on your blocked list and hide their games from you.

Updating the list requires you to submit an update on your Account page and refresh the browser. It appear that we push `user` info only on initial connection.

### DIFF
--- a/server.coffee
+++ b/server.coffee
@@ -522,10 +522,12 @@ app.post '/reset/:token', (req, res) ->
 app.post '/update-profile', (req, res) ->
   if req.user
     db.collection('users').update {username: req.user.username}, {$set: {options: {background: req.body.background,\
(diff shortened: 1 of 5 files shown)


danhut:
Looks fantastic John.  Great feature.

mtgred:
Awesome! 👍

PR #2790, opened 2017-09-27
fixes

### LINKED ISSUE #2717: Employee Strike not clearing properly.
The last few games I have played as Corp with Jinteki: Potential Unleashed the Corp ID has fired perfectly fine. Then when the Runner plays Employee Strike, the ID is turned off, as it should be (no trashing the top card of the deck when net damage occurs). Once Employee Strike has been cleared however, the Corp ID has not triggered properly again even once for the entire of the rest of the game.

This was always with Employee Strike being cleared through an agenda score, rather than playing a Corp current, if that's any help. I thought it was a weird bug the first time it happened, as I've never seen it happen before, but it seems to happen every game now.

### LINKED ISSUE #2751: Cannot cancel "Trash Resource" action as Corp.
When clicking "Done" two credits and a click is still deducted without notice in the Chat.

### LINKED ISSUE #2774: Security Nexus allowed firing multiple times during a turn.

### LINKED ISSUE #2779: Conundrum's first subroutine is incorrect
I think it needs to be something like `(trash-program :player :runner)` (I'm sure that's not exactly correct) rather than just `trash-program`. The runner gets to choose which program is trashed.

### LINKED ISSUE #2781: [Suggestion] Smarter "can play" outlines
Smarter "can play" outlines on cards, taking into account double, tripple, Day Job, priority etc.

danhut:
fix #2774 
fix #2779 
fix #2751 
 fix #2717 
 fix #2781

### DIFF
--- a/src/cljs/netrunner/gameboard.cljs
+++ b/src/cljs/netrunner/gameboard.cljs
@@ -242,10 +242,40 @@
                (get-in @game-state [:runner :rig (keyword (.toLowerCase (:type card)))]))]
     (some #(= (:title %) (:title card)) dest)))
 
+(defn has?
+  "Checks the string property of the card to see if it contains the given value"
+  [card property value]
+  (when-let [p (property card)]
+    (> (.indexOf p value) -1)))
+
+(defn has-subtype?
+  "Checks if the specified subtype is present in the card.
+  Mostly sugar for the has? function."
+  [card subtype]
+  (has? card :subtype subtype))
+
 (defn playable? [{:keys [title side zone cost type uniqueness abilities] :as card}]
   (let [my-side (:side @game-state)
         me (my-side @game-state)]
     (and (= (keyword (.toLowerCase side)) my-side)
+
+         (cond
+
+           (has-subtype? card "Double")
+           (if (>= (:click me) 2) true false)
+
+           (has-subtype? card "Triple")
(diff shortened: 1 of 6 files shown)


src/cljs/netrunner/gameboard.cljs
```diff
@@ -242,10 +242,40 @@
...
+
 (defn playable? [{:keys [title side zone cost type uniqueness abilities] :as card}]
   (let [my-side (:side @game-state)
         me (my-side @game-state)]
     (and (= (keyword (.toLowerCase side)) my-side)
+
+         (cond
+
+           (has-subtype? card "Double")
+           (if (>= (:click me) 2) true false)
```
Saintis: Good job!

The thing that kept me from starting this was an installed Starlight Crusade Funding which messes with this. Can either add a hard-check for that or ignore it since it has rotated.

danhut: I suspect other cards already break the playable? indicator such as Donut Taganes so I think leaving it is ok - and consistent.  And it is rotating ...

Saintis: Agreed. I got too completionist to start working on it 😉

danhut: you can always improve it - i am happy!

mtgred:
👍

PR #2792, opened 2017-09-28
Updated Shard PR

Saintis:
Updates the Shards PR #2303. Resolves merge conflicts.

Also adds a regression test and refactors the shards to use a constructor function for extra DRY-ness.

Thanks to @lostie for the original work!

Supercedes #2303.


mtgred:
👍

PR #2793, opened 2017-09-28
Rebase of #2116.

Saintis:
See #2116 for some details.

This is a rebased and slightly tweaked version of @erbridge's PR that implements some good string related functions: `pluralize` and `quantify`. It also adds a constructor for run events, and standardizes all ICE subtypes as Title Caps, and cleans up the use of "Select" and "Choose".

Select for when you use the target icon to pick something, choose when you pick from a list (I think).

Lots of changes from this. Also some refactors that did not need and work rebasing.

Might need some review to make sure it is all still good and wanted, but hopefully this will get some of the older PRs merged 😄

### DIFF
--- a/src/clj/game/cards.clj
+++ b/src/clj/game/cards.clj
@@ -1,6 +1,6 @@
 (in-ns 'game.core)
 
-(def trash-program {:prompt "Choose a program to trash"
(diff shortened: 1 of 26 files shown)


mtgred:
👍

Issue #2765, opened 2017-09-14
Political Graffiti bonus points issue (+ some others)


PesiTheWizard:
- [x] When an agenda hosting Political Graffiti is forfeited in any way, the point-loss from Political Graffiti is refunded twice.

----
(Less important things)

- [x] This only happens occationally, but using Raman Rai to trash Subliminal Messaging sometimes makes it so that it never HQ's itself again, even when there are other face-down Subliminal Messagings in Archives that do come back. (This may have also happened when using Whampoa Reclamation, though I can't be sure)

- [x] A successful Punitive Counterstrike has 2 log messages that really could be combined into 1.

Saintis:
My guess on Political Graffiti is that the forfeit agenda point loss happens before Political Graffiti is trashed, hence 1 less agenda point is lost, then Political Graffiti is trashed and refunds 1 agenda point as well.

Solution would be to make sure Political Graffiti is trashed before the agenda points are refunded.

Saintis:
Regarding Raman Rai, there is some dark magic going on in the card definition. I think it's to make sure the swapped card enters in the same place as the old card, but since Archives is not ordered I don't think it make a difference. It appears that because of this, the `:moved-zone` registered effect on Subliminal is not triggered when swapping Subliminal into Archives using Raman Rai.

Saintis:
The fix would be to rewrite Raman Rai to use `move` for the card going from HQ to Archives as well.

Saintis:
Political Graffiti is more devious. The issue there is that the when the agenda is forfeit, Political Graffiti is trashed as part of the whole move-to-rfg thing. The effect of trashing Political Graffiti is lost in the process and the agenda point value of the card isn't updated.

Solution is to first trash all hosted cards, then move to rfg.

JoelCFC25:
I probably did Raman Rai using Midori or Toshiyuki Sakai as a template (where swapping _had_ to occur this way) and didn't think it through fully.

PR #2795, opened 2017-09-29
Fixes for Political Graffiti and Punitive Counterstrike.

Saintis:
Fixes part 1 and 3 of #2765.

Makes forfeited agendas trash all hosted cards and update themselves before moving to RFG area.


mtgred:
👍

PR #2799, opened 2017-10-01
Implemented /roll command

### LINKED ISSUE #1195: Feature Request: Random roll command
I think it would be good if it were possible to roll an N-sided die, visible to both players in the chat log. I don't think it deserves a button/prompt in the UI. But a `/roll` command would be nice.

> /roll 3
> _bmatsuo tosses a die [1-3] and rolls a 1_

I think there are some spots where this could provide minor utility.
- Randomizing psi games (or giving the opponent the appearance of randomization).
- Sometimes you don't know what to do.

x3N1GM4x:
User can type "/roll n" to roll an n-sided die and have the result displayed in the chat

Fixes #1195.

### DIFF
--- a/src/clj/game/core/io.clj
+++ b/src/clj/game/core/io.clj
@@ -170,6 +170,9 @@
                                             (rez state side c {:ignore-cost :all-costs :force true}))))}}}
     {:title "/rez-all command"} nil))
 
+(defn command-roll [state side value]
+  (system-msg state side (str "rolls a " value " sided die and rolls a " (+ (rand-int (- value 1)) 1))))
+
 (defn command-close-prompt [state side]
   (when-let [fprompt (-> @state side :prompt first)]
     (swap! state update-in [side :prompt] rest)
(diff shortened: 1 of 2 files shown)


danhut:
Nice work. Worth updating the netrunner-help.cljs too so users know about it. 

Also add Fix #num in the comment so GitHub will auto close the linked issue.

src/clj/game/core/io.clj
```diff
@@ -170,6 +170,9 @@
                                             (rez state side c {:ignore-cost :all-costs :force true}))))}}}
     {:title "/rez-all command"} nil))
 
+(defn command-roll [state side value]
+  (system-msg state side (str "rolls a " value " sided die and rolls a " (+ (rand-int (- value 1)) 1))))
```
nealterrell: Clojure-fu: instead of `(+ ... 1)`, use `(inc ...)`. :) Likewise, `(dec ...)` instead of `(- ... 1)`.

x3N1GM4x: Plus, I now realise my logic was wrong anyway! Consider my wrists slapped for not testing thoroughly enough! 😄

mtgred:
👍

PR #2800, opened 2017-10-01
Enable rejoining of games after a browser refresh / "leave-game"

nealterrell:
Currently a `reconnect` message is sent when a client's websocket experiences a brief disconnect from the web server. This solution is fragile and doesn't work if the disconnection lasts for more than some unknown time. It also does not work if the user accidentally leaves the game or refreshes the browser.

This PR adds a Rejoin button on the game lobby for any game that you've left by either disconnect, page refresh, or Leave Game button. Rejoining causes the game server to emit a full game state, allowing the client to reinitialize the game board.  A rejoinable game shows up first in the game list.

This could use a once-over by anyone with some time.

### DIFF
--- a/server.coffee
+++ b/server.coffee
@@ -75,6 +75,17 @@ removePlayer = (socket) ->
       delete games[k]
       refreshLobby("delete", v.gameid)
 
+rejoinGame = (socket, gameid, user, options) ->
+  game = games[gameid]
+  if game and game.started and game.players.length < 2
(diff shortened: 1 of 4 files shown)


danhut:
This one deserves a big pat on the back. Nice work!

mtgred:
Amazing!

There's a merge conflict with the latest merges. Can you have look at it?

nealterrell:
Fixed.

mtgred:
👍

PR #2810, opened 2017-10-02
Enabled manual trigger ordering for The Archivist

### LINKED ISSUE #2809: Order of The Archivist and Human First triggers is not determined by Runner choice
Raised by @fingerzz in #general on StimSlack - it seems that the order in which the "on agenda score" abilities of these two cards fire cannot be chosen by the Runner (perhaps firing based on install order), meaning sometimes The Archivist fires before Human First, denying the Runner credits they might want to use to boost their link on the trace from The Archivist.

x3N1GM4x:
Fixes #2809.


mtgred:
👍

Issue #2811, opened 2017-10-02
Full Immersion RecStudio (Multiple)


LiteSpecter:
Bug: A trashed Marlyn Campaign will not prompt a reshuffle when it is hosted on Full Immersion RecStudio.

Bug (rather low priority): Breaker Bay Grid cost reduction does not properly apply to cards hosted on a nested RecStudio. It applies properly to the cards hosted on the root RecStudio, but if the corp installs another RecStudio on top of the first, the BBG reduction is lost.

Enhancement: Current implementation of cards like Restore show RecStudio as a possible location for asset installation, is this method possible for installing assets from HQ?

JoelCFC25:
Part one:  `:req (req (= :servers (first (:previous-zone card))))` probably needs to be: 
`:req (req (= :servers (first (:previous-zone (get-nested-host card)))))`

nealterrell:
Part 2: likewise, Breaker Bay Grid should be comparing the zones of its own nested host and the target's nested host. This could probably use a new `nested-zone` function or similar.

nealterrell:
Part 3: probably not. The UI that shows for installing a card from hand is fundamentally different from the UI you get when a card triggers an install effect without specifying a server. I suppose we could just abandon the former.

nealterrell:
@JoelCFC25 hmm, the previous-zone for a Marilyn on RecStudio is probably going to be `[:onhost]`. That would probably be sufficient evidence to conclude it was trashed while installed on a host.

PR #2813, opened 2017-10-02
Refactor "in this server" and "from this server" checks

nealterrell:
Adds a few utilities for:

* Getting the zone of a card's most-nested host.
* Testing two cards to see if they are *in* the same server (game rules: in the content of the same remote server)
* Testing two cards to see if they are *protecting* the same server (in the ice of the same server)
* Testing two cards if they are in-or-protecting the same server.
* Testing two cards if they are *from* the same server (game rules: one card is an upgrade in the root of a central, the other is in that central itself. Example: Red Herrings in the root of HQ, and an agenda in corp's hand.)

Cleans up the logic for *some* of the cards that affect other cards in/from the same server, notably Oaktown and BB Grid, and the "in order to steal an agenda from this server" upgrades (Red Herrings, Old Hollywood Grid, Strongbox, Ben Musashi). Also enables these cards to work even when they or their target are hosted on other cards like RecStudio.

Addresses the concerns of #2811.

### DIFF
--- a/src/clj/game/cards/upgrades.clj
+++ b/src/clj/game/cards/upgrades.clj
@@ -10,7 +10,7 @@
    "Amazon Industrial Zone"
    {:events
      {:corp-install  {:optional {:req (req (and (ice? target)
-                                                (= (card->server state card) (card->server state target))))
+                                                (protecting-same-server? card target)))
                                  :prompt "Rez ICE with rez cost lowered by 3?" :priority 2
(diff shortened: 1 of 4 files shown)


JoelCFC25:
Underway Grid test is bombing...

JoelCFC25:
Ah, because Underway Grid is using `protecting-same-server?` which is for ICE. Would need to be `same-server?`

nealterrell:
Whoops, yep @JoelCFC25.

mtgred:
Amazing!

PR #2814, opened 2017-10-03
Running link_cards script after fetching card data

jwarwick:
Linking cards allows us to find replacements for rotated cards.


mtgred:
👍

PR #2817, opened 2017-10-03
Display card influence dots

### LINKED ISSUE #2815: Deckbuilder - Bug - Unicorn/Restricted Symbol is over-writing influence pips
We should show both right?

jwarwick:
Fixes bug where influence dots were not displayed next to card names.
Also added a rotated symbol.
Displaying status symbols next to cardbrowser text.

Fixes #2815


mtgred:
👍

PR #2823, opened 2017-10-04
Filtering 'Watch' on the server side

jwarwick:
Previously only hiding the `Watch` button on the client side, so you could modify the javascript to join a game as a watcher. Now filtering on the server side and dropping requests based on the game state.


mtgred:
👍

PR #2830, opened 2017-10-04
enable chat logging

danhut:
For @nealterrell to push to dev please!

### DIFF
--- a/server.coffee
+++ b/server.coffee
@@ -130,6 +130,10 @@ requester.connect("tcp://#{clojure_hostname}:1043")
 
 sendGameResponse = (game, response) ->
   diffs = response.runnerdiff
+  console.log("RunnerDiffs:")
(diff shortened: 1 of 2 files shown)


danhut:
Definately looks like a timing problem.

In the capture below for a working case we send a command to the web server "typing" followed by a "say" command.  This works twice below in sequence.

The third time we send "typing" but no "say" so I *think* this is a client side code issue thus far.  But also only occurs due to some timing issue with the ACK from the server.  Maybe due to the client lock design?

screen shot 2017-10-05 at 6 49 29 pm

danhut:
I am guessing the command lock prevents the client sending the "Say" here.  I wonder if a fix is to have the typing indicator command bypass the lock - since it is not a standard command that has any bearing on game state?

PR #2842, opened 2017-10-06
Added req to Indexing for redirected runs

### LINKED ISSUE #2840: Indexing and AgiInfusion redirect bugged
image

x3N1GM4x:
In the same vein as for Account Siphon, ensures that the successful run must be on R&D in order to allow the replacement effect of Indexing to fire.

Fixes #2840.


x3N1GM4x:
I should add a test for this scenario as well, will push another commit later for this.

mtgred:
👍

PR #2844, opened 2017-10-06
Checking JWT token for more actions

jwarwick:
Requiring a valid JWT token to post messages in chat. Using the `username` from the token rather than the one passed in the message request.

Requiring a valid JWT token to `join` or `watch` games. Should prevent one path that was allowing `undefined` users as spectators.


mtgred:
There are some conflicts with the latest merges. Can you have a look?

jwarwick:
@mtgred merged changes

PR #2850, opened 2017-10-08
Consistent scrolling in main chat window

jwarwick:
Think the old method of deciding when to scroll the chat window was based on an old DOM layout.

Now we scroll to the bottom on first load, when changing rooms, when entering from another tab, when sending a message, or when scrolled to the bottom and new messages come in.

If you are manually scrolled away from the bottom, you don't snap to the bottom when new messages arrive.


x3N1GM4x:
Love it, this always annoys me when I'm trying to go back in chat and remember how many cards have been seen off the top of R&D, or what was Celebrity Gifted or seen with Salem's etc etc.

jwarwick:
Actually, I just implemented it for the main chat lobby. If the in-game chat has the same problem, it should be easy to add this behavior.

mtgred:
👍

PR #2852, opened 2017-10-08
Selecting correct card when multiple options exist in deckbuilder

jwarwick:
When typing in the name of a card that was rotated, then replaced in Core 2.0, we were selecting the rotated card. This was causing those cards to show a `rotated` symbol and not handle the MWL symbols correctly. When the name was fully completed, the correct card was selected.

Probably only applied to `Aesop's Pawnshop` and `Magnum Opus` for the MWL symbols.


mtgred:
👍

PR #2854, opened 2017-10-09
Fix issue with removing icons from cards with hosts

### LINKED ISSUE #2846: Dedicated Processor / Femme
Dedicated Processor disappeared from host (Femme Fatale) and game after trashing an ice with a Femme token on it.

nealterrell:
Fix #2846. When calling `remove-icon`, pass the updated version of the card that caused the icon.

Also rework Dedicated Processor to use `:hosting` installs like Parasite and Personal Touch.


mtgred:
👍

PR #2857, opened 2017-10-10
Fixes: Film Critic, Counter Surveillance

JoelCFC25:
Fix a Counter Surveillance bug and undo recent change to Film Critic which tried to address the interaction with The Future Perfect, but in fact unintentionally prevents FC usage after the first psi of a game.


mtgred:
👍

PR #2863, opened 2017-10-11
Fixed `Paige Piper` card trash string with double number

### LINKED ISSUE #2865: Paige Piper causes strange (but harmless) messages
Evening.

Lately, I've been seeing what appears to be a completely harmless bug associated with Paige Piper. When Paige is installed and the Runner plays a card & chooses to trash extra copies of it as per Paige's ability, messages of the following kind tend to appear:

- XXX trashes 11 copy of Maw.

There doesn't seem to be any actual error in the game state, although I might have just been missing something. Honestly it doesn't seem to do much else than give people an opportunity to shout "JUDGE!", but figured I'd report it anyways. I'd not be _opposed_ to just leaving it in as an easter egg ^^'

Oh, and the bug I'm talking about is obviously the lack of pluralization on "copy". Obviously!~

jwarwick:
Fixes #2865

### DIFF
--- a/src/clj/game/cards/resources.clj
+++ b/src/clj/game/cards/resources.clj
@@ -1101,7 +1101,7 @@
                                                      (doseq [c (take target cards)]
                                                        (trash state side c {:unpreventable true}))
                                                      (when (> (int target) 0)
-                                                       (system-msg state side (str "trashes " target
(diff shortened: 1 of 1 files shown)


jwarwick:
Fixes #2865

mtgred:
👍

PR #2867, opened 2017-10-12
Downloading MWL data from NRDB

jwarwick:
Netrunner.db has updated their MWL endpoint with the latest format. Pulling data from them, rather than parsing local file.


mtgred:
👍

PR #2868, opened 2017-10-13
Fixes forfeit requirement.

Saintis:
Should fix so that runners can now forfeit Corp agendas.

We should write a test for this.


mtgred:
👍

PR #2876, opened 2017-10-14
Various minor fixes

### LINKED ISSUE #2820: Rotate symbol should not correspond to faction
Also, I think rotation symbol should not have faction specific colour, but rather just red colour (or similar).

Saintis:
Cleans up card influence display in deck builder.

Refactors Dummy Box.

Fixes #2820.


jwarwick:
I was planning to add tooltips to the symbols which would have needed a very similar `span` refactor, so I'll wait until this goes through, then it's a very simple addition.

jwarwick:
`cardbrowser.clj` also has some symbol logic which we should update to use this refactoring.

Saintis:
Can always base your changes of my branch. I'll go have a look at what `cardbrowser.clj` is using.

PR #2892, opened 2017-10-19
Allow 'Removed From Game' area to be expanded

### LINKED ISSUE #2026: Feature suggestion: "Removed from game" pop-up
I assume this is a simple suggestion for better usability? Is it possible to get a pop-up box when you click on the "Removed from game" pile in the same way as you do for Archives and the Heap?

When you've been nasty to your opponent and removed a major portion of their deck from the game, it can be tricky to flick through the edges of the cards to see exactly what was removed.

jwarwick:
Fixes #2026

### DIFF
--- a/src/cljs/netrunner/gameboard.cljs
+++ b/src/cljs/netrunner/gameboard.cljs
@@ -877,17 +877,25 @@
                    (str total " cards, " (- total face-up) " face-down."))]]
         (for [c discard] (draw-card c))]]))))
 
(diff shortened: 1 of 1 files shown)


nealterrell:
Awesome. We also see requests for this on HQ/Grip. How feasible and/or disruptive would that be?

jwarwick:
This was my test run to figure out how difficult it would be to expand HQ. I don't think it will be too difficult, but we will need to define the clickable area to expand (since clicking a card plays it in many cases).

PR #2893, opened 2017-10-19
Bon cleaners fix

### LINKED ISSUE #2891: Cleaners and Builder of nations.
The extra meat damage doesn't proc with Builder of nations once you've scored a "The Cleaners" agenda.

jacobbillings:
The Cleaners now causes Builder of Nations ability to do 2 meat damage.
Fixes #2891


nealterrell:
Thanks for the submission. Did you reproduce the bug before trying this fix? I honestly can't think of why this change would affect anything, so I'm curious if you saw the bug in action or if you're getting a false positive.

(Since you're new, this would be a good time to create a unit test for this scenario. You should be able to use the individual tests for BoN and Cleaners as a comparison point.)

jacobbillings:
Indeed! Scoring a Cleaners and then activating BoN's ability during a run on an advanced ICE results in 1 meat damage consistently. Thanks for the advice on unit tests; I will get into that.

Saintis:
This should fix it. I think this was an unintentional casualy when I fixed The Clearners increasing meat damage on Cybernetic installs, making in check that the damage source was from `:corp`.

Might be better to change `side` to `:corp` to be extra safe, but it is probably fine as is.

PR #2921, opened 2017-10-30
Limit deck builder add function to legal number of cards

jwarwick:
Saw this as a comment in Jinteki chat about `Astroscript Pilot Program` (retro decks are all the rage). Trying to use the `Add` button in the deck builder with an illegal number of card copies as the quantity creates an `1x Unknown` card in the deck list.

Now checking for `:limited` field (or 3 if not specified) and constraining quantity to that limit.


jwarwick:
Hmm, apparently I based a later pull request off this branch. Closing in favor of that one.

PR #2922, opened 2017-10-30
Reading alt art from filesystem, limit number of cards in deckbuilder

jwarwick:
Instead of listing all the alt art cards in the `promo.json` file, only store the name and suffix of the alt art versions in the file and search the filesystem for all card art that contains that suffix.


jwarwick:
Saw this as a comment in Jinteki chat about Astroscript Pilot Program (retro decks are all the rage). Trying to use the Add button in the deck builder with an illegal number of card copies as the quantity creates an 1x Unknown card in the deck list.

Now checking for :limited field (or 3 if not specified) and constraining quantity to that limit.

PR #2924, opened 2017-11-02
Stats refactoring

jwarwick:
Took a look at how the user and deck statistics were working to see if I could figure out what was causing issues for people.

Didn't see anything obvious, but I did refactor some of the code to make it easier (for me) to follow.

Should be functionally equivalent. Only changed some enforcement on two players and game started before capturing stats. Might catch some of the greater than 100% completion rate cases.

### DIFF
--- a/server.coffee
+++ b/server.coffee
@@ -123,6 +123,83 @@ user_allowed_in_game = (username, game) ->
   not_allowed = username in users_blocked_from_game(game)
   not not_allowed
 
+##
+## User/Deck Stats
+##
+
+save_stats = (option, room) ->
+  (option is "always") or ((option is "competitive") and (room is "competitive"))
+
+inc_deck_start = (deck) ->
+  if deck
+    db.collection('decks').update {_id: mongoskin.helper.toObjectID(deck._id)},
+      {$inc: {"stats.games-started" : 1}}, (err) ->
+        throw err if err
+
+inc_deck_complete = (deck) ->
+  if deck
+    db.collection('decks').update {_id: mongoskin.helper.toObjectID(deck)},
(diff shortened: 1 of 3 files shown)


danhut:
Love your work - nice clean up of my hacked coffeescript which I know nothing about.   Hopefully you found the cause for the glitch as I don't know where else it could be.  

There are a couple of potential areas for issues but these would result in sub-100%

1. When you start a game the counter is incremented, but since you have not yet completed you will be under 100% for any user loading your stats during that window.  I thought about moving the start increment to the end of the game but think there are some abuse angles there.
2. If you start a game, and then don't complete it due to a server restart

jwarwick:
I'm sure I add a whole net set of bugs, but at least they're different bugs 😄

I like the idea of not adding the `started` counter until the game finishes. Maybe we can cache the starting players and do all the updates when the game finishes. I assume the abuse angle would be leaving early and not having it captured.

There may also be a UI updating inconsistency. We update `app-state` outside of the React/Om functions, so maybe there are times when it doesn't update correctly until a refresh happens. I couldn't make it happen in testing though.

danhut:
The idea was to increment the started counter at the end yes.  On pondering i think it is quite straightforward - just increment it for the players contained in the originalPlayers object since the 'players' object is empty at that point.

danhut:
Why have you removed the games commenced vs. completed from deck stats?  I did not have it originally but added it to further deter people dropping from games.

jwarwick:
Utter laziness when I was moving all the accounting to the end of the game. I don't have anything against put the black stain of a rage quitter more places. But since only they can see it, I wasn't convinced it added much extra.

danhut:
You might be right... I was thinking about people taking screenshots.  I say push yourself ;)

PR #2932, opened 2017-11-10
Spoiled card : eXer

PesiTheWizard:
Spoiled in the fan for Council of the Crest

DNM until you-know-when.


danhut:
Nice.  You will need a message to accompany the trashing so the user knows where their card went on the Purge.

PesiTheWizard:
None of the other purge-trash viruses do that.

nealterrell:
👍

PR #2936, opened 2017-11-12
Option to block new games on server announcement

jwarwick:
Added a checkbox to the `admin/announce` page that blocks (on the client side) new games from being started. Added the ability to unblock games as well (if currently blocked).

If games are blocked, an announcement message is displayed at the top of the chat page.


nealterrell:
I'm going to skip merging this and use it as a reference for the Clojure server.

PR #2939, opened 2017-11-15
Custom game board background

jwarwick:
Added the ability to specify a URL to use as your game background. 
Made current background options display as images rather than radio buttons in the Setting page.

Anyone currently using `.lobby-bg` overwriting to specify a custom background in their browser will need to reference `.gameboard-bg` now. Or just use the built-in feature.


nealterrell:
We'll need @mtgred to chime in on this one.

jwarwick:
I would certainly appreciate more eyes on the security aspect as well. Allowing users to specify URLs could lead to bad things. I think I've encoded all the places that we might have an issue, but you guys are a lot more devious than I am.

jwarwick:
This is built on the old coffee script server. Closing for now.

Issue #587, opened 2015-08-23
Street Peddler can duplicate cards


queueseven:
Click on an installed Street Peddler to install a card, but don't click on a card in the prompt yet. Trash Street Peddler manually (by dragging it into your discard).Then select a card from the prompt. That card gets installed, but it's also in your discard.

queueseven:
Duplicate to #290. Same problem I think

queueseven:
Well, since this issue is still open it seems like I have to fix it myself if the issue is still relevant... :-p

queueseven:
Checked this. Issue is still relevant.

Saintis:
Could this be solved by reloading the card before installing so that it will be moved out from the heap?

PR #2952, opened 2017-11-25
Dismiss prompts associated with a trashed card

Snark1994:
This closes issue #587; when a card is manually trashed while it has prompts open, automatically close all prompts associated with the card (more precisely, any card of the same name). This stops an exception being thrown when the user tries to interact with the prompt after trashing the card.


nealterrell:
Well done!

nealterrell:
This has one problem: cards like Spy Camera need their prompt to stay open when they are trashed. I'll revert this and then we can try again.

Snark1994:
Hm yes I see. What would a better solution be? Picking up specifically on the trigger for the runner dragging a card to their heap? Or add error handling to dialogues to fail nicely in case there's a null pointer while enacting the dialogue choice?

PR #2954, opened 2017-11-29
[WIP] NetrunnerDB Integration

### LINKED ISSUE #1874: Use deck from netrunnerDB without copying
It would be nice to have the choise to paste a netrunnerDB link deck in the "deck selection" window.

like:
https://netrunnerdb.com/en/decklist/36976/gramble-scrub-1st-2nd-na-champs-18-0-

I think is unusefull to had to copy and paste a deck (maybe the deck of the week) to try it, maybe only for a few game

jwarwick:
Push and pull a user's decks from NetrunnerDB. Uses OAuth2 for NRDB credentials.

Probably best to push this to the dev server for more testing.

To use, you need to define a number of environment variables (referenced from `config.coffee`):
- `NRDB_CLIENT_ID`
- `NRDB_SECRET`
- `NRDB_CALLBACK_URL`
- `NRDB_ENCRYPTION_KEY`

The ClientID, Secret, and CallbackURL need to be coordinated with the NetrunnerDB maintainer (@Alsciende) for each instance of the server (ie., different credentials for the dev server and the production server).

The EncryptionKey is a 32 byte random string used to encrypt the `access` and `refresh` tokens used by OAuth2 in Mongo.

Fixes #1874 (partially)


jwarwick:
Will rework this for the Clojure server.

PR #2973, opened 2017-12-30
Sovereign Sight cards -- untested

nealterrell:
Could use some more eyes on these cards. Others are still up for grabs.
 
Implementation notes:

* By Any Means: yes, this is supposed to stop ambushes from firing.
* Zamba: the credit gain is technically optional, but I have made it automatic for convenience.
* Puffer: runner must manually adjust memory units.
* Cyberdelia: we do not track how many subroutines you break, so the credit gain is triggered manually by clicking the card.
* Assimilator: resistance is futile.

* Asa Group: click Done if you don't want to install something.
* Reconstruction Contract: another case of "technically it's optional, but why would you turn it down, so now it's automatic".

### DIFF
--- a/src/clj/game/cards/icebreakers.clj
+++ b/src/clj/game/cards/icebreakers.clj
@@ -129,7 +129,8 @@
                                  :msg (msg "swap in " (:title target) " from their Grip")
                                  :effect (req (if-let [hostcard (:host card)]
                                                 (let [hosted (host state side (get-card state hostcard) target)]
-                                                  (card-init state side hosted false))
+                                                  (card-init state side hosted {:resolve-effect false
+                                                                                :init-data true}))
                                                 (let [devavec (get-in @state [:runner :rig :program])
                                                       devaindex (first (keep-indexed #(when (= (:cid %2) (:cid card)) %1) devavec))
                                                       newdeva (assoc target :zone (:zone card) :installed true)
@@ -137,7 +138,8 @@
                                                   (lose state :runner :memory (:memoryunits card))
                                                   (swap! state assoc-in [:runner :rig :program] newvec)
                                                   (swap! state update-in [:runner :hand] (fn [coll] (remove-once #(not= (:cid %) (:cid target)) coll)))
-                                                  (card-init state side newdeva false)))
+                                                  (card-init state side newdeva {:resolve-effect false
+                                                                                 :init-data true})))
                                               (move state side card :hand))}]}))
 
 (defn- conspiracy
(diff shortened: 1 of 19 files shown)


src/clj/game/cards/icebreakers.clj
```diff
@@ -767,6 +773,21 @@
                     {:abilities [(break-sub 1 1 "Sentry")
                                  (strength-pump 2 1 :all-run)]})
 
+   "Puffer"
```
Saintis: Could maybe add functions that increase / decrease MU usage of Runner to help with this? And on trash reduce MU-usage by number of power counters?

src/clj/game/cards/icebreakers.clj
```diff
@@ -884,6 +905,48 @@
...
+                                                            :default (req 1)}
+                                                  :effect (req (let [cost target]
+                                                                 (resolve-ability
+                                                                   state side
+                                                                   {:counter-cost [:virus cost]
+                                                                    :effect (effect (system-msg (str "spends " cost (pluralize " counter" cost) " from " (:title selected-virus)
+                                                                                                     " to break " cost (pluralize " barrier subroutine" cost) " with Yusuf")))}
+                                                                   selected-virus nil)))}
+                                                 yusuf nil)))}]}
+
```
Saintis: My DRY sensense tingle on this card. Could abstract out both functions since they only differ on `(effect ...)` and `:label`.

src/clj/game/cards/identities.clj
```diff
@@ -117,6 +117,24 @@
                             :msg "draw a card"
                             :effect (effect (draw 1))}}}
 
+   "Asa Group: Security Through Vigilance"
+   {:events {:corp-install
+             {:delayed-completion true
+              :req (req (and (first-event? state :corp :corp-install)
+                             ;; doesn't fire when installing ice on a central server
```
Saintis: Should still fire on ICE install on central since you can install a _second_ ice protecting that server right?

src/clj/game/cards/resources.clj
```diff
@@ -139,6 +139,19 @@
                                  (shuffle! :deck)
                                  (runner-install target))}]}
 
+   "Assimilator"
```
Saintis: Might be worth adding code to handle turning face-down Console and Events. This might crash / cause major issues on turning over (`card-init`) Events..?

src/clj/game/cards/icebreakers.clj
nealterrell: I tried, but because the inner effects need to be bound to variables from the outer scope, I couldn't find a way to make it happen. Help me, DRY Master!

src/clj/game/cards/identities.clj
nealterrell: OK, so the inner targeting function needs to disallow selecting an upgrade if the initial target was an ice on a central server.

src/clj/game/cards/icebreakers.clj
Saintis: Merge it like this and I'll find some time to dry it up.

src/clj/game/cards/resources.clj
nealterrell: Right, forgot about those. Knew I could count on you 😄 .

src/clj/game/cards/resources.clj
```diff
@@ -146,11 +146,15 @@
                  :choices {:req #(and (facedown? %)
                                       (installed? %)
                                       (= "Runner" (:side %)))}
-                 :effect (req (let [temp (move state side target :play-area)
-                                    moved (move state side temp (type->rig-zone (:type temp)))]
-                                (card-init state side (make-eid state) moved {:resolve-effect false
-                                                                              :init-data false})))
-                 :msg (msg (str "turn" (:title target) "faceup"))}]}
+                 :effect (req (if (or (is-type? target "Event")
+                                      (has-subtype? target "Console"))
```
Saintis: Consoles are only trashed if another Console is installed
Almost correct 😜

src/clj/game/cards/resources.clj
```diff
@@ -862,6 +866,32 @@
    "Laguna Velasco District"
    {:events {:runner-click-draw {:msg "draw 1 card" :effect (effect (draw))}}}
 
+   "Lewi Guilherme"
+   (let [ability {:once :per-turn
+                  :delayed-completion true
+                  :effect (req (if (zero? (:credit runner))
```
Saintis: If runner has credits on Net Mercur they could still lose a credit I believe, even if their credit count is 0.

src/clj/game/cards/resources.clj
```diff
@@ -862,6 +866,32 @@
...
+                                               :no-ability {:effect (effect (trash card)
+                                                                            (system-msg "trashed Lewi Guilherme"))}}}
+                                   card nil)))}]
+   {:flags {:drip-economy true ;; for Drug Dealer
+            :runner-phase-12 (req (< 1 (count (filter #(card-flag? % :drip-economy true)
+                                                      (all-installed state :runner)))))}
+
+    ;; KNOWN ISSUE: :effect is not fired when Assimilator turns cards over.
+    :effect (effect (lose :corp :hand-size-modification 1))
+    :leave-play (effect (gain :corp :hand-size-modification 1))
```
Saintis: Could maybe extend `:in-play` functionality to deal with modifications applied to the other side, which might play better with Assimilator.

src/clj/game/cards/resources.clj
nealterrell: Bah!

src/clj/game/cards/resources.clj
nealterrell: That's what I get for trying to be cute.

Issue #2945, opened 2017-11-18
Information Sifting + The Turning Wheel


MGillett:
If you make a successful Information Sifting run with The Turning Wheel out, The Turning Wheel gets a power counter when the run is successful - i.e. before the accesses from Information Sifting happen.
Example:

tf34 spends  to use Same Old Thing to play Information Sifting.
tf34 spends 1  to play Information Sifting.
...
tf34 continues the run.
Crunchums has no further action.
tf34 places a power counter on The Turning Wheel.
tf34 accesses Architect from HQ.
tf34 accesses Scarcity of Resources from HQ.
tf34 accesses Vanilla from HQ.
tf34 accesses Bryan Stinson from HQ.
tf34 accesses Biotic Labor from HQ.
tf34 accesses Cyberdex Virus Suite from HQ.
...

danhut:
Does this happen if an Agenda is stolen during sifting?

MGillett:
Yes.

PR #2974, opened 2017-12-30
Fix delayed completion of Information Sifting.

nealterrell:
This card was marking the run as ended before being fully resolved, giving credits for Dirty Laundry while going through one of the piles to access.


MGillett:
I bet this will also fix https://github.com/mtgred/netrunner/issues/2945

nealterrell:
I bet you're right!

PR #2982, opened 2018-01-08
Reconstruction Contract should only get an advancement token for meat damage

### LINKED ISSUE #2981: Reconstruction contract triggers on any type of damage
When the runner suffers brain or net damage with reconstruction contract rezzed, an advancement counter is automatically added to reconstruction contract, even though it should only trigger on meat damage.

jwarwick:
Fixes #2981


nealterrell:
👍

PR #2987, opened 2018-01-11
Block Asa Group from installing operations in servers

### LINKED ISSUE #2979: Asa ability allows you to install operations
If you click an op after installing a card as asa it will be installed in the server upgrade-style.

jwarwick:
Fixes #2979


nealterrell:
:+1:

PR #2991, opened 2018-01-14
Add a test for Officer Frank

butzopower:
Hello,
Big fan of Netrunner and Jinteki.net, been playing for quite a while.  Was interested in contributing and saw that there's maybe a need to backfill some tests.  Since I haven't done much clojure in the past, I figured I'd have a crack at testing a simpler card and specifically one I like to play.  So, here's a test for Officer Frank.

### DIFF
--- a/src/clj/test/cards/resources.clj
+++ b/src/clj/test/cards/resources.clj
@@ -1102,7 +1102,7 @@
                 (default-runner [(qty "Officer Frank" 1) (qty "Skulljack" 1) (qty "Respirocytes" 4)]))
    (play-from-hand state :corp "Swordsman" "Archives")
    (take-credits state :corp)
-   (core/move state :runner (find-card "Respirocytes" (:deck (get-runner))) :hand)
+   (starting-hand state :runner ["Officer Frank" "Skulljack" "Respirocytes" "Respirocytes" "Respirocytes" "Respirocytes"])
    (play-from-hand state :runner "Officer Frank")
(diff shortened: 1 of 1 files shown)


src/clj/test/cards/resources.clj
```diff
@@ -1095,6 +1095,34 @@
...
 
+(deftest officer-frank
+  ;; Officer Frank - meat damage to trash 2 from HQ
+  (do-game
+      (new-game (default-corp [(qty "Swordsman" 1) (qty "Hedge Fund" 2)])
+                (default-runner [(qty "Officer Frank" 1) (qty "Skulljack" 1) (qty "Respirocytes" 4)]))
+   (play-from-hand state :corp "Swordsman" "Archives")
+   (take-credits state :corp)
+   (core/move state :runner (find-card "Respirocytes" (:deck (get-runner))) :hand)
+   (play-from-hand state :runner "Officer Frank")
```
nealterrell: Unless I missed something, this line will fail on occasion when the random starting hand does not contain Officer Frank, since your deck has 6 cards. You should either adjust the deck to only have 5 cards, or use the `(starting-hand)` utility to forcefully set the runner's starting hand to include the cards you need.

butzopower:
Cool, that makes sense.  I think I was originally setting my clicks to 5 and then drawing up, but I think using `starting-hand` makes it more obvious.  Let me know if you want me to squash these commits.

nealterrell:
Nice job! 👍

PR #2993, opened 2018-01-15
Add a test for Zamba

butzopower:
Backfilled test for Zamba.  Seems like the credit gained could be made optional, but maybe it would slow down the pace of the game?


nealterrell:
Until there's an important automated interaction that is ruined by the automatic credit gain, I'm going to keep the gain mandatory to save on UI clicks. Thanks for the test! 👍

PR #2994, opened 2018-01-16
Fix Enhanced Login Protocol trash and reinstall from New Angeles Sol

### LINKED ISSUE #2887: Enhanced Login Protocol and Runner able to make click-less runs
Just played a game against New Angles Sol, where they played an Enhanced Login Protocol, after which I played Hacktivist Meeting. I noticed, after ELP had been trashed, I now did not automatically spend a click to make a run. Once ELP was played again, the run functionality worked normally. But every time ELP was gone, I had clickless runs again. RUNNERS OP!!! Just thought I should bring this to your attention.

screen shot 2017-10-17 at 3 55 49 pm

jwarwick:
Fixes #2887 

Added a test case and verifies that this fixes the bug (and the spelling error caused by the bug). 

But, I don't understand why the original approach didn't work. It appears that the `remove-effect` function was called and `update!` removed the property, but when the card lands in `discard` the `elp-activated` property is back.

### DIFF
--- a/src/clj/game/cards/operations.clj
+++ b/src/clj/game/cards/operations.clj
@@ -1163,24 +1163,26 @@
                    (move state :runner c :hand)))}
 
    "Service Outage"
-   (let [add-effect (fn [state side card]
-                      (update! state side (assoc card :so-activated true))
-                      (run-cost-bonus state side [:credit 1]))
-         remove-effect (fn [state side card]
-                         (run-cost-bonus state side [:credit -1])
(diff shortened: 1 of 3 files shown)


danhut:
It might be an earlier fix where a card is reset when it hits discard? 

From the move function
`(when (#{:discard :hand} to) (reset-card state side moved-card))`

jwarwick:
`Service Outage` had the same card state problem, fixed in the same manner.

nealterrell:
:+1:

Issue #1486, opened 2016-04-29
Feature Request: Psi game "Random" button...


queueseven:
...to randomly spend 0,1 or 2 $

zaroth:
I find such a button a bit restricting, since not very often you want to put equal odds 33% on every chance. What I think would be more flexible is a local  (just JS, no server communication) die roll, with being able to select the number of die sides (e.g. d3, d6 and d10). This way players could mimic real life random decision making.

queueseven:
Related: #1135

queueseven:
Related: #1195

lostie:
Would integrating something like this in the client do the job: https://github.com/noviny/react-dice do the job ?

There's another nice one http://vicentemundim.github.io/dice-roller/examples.html , but this one seems to use angular, which is not using in this app.

queueseven:
A little overkill to use a library for generating a random number... ;-)

lostie:
Well, the idea was more to have an actually dice rolling on the screen, but if it's ok to just generate a random number, yeah, no library needed

queueseven:
Well, if you want to take the time to implement this, feel free to make a pull request.
I just don't know if such a dice animation would fit into the sites appearance, but if you could make it work... ;-)

PR #3005, opened 2018-01-18
Rewrote psi-game to give the user a 'roll d6' choice.

presheaf:
'roll d6' generates a random number and shows it to the user.

Below is an example of the new psi game prompt:
random_psi1

After pressing the 'roll a d6' button,the user is presented with a new prompt where the dice result is shown:
random_psi2

The game state is unaffected by the 'roll d6' button. Its only intended purpose is saving the user a visit to random.org. Implements #1486.

### DIFF
--- a/src/clj/game/core/abilities.clj
+++ b/src/clj/game/core/abilities.clj
@@ -426,6 +426,29 @@
     (swap! state update-in [side :prompt] (fn [pr] (filter #(not= % wait) pr)))))
 
 ;;; Psi games
+
+(defn show-prompt-with-dice
+  "Calls show-prompt normally, but appends a 'roll d6' button to choices.
+  If user chooses to roll d6, reveal the result to user and re-display
+  the prompt without the 'roll d6 button'."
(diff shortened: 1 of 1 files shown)


src/clj/game/core/abilities.clj
```diff
@@ -426,6 +426,29 @@
...
+  ([state side card msg other-choices f]
+   (show-prompt state side card msg other-choices f nil))
+  ([state side card msg other-choices f args]
+   (let [dice-msg "Roll a d6",
+         choices (conj other-choices dice-msg)]
+     (show-prompt state side card msg choices
+                  #(if (not= % dice-msg)
+                     (f %)
+                     (show-prompt state side card
+                                  (str msg
```
nealterrell: If you could, please clean up the style here a little. `str` is a clear enough function that you don't need to put each argument on its own line. Prefer `inc` to `(+ ... 1)`. Collapse a trailing `)` onto the end of the previous line.

PR #3006, opened 2018-01-19
Implemented Falsified Credentials with two tests

presheaf:
Card is fully automated, with user being prompted first for a card type, and then a unrezzed remote server card to expose.

Unit tests verify that guessing correctly gives creds, guessing incorrectly does not, and that guessing rezzed cards does not work. It also verifies that expose prevention prevents cred gain even on a correct guess.

Implementation and test both borrow heavily from Drive By, as the cards are fairly similar. I'm not sure if I used the event system correctly, so any feedback is much appreciated.

### DIFF
--- a/src/clj/game/cards/events.clj
+++ b/src/clj/game/cards/events.clj
@@ -522,6 +522,32 @@
     :msg "look at the top 4 cards of their Stack and add 1 of them to their Grip"
     :effect (effect (move target :hand) (shuffle! :deck))}
 
+   "Falsified Credentials"
+   {:prompt "Choose a type"
+    :choices ["Agenda" "Asset" "Upgrade"]
+    :effect (effect
+             (resolve-ability
+              (let [chosen-type target]
+                {:choices {:req #(let [topmost (get-nested-host %)]
+                                   (and (is-remote? (second (:zone topmost)))
+                                        (= (last (:zone topmost)) :content)
+                                        (not (:rezzed %))))}
+                 :delayed-completion true
+                 :effect (req             ;taken from Drive By - maybe refactor
+                          (when-completed (expose state side target) 
+                            (if async-result ;; expose was successful
+                              (if (= chosen-type (:type target))
+                                (do (resolve-ability
+                                     state :runner
+                                     {:effect (effect (gain :credit 5))
+                                      :msg "gain 5 [Credits] "}
+                                     card nil)
+                                    (effect-completed state side eid))
+                                (effect-completed state side eid))
+                              (effect-completed state side eid))))})
+              card nil))}
+
+   
    "Fear the Masses"
    {:req (req hq-runnable)
     :effect (effect (run :hq {:req (req (= target :hq))
--- a/src/clj/test/cards/events.clj
+++ b/src/clj/test/cards/events.clj
@@ -688,6 +688,31 @@
       (is (= 17 (:credit (get-runner)))
           "Gained 5 credits for guessing agenda correctly, even with rezzed card in server"))))
 
+(deftest falsified-credentials-zaibatsu-loyalty
+  ;; If Falsified Credentials fails to expose, it grants no credits.
+  (do-game
+   (new-game (default-corp [(qty "Zaibatsu Loyalty" 1)
+                            (qty "Project Atlas" 1)])
+             (default-runner [(qty "Falsified Credentials" 2)]))
+   
+    (play-from-hand state :corp "Project Atlas" "New remote")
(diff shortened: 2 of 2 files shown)


src/clj/game/cards/events.clj
```diff
@@ -522,6 +522,32 @@
     :msg "look at the top 4 cards of their Stack and add 1 of them to their Grip"
     :effect (effect (move target :hand) (shuffle! :deck))}
 
+   "Falsified Credentials"
+   {:prompt "Choose a type"
+    :choices ["Agenda" "Asset" "Upgrade"]
```
nealterrell: So we don't currently use delayed completion with the effects of runner cards (only their `:abilities` handlers), because there are no Runner equivalents to AccDiagnostics / Subcontract that exect two card effects in a row and need to wait for the first to finish before the next goes. However, it's not a bad idea to start future proofing runner Events with delayed completion when appropriate (it's also good practice for learning when the system needs to be used). Because this card involves a delay-able effect (expose, which triggers optional abilities like prevention), its root ability must be marked `:delayed-completion true`.

src/clj/game/cards/events.clj
```diff
@@ -522,6 +522,32 @@
...
+    :effect (effect
+             (resolve-ability
```
nealterrell: `resolve-ability` creates a new `eid` for the effect it resolves, and that `eid` will be marked completed when the chained ability completes, which is great... but we also need the `eid` of the original parent ability to complete, which means it needs to be bound to a name and then passed to `effect-completed` by hand. That would be a huge pain in the ass, so instead we have `continue-ability`, which is a drop-in replacement for `resolve-ability`, except it gives the current eid's ability to the new ability, so that the completion of the new ability also marks the completion of the original ability.

src/clj/game/cards/events.clj
```diff
@@ -522,6 +522,32 @@
...
+              (let [chosen-type target]
+                {:choices {:req #(let [topmost (get-nested-host %)]
+                                   (and (is-remote? (second (:zone topmost)))
+                                        (= (last (:zone topmost)) :content)
+                                        (not (:rezzed %))))}
+                 :delayed-completion true
+                 :effect (req             ;taken from Drive By - maybe refactor
+                          (when-completed (expose state side target) 
+                            (if async-result ;; expose was successful
+                              (if (= chosen-type (:type target))
```
nealterrell: Let's merge these two into an `and`.

src/clj/game/cards/events.clj
```diff
@@ -522,6 +522,32 @@
...
+                                (do (resolve-ability
```
nealterrell: This `resolve-ability` and following `effect-completed` can be merged into a `continue-ability` that no longer needs the `do` also!

src/clj/test/cards/events.clj
```diff
@@ -644,6 +644,76 @@
...
     (is (= 3 (count (:discard (get-runner)))))))
 
+(deftest falsified-credentials
+  ;; Falsified Credentials - Expose card in remote
+  ;; server and correctly guess its type to gain 5 creds
+  (do-game
+    (new-game (default-corp [(qty "Eve Campaign" 2)
+                             (qty "Product Placement" 2)
+                             (qty "Project Atlas" 1)])
+              (default-runner [(qty "Falsified Credentials" 5)]))
```
nealterrell: I should've mentioned this on your other tests, and we'll fix those eventually, but it's probably best if we stick to deckbuilding rules as much as possible, particularly re: maximum number of copies. There's no deck validation in the test suite, but just in case...

src/clj/test/cards/events.clj
```diff
@@ -644,6 +644,76 @@
...
+    (play-from-hand state :corp "Project Atlas" "New remote")
+    (play-from-hand state :corp "Product Placement" "HQ")
+    (play-from-hand state :corp "Product Placement" "Server 3")
+    (take-credits state :corp)
+    (let [eve1 (get-content state :remote1 0)
+          eve2 (get-content state :remote2 0)
+          atl (get-content state :remote3 0)
+          pp1 (get-content state :hq 0)
+          pp2 (get-content state :remote3 1)
+          ]
```
nealterrell: Style wise, move the bracket to end of previous line.

src/clj/test/cards/events.clj
presheaf: Point well taken, but I don't think I have any other tests. Unless I'm forgetting some maybe you're thinking of someone else.

nealterrell:
Great job! 👍

PR #3012, opened 2018-01-22
Disable `Draw` button if no cards in deck

jwarwick:


src/cljs/netrunner/gameboard.cljs
```diff
@@ -1260,7 +1260,7 @@
                                                    (or (pos? (:tagged opponent))
                                                        (pos? (:tag opponent))))
                              #(send-command "trash-resource")))
-              (cond-button "Draw" (pos? (:click me)) #(send-command "draw"))
+              (cond-button "Draw" (and (pos? (:click me)) (pos? (count (:deck me)))) #(send-command "draw"))
```
nealterrell: As a Clojure style thing, you can use either `(seq (:deck me))` or `(not-empty (:deck me))` instead of `(pos? (count ...))`. Clojurians prefer `seq` (returns nil if given an empty collection), but I think `not-empty` is a more clear name.

jwarwick: Thanks @nealterrell. I appreciate the feedback.

PR #3014, opened 2018-01-23
Fix #3011 : Allow multiple Respirocytes effects to be active

### LINKED ISSUE #3011: Respirocytes
Hey folks, just wanted to report that Respirocytes aren't working properly.
If you have multiples installed, you get a new card for each copy of Respirocytes installed when you hit 0 cards. Currently, it only gives you 1 even if you have multiples installed. 
Not complaining, just a heads up on an irrelevant rarely used card :P

butzopower:
Fix for #3011.

I'm still super junior on Clojure, so I wasn't sure if I'm using `keyword` correctly here.  Essentially I'm trying to have multiple watches for each Respirocytes that is in play and associate the watch key with the card itself so that it can be unwatched when the card trashes or leaves play.

### DIFF
--- a/src/clj/game/cards/hardware.clj
+++ b/src/clj/game/cards/hardware.clj
@@ -784,15 +784,17 @@
                                (when (= (get-in (get-card state card) [:counter :power]) 3)
                                  (system-msg state :runner "trashes Respirocytes as it reached 3 power counters")
                                  (trash state side card {:unpreventable true})))}]
-   {:effect (req (add-watch state :respirocytes
+   {:effect (req (let [watch-id (keyword "respirocytes" (str (:cid card)))]
+                   (update! state side (assoc card :respirocytes-watch-id watch-id))
+                   (add-watch state watch-id
                             (fn [k ref old new]
                               (when (and (seq (get-in old [:runner :hand]))
                                          (empty? (get-in new [:runner :hand])))
-                                (resolve-ability ref side ability card nil))))
+                                (resolve-ability ref side ability card nil)))))
                  (damage state side eid :meat 1 {:unboostable true :card card}))
     :msg "suffer 1 meat damage"
-    :trash-effect {:effect (req (remove-watch state :respirocytes))}
-    :leave-play (req (remove-watch state :respirocytes))
+    :trash-effect {:effect (req (remove-watch state (:respirocytes-watch-id card)))}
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/hardware.clj
```diff
@@ -784,15 +784,17 @@
                                (when (= (get-in (get-card state card) [:counter :power]) 3)
                                  (system-msg state :runner "trashes Respirocytes as it reached 3 power counters")
                                  (trash state side card {:unpreventable true})))}]
-   {:effect (req (add-watch state :respirocytes
+   {:effect (req (let [watchId (keyword "respirocytes" (str (:cid (get-card state card))))]
```
nealterrell: Just for consistency, we use `-` to separate words in identifier names, rather than camelCasing.  Also, you *probably* don't need to call `get-card` here... `get-card` works by finding a card in the state that matches the given card's `:cid`, so the returned map will have the same `:cid` as the `card` variable you already have access to.

src/clj/test/cards/hardware.clj
```diff
@@ -680,6 +680,18 @@
       (prompt-choice :runner "Yes")  ; 6 installed
       (is (count-spy 6) "6 Spy Cameras installed"))))
 
+(deftest respirocytes-multiple
+  ;; Should draw multiple cards when multiple respirocytes are in play
+  (do-game
+   (new-game (default-corp)
+             (default-runner [(qty "Respirocytes" 9)]))
```
nealterrell: I'd prefer for the test decks to follow deckbuilding rules as much as possible, particularly regarding max copies of cards. You can use the `(starting-hand)` utility to ensure there are 3 copies of Respirocytes in the opening hand.

Saintis: I think the problem is if the first or second damage hits the third respirocyte. Solution would be to just test for installing 2 I think (since then it's only possible you lose one.

butzopower: Makes sense.  I've switch it to 6 cards and doing 2 Respirocytes, since it's effectively testing the same thing.

src/clj/game/cards/hardware.clj
butzopower: Awesome, thanks for the feedback.  Incorporated.

nealterrell:
👍

PR #3016, opened 2018-01-24
Jinteki Biotech text

jwarwick:
Cleaned up text when selecting `Jinteki Biotech: Life Imagined`identities. 
Added tests for `The Tank` and `The Greenhouse`.


nealterrell:
What's the net effect here?

jwarwick:
The identity names were in the form `The Brewery~brewery` when displayed to the user. I assume that's a bad search and replace leftover. Seems to have occurred when cards were split into different files. I changed them to the form `The Brewery`.

PR #3023, opened 2018-01-26
Down the White Nile: SSL Endorsement

jwarwick:


src/clj/game/cards/agendas.clj
```diff
@@ -925,6 +925,21 @@
     :msg "do 2 meat damage"
     :effect (effect (damage eid :meat 2 {:card card}))}
 
+   "SSL Endorsement"
+   (let [add-credits (fn [] (effect (add-counter card :credit 9)))
```
nealterrell: Just out of curiosity, does this not work if you directly do `[add-credits (effect ...)` without the wrapping `fn`?

jwarwick:
Good catch @nealterrell. No need to wrap the macro in a function.

Issue #3025, opened 2018-01-28
Down the White Nile card implementations


jwarwick:
Implementation status of data pack cards.
If you wish to claim a card before working on it, mark up the text below or add a comment to the issue.

- [x] Acacia (#2933)
- [x] Plague (#3028)
- [x] Credit Kiting (#3029)
- [x] Wari (#3046)
- [x] Kabonesa Wu: Netspace Thrillseeker (#3032)
- [x] Takobi (#3040)
- [x] Kongamato (#3038)
- [x] Emergent Creativity (#3047)
- [x] RNG Key (#3043)
- [x] Nightdancer (#3033)
- [x] Jinja City Grid (#3024)
- [x] Aimor (#3031)
- [x] Bacterial Programming (#3053)
- [x] Jua (#3042)
- [x] Threat Assessment (#3037)
- [x] Economic Warfare (#3019)
- [x] Forced Connection (#3036)
- [x] SSL Endorsement (#3023)
- [x] NGO Front (#3027)
- [x] Distract the Masses (#3052 )

butzopower:
@jwarwick I can have a try at NGO Front.

butzopower:
I'll try Plague next @jwarwick

butzopower:
Can do Credit Kiting too.

presheaf:
I'll start out with Kabonesa.

presheaf:
I'll start on Wari now.

PR #3027, opened 2018-01-28
Down the White Nile: NGO Front

butzopower:
Quite an easy one to implement :)

### DIFF
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -979,6 +979,19 @@
                                                    (trigger-event state side :no-trash card))))}
                                card targets))}}
 
+   "NGO Front"
+   (letfn [(builder [cost cred]
+             {
+              :advance-counter-cost cost
+              :effect (effect (trash card {:cause :ability-cost}) (gain :credit cred))
+              :label (str "[Trash]: Gain " cred " [Credits]")
+              :msg (str "gain " cred " [Credits]")
+              })]
+     {
+      :advanceable :always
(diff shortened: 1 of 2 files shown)


butzopower:
#3025

src/clj/game/cards/assets.clj
```diff
@@ -979,6 +979,17 @@
                                                    (trigger-event state side :no-trash card))))}
                                card targets))}}
 
+   "NGO Front"
+   {:advanceable :always
+    :abilities [{:advance-counter-cost 1
+                 :effect (effect (trash card) (gain :credit 5))
```
nealterrell: Add a final parameter `{:cause :ability-cost}` to the `trash` calls to indicate that we're paying an ability cost as the reason for the trash.

src/clj/game/cards/assets.clj
```diff
@@ -979,6 +979,17 @@
...
+                 :label "Gain 5 [Credits]"
```
nealterrell: Let's put `[Trash]` at the beginning of the label string, to show the trashcan icon in the menu item itself: `"[Trash]: Gain 5 [Credits]"`.

Saintis: I'd be very tempted to make both these abilities into an ability-maker function that takes advance-counter-cost and credit-gain to reduce the minor repetition.

butzopower:
Thanks for all the feedback @nealterrell and @Saintis, updated my commit.

PR #3028, opened 2018-01-28
Down the White Nile: Plague

butzopower:
Ref #3025

Mostly cribbed Temujin Contract on this one.  I thought about making sure you don't get counters if the corp destroys the server and immediately starts a new one, but I now see that servers are uniquely named no matter what servers currently exist, and so it doesn't make sense that a runner could run a destroyed server.

### DIFF
--- a/src/clj/game/cards/programs.clj
+++ b/src/clj/game/cards/programs.clj
@@ -646,6 +646,16 @@
                                    :effect (effect (runner-install target {:no-cost true}))} card nil)
                                 (trash state side card)))}]}
 
+   "Plague"
+   {:prompt "Choose a server for Plague" :choices (req servers)
+    :msg (msg "target " target)
+    :req (req (not (:server-target card)))
+    :effect (effect (update! (assoc card :server-target target)))
+    :events {:successful-run
+             {:req (req (= (zone->name (get-in @state [:run :server])) (:server-target (get-card state card))))
+              :msg "gain 2 virus counters"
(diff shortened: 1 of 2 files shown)


src/clj/test/cards/programs.clj
```diff
@@ -568,6 +568,23 @@
         (is (= 1 (count (:discard (get-corp)))) "Enigma trashed")
         (is (= 1 (count (:discard (get-runner)))) "Parasite trashed when Enigma was trashed")))))
 
+(deftest plaque
```
nealterrell: `plague` ;)

src/clj/game/cards/programs.clj
```diff
@@ -646,6 +646,16 @@
                                    :effect (effect (runner-install target {:no-cost true}))} card nil)
                                 (trash state side card)))}]}
 
+   "Plague"
+   {:prompt "Choose a server for Plague" :choices (req servers)
+    :msg (msg "target " target)
+    :req (req (not (:server-target card)))
+    :effect (effect (update! (assoc card :server-target target)))
```
nealterrell: Based on a recent discussion in Slack, let's start saving "misc data" card values inside a `:special` key of the card state. So `assoc-in ... [:special :server-target]` and appropriate changes to the lines that read that value.

butzopower:
Aha, fixed my spelling mistake and updated Plague to store its target under `:special`.

PR #3029, opened 2018-01-28
Down the White Nile: Credit Kiting

butzopower:
#3025

### DIFF
--- a/src/clj/game/cards/events.clj
+++ b/src/clj/game/cards/events.clj
@@ -242,6 +242,17 @@
                                    (effect-completed state side eid)))))}
              :run-ends {:effect (effect (unregister-events card))}}}
 
+   "Credit Kiting"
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/events.clj
```diff
@@ -242,6 +242,17 @@
...
+   "Credit Kiting"
+   {:req (req (some #{:hq :rd :archives} (:successful-run runner-reg)))
+    :prompt "Select a card to install from your Grip"
+    :choices {:req #(and (or (is-type? % "Hardware")
+                             (is-type? % "Program")
+                             (is-type? % "Resource"))
+                         (in-hand? %))}
+    :effect (effect (install-cost-bonus [:credit -8])
+                    (runner-install target)
+                    (gain :tag 1))}
```
nealterrell: I believe the tag should be preventable as it's not a cost. Instead of `gain`, you should call `tag-runner`.

PR #3032, opened 2018-01-30
Implemented Kabonesa and made move to RFG deactivate cards

presheaf:
Behavior is very similar to Test Run.

### DIFF
--- a/src/clj/game/cards/identities.clj
+++ b/src/clj/game/cards/identities.clj
@@ -528,6 +528,25 @@
                                          :effect (effect (add-prop target :advance-counter 4 {:placed true}))} card nil)))
                                 (update! state side (assoc (get-card state card) :biotech-used true))))}]}
 
+   "Kabonesa Wu: Netspace Thrillseeker"
+   {:abilities [{:label "[:click] Install a non-virus program from your stack, lowering the cost by 1 [Credit]"
+                 :cost [:click 1]
+                 :prompt "Choose a program"
+                 :choices (req (cancellable
+                                (filter #(and (is-type? % "Program")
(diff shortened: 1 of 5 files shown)


presheaf:
#3025

src/clj/game/cards/identities.clj
```diff
@@ -528,6 +528,25 @@
...
 
+   "Kabonesa Wu: Netspace Thrillseeker"
+   {:abilities [{:label "[:click] Install a non-virus program from your stack, lowering the cost by 1 [Credit]"
+                 :cost [:click 1]
+                 :prompt "Choose a program"
+                 :choices (req (cancellable
+                                (filter #(and (is-type? % "Program")
+                                              (not (has-subtype? % "Virus")))
+                                        (:deck runner))))
+                 :msg (str"install a non-virus program from their stack, lowering the cost by 1 [Credit]")
```
nealterrell: Can you put a space between `str` and `"`? It seems to throw off GitHub's syntax highlighting, even if it works in Clojure.

src/clj/game/cards/identities.clj
```diff
@@ -528,6 +528,25 @@
...
+                 :effect (effect (trigger-event :searched-stack nil)
+                                 (shuffle! :deck)
+                                 (install-cost-bonus [:credit -1])
+                                 (runner-install (assoc-in target [:special :kabonesa] true) {:no-cost false}))
```
nealterrell: You can just drop the `{:no-cost}` argument at the end, it's only important when that value is true.

presheaf: Whoops. Guess which card I copied this from. :)

PR #3037, opened 2018-01-30
Down the White Nile: Threat Assessment

jwarwick:


src/clj/game/cards/operations.clj
```diff
@@ -1396,6 +1396,28 @@
...
+                   (continue-ability state :runner
+                                     {:prompt (str "Add " (:title chosen) " to the top of the Stack or take 2 tags?")
+                                      :choices [(str "Move " (:title chosen))
+                                                "2 tags"]
+                                      :delayed-completion true
+                                      :effect (req (clear-wait-prompt state :corp)
+                                                   (move state :corp (last (:discard corp)) :rfg)
+                                                   (if (.startsWith target "Move")
+                                                     (do (system-msg state side (str "chooses to move " (:title chosen) " to the Stack"))
+                                                       (move state :runner chosen :deck {:front true}))
```
nealterrell: Looks good! Just need to trigger `effect-completed` in this branch.

jwarwick:
Thanks @nealterrell. Checking my understanding - I need to call `effect_completed` because I marked the effect as `delayed_completion`. I don't need to call it for the `tag_runner` code because it calls it internally somewhere (I think in `resolve_tags`), but I do need to call it for `move`.

Is there a general rule for knowing which actions take care of it themselves versus when I need to handle it in the effect?

nealterrell:
Because the ability is delayed-completion, *someone* is responsible for indicating when the ability is finished. An ability must be marked delayed-completion if it triggers another event, core function, or ability that potentially requires user interaction to resolve (a "delayable action"). Because `tag-runner` is a delayable action (it opens a prevention window), your innermost-nested Threat Assessment effect must be delayed completion, and any of the parent effects that cascade into the nested effect must also be delayed completion.

So every path through the innermost effect must eventually result in `effect-completed` being triggered. The branch with `move` needs to call `effect-completed` manually, because `move` is not a delayable action -- once the `move` function returns to the call site, we know that everything involved with `move` is done, and the card's ability is now finally terminated. But if we flow into the `tag-runner` branch, the card ability's effect is *not* terminated once `tag-runner` returns to the call site, because `tag-runner` is delayable; we must await the end of `tag-runner` before indicating that Threat Assessment is finished. The long way to do this is as such:

```clojure
(if (.startsWith target "Move")
  (do ...)
  (do (system-msg state side "chooses to take 2 tags")
    (when-completed (tag-runner state :runner 2)
```

`tag-runner` is given a new `eid` automatically, and when that `eid` is triggered as complete (after prevention effects), then we resume and mark Threat Assessment as complete. (the symbol `eid` in `(effect-completed ... eid)` refers to the ID of Threat Assessment, not that of the `tag-runner` call. 

This pattern happens frequently enough (my ability is over exactly when this delayable action declares that it is over) that there's a shortcut: passing the `eid` from Threat Assessment to `tag-runner` as the third argument. Any delayable action has this ability: if given an eid directly, the action will use that eid as its own; otherwise a new one will be generated for it.  You can see this in the function definition of any delayable action: typically there is one (final) overload that explicitly takes an `eid` as the third parameter, and the other overloads generate an `eid` (with `(make-eid state)`) and pass it to the final overload.

This is how you can recognize any function that is delayable: there will be an overload that takes an `eid` parameter.

PR #3038, opened 2018-01-31
Down the White Nile: Kongamato

jwarwick:


src/clj/game/cards/resources.clj
```diff
@@ -864,6 +864,12 @@
       :msg "gain 2 [Credits]"
       :effect (effect (gain :credit 2))}}}
 
+   "Kongamato"
+   {:abilities [{:label "[Trash]: Break the first subroutine on the encountered piece of ice"
+                 :req (req (and (:run @state) (rezzed? current-ice)))
+                 :effect (effect (trash card {:cause :ability-cost})
+                                 (system-msg :corp "trashes Kongamato to break the first subroutine on the encountered piece of ice"))}]}
```
nealterrell: Let's use the title of the current ice, instead of "on the encountered piece of ice".

PR #3040, opened 2018-01-31
Down the White Nile: Takobi

jwarwick:


src/clj/game/cards/programs.clj
```diff
@@ -826,6 +826,23 @@
                    :label "Swap the Barrier ICE currently being encountered with a piece of ICE directly before or after it"
                    :effect (effect (resolve-ability (surf state current-ice) card nil))}]})
 
+   "Takobi"
+   {:implementation "Adding power counter is manual"
+    :abilities [{:label "Add power counter after breaking all subroutines on a piece of ice"
```
nealterrell: We typically don't spell out requirements for abilities like this (see Quicksand as an example), so for brevity I think "Add 1 power counter" is sufficient.

PR #3042, opened 2018-02-01
Down the White Nile: Jua

jwarwick:


src/clj/game/cards/ice.clj
```diff
@@ -994,6 +994,38 @@
...
+                   :choices {:req #(and (= (:side %) "Runner") (installed? %)) :max 2 :all true}
+                   :msg (msg "add either " (card-str state (first targets)) " or " (card-str state (second targets)) " to the Stack")
+                   :effect (req (when (= (count targets) 2)
+                                     (show-wait-prompt state :corp "Runner to decide which card to move")
+                                     (continue-ability
+                                       state
+                                       :runner
+                                        {:player :runner
+                                         :priority 1
+                                         :delayed_completion true
```
JoelCFC25: should be `:delayed-completion`

jwarwick: Good catch!

src/clj/game/cards/ice.clj
```diff
@@ -994,6 +994,38 @@
...
+                   :choices {:req #(and (= (:side %) "Runner") (installed? %)) :max 2 :all true}
+                   :msg (msg "add either " (card-str state (first targets)) " or " (card-str state (second targets)) " to the Stack")
+                   :effect (req (when (= (count targets) 2)
+                                     (show-wait-prompt state :corp "Runner to decide which card to move")
+                                     (continue-ability
+                                       state
+                                       :runner
+                                        {:player :runner
+                                         :priority 1
+                                         :delayed-completion true
```
nealterrell: I don't think this final ability needs to be delayed completion, as it doesn't call or cause a delayable action from this point on. You then wouldn't have to manually trigger `effect-completed`.

PR #3043, opened 2018-02-01
Down the White Nile: RNG Key

jwarwick:
Need to implement test cases still.
Need someone to check my understanding of `delayed-completion`.
Need insight on how to block the `accesses and unseen card from R&D` message after I reveal it.

### DIFF
--- a/src/clj/game/cards/programs.clj
+++ b/src/clj/game/cards/programs.clj
@@ -698,9 +698,10 @@
 
    "RNG Key"
    {:events {:pre-access-card {:req (req (get-in card [:special :rng-guess]))
+                               :delayed-completion true
                                :msg (msg "to reveal " (:title target))
-                               :effect (req (when-let [guess (get-in card [:special :rng-guess])]
-                                              (when (or (= guess (:cost target))
+                               :effect (req (if-let [guess (get-in card [:special :rng-guess])]
+                                              (if (or (= guess (:cost target))
                                                         (= guess (:advancementcost target)))
                                                 (continue-ability state side
                                                                   {:prompt "Choose RNG Key award"
@@ -713,7 +714,9 @@
                                                                                   (do (gain state :runner :credit 3)
                                                                                     (system-msg state :runner "uses RNG Key to gain 3 [Credits]"
                                                                                     (effect-completed state side eid card)))))}
-                                                                  card nil))))}
+                                                                  card nil)
+                                                (effect-completed state side eid card))
+                                              (effect-completed state side eid card)))}
              :post-access-card {:effect (effect (update! (assoc-in card [:special :rng-guess] nil)))}
              :successful-run {:req (req (let [first-hq (first-successful-run-on-server? state :hq)
                                               first-rd (first-successful-run-on-server? state :rd)]
--- a/src/clj/game/cards/programs.clj
+++ b/src/clj/game/cards/programs.clj
@@ -716,15 +716,12 @@
                                                         (= guess (:advancementcost target)))
                                                 (continue-ability state side
                                                                   {:prompt "Choose RNG Key award"
-                                                                   :delayed-completion true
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/programs.clj
```diff
@@ -696,6 +696,34 @@
                             :effect (effect (draw :runner 1))
                             :msg "draw 1 card"}}}
 
+   "RNG Key"
+   {:events {:pre-access-card {:req (req (get-in card [:special :rng-guess]))
+                               :msg (msg "to reveal " (:title target))
+                               :effect (req (when-let [guess (get-in card [:special :rng-guess])]
```
JoelCFC25: Slot in `:delayed-completion true` here

src/clj/game/cards/programs.clj
```diff
@@ -696,6 +696,34 @@
                             :effect (effect (draw :runner 1))
                             :msg "draw 1 card"}}}
 
+   "RNG Key"
+   {:events {:pre-access-card {:req (req (get-in card [:special :rng-guess]))
+                               :msg (msg "to reveal " (:title target))
```
JoelCFC25: Slot in `:delayed-completion true` here

src/clj/game/cards/programs.clj
```diff
@@ -696,6 +696,37 @@
...
+   "RNG Key"
+   {:events {:pre-access-card {:req (req (get-in card [:special :rng-guess]))
+                               :delayed-completion true
+                               :msg (msg "to reveal " (:title target))
+                               :effect (req (if-let [guess (get-in card [:special :rng-guess])]
+                                              (if (or (= guess (:cost target))
+                                                        (= guess (:advancementcost target)))
+                                                (continue-ability state side
+                                                                  {:prompt "Choose RNG Key award"
+                                                                   :delayed-completion true
```
nealterrell: Because neither path of this ability's effect causes a delayable action, the ability doesn't need delayed completion.

src/clj/game/cards/programs.clj
```diff
@@ -696,6 +696,37 @@
...
+                                                                  card nil)
+                                                (effect-completed state side eid card))
+                                              (effect-completed state side eid card)))}
+             :post-access-card {:effect (effect (update! (assoc-in card [:special :rng-guess] nil)))}
+             :successful-run {:req (req (let [first-hq (first-successful-run-on-server? state :hq)
+                                              first-rd (first-successful-run-on-server? state :rd)]
+                                          (and first-hq first-rd (or (= target :hq) (= target :rd)))))
+                              :optional {:prompt "Fire RNG Key?"
+                                         :yes-ability {:prompt "Guess a number"
+                                                       :choices {:number (req 20)}
```
nealterrell: Watch the next release to contain a 21-cost upgrade ;). Doesn't need a fix.

jwarwick:
Removed `delayed-completion` from the `continue-ability` (and the corresponding `effect-completed` calls. Does this all mean I should remove `delayed-completion` from the `pre-access` event, since nothing under it is delayable?

Or am I think about it wrong and need to add `effect-completed` back into both paths in `continue-ability` to catch the `delayed-completion` from the upper level?

nealterrell:
You seem to have everything down for this card. The outer ability still needs delayed completion because it uses `continue-ability`, so it itself is a delayable action. So now we double check that all paths eventually trigger an `effect-completed`. The innermost "success" ability is not delayed-completion, so the engine will trigger `effect-completed` immediately after executing the effect. Good! In all cases where we don't end up continuing into that ability, you do correctly call `effect-completed` manually. Piece of cake!

jwarwick:
Thanks @nealterrell. That explanation makes sense again. Hopefully I can hold all of that in my head for the next card.

I think this PR is good to merge now.

PR #3046, opened 2018-02-02
Wari

presheaf:
Implemented Wari. (#3025) On the first successful HQ run each turn, if there are unrezzed ice, Wari prompts the user for whether they want to trigger Wari. If yes, Wari is trashed, the user is prompted for a subtype, and then to select a piece of ice to expose.

This got more intricate than I anticipated, so there's probably some abuse of the effect system going on here. Comments about how to simplify this are appreciated. :)

### DIFF
--- a/src/clj/game/cards/programs.clj
+++ b/src/clj/game/cards/programs.clj
@@ -880,45 +880,48 @@
                    :effect (effect (gain :click 2))}]}
 
    "Wari"
-   {:events {:successful-run
-             {:interactive (req )
+   (letfn [(prompt-for-subtype []
+             {:prompt "Choose a subtype"
+              :choices ["Barrier" "Code Gate" "Sentry"]
               :delayed-completion true
-              :req (req (and (= target :hq)
-                             (first-successful-run-on-server? state :hq)
-                             (some #(and (ice? %)
-                                         (not (rezzed? %)))
-                                   (all-installed state :corp))))
-              :effect (effect
-                       (continue-ability
-                        {:prompt "Use Wari?"
-                         :choices ["Yes" "No"]
-                         :delayed-completion true
-                         :effect (effect 
-                                  (continue-ability
+              :effect (req
+                       (when-completed (trash state side card {:unpreventable true})
+                         (continue-ability
+                          state side
+                          (expose-and-maybe-bounce target)
+                          card nil)))})
+           
+           (expose-and-maybe-bounce [chosen-subtype]
+             {:choices {:req #(and (ice? %)
+                                   (not (rezzed? %)))}
+              :delayed-completion true
+              :msg (str "name " chosen-subtype)
+              :effect (req
+                       (when-completed (expose state side target)
+                         (do (if (and async-result
+                                      (has-subtype? target chosen-subtype))
+                               (do (move state :corp target :hand)
+                                   (system-msg state :runner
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/programs.clj
```diff
@@ -877,4 +877,48 @@
                    :counter-cost [:power 3]
                    :once :per-turn
                    :msg "gain [Click][Click]"
-                   :effect (effect (gain :click 2))}]}})
+                   :effect (effect (gain :click 2))}]}
+
+   "Wari"
+   {:events {:successful-run
+             {:interactive (req )
```
nealterrell: This should be `(req true)` if you want this event to always force the Runner into using the simultaneous event resolution prompt.

src/clj/game/cards/programs.clj
```diff
@@ -877,4 +877,48 @@
...
+                                         (not (rezzed? %)))
+                                   (all-installed state :corp))))
+              :effect (effect
+                       (continue-ability
+                        {:prompt "Use Wari?"
+                         :choices ["Yes" "No"]
+                         :delayed-completion true
+                         :effect (effect 
+                                  (continue-ability
+                                   (if (= target "Yes")
```
nealterrell: This is a little messy. I think you meant for the `if` to be before the `continue-ability`, as in, if the Runner clicks Yes, then continue this ability, otherwise trigger `effect-completed`. This approach *might* have the same overall effect, but for clarity we should switch it around. You'll have to use `(req` instead of `(effect`.

src/clj/game/cards/programs.clj
```diff
@@ -877,4 +877,48 @@
...
+                                     {:prompt "Choose a subtype"
+                                      :choices ["Barrier" "Code Gate" "Sentry"]
+                                      :delayed-completion true
+                                      :effect (effect
+                                               (trash card)
```
nealterrell: This needs to be `{:unpreventable true}`, as it's a "trash __to__ do this" wording. To handle chain effects from the trashing (say... Q-Coherence Chip, causing trash-prevention from SacCon), you will want to await `(trash state side card {:unpreventable true})` with `when-completed`, the continue the ability. Getting tricky now!

src/clj/game/cards/programs.clj
```diff
@@ -877,4 +877,48 @@
...
+                                               (continue-ability
+                                                (let [chosen-subtype target]
+                                                  {:choices {:req #(and (ice? %)
+                                                                        (not (rezzed? %)))}
+                                                   :delayed-completion true
+                                                   :msg (str "name " chosen-subtype)
+                                                   :effect (req
+                                                            (when-completed (expose state side target)
+                                                              (if (and async-result
```
nealterrell: Very nice :).

src/clj/game/cards/programs.clj
```diff
@@ -877,4 +877,48 @@
...
+                                                                       (has-subtype? target chosen-subtype))
+                                                                (continue-ability
```
nealterrell: This final effect doesn't need a continue-ability, you can just directly code the effect functions (`move`, `system-msg`, `effect-completed`) here inside of a `do`. You _typically_ only need `continue-ability` if you need to show a new prompt,.

presheaf:
Hope I didn't overstep (or mess up) by merging here

PR #3052, opened 2018-02-03
Down the White Nile: Implement Distract the Masses

butzopower:
> _The Great Work was completed on a rainy Saturday afternoon._ 

This was a bit more challenging than I thought and took me a good couple of days of playing around with it.  Definitely helped really wrap my head around the code base, though, so I was glad it was challenging.  The biggest bit I got stuck on was trying to move DtM to `:rfg` from `:play-area`, but no matter what I did, it always seemed to have made its way to `:discard`.  I'm not sure if this is due to something completing too soon, or what, but I ended up having to work around it and grabbing it from Archives via `:cid`.  I noticed _Wake Up Call_ doing something similar, so I think it might be more an issue with delayed RFG operations.

I imagine there's potentially some optimisations to be made, let me know if I'm doing anything completely wrong!

### DIFF
--- a/src/clj/game/cards/operations.clj
+++ b/src/clj/game/cards/operations.clj
@@ -332,11 +332,9 @@
                                              (in-hand? %))}
                         :msg (msg "trash " (quantify (count targets) "card") " from HQ")
                         :effect (req (when-completed
-                                       (trash-cards state side eid targets nil)
+                                       (trash-cards state side targets nil)
                                        (continue-ability state side shuffle-two card nil)))
-                        :cancel-effect (req (continue-ability state side shuffle-two card nil))
-      }]
-
+                        :cancel-effect (req (continue-ability state side shuffle-two card nil))}]
      {:delayed-completion true
       :msg "give The Runner 2 [Credits]"
       :effect (effect (gain :runner :credit 2)
--- a/src/clj/game/cards/operations.clj
+++ b/src/clj/game/cards/operations.clj
@@ -322,6 +322,26 @@
    {:events {:pre-damage {:req (req (= target :brain)) :msg "do 1 additional brain damage"
                           :once :per-turn :effect (effect (damage-bonus :brain 1))}}}
 
+   "Distract the Masses"
+   (let [shuffle-two {:delayed-completion true
+                      :effect (effect (rfg-and-shuffle-rd-effect (find-cid (:cid card) (:discard corp)) 2))}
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/operations.clj
```diff
@@ -322,6 +322,26 @@
...
+   (let [shuffle-two {:delayed-completion true
+                      :effect (effect (rfg-and-shuffle-rd-effect (find-cid (:cid card) (:discard corp)) 2))}
+         trash-from-hq {:delayed-completion true
+                        :prompt "Select up to 2 cards in HQ to trash"
+                        :choices {:max 2
+                                  :req #(and (= (:side %) "Corp")
+                                             (in-hand? %))}
+                        :msg (msg "trash " (quantify (count targets) "card") " from HQ")
+                        :effect (req (when-completed
+                                       (trash-cards state side eid targets nil)
```
nealterrell: Don't pass an `eid` to a function if you're using it in `when-completed`; the macro will automatically generate a new eid to associate with the command. Right now DtM is being marked complete as soon as `trash-cards` finishes, because you're handing over DtM's eid to trash-cards. This would only be noticeable with Subcontract or AccDiag, but it's there.

src/clj/game/cards/operations.clj
```diff
@@ -322,6 +322,26 @@
...
+                                       (continue-ability state side shuffle-two card nil)))
+                        :cancel-effect (req (continue-ability state side shuffle-two card nil))
+      }]
```
nealterrell: I know it fights your C-syntax instincts (mine too!!!), but the preferred Clojure style is for these to be on the previous line.

src/clj/game/cards/operations.clj
butzopower: Makes sense, I had a feeling it was the wrong thing to do, but I kept getting arity issues and the only thing that seemed to work was passing through the eid.  Still not sure why passing a nil at the end calls the `trash-cards` function that receives arguments `[state side cards]`, need to read up more on Closure.

nealterrell: It doesn't! 😄  `when-completed`, as a macro, will auto-insert an `eid` parameter that it generates itself as the third parameter if none is present, so you are indeed calling the 5-parameter `trash-cards` overload. The difference compared to your earlier approach is that `when-completed` will create a new unique `eid` intead of using the UI already assigned to the "current" ability.

PR #3053, opened 2018-02-04
[WIP] Down the White Nile: Bacterial Programming

jwarwick:
Attempting to do an implementation of `Bacterial Programming`.
But it's broken - everything works great up to the point where the `reorder-choice` ability is called, then the prompts hang, though `(:prompt (:corp @state))` shows what I would expect to see on the screen.

I'm sure it's got to do with `delayed-completion` somehow, but I don't see it.

If I remove the `move` and `trash` calls in the `doseq` it reorders as expected.

Also need to handle cases where there are no cards left for any given step.
Also need to write unit tests.

### DIFF
--- a/src/clj/game/cards/agendas.clj
+++ b/src/clj/game/cards/agendas.clj
@@ -140,6 +140,56 @@
                                                                   (clear-wait-prompt :runner))} card nil)))}
                               card nil))}}
 
+   "Bacterial Programming"
+   (letfn [(hq-step [remaining to-trash to-hq]
+             {:delayed-completion true
+              :prompt "Select a card to move to HQ"
+              :choices (conj (vec remaining) "Done")
+              :effect (req (if (= "Done" target)
+                             (do
+                               (doseq [t to-trash]
+                                 (println "Moving" (:title t) "to trash")
+                                 (trash state :corp t {:unpreventable true}))
+                               (doseq [h to-hq]
+                                 (println "Moving" (:title h) "to hand")
+                                 (move state :corp h :hand))
+                               (println "Remaining cards:" (map :title remaining))
+                               (continue-ability state :corp (reorder-choice :corp :runner remaining '()
+                                                                            (count remaining) remaining) card nil)
+                               (system-msg state :corp (str "uses Bacterial Programming to add " (count to-hq) " cards to HQ,  discard "
+                                                            (count to-trash) ", and arrange the top cards of R&D")))
+                             (do
+                               (system-msg state :corp (str "selected " (:title target) " to move to HQ"))
+                               (continue-ability state :corp (hq-step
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/agendas.clj
```diff
@@ -140,6 +140,56 @@
...
+              :choices (conj (vec remaining) "Done")
+              :effect (req (if (= "Done" target)
+                             (continue-ability state :corp (hq-step remaining to-trash `()) card nil)
+                             (do
+                               (system-msg state :corp (str "selected " (:title target) " to trash"))
+                               (continue-ability state :corp (trash-step
+                                                               (clojure.set/difference (set remaining) (set [target]))
+                                                               (conj to-trash target)) card nil))))})]
+     (let [arrange-rd (effect (continue-ability
+                                {:delayed-completion true
```
nealterrell: I think this actually goes inside the `:optional` map, next to the `:prompt`. No keys that are "parallel" to an `:optional` are actually used in the resolution of an optional ability. (See `check-optional` in abilities.clj -- we read the `:optional` key of the given `ability`, then completely ignore anything else that might have also been in `ability`.)

src/clj/game/cards/agendas.clj
```diff
@@ -140,6 +140,56 @@
...
+                                 :optional
+                                 {:prompt "Arrange top 7 cards of R&D?"
+                                  :no-ability {:effect (req (effect-completed state side eid card))}
```
nealterrell: Just FYI, the `card` parameter to `effect-completed` is actually deprecated and does nothing.

src/clj/game/cards/agendas.clj
```diff
@@ -140,6 +140,56 @@
...
+                                  :yes-ability {:effect (req (let [c (take 7 (:deck corp))]
```
nealterrell: This ability needs to be marked `:delayed-completion` because it continues into `trash-step`.

src/clj/game/cards/agendas.clj
nealterrell: Also, could you experiment and report back about not having this `:no-ability` at all? I'm looking at the engine and it LOOKS like a missing `:no-ability` will trigger `effect-completed` automatically, but for some reason I've been scattering around `:no-ability` values like yours without checking if they were really necessary.

jwarwick:
Can confirm that a `no-ability` with just `effect-completed` is unnecessary  in my experiments. Adjusting `delayed-completion` into the `optional` map fixed a prompt for the runner, but didn't appear to have any effect on `reorder-choice`.

After more futzing around, it looks like recursively calling `continue-ability` is part of the issue. If you choose `Done` (without selecting cards), things appear to work correctly. As soon as you select any card in the `HQ` or `Discard` branches, it hangs when `reorder-choice` prompt should be drawn.

PR #3070, opened 2018-02-10
Automate Kit's ability

butzopower:
Noticed that Kit's ability was manually triggered and maybe not super obvious that it wasn't automatically taking effect.  I tried moving over what the old ability did just to events, but it didn't work correctly, so I reimplemented it a bit.  Making sure to grab the latest version of the card when reapplying its old subtypes so that additonal effects from things like Peregine still persist.

### DIFF
--- a/src/clj/game/cards/identities.clj
+++ b/src/clj/game/cards/identities.clj
@@ -784,16 +784,16 @@
     {:encounter-ice {:once :per-turn
                      :effect (req (let [stypes (:subtype target)]
                                     (update! state :runner (-> card
-                                                               (assoc :kit-target target)
-                                                               (assoc :kit-target-stypes stypes)))
+                                                               (assoc-in [:special :kit-target] target)
+                                                               (assoc-in [:special :kit-target-stypes] stypes)))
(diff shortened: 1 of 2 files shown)


JoelCFC25:
There is a _super_ niche corner case here (and probably existed anyway already):  if Kit runs an unprotected R&D and hits one of the ICE that gets encountered when accessed (Archangel, Sapper, et al), the ability won't be used up and will still trigger when you subsequently encounter installed ICE later that turn.

butzopower:
Does it make sense for those trap ICE to trigger the `:encounter-ice` event, since there's a few other cards that would trigger in that scenario as well (`Scrubbed` probably being the only other one that is relevant)?  Would the access happen before `:run-ends` is triggered so that the effect would clear properly?

nealterrell:
It's another super niche problem (I doubt anyone is ever going to run 4-influence Charlatan in Kit), but because Charlatan bypasses at approach and not encounter, Kit's ability should not be consumed by the bypassed ice. We currently always trigger the encounter event for each step in the run.

NoahTheDuke:
Closing cuz it's very stale and will require work across the board to be usable.

PR #3076, opened 2018-02-12
The Devil and the Dragon: Implement City Works Project

NoahTheDuke:
I don't know Clojure super well, so hopefully I've implemented this correctly (as I currently can't get the server to compile/run on my Windows machine).

### DIFF
--- a/src/clj/game/cards/agendas.clj
+++ b/src/clj/game/cards/agendas.clj
@@ -252,12 +252,12 @@
     :effect (effect (move-zone :runner :discard :rfg))}
 
    "City Works Project"
-   {:install-state :face-up
-    :access {:req (req installed)
-             :msg (msg "do " (+ 2 (:advance-counter (get-card state card) 0)) " meat damage")
-             :delayed-completion true
-             :effect (effect (damage eid :meat (+ 2 (:advance-counter (get-card state card) 0))
-                                     {:card card}))}}
+   (let [meat-damage (+ 2 (:advance-counter (get-card state card) 0))]
+     {:install-state :face-up
+      :access {:req (req installed)
+               :msg (msg "do " meat-damage " meat damage")
+               :delayed-completion true
+               :effect (effect (damage eid :meat meat-damage {:card card}))}})
 
    "Clone Retirement"
    {:msg "remove 1 bad publicity" :effect (effect (lose :bad-publicity 1))
--- a/src/clj/game/cards/agendas.clj
+++ b/src/clj/game/cards/agendas.clj
@@ -251,6 +251,14 @@
     :interactive (req true)
     :effect (effect (move-zone :runner :discard :rfg))}
 
+   "City Works Project"
+   {:install-state :face-up
+    :access {:req (req installed)
+             :msg (msg "do " (+ 2 (:advance-counter (get-card state card) 0)) " meat damage")
+             :delayed-completion true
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
Anyone want to take a look at this and let me know if I've done it correctly?

test/clj/game_test/cards/agendas.clj
```diff
@@ -158,7 +158,20 @@
       (prompt-select :corp kati)
       (is (empty? (:prompt (get-runner))) "Fall Guy prevention didn't occur")
       (is (= 1 (count (:discard (get-runner)))) "Kati Jones trashed"))))
-	  
+
+(deftest city-works-project
+  ;; City Works Project - do 2 + advancement counters meat damage on access
+  (do-game
+    (new-game (make-deck "Weyland Consortium: Building a Better World" [(qty "City Works Project" 1)])
```
Saintis: You don't need to make sure ID agrees with agenda for the tests. Tests care not for legal decks ;)

src/clj/game/cards/agendas.clj
```diff
@@ -251,6 +251,14 @@
     :interactive (req true)
     :effect (effect (move-zone :runner :discard :rfg))}
 
+   "City Works Project"
+   {:install-state :face-up
+    :access {:req (req installed)
+             :msg (msg "do " (+ 2 (:advance-counter (get-card state card) 0)) " meat damage")
```
Saintis: I'd be tempted to extract the 2 + advancements function to reduce the minor repetition.

test/clj/game_test/cards/agendas.clj
NoahTheDuke: That's great to know.

src/clj/game/cards/agendas.clj
NoahTheDuke: Something like this?

```clojure
   "City Works Project"
   (let [ability (+ 2 (:advance-counter (get-card state card) 0))]
     {:install-state :face-up
      :access {:req (req installed)
               :msg (msg "do " ability " meat damage")
               :delayed-completion true
               :effect (effect (damage eid :meat ability {:card card}))}})
```

Saintis: Exactly! I'd name it something like `meat-damage` or `damage-strength`.

src/clj/game/cards/agendas.clj
```diff
@@ -251,6 +251,14 @@
     :interactive (req true)
     :effect (effect (move-zone :runner :discard :rfg))}
 
+   "City Works Project"
+   (let [meat-damage (+ 2 (:advance-counter (get-card state card) 0))]
```
nealterrell: Hmmm... The `let` here is evaluated once and only once at the start of the server lifecycle. At that point there is no state, card, etc; so I'd be surprised if this actually compiled, as you're calling `+` on a function that can't evaluate because there's no variable named `state` or `card` in the given context.

The way you're using `meat-damage`, it needs to be a function that you pass `state` and `card` to from the `:msg` and `:effect` below.

NoahTheDuke: Good to know. I'll try again. Thanks for the help.

NoahTheDuke:
Okay, I messed up and made this branch point towards dev, not master, so I'm going to close and reopen from a better named branch, pointing at master.

PR #3077, opened 2018-02-12
Made Assimilator charge for flipped program mu

### LINKED ISSUE #2990: Assimilated programs don't cost MU

jwarwick:
Decrements runner memory and displays a toast message if memory goes negative.

Fixes #2990


nealterrell:
I'd like to abstract this and clean up the install code overall. Maybe you could look into separating things that happen when runner installs a card (costs, zone restrictions, card reqs, ...?) vs. things that must happen when a card enters play (register events, trigger effect, adjust MU). There's a number of interactions that are broken because we use `:effect` of cards to both trigger their "when you install" ability, and to do miscellaneous tracking and adjustment (such as via `watch`). Separating the engine functions for on-install vs. on-activated would be the first step towards cleaning that up.

Issue #2942, opened 2017-11-16
The Gauntlet and Rip Deal do not work together as intended


marc-jones:
I believe that The Gauntlet and Rip Deal should work together so you can return min(#cards in HQ, 1 + #ICE completely broken) to hand. However, Rip Deal currently fires first, returning 1 card, then The Gauntlet triggers to no effect.

Issue #3082, opened 2018-02-14
The Gauntlet not activating on HQ runs when Crisium Grid active.


Zerothmaxima:
The wording from The Gauntlet should trigger on card accesses, not on successful run.

danhut:
You are right this currently fires on success- does that cause a problem in terms of game play?

Zerothmaxima:
Thanks for looking into it!

Yes, Gauntlet specifies on access, so run success as defined by Crisium
should not impact its trigger. Maybe I'm not aware of a command to enable
multiaccess on runs, that could be a substitute. I could run this by Jako
if we need confirmation, but the language on the cards seems pretty clear.

danhut:
Makes sense - thanks for the reasoning - it helps understand the situation.  Text is clear - just implementation not.

PR #3088, opened 2018-02-17
Fix for Feb

### LINKED ISSUE #3085: Weird interaction between Sac Con + Severnius Stim Implant
... where, if programs are trashed from the grip for the Severnius run, it gives the "Runner has option to prevent trashing" dialogue once for each program. This happens after the cards have been selected and "Done" clicked on the Severnius prompt.

(Not a huge problem but, since the "Done" and "Jack Out" buttons appear in the same place, it's easy enough to accidentally click one too many times and end your Severnius run early.)

danhut:
fix #3085

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -11,6 +11,7 @@
                  [cheshire "5.6.3"]
                  [org.omcljs/om "0.9.0"]
                  [sablono "0.3.4"]
+                 [reagent "0.8.0-alpha2"]
                  [com.novemberain/monger "3.1.0"]
                  [org.clojure/core.match "0.3.0-alpha4"]
                  [differ "0.3.1"]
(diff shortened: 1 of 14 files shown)


nealterrell:
RE: #3082, the handler definitely should be `:pre-access`, but I changed it to `:successful-run` to fix #2942. Another solution will need to be constructed to deal with both issues.

danhut:
How about a new event which The Gauntlet fires from?
:post-successful-run 

Assuming you had all 3 cards mentioned in play:

1/ Runner plays Rip Deal
2/ Breaks all subs on an ICE
3/ :successful-run event fires   (if Crisium is rezzed this is suppressed but no longer effect The Gauntlet)
4/ :post-succesful-run fires - The Gauntlet happens
5/ Rip Deal replaces access

PR #3107, opened 2018-02-21
CotC: Friday Chip

jwarwick:
Implemented Friday Chip

### DIFF
--- a/src/clj/game/cards/hardware.clj
+++ b/src/clj/game/cards/hardware.clj
@@ -404,8 +404,9 @@
      {:events {:runner-turn-begins ability
                :runner-trash {:optional
                               {:prompt "Gain a virus counter on Friday Chip?"
-                               :yes-ability {:msg "to gain a virus counter"
-                                             :effect (effect (add-counter card :virus 1))}}}}})
+                               :yes-ability
+                               {:effect (effect (add-counter card :virus 1)
+                                                (system-msg :runner (str "places 1 virus counter on Friday Chip")))}}}}})
 
    "GPI Net Tap"
    {:implementation "Trash and jack out effect is manual"
--- a/src/clj/game/cards/hardware.clj
+++ b/src/clj/game/cards/hardware.clj
@@ -394,6 +394,18 @@
                 {:msg "remove 1 tag" :label "[Trash]: Remove 1 tag"
                  :effect (effect (trash card {:cause :ability-cost}) (lose :tag 1))}]}
 
+   "Friday Chip"
+    (let [ability {:msg (msg "move 1 virus counter to " (:title target))
+                   :req (req (pos? (get-in card [:counter :virus] 0)))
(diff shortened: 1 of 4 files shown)


src/clj/game/cards/hardware.clj
```diff
@@ -394,6 +394,18 @@
                 {:msg "remove 1 tag" :label "[Trash]: Remove 1 tag"
                  :effect (effect (trash card {:cause :ability-cost}) (lose :tag 1))}]}
 
+   "Friday Chip"
+    (let [ability {:msg (msg "move 1 virus counter to " (:title target))
+                   :req (req (pos? (get-in card [:counter :virus] 0)))
+                   :choices {:req #(and (has-subtype? % "Virus")
+                                        (pos? (get-in % [:counter :virus] 0)))}
```
nealterrell: I think this is out of place -- Friday Chip does not require the target to have a virus counter already. We should also enforce "virus _program_"... so far only programs are viruses, but you never know.

src/clj/game/cards/hardware.clj
```diff
@@ -394,6 +394,18 @@
...
+                   :effect (req (when (pos? (get-virus-counters state side target))
+                                  (add-counter state side card :virus -1)
+                                  (add-counter state side target :virus 1)))}]
+     {:events {:runner-turn-begins ability
+               :runner-trash {:msg "to gain a virus counter"
+                              :effect (effect (add-counter card :virus 1))}}})
```
nealterrell: Should we make this optional? There are now reasons to not gain virus counters (Reverse Infection)... though I suppose the runner has plenty of time to manually remove a counter after trashing a corp card, before ending their turn and letting the corp act?

src/clj/game/cards/hardware.clj
```diff
@@ -398,13 +398,14 @@
...
-                                  (add-counter state side target :virus 1)))}]
+                                        (is-type? % "Program"))}
+                   :effect (req (add-counter state :runner card :virus -1)
+                                (add-counter state :runner target :virus 1))}]
      {:events {:runner-turn-begins ability
-               :runner-trash {:msg "to gain a virus counter"
-                              :effect (effect (add-counter card :virus 1))}}})
+               :runner-trash {:optional
+                              {:prompt "Gain a virus counter on Friday Chip?"
+                               :yes-ability {:msg "to gain a virus counter"
```
JoelCFC25: Keep in mind this `:msg` will result in a message like "Player uses Friday Chip to to gain a virus counter"

jwarwick: Hmm. Maybe I should do it with a `system-msg` as `places 1 virus counter on Friday Chip`.

JoelCFC25: My point was more about the doubled "to", but I've used `system-msg` like that often to indicate results of automatic behavior rather than active usage.

PR #3122, opened 2018-02-26
Fix Leave No Trace derezzing pre-rezzed modified ICE

### LINKED ISSUE #3121: Leave No Trace derezzes ICE advanced via Reconstruction Contract
This one is a bit weird to reproduce.

I had a rezzed Firewall on HQ and my opponent used Leave No Trace to make a run on a remote server.  I trashed a Reconstruction Contract with some counters on it and advanced the Firewall.  Once the run ended, the Firewall derezzed due to Leave No Trace.  I'm guessing it's not specifically Reconstruction Contract, might be easier to reproduce with Matrix Analyzer or Space Camp.

butzopower:
Fixes #3121 

Compare previously rezzed ICE with post-run rezzed ICE via their :cids instead of the cards themselves.

There's probably still some edge cases for LNT if the runner derezzes midrun with a bird breaker and the corp loops them back somehow and rezzes, but that seems like a pretty crazy edge case and would have been the case with the old implementation as well.

### DIFF
--- a/src/clj/game/cards/events.clj
+++ b/src/clj/game/cards/events.clj
@@ -948,24 +948,27 @@
                  (game.core/run state side (make-eid state) target nil card))}
 
    "Leave No Trace"
-   {:prompt "Choose a server"
-    :msg "make a run and derez any ICE that are rezzed during this run"
-    :choices (req runnable-servers)
-    :delayed-completion true
-    :effect (req
-              (let [old-ice (filter #(and (rezzed? %) (is-type? % "ICE")) (all-installed state :corp))]
-                (swap! state assoc :lnt old-ice)
-                (register-events state side (:events (card-def card)) (assoc card :zone '(:discard)))
-                (game.core/run state side (make-eid state) target nil card)))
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/events.clj
```diff
@@ -948,24 +948,27 @@
...
-                                           old (set (:lnt @state))
-                                           diff (seq (clojure.set/difference new old))]
-                                       (doseq [ice diff]
-                                         (derez state side ice))
-                                       (when-not (empty? diff)
-                                         (system-msg state side (str "derezzes " (join ", " (map :title diff)) " via Leave No Trace")))
-                                       (swap! state dissoc :lnt)
-                                       (unregister-events state side card)))}}}
+   (letfn [(get-rezzed-cids [ice]
+             (map #(:cid %) (filter #(and (rezzed? %) (is-type? % "ICE")) ice)))]
```
nealterrell: If you ever find yourself writing `#(_some_function %)`, just use `_some_function` instead, as in `(map :cid ...)`.

butzopower: Ah, makes sense, always forget that `:xxx` is a function.  Should be fixed now.

Issue #3096, opened 2018-02-19
Council of the Crest card implementations


jwarwick:
Rumored to be coming March 1st.

- [x] eXer (#2932)
- [x] Friday Chip (#3107)
- [x] Crypt (#3134)
- [x] Corporate "Grant" (#3097)
- [x] No One Home (#3148)
- [x] Marathon (#3158)
- [x] Gbahali (#3137)
- [x] White Hat (#3133)
- [x] Kuwinda K4H1U3 (#3129)
- [x] NEXT Sapphire (#3132)
- [x] Anansi (#3139 )
- [x] Code Replicator (#3138)
- [x] Reverse Infection (#3128)
- [x] Azmari EdTech: Shaping the Future (#2933)
- [x] Degree Mill (#3141)
- [x] Personalized Portal (#3098)
- [x] Armed Intimidation (#3127)
- [x] Death and Taxes (#3127)
- [x] Trojan Horse (#3157)
- [x] TechnoCo (#3126)

Saintis:
More cards [here](https://www.reddit.com/r/Netrunner/comments/7zdihf/data_breach_textonly_spoilers_of_the_remaining/).

butzopower:
I can do TechnoCo.

presheaf:
I can do Marathon.

jwarwick:
All card implementations merged.

Issue #3125, opened 2018-02-27
Acacia not giving credits for tokens purged from Virus Breeding Ground


Pirsig:
When the corp purges and Acacia is used credits are not gained for the tokens lost from Virus Breeding Ground.

butzopower:
Think this is fixed now, and after looking into it, I think it makes sense to check all Runner cards given this is coming out: https://netrunnerdb.com/en/card/21083

PR #3128, opened 2018-02-28
Implement Reverse Infection

butzopower:
#3096

Due to things like Progenitor I figure out removed tokens by counting pre-purge tokens and taking the difference post-purge to see how many tokens total were purged.


butzopower:
Realized I could do a refactor that would fix the Acacia bug (#3125) as well.

PR #3141, opened 2018-03-03
CotC: Degree Mill

jwarwick:
Basic implementation and tests for Degree Mill.

Can't figure out how to check if the runner actually selected two cards to shuffle back into their deck. The agenda steal happens, then the runner is prompted to select two cards, but they can click `Done` and I don't see how to catch that to prevent the steal.


danhut:
you can use this to ensure two cards are selected via a prompt-select

`:choices {:max 2 :all true}`

jwarwick:
Thanks @danhut! That fixes the problem.

PR #3152, opened 2018-03-06
Spoiled card: Too Big to Fail

PesiTheWizard:
From "reign and reverie"

_(Not sure about the BP. Some such things are silent, some not; some are interactive, some not; some are delayed, some not.)_

### DIFF
--- a/src/clj/game/cards/events.clj
+++ b/src/clj/game/cards/events.clj
@@ -1083,7 +1083,7 @@
 
                                       (= target "Take 1 Bad Publicity")
                                       (do (gain state :corp :bad-publicity 1)
-                                          (system-msg state side "takes 1 Bad Publicity from Mining Accident")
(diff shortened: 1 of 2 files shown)


JoelCFC25:
There's only one card printed to date that can prevent gaining bad publicity (Broadcast Square), and it uses a crude workaround (removing it after the fact, not true prevention). 

Minor point: I don't think any other cards with bad publicity in the message ever use title case for it, so all-lower is probably the way to go for consistency.

PesiTheWizard:
Man, a lot has sure happened around here in not a lot of time.

Looks like someone else is changing how BP is granted. Let's be vigilant.

NoahTheDuke:
All that's changed is `(gain :corp :bad-publicity n)` is now `(gain-bad-publicity :corp n)`.

PR #3163, opened 2018-03-08
Friday Chip Fix

### LINKED ISSUE #3161: Friday Chip triggering on Runner discards
When the Runner discards down to their max hand size at the end of their turn, each trash is triggering the prompt to place a virus counter on Friday Chip.

NoahTheDuke:
Solution developed from Gravedigger's method

Fixes #3161

### DIFF
--- a/src/clj/game/cards/hardware.clj
+++ b/src/clj/game/cards/hardware.clj
@@ -396,17 +396,18 @@
                  :effect (effect (trash card {:cause :ability-cost}) (lose :tag 1))}]}
 
    "Friday Chip"
-    (let [ability {:msg (msg "move 1 virus counter to " (:title target))
-                   :req (req (pos? (get-in card [:counter :virus] 0)))
-                   :choices {:req #(and (has-subtype? % "Virus")
-                                        (is-type? % "Program"))}
-                   :effect (req (add-counter state :runner card :virus -1)
-                                (add-counter state :runner target :virus 1))}]
+   (let [ability {:msg (msg "move 1 virus counter to " (:title target))
+                  :req (req (pos? (get-in card [:counter :virus] 0)))
+                  :choices {:req #(and (has-subtype? % "Virus")
+                                       (is-type? % "Program"))}
+                  :effect (req (add-counter state :runner card :virus -1)
+                               (add-counter state :runner target :virus 1))}]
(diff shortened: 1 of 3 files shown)


NoahTheDuke:
Honestly, I have no idea what I'm doing, but this looks about right?

src/clj/game/core/cards.clj
```diff
@@ -197,6 +197,11 @@
   (let [hiveminds (filter #(= (:title %) "Hivemind") (all-installed state :runner))]
     (reduce + (map #(get-in % [:counter :virus] 0) (cons card hiveminds)))))
 
+(defn count-virus-programs
+  "Calculate the number of virus programs in play"
+  [state]
+  (count (filter #(= false (has-subtype? % "Virus")) (all-installed state :runner))))
```
jwarwick: Instead of `filter` you could try `remove` and drop the `= false` comparison.

NoahTheDuke: Yeah, I realized that I had the filter predicate incorrect in my head. Needs to return a sequence of cards where the predicate `:subtype == "Virus"` is true, not the other way around.

Fixed, in an accidental force-push.

src/clj/game/cards/hardware.clj
```diff
@@ -396,17 +396,19 @@
...
-                                (add-counter state :runner target :virus 1))}]
-     {:events {:runner-turn-begins ability
-               :runner-trash {:optional
+   (let [ability {:msg (msg "move 1 virus counter to " (:title target))
+                  :req (req (pos? (get-in card [:counter :virus] 0)))
+                  :choices {:req #(and (has-subtype? % "Virus")
+                                       (is-type? % "Program"))}
+                  :effect (req (add-counter state :runner card :virus -1)
+                               (add-counter state :runner target :virus 1))}]
+     {:events {:runner-turn-begins {:req (req (pos? (count-virus-programs state)))
```
Saintis: I'm not sure this will work, since `ability` is not in the form the game engine expects (a 5 (6?) function (it's what the `req` and `effect` macros creates). 

It's better if you just add the new req clause to the `ability` map.

NoahTheDuke: Ah, okay. Good to know.

Issue #3144, opened 2018-03-04
Decks not loading in deckbuilder tab and unable to start new games


Sythic:
I get stuck with an endless "loading deck collection" and when starting a new game I get strange glitches in the interface that makes it seem like I've joined someone else's game that was in progress.  I also cant watch any games.

have tested on 3 devices, in chrome, Edge and opera.

Saintis:
Probably related to #3143. Try hard-reloading / refreshing the cache.

Sythic:
I have, and now tried on a 4th PC.
I have created a second account and it works fine, so it might be that I have too many decks saved... but cant get to a screen to remove the older ones

Saintis:
Could you have a look on the developer tools / web inspector / debug mode to see if there are any javascript errors?

Sythic:
will try later, currently on an andriod tablet and not sure if the debug tools are available in this version?

Sythic:
Ok!  So I get: 

quite a few: 

Uncaught TypeError: Cannot read property 'toLowerCase' of null
    at SX (app.js?v=1.0-RC2:1396)
    at UX (app.js?v=1.0-RC2:1397)
    at app.js?v=1.0-RC2:1401
    at app.js?v=1.0-RC2:1401
    at Vf (app.js?v=1.0-RC2:187)
    at Uf.k.ua (app.js?v=1.0-RC2:190)
    at n (app.js?v=1.0-RC2:111)
    at app.js?v=1.0-RC2:247
    at Vf (app.js?v=1.0-RC2:187)
    at Uf.k.ua (app.js?v=1.0-RC2:190)

and lots of: 

app.js?v=1.0-RC2:28 Uncaught Error: Minified exception occurred; use the non-minified dev environment for the full error message and additional helpful warnings.
    at r (app.js?v=1.0-RC2:28)
    at Object.findComponentRoot (app.js?v=1.0-RC2:27)
    at Object.findReactNodeByID (app.js?v=1.0-RC2:27)
    at Object.s [as getNode] (app.js?v=1.0-RC2:27)
    at Object.updatePropertyByID (app.js?v=1.0-RC2:25)
    at a._updateDOMProperties (app.js?v=1.0-RC2:25)
    at a.updateComponent (app.js?v=1.0-RC2:25)
    at a.receiveComponent (app.js?v=1.0-RC2:25)
    at Object.receiveComponent (app.js?v=1.0-RC2:27)
[...]

PR #3165, opened 2018-03-08
Handling decks with missing identities

jwarwick:
Catching decks with missing identities when loaded. Leaves the deck in an unplayable state, but a user can still get to the included cards (to rebuild).

Checking for an `:identity` on deck save as well.


jwarwick:
Fixes #3144

PR #3166, opened 2018-03-08
Open Forum fix

JoelCFC25:
Tweak `end-turn-12` to let multiple Open Forums process one by one.


nealterrell:
Let's also make this compatible with Political Dealings / Jinja City Grid. `draw` is now an awaitable function that you can use with `when-completed` to wait for on-draw triggers to resolve before moving to the "add 1 to R&D" routine.

nealterrell:
NEVERMIND, this card does not draw, it adds to HQ.

PR #3177, opened 2018-03-11
fix rebirth kate test

Giesch:
NetrunnerDB has added spoiled cards, so this test breaks on a fresh 'lein fetch'.


JoelCFC25:
Neal beat you to the punch in #3173. :)

Giesch:
so sad

PR #3179, opened 2018-03-11
kakugo swap fix

### LINKED ISSUE #3164: Kakugo and Inversificator
Hi,

I have just played a game against a runner with Inversificator and a few Kakugoes floating around on my side. The usual net damage firing (when the Runner "continues" past the ICE) happened as usual, until Kakugo was swapped with Inversificator. It then never fired again for the rest of the game in its new position. This happened for every individual copy of Kakugo, working fine until it was swapped somewhere else, so it looks like the switching positions with Inversificator turns off the function somehow.
 
Not sure if this relates to #2835 in some way too?

Giesch:
Fixes #3164 with what Saintis suggested.


Saintis:
Looks good. Test fail is due to new shaper runner added on NRDB which messes up the Rebirth test, see #3173.

PR #3180, opened 2018-03-11
Check that Akiko is first in Shaper rebirth list

butzopower:
Akiko is now ahead of Ayla alphabetically, so I updated the test to reflect that.  Hopefully they don't release anything like `Aardvark Jones: Mutant Celebrity` or something at some point.


JoelCFC25:
Already in at #3173...

PR #3181, opened 2018-03-12
Block users from the lobby chat screen

jwarwick:
**Need some help with styling**, but the functionality is in place. Click a button by the username in a chat message, they get added to your block list. You get a toast message with confirmation and reminder to refresh your browser.

It should be possible to update the Om state and get it to redraw automatically, but I can't seem to make that happen.

You don't get a button by your username (`test2` in the screenshot) or when not logged in.

screen shot 2018-03-11 at 9 07 59 pm

### DIFF
--- a/src/cljs/netrunner/chat.cljs
+++ b/src/cljs/netrunner/chat.cljs
@@ -20,13 +20,15 @@
   :chat/message
   (partial put! chat-channel))
 
-(defn current-block-list [] (get-in @app-state [:options :blocked-users] nil))
+(defn current-block-list
+  []
+  (if-let [curr (get-in @app-state [:options :blocked-users] nil)]
+    curr
+    []))
 
 (defn filter-blocked-messages
   [messages]
-  (if-let [blocked-users current-block-list]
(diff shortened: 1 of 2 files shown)


nealterrell:
I'd rather this be tied to a menu that appears when clicking the user's name. Eventually that menu could be expanded with Add Friends, PM, etc.

jwarwick:
Makes sense. Do you imagine a context menu? That might be confusing, since we can't trigger it with a right click. Doing a full page overlay (like the login prompt) is probably easier.

nealterrell:
I had pictured something similar to the card ability pop-in menu.

jwarwick:
Okay, that's more the context menu I was envisioning. I'll take a stab at that and see how it looks.

jwarwick:
Reworked so that when you click on a username in chat you get a menu. The cursor changes when you mouse over another user (but not your name). Clicking the name when the menu is open closes the menu. Selecting `Cancel` closes the menu.

screen shot 2018-03-17 at 5 43 44 pm

Issue #3183, opened 2018-03-12
Persephone + AR-Enhanced Security


JoelCFC25:
With a copy of AR-Enhanced scored, Runner usage of Persephone should result in gaining a tag if it's the first Runner trash of a Corp card that turn. Currently Persephone is doing: `(mill :corp target)`, which only moves cards from R&D to the discard and never generates a trash event.

NoahTheDuke:
This seems like an issue with `mill` moving cards directly into the discard instead of doing some sort of short-form trashing. I'd try switching it out for `trash`, but that then interacts with the entire lattice-work that is "the Runner trashing things they access", which isn't directly what this is. Maybe borrow the `trash-no-cost` functionality, without the scoring agendas bit? Or `resolve-trash-no-cost` without the `card :seen true` bit, because the cards are trashed facedown.

PR #3187, opened 2018-03-13
FIX: Milling doesn't trigger trash effects

NoahTheDuke:
~~**NOTE:** Not actually intended to be a fix, but a good start on a fix for someone who can actually test the code (as I can't get my dev environment working).~~ Just kidding, see below.

If I understand things correctly, this should solve the issue in #3183 where milling doesn't count as "trashing cards" for each side (meaning that AR-Enhanced Security doesn't fire when Persephone fires) because `:trashed-card` isn't being set.

I've added an additional parameter to `resolve-trash-no-cost`, so that by default it sets all trashed cards to `:seen true`, but with milling, we can call `(resolve-trash-no-cost state side card false)` and they *should* be discarded facedown. Additionally, I moved the `closed-access-prompt` to `trash-no-cost`, as it can't be closed if it's not opened when `mill` is called. 

Honestly, I am only partially sure I got this right. Someone come steal my thunder and fix my mistakes!

### DIFF
--- a/src/clj/game/cards/agendas.clj
+++ b/src/clj/game/cards/agendas.clj
@@ -1087,23 +1087,15 @@
                                                                  (fn [_ _ c] (not= (:cid c) (:cid card)))))}}}}
 
    "Underway Renovation"
-   {:install-state :face-up
-    :events {:advance {:req (req (= (:cid card) (:cid target))) 
-                       :msg (msg (let [deck (:deck runner)
-                                       anydeck? (pos? (count deck)) 
-                                       adv4? (>= (:advance-counter (get-card state card)) 4)] 
-                         (cond
-                           (and anydeck? adv4?)
-                           (str "trash " (join ", " (map :title (take 2 deck))) " from the Runner's stack")
-
-                           (and anydeck? (not adv4?) )
-                           (str "trash " (:title (first deck)) " from the Runner's stack")
-
-                           (false? anydeck?)  
-                           "trash from the Runner's stack but it is empty")))
-
-                       :effect (effect (mill :runner
-                                             (if (>= (:advance-counter (get-card state card)) 4)  2 1)))}}}
+   (letfn [(adv4? [s c] (if (>= (:advance-counter (get-card s c)) 4) 2 1))]
+     {:install-state :face-up
+      :events {:advance {:req (req (= (:cid card) (:cid target)))
+                         :msg (msg (if (pos? (count (:deck runner)))
+                                     (str "trash "
+                                          (join ", " (map :title (take (adv4? state card) (:deck runner))))
+                                          " from the Runner's stack")
+                                     "trash from the Runner's stack but it is empty"))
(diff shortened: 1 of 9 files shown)


jwarwick:
What's up with your dev environment? I can't even imagine trying to do this without seeing it live.

NoahTheDuke:
When I load localhost, I get [this](https://screenshots.firefox.com/MuM3oUb1DLm0CaId/localhost), and after clicking sign up or login the figwheel console says:
```
unknown socket msg {:type :ws, :open? false, :ever-opened? false} {:type :ws, :open? true, :ever-opened? true, :uid 60b8ca33-33e0-4325-b597-719b87d4afe0, :csrf-token nil, :handshake-data nil, :first-open? true}
unknown socket msg 60b8ca33-33e0-4325-b597-719b87d4afe0 nil
unknown socket msg :games/list []
```

I'm on Windows, if that means anything.

JoelCFC25:
Leaving aside Figwheel for a moment (I don't use it and don't understand it, haha)...do you mean you are not able to create a new user account by doing "Sign up" in your dev environment?

NoahTheDuke:
Well, I'm not sure what changed, but I tried again and it seems to be working! Boy, this is exciting! I'll let you know when I figure out how to get this working.

NoahTheDuke:
In writing a test for this, how do I trigger Persephone's ability? `(prompt-choice :runner "Yes")` or `(prompt-choice :runner 1)` throws an error.

JoelCFC25:
Does your test include a run in which the Runner has passed a rezzed sentry?

NoahTheDuke:
Current test:

```clojure
(deftest persephone
  ;; Persephone's ability trashes cards from R&D, triggering AR-Enhanced Security
  (do-game
    (new-game
      (default-corp [(qty "Data Raven" 1) (qty "AR-Enhanced Security" 1)])
      (default-runner [(qty "Persephone" 1)]))
    (play-from-hand state :corp "AR-Enhanced Security" "New remote")
    (score-agenda state :corp (get-content state :remote1 0))
    (play-from-hand state :corp "Data Raven" "Archives")
    (core/rez state :corp (get-ice state :archives 0))
    (take-credits state :corp)
    (play-from-hand state :runner "Persephone")
    (run-on state "Archives")
    (run-continue state)
    ; (prompt-choice :corp "No")
    ; (prompt-choice :runner "Continue")
    ; (prompt-choice :runner "Yes")
    ; (prompt-choice :runner 1)
    (is (= 1 (:tag (get-runner))) "Runner took 1 tag from using Persephone's ability while AR-Enhanced Security is scored")))
```

How do I trigger the Persephone ability? If I uncomment any or all of the prompt-choices, it throws an error.

JoelCFC25:
Ah, here's why:  Persephone requires the count of the Stack to be a positive number. If there's nothing you can trash from it, the optional ability doesn't prompt you at all.

NoahTheDuke:
I think I got it right! Finally figure out how to make the test behave, and have now cleaned up the rest.

PR #3192, opened 2018-03-15
Awaitable costs for accessing and stealing

### LINKED ISSUE #3167: Degree Mill grants extra accesses
Degree Mill grants the runner another access (including any multiaccess modifiers) when it's stolen. It also erroneously specifies "programs" in the prompt, but still allows the runner to choose any installed cards.

nealterrell:
I know @danhut and @butzopower were interested in looking at this, but it was more complicated than I anticipated so I thought to do it myself.

Adds an awaitable function `pay-sync`, which takes the same parameters as `pay` but can be awaited with `when-completed`. Integrated this functionality into access and steal costs, so Degree Mill can work again.

As a brief walkthrough:

1. Each of the individual cost-payment methods had to be made awaitable.  (`pay-trash`, `pay-forfeit`, `pay-shuffle-into-stack`.) To do this, we:
    1. Add a new third parameter `eid`.
    2. Pass that eid to the prompt we are using to resolve the cost with `continue-ability`. 
    3. Each of those prompts must be delayed completion, so that we can...
    4. Complete the payment effect with `effect-completed`, "returning" the string we use to describe the cost was paid in the user interface. This is the tricky part. The non-awaited version of these functions returns that string (`cost-name`) immediately, but an awaitable function does not participate in the normal function-return sequence. (The call site continues executing while the awaitable function is still active.) To return something from an awaitable function, we use `make-result` to wrap the `eid` we trigger as completed with a "return value", in this case the string description. Whoever calls `when-completed` can then capture this returned value with the variable `async-result`, automatically bound in the body of the `when-completed` follow-up function.

2. `cost-handler` is also made awaitable. For complicated costs, its `eid` is passed to the appropriate handler (`pay-X` above), so when that function is complete, so too is `cost-handler`. For simple costs, `effect-completed` is triggered manually. In all cases, the return value from the handler is the string description returned from the `pay-X` functions (or from `deduce`), and we wrap that as the async result for `cost-handler`.  We also return that value (called `r` in the code) so that `cost-handler` can continue to be used in a non-awaitable fashion with the existing `pay` function.

3. `pay-sync` then triggers `cost-handler` for each cost in its list, and awaits the completion of the handler. The next cost is then recursively handled. Each async result (the string description of the cost) from a handler is added to a vector, and all those strings are joined together to make a single description as in `pay` itself.

4. To use the new function, we do something like this:

```clojure
(when-completed 
  (pay-sync state side card costs)
  (if async-result
    ;; cost was paid
    ;; cost was not paid
  )
)
```

Whew.

Fixes #3167

### DIFF
--- a/src/clj/game/core/costs.clj
+++ b/src/clj/game/core/costs.clj
@@ -1,6 +1,6 @@
 (in-ns 'game.core)
 
-(declare forfeit prompt! toast damage mill installed? is-type? is-scored? system-msg)
+(declare forfeit prompt! toast damage mill installed? is-type? is-scored? system-msg facedown? make-result)
 
 (defn deduce
   "Deduct the value from the player's attribute."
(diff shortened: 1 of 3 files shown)


JoelCFC25:
Hail to the king.

PR #3196, opened 2018-03-15
Bad Publicity triggers and The Outfit

NoahTheDuke:
Changes out bad publicity is gained to mirror taking tags: in fact, this is a copy-paste of the tagging code, with the necessary changes, and all gain bad publicity references modified to follow suit.

This allows for The Outfit to work as needed: when gaining any amount of bad publicity, the effect triggers `:corp-gain-bad-publicity`, and the corp gains 3 credits.

### DIFF
--- a/src/clj/game/cards/agendas.clj
+++ b/src/clj/game/cards/agendas.clj
@@ -784,8 +784,10 @@
    {:interactive (req true)
     :choices ["0" "1" "2" "3"] :prompt "How many bad publicity?"
     :msg (msg "take " target " bad publicity and gain " (* 5 (str->int target)) " [Credits]")
-    :effect (final-effect (gain-bad-publicity :corp eid (str->int target) {:unpreventable true})
-                          (gain :credit (* 5 (str->int target))))}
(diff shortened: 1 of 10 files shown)


NoahTheDuke:
Note that this will require a change in Too Big To Fail (#3152), so as to use the new bad publicity gaining function.

JoelCFC25:
Pretty cool. Shouldn't Broadcast Square have its ability hooked to a `:pre-bad-publicity` listener? And it doesn't have the capability to prevent just 1, it would prevent all inbound BP if there is ever a card that gives more than 1 at a time. For Profiteering, if you opted to take 2 or 3, then Broadcast Square kicked in and you prevented it all, you'd gain no Profiteering credits. Weird one there.

NoahTheDuke:
Yeah, good call. I'll see if I can auto-wire that.

NoahTheDuke:
There it is, all done. Broadcast Square, Profiteering, and The Outfit all work correctly. All necessary tests have been updated and work properly.

PR #3202, opened 2018-03-18
Alt art re-enablement

jwarwick:
When we upgraded to the Clojure server a few of the alt art functionalities stopped working. Now alt-art cards show up in the cardbrowser and may be selected there.

I also put @nealterrell's code back in that allows you to select `Original Core Set` art as an alt for `Revised Core Set` cards.

So basically I just uncommented a bunch of code.

### DIFF
--- a/src/cljs/netrunner/account.cljs
+++ b/src/cljs/netrunner/account.cljs
@@ -18,11 +18,11 @@
             cards (->> @all-cards
                     (filter #(not (:replaced_by %)))
                     (map #(select-keys % [:title :setname :code :alt_art :replaces :replaced_by]))
-                    ;(map #(let [replaces (:replaces %)
-                    ;            setname (:setname %)]
-                    ;        (if (and replaces (= "Revised Core Set" setname))
-                    ;          (update-in % [:alt_art] assoc :core replaces)
-                    ;          %)))
+                    (map #(let [replaces (:replaces %)
+                                setname (:setname %)]
+                            (if (and replaces (= "Revised Core Set" setname))
+                              (update-in % [:alt_art] assoc :core1 replaces)
+                              %)))
                     (filter :alt_art)
                     (into {} (map (juxt :code identity))))]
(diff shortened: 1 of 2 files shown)


kevkcc:
I haven't played for awhile but I notice some alt-art has disappeared such as the full bleed alt art hedge fund that looked like this: https://www.fantasyflightgames.com/en/news/2016/10/20/make-your-run/
Any plans to bring those back?

jwarwick:
Interesting. I do have that particular alt-art on my dev system, but I don't see it on the production server. This alt works: https://jinteki.net/img/cards/01110-wc2015.png
But this one (that your referred to), doesn't: https://jinteki.net/img/cards/01110-alt.png

We now build the list of alts based on the images on the filesystem, so either that system is broken (but not on my dev machine) or the file really is gone.

Don't know if @nealterrell has filesystem access or can check the card in the db and see what alts it lists.

nealterrell:
So I'm a little confused by this, because alt arts seem to be working OK on live? What exactly is this fixing?

Also, I went a different direction with Core 1.0 arts. Since not all Core 2.0 replacements have new artwork, I manually copied old card arts into a new art category that's treated the same as the other categories, hence the commented out code for that.

jwarwick:
@nealterrell I don't think the alts are working correctly on the live server. In the card browser, search for `Adonis Campaign` (with alts turned on and rotated cards turned off). For me, you don't get all the cards - at least one of the World Champ or Alt art is missing. All three appear in the `Settings page`.

Second thing that I fixed was the ability to select which alt art to use from the card browser. That endpoint disappeared in the server move. Just re-enabled it.

jwarwick:
I see your point about the original core set to revised core set art. If you've made copies of the old cards and renamed them on the server, then I shouldn't do that in the code. Agree that that approach is likely better and more future-proof. I'll update.

nealterrell:
OK, thanks for clearing that up!

PR #3207, opened 2018-03-19
add Diversion of Funds

Giesch:
Basically cribbed from vamp & siphon


msbeck:
Do we need an effect that jacks the runner out instead of it being manual?

NoahTheDuke:
Did you mean to post this to another card?

Giesch:
It's bugged because it shouldn't be delayed-completion. I can make a pr tonight if no one beats me to it.

PR #3209, opened 2018-03-20
Added support for Hivemind to Aumakua.

nicohasa:
Used get-virus-counters function instead of checking manually for number of counters.


JoelCFC25:
Nice catch!

PR #3212, opened 2018-03-21
Agenda Tests and Testing Updates

NoahTheDuke:
Testing is good and I need practice, so this is a branch for writing at least 1 test for every single untested agenda. Includes some small cosmetic changes to the agenda definitions.

## General notes:
* Added `play-and-score`, a 2-function that takes state and the title of an agenda in hand. It plays the agenda using `play-from-hand` to a new remote and then scores that agenda with `score-agenda`. This consumes 1 click and no credits, and increments the remote server count.
* Added `get-hand-size`, a 1-function that takes side, to the `do-game` macro. It sums `:hand-size-base` and `:hand-size-modification` for easy checking and usage.
* Moved `(trigger-event state :corp :pre-init-trace card)` from in `init-trace` to `corp-trace-prompt` so that trace bonuses are triggered at the actual start of a trace, before the Corp has made any decisions. This fixes the issue of the Corp not seeing the trace bonus.
* Changed all `:agendapoints-runner` from `(req (do X))` to `(req X)`. This is both cleaner and more explicit in what is happening. `(do X)` gives the impression the `do` is taking action when it is not. I experimented with converting `:agendapoints-runner` from a 5-function to a simple integer, but it's not necessarily a bad idea to future-proof (heh) the design this way.
* Rearranged a lot of agenda internals for visual clarity. Tried to keep the line lengths shorter by moving all hashes to one entry per line.

## Agenda rewrites:
* Meteor Mining: Hardcoded the options, so that they conform to current standards of "Actions" above "No action". Reworded prompt and options so they used the language on the card (give damage -> do damage, take credits -> gain credits, take nothing -> no action). Switched from `cond` to `case` for clarity.
* Voting Machine Initiative: Honestly not sure how this worked before, so I rewrote it to function how I'd expect: a prompt triggers at the start of the runners turn. If the corp chooses yes, an agenda counter is removed and the runner loses 1 click. ~~The "Yes" and "No" choices are hardcoded so as to avoid repeating `clear-wait-prompt` and `effect-completed`~~. Now properly uses `delayed-completion` and `:optional` instead of hardcoding anything.
* Standoff: I tried twice to write a test and got frustrated, so it's currently lacking a test. Someone else can tackle this one, I'm tired of writing tests. 😄


NoahTheDuke:
Nevermind, fixed that issue!

NoahTheDuke:
@nealterrell, ready for review!

Notes added to the original PR above.

src/clj/game/cards/agendas.clj
```diff
@@ -1128,16 +1186,25 @@
...
-                            :prompt "Cause the Runner to lose [Click] at the start of their next turn?"
-                            :yes-ability {:effect (effect (toast (str "The Runner will lose " (inc (:vmi-count card 0))
-                                                                      " [Click] at the start of their next turn") "info")
-                                                    (update! (update-in card [:vmi-count] #(inc (or % 0)))))}}}]
-    :events {:runner-turn-begins {:req (req (pos? (:vmi-count card 0)))
-                                  :msg (msg "force the Runner to lose " (:vmi-count card) " [Click]")
-                                  :effect (effect (lose :runner :click (:vmi-count card))
-                                                  (add-counter (dissoc card :vmi-count) :agenda (- (:vmi-count card))))}}}
+    :events {:runner-turn-begins
+             {:req (req (pos? (get-in card [:counter :agenda] 0)))
```
nealterrell: This needs to be `:delayed-completion true` if you're going to use continue-ability later (which you want to do).

src/clj/game/cards/agendas.clj
```diff
@@ -1128,16 +1186,25 @@
...
+              :effect (effect (show-wait-prompt :runner "Corp to use Voting Machine Initiative")
+                              (continue-ability
+                                {:player :corp
+                                 :prompt "Use Voting Machine Initiative to make the Runner lose 1 [Click]?"
+                                 :once :per-turn
```
nealterrell: Should be unnecessary, this event only fires once per turn anyways.

src/clj/game/cards/agendas.clj
```diff
@@ -1128,16 +1186,25 @@
...
+                                 :choices ["Yes" "No"]
```
nealterrell: This whole part should be an `:optional` ability.

src/clj/game/cards/agendas.clj
```diff
@@ -1128,16 +1186,25 @@
...
+                                 :effect (req (case target
+                                                "Yes"
+                                                (do (lose state :runner :click 1)
+                                                    (add-counter state side card :agenda -1)
+                                                    (system-msg state :corp
+                                                                (str "uses Voting Machine Initiative to make the Runner lose 1 [Click]")))
+                                                "No"
+                                                (system-msg state :corp (str "doesn't use Voting Machine Initiative")))
+                                              (clear-wait-prompt state :runner)
+                                              (effect-completed state side eid card))}
```
nealterrell: You won't need the effect-completed once you've converted this to an optional ability.

src/clj/game/core/rules.clj
```diff
@@ -463,11 +463,15 @@
   (let [base-points (:agendapoints card)
         runner-fn (:agendapoints-runner (card-def card))
         corp-fn (:agendapoints-corp (card-def card))]
-    (if (and (= side :runner) (not (nil? runner-fn)))
-      (runner-fn state side (make-eid state) card nil)
-      (if (and (= side :corp) (not  (nil? corp-fn)))
+    (cond
+      (and (= side :runner)
+           (some? runner-fn))
+        (runner-fn state side (make-eid state) card nil)
```
nealterrell: This should be aligned with the `and`.

src/clj/game/core/rules.clj
```diff
@@ -463,11 +463,15 @@
...
+      (and (= side :corp)
+           (some? corp-fn))
         (corp-fn state side (make-eid state) card nil)
```
nealterrell: Likewise.

src/clj/game/core/rules.clj
```diff
@@ -463,11 +463,15 @@
...
-        base-points))))
+      :else
+        base-points)))
```
nealterrell: Likewise.

src/clj/game/cards/agendas.clj
NoahTheDuke: Yeah, I wasn't sure about that one. Good call.

NoahTheDuke:
I think that's it, then! All completed. Can't wait to get to work on the assets, hahaha 😭

nealterrell:
Great job, Noah 👍 .

PR #3214, opened 2018-03-21
Added all-active-installed to differentiate between installed and active+installed runner cards. Fixes bugs with Degree Mill + Jua (and more!).

### LINKED ISSUE #3206: Can't pay for Degree Mill steal with facedown card
playing Kabonesa Wu vs Edtech
I had an installed Dhegdeer and a facedown installed card (Same old thing)
Accessed a Degree Mill from R&D.
"Would you like to shuffled two installed cards into your deck to steal", prompt came up and answered yes, but it said that I couldn't pay for it and denied the steal.
I think facedown cards do count as installed cards, so should have been able to steal.

nicohasa:
Changed all-installed to also count facedown cards. This introduced a bug in Apocalypse where hosted cards got facedowned twice, so had to change apocalypse slightly. Also made a test for facedown cards for Degree Mill.

Closes #3206

### DIFF
--- a/test/clj/game_test/cards/agendas.clj
+++ b/test/clj/game_test/cards/agendas.clj
@@ -258,8 +258,8 @@
 (deftest degree-mill
   ;; Degree Mill - runner must shuffle two installed cards into stack to steal
   (do-game
-    (new-game (default-corp [(qty "Degree Mill" 1)])
-              (default-runner [(qty "Ice Analyzer" 1) (qty "All-nighter" 1)]))
+    (new-game (default-corp [(qty "Degree Mill" 2)])
+              (default-runner [(qty "Ice Analyzer" 1) (qty "All-nighter" 1) (qty "Hunting Grounds" 1)]))
     (play-from-hand state :corp "Degree Mill" "New remote")
     (take-credits state :corp)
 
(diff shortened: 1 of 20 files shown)


test/clj/game_test/cards/agendas.clj
```diff
@@ -281,7 +281,36 @@
...
+        (prompt-choice :runner "Yes")
+        (prompt-select :runner fd1)
+        (prompt-select :runner fd2)
+        (is (= 6 (:agenda-point (get-runner))) "Runner failed to steal Degree Mill with facedown cards")
+        (is (empty? (get-in (get-runner)  [:rig :facedown])) "Degree Mill didn't remove facedown cards")
+        (is (= 2 (count (:deck (get-runner)))) "Degree Mill didn't put cards back in deck")
+        )
+                     )
+
+    ))
```
NoahTheDuke: Might want to clean up the final parentheses.

nicohasa:
Fixed, also added one for facedown runner cards.

NoahTheDuke:
I love `get-runner-facedown`. That's sweet. Every chance to reduce visual noise like that is good in my book.

JoelCFC25:
Very nice--thanks. Many moons ago I did some updates to `all-installed`...not sure how facedown cards were missed!

nealterrell:
I would feel better about this if someone could review all uses of `all-installed` to see if there will be issues with including facedown cards in the result. For example, the "start of turn" code looks at `all-installed` (IIRC) to see if any cards want to act in the window between start-of-turn and gaining clicks/mandatory draw. We don't want that code seeing a facedown Aesop's Pawnshop and prompting the Runner that they have effects they can trigger before they take their clicks.

nicohasa:
Very good point! Was planning on doing this at some point but then forgot about it.

What I wanna do is make an all-active by doing something like (remove :facedown all-installed) or something, and then swap to all-active where applicable. Don't have time to do this until after easter though, so if someone else wants to do it before then go ahead 😄

nicohasa:
Looked over the places all-installed are used, was planning on changing to all-active where it's needed. Sometimes this means the program'll check if the current is an icebreaker or a program or w/e, which isn't ideal. Was thinking of making all-active-installed or something, being only the active AND installed cards.

However, as far as I can see, all-active seems to fetch ALL installed corp-cards, not just rezzed cards. Is this an error or is there something else I'm missing?

nicohasa:
Introduced some errors in testing, don't have time to fix them now.

Would have been waaaay easier probably to just add facedown cards to what Degree Mill checked, but a function called all-installed should get all installed cards dangnabbit.

nicohasa:
Fix should totes work now, the added all-active-installed should be used on all the correct places.

There's still places where (filter (:rezzed %) (all-installed state :corp)) happens I think, even though you could use (all-active-installed state :corp) for that. This does not change behavior however.

Jua did not work previously if the corp chose a facedown card as one (or both) of the cards, this is now fixed as well. Jua searched for the card-str of the card to shuffle back in, and therefore searched for "a facedown card" in this case. Now it still shows "a facedown card" in chat, but runner is shown actual card name. See commit text for problems with this approach.

nealterrell:
Thanks for the thorough work!

JoelCFC25:
Something in this PR seems to have broken something that used to work--you can see it in my Heartbeat test in #3222 that was passing at the time I wrote it and is now failing. In testing, what's happening is that Apex installs a copy of Heartbeat facedown at the start of the turn, and then tries to install a copy the normal faceup way--and it won't do it, giving a toast about the uniqueness rule.

nealterrell:
Yeah I'm tracking that issue too... should have a fix.

nicohasa:
found the bug, will commit a fix soon-ish. Will just make sure I don't create more bugs by doing so.

Uniqueness uses the in-play? function, which uses installed-byname, which still uses all-installed over all-active installed. I was unsure which one to use in this case, but I thought I checked where installed-byname was used and only found Jua. Will make sure in-play? does not break somewhere else before I change installed-byname to use all-active-installed.

PR #3223, opened 2018-03-24
TDatD: 419: Amoral Scammer

jwarwick:


NoahTheDuke:
Damn, that's way more complicated than I expected. Nicely done!

jwarwick:
Thanks, there were certainly more cases than I expected. Luckily exposure prevention via `Underway Grid` (and others) was already handled.

And I'm sure I messed up `delayed-completion` related things, but that's just a given.

src/clj/game/cards/identities.clj
```diff
@@ -29,7 +29,46 @@
...
+                         {:optional
+                          {:prompt "Expose installed card unless Corp pays 1 [Credits]?"
+                           :player :runner
+                           :no-ability {:effect (req (clear-wait-prompt state :corp))}
+                           :yes-ability
+                           {:effect (req (clear-wait-prompt state :corp)
+                                         (if (not (can-pay? state :corp nil :credit 1))
+                                           (do
+                                             (toast state :corp "Cannot afford to pay 1 [Credits] to block card exposure" "info")
+                                             (expose state side itarget))
```
nealterrell: Since expose is a preventable (awaitable) effect, you should pass `eid` to `expose`, which will also mark the overall effect as completed in this branch.

src/clj/game/cards/identities.clj
```diff
@@ -29,7 +29,46 @@
...
+                                           (do
+                                             (show-wait-prompt state :runner "Corp decision")
+                                             (continue-ability
+                                               state side
+                                               {:optional
+                                                {:prompt "Pay 1 [Credits] to prevent exposure of installed card?"
+                                                 :player :corp
+                                                 :no-ability
+                                                 {:effect (req (expose state side itarget)
```
nealterrell: Same as before; this no-ability should be delayed completion and then pass eid to expose.

jwarwick:
@nealterrell - think I got the `delayed-completion` handling caught in all cases. I either pass an `eid` to another function or I call `effect-completed`

src/clj/game/cards/identities.clj
```diff
@@ -29,7 +29,49 @@
...
+              :req (req (first-event? state :corp :corp-install))
+              :effect
+              (req (show-wait-prompt state :corp "Runner to use 419: Amoral Scammer")
+                     (let [itarget target]
+                       (continue-ability
+                         state side
+                         {:optional
+                          {:prompt "Expose installed card unless Corp pays 1 [Credits]?"
+                           :player :runner
+                           :delayed-completion true
```
nealterrell: IIRC, this key is actually ignored in optional prompts. The eid of the optional itself is automatically passed to the no- and yes-abilities.

src/clj/game/cards/identities.clj
```diff
@@ -29,7 +29,49 @@
...
+                           :no-ability {:effect (req (clear-wait-prompt state :corp)
+                                                     (effect-completed state side eid))}
```
nealterrell: This should be unneeded. The no-ability will assume the parent's eid, and will be automatically triggered as completed because the no-ability is not delayed completion.

src/clj/game/cards/identities.clj
```diff
@@ -29,7 +29,49 @@
...
+                           :yes-ability
+                           {:effect (req (clear-wait-prompt state :corp)
```
nealterrell: This ability does need to be delayed-completion because of the expose/continue calls.

src/clj/game/cards/identities.clj
```diff
@@ -29,7 +29,49 @@
...
+                                             (toast state :corp "Cannot afford to pay 1 [Credits] to block card exposure" "info")
+                                             (expose state side eid itarget))
+                                           (do
+                                             (show-wait-prompt state :runner "Corp decision")
+                                             (continue-ability
+                                               state side
+                                               {:optional
+                                                {:prompt "Pay 1 [Credits] to prevent exposure of installed card?"
+                                                 :player :corp
+                                                 :delayed-completion true
```
nealterrell: Same comments as above.

src/clj/game/cards/identities.clj
```diff
@@ -29,7 +29,49 @@
...
+                                                 :no-ability
+                                                 {:effect (req (expose state side eid itarget)
```
nealterrell: This no-ability does need to be delayed-completion because of expose.

src/clj/game/cards/identities.clj
```diff
@@ -29,7 +29,49 @@
...
+                                                               (clear-wait-prompt state :runner))}
+                                                 :yes-ability
+                                                 {:effect (req (lose state :corp :credit 1)
```
nealterrell: This does not need delayed completion, nor the effect-completed later.

PR #3229, opened 2018-03-25
add False Flag

Giesch:
going for more rspec/jasmine style tests

### DIFF
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -528,6 +528,21 @@
                  :msg (msg "remove " (:advance-counter card) " bad publicity")
                  :effect (effect (trash card) (lose :bad-publicity (:advance-counter card)))}]}
 
+   "False Flag"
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
I like the look of that test _a lot_. I'm gonna have to step up my game dramatically to match!

Have you seen the [are macro](https://clojure.github.io/clojure/clojure.test-api.html#clojure.test/are)? It's like `is`, but works across a bunch of possibilities. I don't know if it'd help at all, but I thought of it when I saw your final "when the Runner accesses False Flag" `letfn`.

Giesch:
Thanks!
I'm happy to go with whatever version people think is more readable. I think I'd rather write out the cases than use the macro though, if it breaks line numbers in reporting.

One thing I didn't expect is the formatting. A failed test looks like:
```
top_context sub_context
output
```
rather than:
```
top_context
    sub_context
        output
```

nealterrell:
Whoops wrong button. Good job!

PR #3231, opened 2018-03-25
add Market Forces

Giesch:
I did named parameters in the examples on this one.

Is there a nice way to avoid calling credit-diff twice in the effect here? It seems like it would need a new macro, but idk if I'm missing something.

### DIFF
--- a/src/clj/game/cards/operations.clj
+++ b/src/clj/game/cards/operations.clj
@@ -735,6 +735,16 @@
                                       :delayed-completion true
                                       :effect (effect (tag-runner :runner eid 1))}}}}
 
+   "Market Forces"
+   (letfn [(credit-diff [runner]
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
May define a let instead of the letfn, where you do the calculation?

Giesch:
Yeah, what I mean is doing something like what i did in msg in effect. But if you try it it blows up because the effect macro tries to turn the let binding into game engine 5-fn.
I think we'd need a new effect-let macro, but at the same time, counting tags an extra time probably isn't going to bring down the server i guess.

Saintis:
In that case you’ll want to replace the effect macro with a req macro. Then it should not blow up (you’ll have to manually add the state and side parameters to the inner functions though)

Giesch:
Thanks! I should read the wiki more.

PR #3232, opened 2018-03-26
Fixes uniqueness-issues with facedown cards

nicohasa:
changes installed-byname to check only faceup cards, which is again used by in-play? which is used to check if another card is installed when checking for uniqueness.

I have checked where in-play? and installed-byname is used, and this should be the correct behaviour as far as I can see.

Mentioned in #3214, should make the heartbeat test in #3222 work again.


nealterrell:
This is a better fix than my hack in 065e25ca96132a6ec477db472b2e8e0be94c83b5.

nealterrell:
Speaking of, could you revert my change?

PR #3250, opened 2018-03-31
fixes, update to card-reset, update to pay to have blocked payment methods

### LINKED ISSUE #3193: Tinkering + Ankusa bug (Tinkering counter does not clear at end of turn if in grip)
If Tinkering is played on a piece of ICE that is bounced by Ankusa, it doesn't clear the Tinkering counter at the end of the turn and still has it if it is reinstalled.  It disappears only if installed at the end of turn.

danhut:


danhut:
@NoahTheDuke Given unpreventable will be nil I don't think making it false matters as it is equivalent in this case.

NoahTheDuke:
Okay, cool. Good to know!

PR #3251, opened 2018-04-02
Updated fetch task to optionally use local json files

jwarwick:
Added a `--local <path>` option to the `lein fetch` script to use a local copy of the `netrunner-cards-json` project (https://github.com/Alsciende/netrunner-cards-json).


Saintis:
Looks useful, but I don't understand that bit of the code so I'd suggest that Neal has a look 😄

nealterrell:
@jwarwick wrote this code in the first place so I'm sure it's correct 😄

PR #3252, opened 2018-04-02
TDatD: Knobkierie

jwarwick:


src/clj/game/cards/hardware.clj
```diff
@@ -437,6 +437,26 @@
...
+              :optional
+              {:prompt "Place a virus counter?"
+               :yes-ability
+               {:delayed-completion true
+                :effect (effect (continue-ability
+                                  {:prompt "Select an installed virus program"
+                                   :choices {:req #(and (installed? %)
+                                                        (has-subtype? % "Virus")
+                                                        (is-type? % "Program"))}
+                                   :msg (msg "places 1 virus counter on " (:title target))
```
JoelCFC25: You want "place" singular so you don't get "Player uses Knobkierie to places 1 virus..."

src/clj/game/cards/hardware.clj
```diff
@@ -437,6 +437,26 @@
    "HQ Interface"
    {:in-play [:hq-access 1]}
 
+   "Knobkierie"
+   {:implementation "MU usage restriction not enforced"
+    :in-play [:memory 3]
+    :events {:successful-run
+             {:delayed-completion true
```
nealterrell: This tag is ignored when it's the same map as an `:optional`, `:psi`, or `:trace`.

jwarwick:
@nealterrell good catch, best to remove it otherwise I'll copy it into something else in the future and be confused as to why its not working

PR #3254, opened 2018-04-03
TDatD: Endless EULA

msbeck:
I used Pup as an archetype for Endless EULA because it has subroutines which do not fire if the runner pays a credit.

```
   "Pup"
   {:subroutines [(do-net-damage 1)]
    :runner-abilities [(runner-break [:credit 1] 1)]}
```

### DIFF
--- a/src/clj/game/cards/ice.clj
+++ b/src/clj/game/cards/ice.clj
@@ -724,8 +724,10 @@
     :runner-abilities [(runner-break [:click 2] 2)]}
 
    "Endless EULA"
-   {:subroutines [end-the-run]
-    :runner-abilities [(runner-break [:credit 1] 1)]}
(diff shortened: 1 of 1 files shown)


src/clj/game/cards/ice.clj
```diff
@@ -723,6 +723,12 @@
                   end-the-run]
     :runner-abilities [(runner-break [:click 2] 2)]}
 
+   "Endless EULA"
+   {:implementation "Encounter effect is manual. Runner choice is not implemented"
```
NoahTheDuke: Maybe reword as, "Runner choice is manual."

msbeck:
Yes, I can update the text to more appropriately reflect the runner choice. I pulled this from Data Raven which is a different occasion.

msbeck:
I just played a game where it said Endless EULA was "unimplemented"

PR #3255, opened 2018-04-03
TDatD: Bio Vault

jwarwick:


src/clj/game/cards/upgrades.clj
```diff
@@ -103,6 +103,21 @@
...
+   :abilities [{:label "[Trash], 2 hosted advancement tokens: End the run."
+                :advance-counter-cost 2
+                :req (req (:run @state))
+                :msg "end the run. Bio Vault is trashed"
+                :delayed-completion true
+                :effect (effect (continue-ability
+                                  {:delayed-completion true
+                                   :effect (effect
+                                             (end-run)
+                                             (trash eid card {:cause :ability-cost}))}
```
nealterrell: I think this can all be in the root `:effect`, without the continue-ability.

jwarwick: @nealterrell not quite sure why I nested it like that in the first place. I think I must have thought that the `eid` being passed to `trash` needed a `continue-` or `resolve-ability`.

nealterrell:
I shortened the label, as we don't explicitly list counter costs in other cards.

PR #3256, opened 2018-04-03
TDatD: Sandman

jwarwick:


src/clj/game/cards/ice.clj
```diff
@@ -1604,6 +1604,18 @@
                                        #(assoc % :position (count (get-in corp (conj dest :ices)))
                                                  :server (rest dest)))))}]}
 
+   "Sandman"
+   {:subroutines [{:label "Add an installed Runner card to the grip"
+                   :req (req (pos? (count (all-installed state :runner))))
```
nealterrell: I like `not-empty` here instead.

PR #3257, opened 2018-04-03
Because I Can

presheaf:
Card implementation is very plain, but should work with redirection. PR also includes a bugfix (and test) for Sand Storm.


src/clj/game/cards/ice.clj
```diff
@@ -1599,10 +1599,10 @@
                    :choices (req (cancellable servers))
                    :msg (msg "move Sand Storm and the run.  The Runner is now running on " target ". Sand Storm is trashed")
                    :effect (req (let [dest (server->zone state target)]
-                                (trash state side card {:unpreventable true})
-                                (swap! state update-in [:run]
-                                       #(assoc % :position (count (get-in corp (conj dest :ices)))
-                                                 :server (rest dest)))))}]}
+                                  (swap! state update-in [:run]
+                                         #(assoc % :position (count (get-in corp (conj dest :ices)))
```
nealterrell: Let's clean this a little while we're at it. Try `(swap! state assoc-in [:run :position] (count (...))`.

presheaf: Maybe I'm misunderstanding, but that doesn't seem like it would be cleaner. Both `:position` and `:server` need to be modified, so you'd need two calls to assoc-in. One option might be to drop the lambda function and just have `(swap! state update-in [:run] :position (count ...) :server (rest dest)`, though. Would that be okay?

nealterrell: You're right, I misread the code. Carry on!

PR #3258, opened 2018-04-04
Implemented Amani Senai

presheaf:
Ignore branch name typo. Implementation allows corp to trigger trace even if there are no installed Runner cards on score/steal, which is intentional. I was unsure if anything here needed to be delayed, but I think maybe not? Or do traces need to?

### DIFF
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -128,6 +128,19 @@
                  :effect (effect (trash card))
                  :msg (msg "swap " (:advance-counter card 0) " cards in HQ and Archives")}]}
 
+   "Amani Sendai"
+   (let [get-last-stolen-pts (fn [state] (:agendapoints (last (get-in @state [:runner :scored]))))
+         get-last-scored-pts (fn [state] (:agendapoints (last (get-in @state [:corp :scored]))))
+         sendai-ability (fn [trace-base-func]
+                          {:optional {:prompt "Trace with Amani Sendai?" :player :corp
+                                      :yes-ability {:trace {:base (req (trace-base-func state))
+                                                            :choices {:req #(and (installed? %)
+                                                                                 (card-is? % :side :runner))}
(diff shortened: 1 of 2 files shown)


test/clj/game_test/cards/assets.clj
```diff
@@ -79,6 +79,42 @@
       (card-ability state :corp alix 0)
       (is (= 8 (get-in @state [:corp :credit]))))) "Gain 4 credits from Alix")
 
+(deftest amani-senai
+  ;; Team Sponsorship - Install from HQ
```
nealterrell: Wat :wink:

src/clj/game/cards/assets.clj
```diff
@@ -128,6 +128,19 @@
                  :effect (effect (trash card))
                  :msg (msg "swap " (:advance-counter card 0) " cards in HQ and Archives")}]}
 
+   "Amani Senai"
+   (let [get-last-stolen-pts (fn [state] (:agendapoints (last (get-in @state [:runner :scored]))))
+         get-last-scored-pts (fn [state] (:agendapoints (last (get-in @state [:corp :scored]))))
+         senai-ability (fn [trace-base-func]
+                          {:optional {:prompt "Trace with Amani Senai?" :player :corp
```
nealterrell: This is indented one space too far. Not huge, but let's be consistent. Let's also put the `:player :corp` on its own line for readability.

test/clj/game_test/cards/assets.clj
presheaf: Haha, oops. :) You'll never guess which test I modified for this one.

presheaf:
Thanks for the kind words! I read the card closer, though, and discovered that I'd mistaken advancement cost for printed points. Oops. This frees us from worrying about effects like The Board, but means we must worry about effects like Medical Breakthrough and NAPD. I replaced the test to at least look at Med Breakthrough, but should maybe also look at NAPD and/or The Source at some later date.

PR #3261, opened 2018-04-04
TDatD: Tempus

jwarwick:


src/clj/game/cards/upgrades.clj
```diff
@@ -804,6 +804,38 @@
                                                              (rez target))}}}
                             card nil))}}}
 
+   "Tempus"
+   {:access {:req (req (not= (first (:zone card)) :discard))
+             :interactive (req true)
+             :effect (req (when (= (first (:zone card)) :deck)
```
nealterrell: Did you base this on another card? I'm surprised it does anything in the presence of the `:trace` key.

jwarwick: Yup, I think that was from `Forced Connection`. Otherwise you just get the `accessed an unknown card` and then the trace prompt. It's probably exploiting a bug...

PR #3266, opened 2018-04-05
tDatD: Rogue Trading

msbeck:
Used Armitage Codebusting as the card archetype for this. Prisec had an example of a tag provided to the runner within a `req`.

screen shot 2018-04-05 at 1 00 58 pm
screen shot 2018-04-05 at 1 01 05 pm
screen shot 2018-04-05 at 1 01 15 pm


NoahTheDuke:
Looks sweet! Feel like whipping up a test or two? If not, I can do it.

msbeck:
If you would make a test and then walk me through it, I'd love that! I've never done test dev before.

PR #3268, opened 2018-04-05
kibit fixes and safe-zero

NoahTheDuke:
[kibit](https://github.com/jonase/kibit/) is a really cool linting and idiomatic clojure tool, and I've been using it on my code. Seems like it might be helpful to upstream some of the suggestions!

This also adds ~`s-zero?`~ `safe-zero?`, a safe version of the built-in `zero?`, meaning you can check `(safe-zero? nil)` and it will return false as expected.


src/clj/game/core/abilities.clj
```diff
@@ -15,12 +15,12 @@
...
-     (let [should-trigger? (partial should-trigger? state side eid card targets)]
+     (let [s-t? (partial should-trigger? state side eid card targets)]
        (cond
          req (req state side eid card targets)
-         optional (should-trigger? optional)
-         psi (should-trigger? psi)
-         trace (should-trigger? trace)
+         optional (s-t? optional)
+         psi (s-t? psi)
+         trace (s-t? trace)
```
Saintis: I'd rather the partial `should-trigger` be called `partial-should-trigger?` than `s-t?` since the latter give no information what it means (even though it should be clear from the let).

src/clj/game/core/abilities.clj
```diff
@@ -247,17 +247,17 @@
 (defn- print-msg
   "Prints the ability message"
   [state side {:keys [eid] :as ability} card targets cost-str]
-  (when-let [msg (:msg ability)]
-    (when-let [desc (if (string? msg) msg (msg state side eid card targets))]
+  (when-let [a-msg (:msg ability)]
+    (when-let [desc (if (string? a-msg) a-msg (a-msg state side eid card targets))]
```
Saintis: I'd also prefer if this was `ability-msg`

src/clj/game/core/abilities.clj
```diff
@@ -247,17 +247,17 @@
...
                   (str (build-spend-msg cost-str "use")
                        (:title card) (when desc (str " to " desc)))))))
 
 (defn- do-effect
   "Trigger the effect"
   [state side {:keys [eid] :as ability} card targets]
-  (when-let [effect (:effect ability)]
-    (effect state side eid card targets)))
+  (when-let [a-effect (:effect ability)]
+    (a-effect state side eid card targets)))
```
Saintis: And if this was `ability-effect`

src/clj/game/core/abilities.clj
```diff
@@ -279,9 +279,9 @@
 ;;; Optional Ability
 (defn optional-ability
   "Shows a 'Yes/No' prompt and resolves the given ability if Yes is chosen."
-  ([state side card msg ability targets] (optional-ability state side (make-eid state) card msg ability targets))
-  ([state side eid card msg ability targets]
-   (show-prompt state side eid card msg ["Yes" "No"]
+  ([state side card a-msg ability targets] (optional-ability state side (make-eid state) card a-msg ability targets))
+  ([state side eid card a-msg ability targets]
+   (show-prompt state side eid card a-msg ["Yes" "No"]
```
Saintis: I don't remember what `msg` is here, if it's a 5-fn I'd call it `msg-fn` otherwise I'd call it `message`

src/clj/game/core/actions.clj
```diff
@@ -163,7 +163,7 @@
   (let [card (get-card state card)
         r (get-in @state [side :selected 0 :req])
         cid (get-in @state [side :selected 0 :not-self])]
-    (when (and (not= (:cid card) cid) (or (not r) (r card))
+    (when (and (not= (:cid card) cid) (or (not r) (r card)))
```
Saintis: How has this not caused heaps of issues??
Ah, probably lazy-evaluation. At least it's clearer this way

src/clj/game/core/actions.clj
```diff
@@ -205,7 +205,7 @@
   "Use the 'match strength with ice' function of icebreakers."
   [state side args]
   (let [run (:run @state) card (get-card state (:card args))
-        current-ice (when (and run (> (or (:position run) 0) 0)) (get-card state ((get-run-ices state) (dec (:position run)))))
+        current-ice (when (and run (pos? (or (:position run) 0))) (get-card state ((get-run-ices state) (dec (:position run)))))
```
Saintis: you could also lose another nested bracket by using `(:position run 0)`.

src/clj/game/core/cards.clj
```diff
@@ -52,7 +52,7 @@
       (update-hosted! state side card)
       (let [z (cons (to-keyword (or (get-scoring-owner state card) (:side card))) zone)
             [head tail] (split-with #(not= (:cid %) cid) (get-in @state z))]
-        (when-not (empty? tail)
+        (when (seq tail)
```
Saintis: See @nealterrell regarding a discussion why we choose to use not-empty instead of `seq`...

src/clj/game/core/costs.clj
```diff
@@ -19,19 +19,19 @@
 (defn toast-msg-helper
   "Creates a toast message for given cost and title if applicable"
   [state side cost]
-  (let [type (first cost)
+  (let [c-type (first cost)
```
Saintis: I'd call this `cost-type`

Saintis: Also, I wonder if we can use expansion here instead of `first` and `last`. So instead `[[cost-type amount] cost]`

src/clj/game/core/installing.clj
```diff
@@ -292,7 +292,7 @@
    Returns :req if card-def :req check fails
    !! NB: This should only be used in a check with `true?` as all return values are truthy"
   [state side card facedown]
-  (let [req (:req (card-def card))
+  (let [c-req (:req (card-def card))
```
Saintis: I'd name this `card-req`

src/clj/game/core/io.clj
```diff
@@ -194,7 +194,7 @@
         value (if-let [n (string->num (first args))] n 1)
         num   (if-let [n (-> args first (safe-split #"#") second string->num)] (dec n) 0)]
     (when (<= (count args) 2)
-      (if (= (first (first args)) \#)
+      (if (= (ffirst args) \#)
```
Saintis: Woah neat!

src/clj/game/core/io.clj
```diff
@@ -208,17 +208,17 @@
...
+          "/close-prompt" command-close-prompt
           "/counter"    #(command-counter %1 %2 args)
           "/credit"     #(swap! %1 assoc-in [%2 :credit] (max 0 value))
           "/deck"       #(toast %1 %2 "/deck number takes the format #n")
           "/discard"    #(toast %1 %2 "/discard number takes the format #n")
           "/discard-random" #(move %1 %2 (rand-nth (get-in @%1 [%2 :hand])) :discard)
           "/draw"       #(draw %1 %2 (max 0 value))
           "/end-run"    #(when (= %2 :corp) (end-run %1 %2))
-          "/error"      #(show-error-toast %1 %2)
+          "/error"      show-error-toast
```
Saintis: I do somewhat miss the symmetry / pattern of the old version... 😛

src/clj/game/core/rules.clj
```diff
@@ -34,8 +34,8 @@
                                          [:credit (:cost card)]))
            eid (if-not eid (make-eid state) eid)]
        ;; ensure the instant can be played
-       (if (and (if-let [req (:req cdef)]
-                  (req state side eid card targets) true) ; req is satisfied
+       (if (and (if-let [c-req (:req cdef)]
```
Saintis: I'd call this `cdef-req` or just `card-req`

src/clj/game/core/rules.clj
```diff
@@ -102,7 +102,7 @@
    (swap! state update-in [side :register] dissoc :most-recent-drawn) ;clear the most recent draw in case draw prevented
    (trigger-event state side (if (= side :corp) :pre-corp-draw :pre-runner-draw) n)
    (let [active-player (get-in @state [:active-player])
-         n (-> n (+ (or (get-in @state [:bonus :draw]) 0)))
+         n (+ n (or (get-in @state [:bonus :draw]) 0))
```
Saintis: Less nesting with `(get-in ... 0)`

src/clj/game/core/runs.clj
```diff
@@ -370,10 +370,10 @@
   title-fn: a function taking a card map being accessed and returning a string to print as the card's title, e.g.,
       'an unseen card from R&D' for an R&D run.
   already-accessed: a set of cards already accessed from this zone or its root."
-
+  [state c-zone label amount select-fn title-fn already-accessed]
   (let [get-root-content (fn [state]
-                           (filter #(not (contains? already-accessed %)) (get-in @state [:corp :servers zone :content])))
-        server-name (central->name zone)
+                           (filter #(not (contains? already-accessed %)) (get-in @state [:corp :servers c-zone :content])))
+        server-name (central->name c-zone)
```
Saintis: I'd call this `card-zone`, unless that's also a fn... in which case `c-zone` is acceptable I guess.

src/clj/game/core/runs.clj
```diff
@@ -724,7 +725,7 @@
 
 (defn jack-out-prevent
   [state side]
-  (swap! state update-in [:jack-out :jack-out-prevent] #(+ (or % 0) 1))
+  (swap! state update-in [:jack-out :jack-out-prevent] #(inc (or % 0)))
```
Saintis: Could use `fnil` here even

src/clj/game/utils.clj
```diff
@@ -10,6 +10,8 @@
 
 (defn abs [n] (max n (- n)))
 
+(defn s-zero? [n] ((fnil zero? 1) n))
```
Saintis: I'd call this `safe-zero?` since that's what it is.

test/clj/game_test/utils.clj
```diff
@@ -14,12 +14,9 @@
...
   (let [conn (mg/connect {:host "127.0.0.1" :port 27017})
         db (mg/get-db conn "netrunner")
-        cards (mc/find-maps db "cards")
-        ret (take 99999 cards)]
-    ;; Doing this to materialize the list. I'm sure there's a better way. The take above might be useless.
-    (count ret)
+        cards (doall (mc/find-maps db "cards"))]
     (mg/disconnect conn)
-    ret))
+    cards))
```
Saintis: Look at that, there was a better way! 😄

src/clj/game/core/abilities.clj
NoahTheDuke: I'll verify!

src/clj/game/core/actions.clj
NoahTheDuke: haha Yeah, I have no idea how this hasn't messed everything up.

src/clj/game/core/actions.clj
NoahTheDuke: Oh damn, if you can do that, there's _a lot_ of fixes available.

src/clj/game/core/io.clj
NoahTheDuke: Yeah, either way is fine by me for this one. Makes sense to be explicit, makes sense to curry. I have no preference either way!

src/clj/game/core/runs.clj
NoahTheDuke: How would you write it?

test/clj/game_test/cards/icebreakers.clj
```diff
@@ -272,45 +272,43 @@
...
-      (prompt-card :runner (find-card "Astrolabe" (:hand (get-runner))))
-      (is (empty? (:prompt (get-runner))) "No trash-prevention prompt for hardware")
-      (card-ability state :runner faust 1)
-      (prompt-card :runner (find-card "Gordian Blade" (:hand (get-runner))))
-      (is (empty? (:prompt (get-runner))) "No trash-prevention prompt for program")
-      (card-ability state :runner faust 1)
-      (prompt-card :runner (find-card "Armitage Codebusting" (:hand (get-runner))))
-      (is (empty? (:prompt (get-runner))) "No trash-prevention prompt for resource"))))
+(deftest faust
+  (testing "Basic test: Pump by discarding"
```
Saintis: Does this mean one cannot use `lein test :only` to test just the "basic test" here?

src/clj/game/core/io.clj
Saintis: @nealterrell  is the head honco these days. Any preference?

src/clj/game/core/abilities.clj
NoahTheDuke: It can be both. What do you suggest?

Saintis: Probably just `message` then. Or maybe `prompt-msg` since you use it with `show-prompt`.

src/clj/game/core/cards.clj
NoahTheDuke: What's up with this?

src/clj/game/core/costs.clj
NoahTheDuke: Note for others who read this and can't find anything about "expansion": it's called destructuring, and it's hella cool! I found the docs here: https://clojure.org/guides/destructuring

src/clj/game/core/runs.clj
NoahTheDuke: In this instance, it's just `:rd` or `:hq`, so I think `card-zone` works.

test/clj/game_test/cards/icebreakers.clj
NoahTheDuke: To test only the `testing` section of this test? Yeah, that's what these changes do. I've been refactoring the tests like this so that there's a one-to-one relationship between the cards and the test functions. The equivalent would be a single test class with multiple methods for a given class in Java.

I feel this is a positive for development, because it means you're testing _all_ functionality against a card when making changes instead of having to remember to test each interaction individually or running the whole card type test suite.

Saintis: Sounds good. I like the reasoning.

NoahTheDuke:
I'll get to the `a-msg` and `c-type`s in a bit! Thanks so much!

PR #3269, opened 2018-04-05
remove-once fix

NoahTheDuke:
Reverses `remove-once`, so that it works as expected: Removes the first element which is true for the given predicate. Previously, you had to wrap your predicate in a `(not= ...)` for it to work.


src/clj/game/utils.clj
```diff
@@ -266,6 +266,11 @@
 (defn get-server-type [zone]
   (or (#{:hq :rd :archives} zone) :remote))
 
+(defn get-cid
+  "Gets the cid of a given card"
+  [card]
+  (:cid (:card card)))
```
danhut: `(get-in card [:card :cid]` would be better?

NoahTheDuke: Good call!

PR #3270, opened 2018-04-05
tDatD: Nyashia

NoahTheDuke:
Another simple one. Tests could be beefed up, but I'm okay with it for now.


nealterrell:
If I'm not mistaken, this is charging a counter every time it fires, before the runner makes their decision.

NoahTheDuke:
Oh good call. Yep, need to move that into the prompt effect

nealterrell:
Looks great, Noah.

Issue #2968, opened 2017-12-22
Maw triggers automatically on un-trashable cards while Imp is installed


mikeprosser:
When accessing a card that can't be stolen or normally trashed (for example, an agenda that has been protected by Lakshmi Smartfabrics or an operation/ice out of a central), Maw triggers immediately on access trashing a card from HQ. The runner can then trigger Imp to trash the card forcing the corp to manually put the card back into HQ, and causing some confusion later if Maw should then resolve further into a turn.

Steps to reproduce
----
1. Have the corp install Lakshmi Smartfabrics, a second asset and a Project Vitruvius, with a second PV in HQ
2. Have the runner install Maw and Imp
3. Make a run on the Vitruvius
4. The corp rezzes Lakshmi, the asset and reveals PV from HQ
5. The runner accesses PV, maw triggers trashing a card from HQ, then the runner can Imp PV

danhut:
I think this needs handling like Mumbad Virtual Tour where if Imp or another "Slow-trash" card is installed we don't trigger the :no-steal event that makes Maw fire.  If the players chooses to NOT use IMP they would need to fire Maw manually.

Maybe... IMP could be changed to :optional so we know when a user has declined to use it?

NoahTheDuke:
Imp is written kinda funny, having come from the agendas. Maybe it's how all the programs are written? I rewrote it, but it doesn't quite work as it needs additional changes elsewhere to properly work:

```clojure
   {:data {:counter {:virus 2}}
    :events {:access {:delayed-completion true
                      :req (req (pos? (get-in card [:counter :virus] 0)))
                      :effect (effect (show-wait-prompt :corp "Runner to use Imp")
                                      (continue-ability
                                        {:optional {:prompt "Trash accessed card at no cost?"
                                                    :once :per-turn
                                                    :yes-ability {:msg "trash at no cost"
                                                                  :effect (effect (add-counter card :virus -1)
                                                                                  (trash-no-cost)
                                                                                  (clear-wait-prompt :corp))}
                                                    :no-ability {:effect (effect (clear-wait-prompt :corp))}}}
                                        card nil))}}}
```

NoahTheDuke:
Closed via #3271.

Issue #1844, opened 2016-08-03
Future Perfect and Imp Weird Error


finnbar:
Okay, so I ran R&D with an Imp, and encountered a Future Perfect, which I Imped. On Imping it, the psigame happened, which was fine. I lost this psigame, and then another psigame triggered? And then I won that psigame, so Jinteki Personal Evolution fired. And then it fired again. And it kept firing. I have no idea how this happened, but it's slightly crazy.

Also, looking online, it seems that in a real game I'd be able to Imp it after failing the psigame rather than before. Is there any way to do this on Jinteki.net? Would it be possible to add a trash step for on-access Psi games? This sounds like a nightmare to implement though.

Stack Trace:
```clojure
#error {
 :cause nil
 :via
 [{:type java.lang.NullPointerException
   :message nil
   :at [clojure.lang.Numbers ops "Numbers.java" 1013]}]
 :trace
 [[clojure.lang.Numbers ops "Numbers.java" 1013]
  [clojure.lang.Numbers gt "Numbers.java" 229]
  [clojure.lang.Numbers gt "Numbers.java" 3864]
  [game.core$steal$fn__5034 invoke "core-runs.clj" 50]
  [game.core$effect_completed invokeStatic "core-events.clj" 224]
  [game.core$effect_completed invoke "core-events.clj" 220]
  [game.core$trigger_event_simult$fn__3902$fn__3903$fn__3904 invoke "core-events.clj" 177]
  [game.core$effect_completed invokeStatic "core-events.clj" 224]
  [game.core$effect_completed invoke "core-events.clj" 220]
  [game.core$trigger_event_simult_player$choose_handler__3837$fn__3846$fn__3847 invoke "core-events.clj" 96]
  [game.core$effect_completed invokeStatic "core-events.clj" 224]
  [game.core$effect_completed invoke "core-events.clj" 220]
  [game.core$resolve_damage$fn__4126 invoke "core-rules.clj" 174]
  [game.core$effect_completed invokeStatic "core-events.clj" 224]
  [game.core$effect_completed invoke "core-events.clj" 220]
  [game.core$trigger_event_sync$fn__3828 invoke "core-events.clj" 64]
  [game.core$effect_completed invokeStatic "core-events.clj" 224]
  [game.core$effect_completed invoke "core-events.clj" 220]
  [game.core$trigger_event_sync_next invokeStatic "core-events.clj" 52]
  [game.core$trigger_event_sync_next doInvoke "core-events.clj" 39]
[...]

PR #3271, opened 2018-04-05
Access Prompts and Imp Fix

NoahTheDuke:
Imp is written in an odd style, as an ability instead of an access trigger. I've rewritten it, and have discovered a heap of interesting and tough issues, namely #2968 but also #1844. 

This is my attempt to solve the issue, but we'll see how it goes. Imp works fine as it is, I just don't like that it's not triggered like more recent cards, lol. We'll see!

--

Update 2018-04-26: Decided to overhaul access prompts, as that made for a nicer experience. #yolo


src/clj/game/cards/programs.clj
```diff
@@ -435,10 +435,20 @@
    "Imp"
    {:flags {:slow-trash (req (pos? (get-in card [:counter :virus] 0)))}
     :data {:counter {:virus 2}}
-    :abilities [{:counter-cost [:virus 1]
-                 :msg "trash at no cost"
-                 :once :per-turn
-                 :effect (effect (trash-no-cost))}]}
+    :events {:access {:delayed-completion true
+                      :req (req (pos? (get-in card [:counter :virus] 0)))
```
danhut: The trailing 0/zero in get-in does nothing here.  Not a problem but not needed.

src/clj/game/cards/programs.clj
```diff
@@ -435,10 +435,20 @@
...
+                      :once :per-turn
```
danhut: I don't think this belongs here?  The once per turn relates to actual firing of the trash effect which is in the effect clause below.  This should be an additional :req check to ensure the trash has not been used yet.

test/clj/game_test/cards/events.clj
```diff
@@ -393,7 +393,9 @@
       (prompt-choice :runner "HQ")
       (is (= 4 (:rec-counter (find-card "Cold Read" (get-in @state [:runner :play-area])))) "Cold Read has 4 counters")
       (run-successful state)
-      (card-ability state :runner (get-program state 0) 0)
+      (prompt-choice :runner "Yes")
+      (prompt-choice :runner "Yes")
+      ; (card-ability state :runner (get-program state 0) 0)
```
danhut: Delete this comment if not needed now?

test/clj/game_test/cards/programs.clj
```diff
@@ -304,7 +304,8 @@
     (testing "Trash before access click"
       (run-empty-server state "HQ")
       ;; Should access TFP at this point
-      (card-ability state :runner (get-program state 0) 0)
+      (prompt-choice :runner "Yes")
+      ; (card-ability state :runner (get-program state 0) 0)
```
danhut: delete?

src/clj/game/cards/programs.clj
NoahTheDuke: The goal is to give a default value if the card returns `nil`. Will it not ever return `nil`?

test/clj/game_test/cards/events.clj
NoahTheDuke: I'd definitely remove it for the final PR. This was an "I'm annoyed and don't want to work on this at the moment." commit, thus commented code instead of flatly removed.

src/clj/game/cards/programs.clj
danhut: you are right - i retract this one ;)

NoahTheDuke:
Here's a bunch of images of the prompts as they currently exist. All wording is open to debate and will likely be adjusted, so please speak up now.

agenda
ambush
cant-steal-future-perfect
operation
steal-future-perfect
upgrade

Saintis:
I think I would prefer that the prompt title to just be "You accessed [card name]." And then each option be "Trash with Imp" (or "Imp ability"), "Pay X to trash", "Steal", "No action" etc.

NoahTheDuke:
That would certainly make the prompts much simpler. How do you feel about making all "OK"s to "No action" as well?

src/clj/game/core/runs.clj
```diff
@@ -195,47 +211,84 @@
...
+                                        (when-completed (pay-sync state side nil cost {:action :steal-cost})
+                                                        (do (system-msg
+                                                              state side
+                                                              (str "pays " cost-as-symbol " to steal " card-name))
+                                                            (resolve-steal state side eid c)))
+
+                                        ;; Don't pay single additional cost to steal
+                                        (.contains target "Don't")
+                                        (do (trigger-event state side :no-steal c)
+                                            (effect-completed state side eid))
```
nealterrell: I think we want to call `access-end` here instead of `effect-completed`.

src/clj/game/core/runs.clj
```diff
@@ -195,47 +211,84 @@
...
+                                        (.contains target "ability")
+                                        (let [idx (.indexOf ability-strs target)
+                                              trash-ab-card (nth trash-ab-cards idx)
+                                              cdef (-> (card-def trash-ab-card)
+                                                       :interactions
+                                                       :trash-ability
+                                                       (assoc :eid eid))]
+                                          (resolve-ability state side cdef trash-ab-card [c])
+                                          (trigger-event state side :no-steal c)
+                                          (effect-completed state side eid))))})
```
nealterrell: Don't need this, as the `eid` has been passed to the cdef you are resolving above.

src/clj/game/core/runs.clj
NoahTheDuke: Oh whoops, good call.

src/clj/game/core/runs.clj
NoahTheDuke: Ah, perfect. Thanks!

NoahTheDuke:
I'm sorry, life caught up with me once I got home on Monday. Should be ready for merging in an hour.

NoahTheDuke:
No worries, holmes. I know it can be tough to wait, and I did promise I'd have it out sooner than I delivered.

Issue #3219, opened 2018-03-23
The Devil and the Dragon Card Implementations


jwarwick:
Rumored to be released 5 April 2018 in the US.

Checked off means that the card is merged into `master`.
Place a link to the pull request next to the card when it is pushed.
If you wish to claim a card, make a comment on this issue or put your name next to the card.

- [x] [Glut Cipher](https://netrunnerdb.com/en/card/21061) (#3272)
- [x] [Knobkierie](https://netrunnerdb.com/en/card/21062) (#3252)
- [x] [419: Amoral Scammer](https://netrunnerdb.com/en/card/21063) (#3223)
- [x] [Falsified Credentials](https://netrunnerdb.com/en/card/21064) (#3006)
- [x] [Rogue Trading](https://netrunnerdb.com/en/card/21065) (#3266)
- [x] [Because I Can](https://netrunnerdb.com/en/card/21066) (#3257)
- [x] [Nyashia](https://netrunnerdb.com/en/card/21067) (#3270)
- [x] [Consume](https://netrunnerdb.com/en/card/21068) (#3263)
- [x] [Malia Z0L0K4](https://netrunnerdb.com/en/card/21069) (#3275)
- [x] [Kill Switch](https://netrunnerdb.com/en/card/21070) (#3265)
- [x] [Tempus](https://netrunnerdb.com/en/card/21071) (#3261)
- [x] [Bio Vault](https://netrunnerdb.com/en/card/21072) (#3255)
- [x] [Sadaka](https://netrunnerdb.com/en/card/21073) (#3262)
- [x] [Endless EULA](https://netrunnerdb.com/en/card/21074) (#3254)
- [x] [Sandman](https://netrunnerdb.com/en/card/21075) (#3256)
- [x] [Amani Senai](https://netrunnerdb.com/en/card/21076) (#3258)
- [x] [SSO Industries: Fueling Innovation](https://netrunnerdb.com/en/card/21077) (#3228)
- [x] [City Works Project](https://netrunnerdb.com/en/card/21078) (#3195)
- [x] [Oduduwa](https://netrunnerdb.com/en/card/21079) (#3210)
- [x] [Rashida Jaheem](https://netrunnerdb.com/en/card/21080) (#3267)

NoahTheDuke:
Yeah, they're not leaving us a lot of time to get these implemented, heh. We need them leaks!

msbeck:
[...]

PR #3272, opened 2018-04-05
Glut Cipher implementation

nicohasa:
For #3219

### DIFF
--- a/test/clj/game_test/core.clj
+++ b/test/clj/game_test/core.clj
@@ -111,6 +111,10 @@
   [state pos]
   (get-in @state [:runner :rig :facedown pos]))
 
+(defn get-discarded
+  [state side pos]
+  (get-in @state [side :discard pos]))
+
 (defn get-scored
   "Get a card from the score area. Can find by name or index.
   If no index or name provided, get the first scored agenda."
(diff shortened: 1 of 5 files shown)


test/clj/game_test/macros.clj
```diff
@@ -6,6 +6,7 @@
   `(let [~'state ~s
          ~'get-corp (fn [] (:corp @~'state))
          ~'get-runner (fn [] (:runner @~'state))
+         ~'get-run (fn [] [:run @~'state])
```
Saintis: I think this should be `()` and not `[]`.

test/clj/game_test/core.clj
```diff
@@ -111,6 +111,10 @@
   [state pos]
   (get-in @state [:runner :rig :facedown pos]))
 
+(defn get-discarded
+  [state side pos]
+  (get-in @state [side :discard pos]))
```
Saintis: I'd add a "default" value for pos as well for when you get the top discarded card.

nealterrell:
Does this  handle the UFAQ rule that using this on Archives with <5 cards skips the access phase entirely?

nicohasa:
Yep, it does. Will fix the other comments soon-ish.

PR #3279, opened 2018-04-06
Revealing Unseen Cards upon Firing Kill Switch

msbeck:
Added a message to reveal an agenda upon access per the UFAQ: http://ancur.wikia.com/wiki/The_Devil_and_the_Dragon_UFAQ
screen shot 2018-04-06 at 8 33 32 am

### DIFF
--- a/src/clj/game/cards/operations.clj
+++ b/src/clj/game/cards/operations.clj
@@ -679,7 +679,8 @@
    {:msg "gain 13 [Credits]" :effect (effect (gain :credit 13))}
 
    "Kill Switch"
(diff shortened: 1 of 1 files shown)


Saintis:
Actually, I would prefer "[Runner] must reveal they accessed [Agenda] due to Kill Switch". (makes it same wording as my R&D reveal PR)

msbeck:
Symmetry with the other RnD reveal text does seem favorable. This ticket was addressed before that came up.

nealterrell:
Agree w/ Saintis.

PR #3283, opened 2018-04-07
implement Undo - To Turn and Click

### LINKED ISSUE #1563: Feature Request: Undo Button
People make the odd mistake all the time or miss click and although we have the commands to rectify these mistakes it can take sometime and newer players don't necessarly know the / commands. So an undo button for the last action you took I feel would be a welcome addition to jinteki.net

### LINKED ISSUE #2117: create a console command to start turn or reset turn when it hangs
Would be useful to have  a start/reset turn console command when the game sometimes bugs out and you can't start a turn.

danhut:
This implements undo features
1/ Undo back to start of turn.  Has to be requested by both players with the command `undo-turn`.  Only keeps a single turn state so you can go only back to the start of the latest turn.

2/ Undo back to start of click using the command `undo-click`.  Only available to the player whose turn it is.  Only one click worth of state is saved.  Only allowed to rewind one click.  Tested to work ok with things like celebrity gift, same old thing, etc.  The saving of state takes place when the player spends 1 or more clicks.  

Both of these *could* be upgraded to save a stack of states and revert one by one if that was desirable.

Closes #1563
Closes #2117


Saintis:
Neat!
A start of click that is saved "before" every click is spent would also be neat.

Saintis:
Maybe should have a "confirm" prompt for the other player.

danhut:
@Saintis  thx -
1/ yes - i will add "undo-to-click" once we agree this approach or if the new atom is a better way
2/ I did something like a confirm when i added "reset-win" where both players have to send the command to make it work.  Trivial to add if we think it wise.

src/clj/web/game.clj
```diff
@@ -1,6 +1,6 @@
 (ns web.game
   (:require [web.ws :as ws]
-            [web.lobby :refer [all-games old-states] :as lobby ]
+            [web.lobby :refer [all-games old-states] :as lobby]
```
nealterrell: Hey that space was important to me!

danhut: Thanks Neal.   The other other way would be to store incremental diff but I think that is prone to a lot of issues. Or the data could be stored in app-state on the clients.

danhut: Would add - I can (dissoc) the :log key from the state before saving it - as you rightly say we don't want to over-write the chat log to we don't need to keep it.  We can grab it from the existing state atom.  This is probably a big chunk saved.  

Also the other big(-ish) key would usually be :turn-events which will be blank for this. 

Finally noticed a bit of cleaning up to be done from the stats implementation.  Game stats are kept in every message in the log right now.  Who the hell did that work?
       {:user {:_id "5a91b8d9f736882837e03f41",
               :username "wozzit",
               :emailhash "201c6d26d16065a4f678b8340c4625ec",
               :options {},
               :stats {:games-started 69,
                       :games-started-runner 68,
                       :games-completed 48,
                       :games-completed-runner 47,
                       :games-started-corp 1,
                       :games-completed-corp 1}},
        :text "sfd"}

danhut: This looks useful to see how much space the atom is using.  http://clojure-goes-fast.com/blog/introspection-tool-object-memory-meter/   Will play around on dev... my feel is the game-state is tiny compared to all the other stuff running on the JVM

danhut: Brand new game with no :turn-state added or any user chat:
```(clj-memory-meter.core/measure state)
=> "183.1 KB"```

With turn started, whole game-state grafted into the :turn-state key:
```(clj-memory-meter.core/measure state)
=> "187.8 KB"```

At end of corp turn 1 having taken 3 actions (install, play action, play action).  No user chat.  Noting this will add a bunch of data to the :turn-events and :log keys.
```(clj-memory-meter.core/measure state)
=> "195.1 KB"```

Runner hits Start Turn: - noting this includes :log from the previous turn still in this build.
```(clj-memory-meter.core/measure state)
=> "197.0 KB"```

One of the players says "I AM A PONY!!" x 10...
```(clj-memory-meter.core/measure state)
=> "199.4 KB"```

Runner does 4 actions then ends-turn:
```(clj-memory-meter.core/measure state)
=> "206.3 KB"```

Corp hits "Start turn"
```(clj-memory-meter.core/measure state)
=> "210.8 KB"```

And probably the most useful.  Reset! the atom at this point by removing the :turn-state data:
```(reset! state (dissoc @state :turn-state))

(clj-memory-meter.core/measure state)
=> "196.1 KB"```

So the diff is ~10-15% vs. current usage.  Likely a bit lower once we drop the 👍 log key

I am thinking to combine the turn undo and click undo - so we only allow one rollback.  Thoughts?

nealterrell: Right, the miracles of persistent hashmap data structures. Looking great. Full steam ahead!

danhut: Should be good to merge.  I could not yet find a reason to allow replaying back multiple turns but not hard to add if we want to. 

Undo-click will be via another PR.  Actually harder... as you can spend clicks in multiple ways in the engine.

nealterrell:
Sorry, one more request. Please remove your undo keys from the game state that is passed to the players. You'll find the relevant function named `strip` in `netrunner/main.clj`

danhut:
Will do. Thought of that last night too but had no laptop home. The review is v welcome always.

Issue #2871, opened 2017-10-13
Net damage fails to take effect after Titanium Ribs is turned facedown after Apocalypse


marc-jones:
The Runner had Earthrise Hotel, Daily Casts, Vigil, Eater, and Titanium Ribs installed. The Runner played Apocalypse, turning these five cards facedown. The Corp was Jinteki: Personal Evolution. Any net damage taken by the Runner, including the /take-net command, did not cause a card to be lost from hand. No error message was raised. The problem persevered even after the facedown Titanium Ribs was trashed by the Runner.

PesiTheWizard:
It's known that :leave-play clauses don't stop when the cards are turned face-down. Titanium Ribs uses such to stop working. If this happens again, try - if possible - to install Titanium Ribs again and trash it while it's face up.

JoelCFC25:
This all worked at one point in time, when it changed is anyone's guess. The (incorrect) order of effects seems to currently be: 

1. Apocalypse does a `move` on all Runner cards to `[:rig :facedown]`
2. `move` assocs `:facedown true` with the card being moved *before* any attempt to `deactivate`
3. `deactivate` tries `trigger-leave-effect`...which has logic to not operate on cards that are facedown

*** headdesk ***

NoahTheDuke:
Using git bisect, you should be able to narrow it down somewhat quickly.

jwarwick:
Fixed by #3284

PR #3284, opened 2018-04-07
Better trashing-in-bulk, better facedowning of cards, a lot of tiny bugs fixed

### LINKED ISSUE #3213: Apocalypse destroys host relations and does not keep counters on facedowned runner cards.
If you apocalypse with an Aumakua on Progenitor, the cards will be facedowned but the host relation will be destroyed and Aumakua will lose its counters.

nicohasa:
Changed move to better work with facedown cards
Now cards are deactivated before they are turned facedown, and
facedown cards will keep their host relations and counters.

Changed trash to split prevention into its own function
Now you can do prevention checks independently of trashing (good for next bit)

Changed trash-cards so it checks prevention before it trashes cards
Now we do all prevention-effects all at once before the trash starts, and
also inform the events correctly about how many cards are trashed at once.
Makes apocalypse behave with Hostile Infrastructure, makes All-Seeing I
not trash Jarogniew Mercs even if it's the last card installed, ++

Closes #3213, #2871, #3291 and possibly more???


src/clj/game/cards/events.clj
```diff
@@ -40,38 +40,40 @@
...
+                     (some #{:archives} (:successful-run runner-reg))))
+      :delayed-completion true
+      ;; trash cards from right to left
+      ;; otherwise, auto-killing servers would move the cards to the next server
+      ;; so they could no longer be trashed in the same loop
+      :msg "trash all installed Corp cards and turn all installed Runner cards facedown"
+      :effect (req (when-completed
+                     (resolve-ability state side corp-trash card nil)
+                     (continue-ability state side runner-facedown card nil)
+                     )
```
danhut: Great job on this.  Some rogue brackets here...

src/clj/game/core/rules.clj
```diff
@@ -423,21 +423,44 @@
...
 
+(defn trash
+  "Attempts to trash the given card, allowing for boosting/prevention effects."
+  ([state side card] (trash state side (make-eid state) card nil))
+  ([state side card args] (trash state side (make-eid state) card args))
+  ([state side eid {:keys [zone type] :as card} {:keys [unpreventable cause suppress-event] :as args}]
+   (swap! state update-in [:trash] dissoc :trash-list)
+   (when-completed (prevent-trash state side card args)
+                   (let [c (first (get-in @state [:trash :trash-list]))]
+                     (if c
```
danhut: This should probably be an `if-let`  so `(if-let [c (first (get-in @state [:trash :trash-list]))] `etc

src/clj/game/cards/events.clj
```diff
@@ -40,38 +40,37 @@
...
+                                                     (reverse))
+                                        allcorp (concat onhost unhosted)] 
+                                    (trash-cards state :runner eid allcorp)))}
+         runner-facedown {:effect (req (let [installedcards (all-active-installed state :runner)
+                                             ishosted (fn [c] (or (= ["onhost"] (get c :zone)) (= '(:onhost) (get c :zone))))
+                                             hostedcards (filter ishosted installedcards)
+                                             nonhostedcards (remove ishosted installedcards)]
+                                         (doseq [oc hostedcards :let [c (get-card state oc)
+                                                                      c (deactivate state side c true)
+                                                                      c (assoc-in c [:facedown] true)]]
```
nealterrell: Why is `move` unsuitable for these cards?

If `move` doesn't work, then It's probably best if these effects go into a function that can be reused, in case we get future effects that turn installed cards facedown.

src/clj/game/cards/events.clj
```diff
@@ -40,38 +40,37 @@
...
+                                           (update! state side c))
+                                         (doseq [oc nonhostedcards :let [c (get-card state oc)]]
+                                           (move state side c [:rig :facedown]))
+                                         (effect-completed state side eid)))}]
```
nealterrell: This is only necessary if the ability is `:delayed-completion true`, which this ability is not.

src/clj/game/cards/events.clj
```diff
@@ -40,38 +40,37 @@
...
+         runner-facedown {:effect (req (let [installedcards (all-active-installed state :runner)
+                                             ishosted (fn [c] (or (= ["onhost"] (get c :zone)) (= '(:onhost) (get c :zone))))
+                                             hostedcards (filter ishosted installedcards)
+                                             nonhostedcards (remove ishosted installedcards)]
+                                         (doseq [oc hostedcards :let [c (get-card state oc)
+                                                                      c (deactivate state side c true)
+                                                                      c (assoc-in c [:facedown] true)]]
+                                           (update! state side c))
+                                         (doseq [oc nonhostedcards :let [c (get-card state oc)]]
+                                           (move state side c [:rig :facedown]))
```
nealterrell: Does `move` handle deactivating these cards?

src/clj/game/core/installing.clj
```diff
@@ -12,8 +12,7 @@
   (let [c (dissoc card :current-strength :abilities :subroutines :runner-abilities :rezzed :special :new
                   :added-virus-counter :subtype-target :sifr-used :sifr-target)
         c (if keep-counter c (dissoc c :counter :rec-counter :advance-counter :extra-advance-counter))]
-    (if (and (= (:side c) "Runner") (not= (last (:zone c)) :facedown))
-      (dissoc c :installed :facedown :counter :rec-counter :pump :server-target) c)))
+    c))
```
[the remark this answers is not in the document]
nicohasa: This part was in dissoc-card basically to "fix" Apocalypse, as move used to put a card facedown before deactivating it. Now it's deactivated first, then moved facedown, so this is already fixed inside of 'move'.

(Specifically, disassociating :installed is done by checking if a card is moved to an "install"-zone in move, :facedown by checking if a card is moved to facedown, and :counter should be kept or not depending on keep-counter. :pump and :server-target should probably be disassociated as well in the main thing tbh.)

src/clj/game/cards/events.clj
nicohasa: These cards are hosted (and should remain so), move does not handle moving to hosted position.

Can make a `flip-facedown` function, that checks if a card is hosted, and either does this or moves it in the normal manner, np.

src/clj/game/cards/events.clj
nicohasa: Yep, move checks if a card should be deactivated (trashed, moved to grip, heap, turned facedown), and then deactivates it.

src/clj/game/cards/events.clj
nicohasa: yepyep you're right, trying to think if this should do some delayed access shenanigans, but at least atm nothing triggers on facedowning cards so shouldn't matter (also they're facedowned one at a time, which is not what they should be rules-wise so would have to do more work at that point anyway)

nealterrell:
Nice work!

PR #3286, opened 2018-04-07
Adds `:rd-reveal` flags to automate reveal-from-r&d.

### LINKED ISSUE #3278: rnd-reveal key engine enhancement
Many cards which must be revealed have explicit things that occur in response that make it obvious to both sides which card was revealed. Some do not.

For Snare! the runner sees "Waiting for corp..." and the corp gets asked if they want to trigger it.

The downside is that if the corp declines to play and the runner does not trash, nothing is logged in chat about the card interactions that occurred. See below:

screen shot 2018-04-06 at 9 10 02 am

We saw with Forced Connection https://github.com/mtgred/netrunner/issues/3238 that a trace will fire before the corp knows what card is accessed. All they saw was the trace prompt while the runner sees, "waiting for trace from Forced Connection". No card name was printed for the corp to hover in order review decision making points on the card's text.

This was handled as a one-off, but it sparked a discussion about handling things like this better. Perhaps for a new card property to be available for cards which have the "must reveal" text on them.

@Saintis has proposed adding `:rnd-reveal (req true)` as an option for adding to any cards which must be revealed. The engine would print its normal 
> Runner accesses an unseen card from R&D

followed immediately by
> runner must reveal {:title target}

@nealterrell would like this implementation to show up in `:flags`

@NoahTheDuke has an idea about mapping a list of standard responses that he will share with us shortly.

Saintis:
Fixes #3278.

Screenshot:
screen shot 2018-04-07 at 17 44 29
screen shot 2018-04-07 at 16 23 45


danhut:
very nice!

PR #3288, opened 2018-04-08
KA: Amina - Card Implemented

msbeck:
I used Ankusa as the archetype for this card. Simple icebreakers like this turned out to be easy. Just took out a few additional things Ankusa had and added the corp losing 1[c].

Unsure of what is typical - do we always refer to the Corp by name, or do we refer to it like the card says "the Corp"?


msbeck:
Not for implementation until pack release.

PR #3296, opened 2018-04-09
Catch Pheromones crash when firing ability with no virus counters

jwarwick:
Actually a bug in `core.actions.clj` when attempting to access unset `rec-counters`.


src/clj/game/core/actions.clj
```diff
@@ -192,9 +192,9 @@
         abilities (:abilities cdef)
         ab (if (= ability (count abilities))
              ;; recurring credit abilities are not in the :abilities map and are implicit
-             {:msg "take 1 [Recurring Credits]" :req (req (> (:rec-counter card) 0))
+             {:msg "take 1 [Recurring Credits]" :req (req (> (get card :rec-counter 0) 0))
```
NoahTheDuke: You could also do `(:rec-counter card 0)`, as invoking a `:keyword` has an optional parameter which is the default: https://clojure.org/reference/data_structures#Keywords

PR #3317, opened 2018-04-13
FIX: 419 - Ixodidae interaction

NoahTheDuke:
Changes `lose` to `pay`! Adds a test to handle the interaction.


NoahTheDuke:
If only all fixes were this easy.

PR #3320, opened 2018-04-13
Fix list item formatting in cardbrowser text

jwarwick:
Cards like `Deuces Wild` and `Reverse Infection` had their lists of options shifted too far to the left of the text box.

### DIFF
--- a/src/css/base.styl
+++ b/src/css/base.styl
@@ -824,6 +824,10 @@ nav ul
   .alt-art-selector
     margin-top: 10px
 
+  ul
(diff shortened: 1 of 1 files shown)


Saintis:
I'm sure this is good, but do you have any screenshots of before / after?

jwarwick:
Before:
screen shot 2018-04-14 at 4 16 44 pm

After:
screen shot 2018-04-14 at 4 17 27 pm

Saintis:
Much better looking! 💯

PR #3331, opened 2018-04-15
KA: Better Citizen Program

msbeck:
I used Ken as an archetype for "first time playing a run event", LLDS Processor for "installs an icebreaker", and Zealous Judge for giving a tag.

(Not intended for implementation until card release or for testing shortly beforehand.)

### DIFF
--- a/src/clj/game/cards/agendas.clj
+++ b/src/clj/game/cards/agendas.clj
@@ -204,6 +204,17 @@
                  :effect arrange-rd}
         :interactive (req true)}))
 
+   "Better Citizen Program"
+   {:events {:play-event {:req (req (has-subtype? target "Run"))
+                          :msg "give the Runner a tag for playing a run event"
+                          :effect (effect (tag-runner :runner eid 1))
+                          :once :per-turn}
+             :runner-install {:silent (req true)
+                              :req (req (has-subtype? target "Icebreaker"))
+                              :msg "give the Runner a tag for installing an icebreaker"
+                              :effect (effect (tag-runner :runner eid 1))
+                              :once :per-turn}}}
+
    "Bifrost Array"
    {:req (req (not (empty? (filter #(not= (:title %)
                                           "Bifrost Array")
--- a/src/clj/game/cards/agendas.clj
+++ b/src/clj/game/cards/agendas.clj
@@ -210,11 +210,11 @@
                           :msg "give the Runner a tag for playing a run event"
                           :effect (effect (tag-runner :runner eid 1))}
              :runner-install {:silent (req true)
-                              :req (req (has-subtype? target "Icebreaker"))
+                              :req (req (and (has-subtype? target "Icebreaker")
+                                             (first-event? state :runner :runner-install #(has-subtype? (first %) "Icebreaker"))))
                               :delayed-completion true
                               :msg "give the Runner a tag for installing an icebreaker"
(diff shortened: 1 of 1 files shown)


msbeck:
You mean to give them the tag after the run instead of before?

src/clj/game/cards/agendas.clj
```diff
@@ -204,6 +204,19 @@
...
+                          :msg "give the Runner a tag for playing a run event"
+                          :effect (effect (tag-runner :runner eid 1))
+                          :once :per-turn}
+             :runner-install {:silent (req true)
+                              :req (req (has-subtype? target "Icebreaker"))
+                              :delayed-completion true
+                              :msg "give the Runner a tag for installing an icebreaker"
+                              :effect (effect (tag-runner :runner eid 1))
+                              :once :per-turn}}}
+
```
nealterrell: I saw you asking on Slack about DRY'ing this. There's one thing we can do:

```clojure
(let [ability { the delayed-completion, effect, and once key/values go here }]
  {:events {:play-event (assoc ability :req (req (has-subtype ...))
                                       :msg "give the runner...")
            :runner-install (assoc ability :req (...)
                                           :msg ...
                                           :silent ...)})
```

but that arguably isn't much better.

danhut: Good idea though! - somehow eluded me in my fog yesterday

msbeck: @nealterrell would you like me to give your suggested change a shot?

Saintis: I think it's alright in its current less-DRY version. What you could do is make a function that takes `"Run"` or `"Icebreaker"` and generate the correct ability (with the particulat`req` and `msg`) from that.

msbeck:
Ok, I see what Neal is talking about now. I'll revisit this card after I get a few others functional.

@saintis
Do you have any card examples for your approach? It's good to learn new ways of tackling things like this.

nealterrell:
@msbeck this implementation is fine, I don't think there's anything to gain from my approach. Your code is perfectly readable.

src/clj/game/cards/agendas.clj
```diff
@@ -204,6 +204,19 @@
...
+                          :delayed-completion true
+                          :msg "give the Runner a tag for playing a run event"
+                          :effect (effect (tag-runner :runner eid 1))
+                          :once :per-turn}
+             :runner-install {:silent (req true)
+                              :req (req (has-subtype? target "Icebreaker"))
+                              :delayed-completion true
+                              :msg "give the Runner a tag for installing an icebreaker"
+                              :effect (effect (tag-runner :runner eid 1))
+                              :once :per-turn}}}
```
nealterrell: This should technically use `first-event` instead of `:once`, as it's possible that a runner could have already installed an icebreaker on the turn the corp scores BCP, then installs another during the end-of-turn window (which should not trigger BCP). The same would apply to `:play-event` if the Runner could play events on corp's turn, but alas...

NoahTheDuke:
Would you be willing to write a test for this?

src/clj/game/cards/agendas.clj
msbeck: swapped `{:events {:play-event {:req (req (has-subtype? target "Run"))`

with `{:events {:play-event {:req (req (first-event? state :runner :run))`

PR #3334, opened 2018-04-16
Folder structure

NoahTheDuke:
What if we were to store every card in it's own file? Here's an attempt to make it work!

Inspired by [Ringteki](https://github.com/gryffon/ringteki/), and pushed forward by Eatwood throwing up over the large card definitions.


NoahTheDuke:
I still think this is a great idea, and am in full support of it, but I'm going to close this PR for now as I'd rather focus my efforts on finalizing #3343 first. Once that's done, I'll restart this and get it within that set-up. Until then, pursuing this is valuable for exploration but not much else. Anyways, I don't like messy histories, so I'll just redo all this work next time much cleaner.

Saintis:
Great groundwork!

> Anyways, I don't like messy histories

interactive rebase is your friend 😉

PR #3338, opened 2018-04-17
Ka: Kasi String (COMPLETE)

msbeck:


src/clj/game/cards/resources.clj
```diff
@@ -920,6 +920,19 @@
                  :effect (req (gain state side :credit (get-in card [:counter :credit] 0))
                               (add-counter state side card :credit (- (get-in card [:counter :credit] 0))))}]}
 
+ "Kasi String"
+ {:implementation "Add counters manually for access outside of a run or cards that replace access like Ash"
```
NoahTheDuke: The effect only fires during a run. Why this note?

msbeck:
its a WIP. that was copied from turtle.

src/clj/game/cards/resources.clj
```diff
@@ -923,6 +923,17 @@
...
                               (add-counter state side card :credit (- (get-in card [:counter :credit] 0))))}]}
 
+ "Kasi String"
+ {:events {:run-ends {:once :per-turn
+                      :req (req (and (not (get-in @state [:run :did-steal]))
+                                     (get-in @state [:run :did-access])
+                                     (is-remote? (:server run))))
+                      :effect (effect (add-counter card :power 1))
+                      :msg "add a power counter to itself"}
+           :counter-added {:req (req (>= (get-in (get-card state card) [:counter :power]) 4))
```
nealterrell: This event triggers whenever ANY card gains a counter. That will cause a failure if another card gains a counter and KS has none (you'll be comparing `nil` vs 4). Add a default value of `0` to the `get-in` for safety, and even better, ensure that the `counter-added` event was triggered on this KS in the first place.

NoahTheDuke:
Would you be willing to write a test for this card?

src/clj/game/cards/resources.clj
msbeck: added default 0

msbeck:
@NoahTheDuke I need to step back from dev this week. I'm closing up a few loose ends and ducking out.

PR #3342, opened 2018-04-18
Better logging for invalid game messages

jwarwick:
Added the `gameid` to a few messages and better error handling around messages that do not seem intended for the current game. Need to catch an actual failure to start tracking down the root cause.


jwarwick:
This should wait until #3360 gets merged, since it fixes the root problem (merging state game state). Then this can layer on top to attach `gameids` to messages and be a bit smarter about looking up games.

nealterrell:
Can you check out the merge conflict?

PR #3349, opened 2018-04-20
Asset tests plus counter refactor initial work

NoahTheDuke:
Ya boi's back at it again, this time with tests for ASSETS! NetrunnerDB says there's 142 assets, vim says there's 85 `deftest`s, this shouldn't be too hard, lmao. FAMOUS LAST WORDS.

### Bug Fixes
* Bank Job
* Constellation Protocol
* Lily Lockwell
* News Team
* Shi.Kyu
* Urban Renewal

### Card Refactors:
* City Surveillance
* Constellation Protocol
* Contract Killer
* Corporate Town
* CPC Generator
* Ibrahim Salem
* Kuwinda K4H1U3
* Lily Lockwell
* Paige Piper
* Shattered Remains

### Card Rewrites:
* Clone Suffrage Movement
* Clyde Van Rite
* Test Ground

### Engine touches
* Quantified tag avoidance string, bad publicity string.

### Card Definition Notes:
* Added `{:cause :ability-cost}` to all necessary `(trash card)` calls in the assets (nothing else).
* Converted all counter accessing to new `get-counters` function, moved from test utils to core utils. Handles all counter types, defaults to 0 in every case.
* Converted all `(< 0 X)` to `(pos? X)`, and all `(= 0 X)` to `(zero? X)`.
* Generally cleaned up all assets. Moved every key-value pair to it's own line, reordered keys to be more logical, reformatted bits to make it read nicer, etc.

### Test Notes:
* Combined all tests for a given card into a single `deftest` with each subtest wrapped in a `testing` block.
* Converted all `= 0` to `zero?`, fixing where needed to handle `nil`. (Man, fuck `nil`.)
* Converted all `get-in @state [:X ...]` calls to the appropriate `(get-X)` call.
* Accidentally rewrote the Mark Yale test, cuz it was missing a blank line between tests. I like my test more, so I've kept it, lol.

Counter: 142/142


nealterrell:
Is this WIP?

NoahTheDuke:
Yep! Plan on covering all of the currently implemented cards. I'll remove WIP once it's ready.

NoahTheDuke:
I feel like I should explain that last commit.

It started off with me being annoyed by the `NPE`s that @jwarwick kept fixing, all these little niggling areas where we were passing `nil` when we just want `0`. Then, when reading through other tests to remind myself of good practices, I came across `deftest astro-script-token`, which contained the lines:
```clojure
(is (or (= nil (:advance-counter (refresh to)))
        (= 0 (:advance-counter (refresh to))))
```
and that pushed me over the edge. I had already played around with the idea of moving `:advance-counter` to `[:counter :advancement]` (along with `:rec-counter` to `[:counter :recurring]`), but that would require a serious effort which I didn't feel worth spending the effort for up front.

What I wanted to do was finish Asset Tests (this PR), but with a small QoL upgrade, namely, using `(get-counters card :X)` generically for all counter types. Moving it from `test/utils` to `core/utils` and expanding to handle the different kinds of counters wasn't that hard, surprisingly. Fixing all of the now-broken and incorrect tests was annoying, and then of course for consistency I had to change all counter accesses in the entire codebase, which required further test modifications, and now here we are.

Well, almost. Because I could rely on these things to be `nil`-safe, this let me fix all places where we were checking against `<= 1` instead of `pos?` and then because I'd already put in the work, why not fix all the places where we checked against `= 0` instead of `zero?`?

Along the way, I also reformatted various non-asset cards, cleaned up some cards, and overall tried to just leave the code looking better than it was when I started.

Lemme know what you think!

PR #3350, opened 2018-04-20
implement Standard Procedure

JoelCFC25:
Delayed completion only needed because of #%@#$# Accelerated Diagnostics.

### DIFF
--- a/src/clj/game/cards/operations.clj
+++ b/src/clj/game/cards/operations.clj
@@ -1358,15 +1358,13 @@
 
    "Standard Procedure"
    {:req (req (last-turn? state :runner :successful-run))
-    :delayed-completion true
     :prompt "Choose a card type"
     :choices ["Event" "Hardware" "Program" "Resource"]
     :effect (req (let [n (* 2 (count (filter #(is-type? % target) (:hand runner))))]
                    (gain state :corp :credit n)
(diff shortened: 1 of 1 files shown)


src/clj/game/cards/operations.clj
```diff
@@ -1356,6 +1356,18 @@
                  (shuffle! state side :deck)
                  (draw state side eid (count targets) nil))}
 
+   "Standard Procedure"
+   {:req (req (last-turn? state :runner :successful-run))
+    :delayed-completion true
```
nealterrell: This doesn't need delayed completion, because it does not trigger any awaitable actions. Did you try it with AccDiag and see an error?

src/clj/game/cards/operations.clj
```diff
@@ -1356,6 +1356,18 @@
...
+    :prompt "Choose a card type"
+    :choices ["Event" "Hardware" "Program" "Resource"]
+    :effect (req (let [n (* 2 (count (filter #(is-type? % target) (:hand runner))))]
+                   (gain state :corp :credit n)
+                   (system-msg state side (str "uses Standard Procedure to name " target ", reveal "
+                                               (join ", " (map :title (:hand runner)))
+                                               " in the Runner's Grip, and gain " n " [Credits]"))
+                   (effect-completed state side eid)))}
```
nealterrell: This won't be needed.

PR #3352, opened 2018-04-20
WiN: Contaminate

jwarwick:


nealterrell:
So... what's the interaction here with Hivemind? Do we need to account for Hivemind when selecting a target?

src/clj/game/cards/hardware.clj
```diff
@@ -399,8 +399,7 @@
    (let [ability {:msg (msg "move 1 virus counter to " (:title target))
                   :req (req (and (pos? (get-in card [:counter :virus] 0))
                                  (pos? (count-virus-programs state))))
-                  :choices {:req #(and (has-subtype? % "Virus")
-                                       (is-type? % "Program"))}
+                  :choices {:req #(is-virus-program? %)}
```
Saintis: Just `:req is-virus-program?` should be enough here.

src/clj/game/cards/events.clj
```diff
@@ -215,11 +215,14 @@
...
-    :choices {:req #(and (installed? %)
-                         (= (:side %) "Runner")
-                         (= 0 (get-in % [:counter :virus] 0)))}
-    :effect (req (add-counter state :runner target :virus 3))}
+   {:effect (req (resolve-ability
+                   state side
+                   {:msg (msg "place 3 virus tokens on " (:title target))
+                    :choices {:req #(and (installed? %)
+                                         (= (:side %) "Runner")
+                                         (= 0 (get-virus-counters state side %)))}
```
Saintis: This should be `(zero? (get-virus-counters state side %))` or `safe-zero?` if `get-virus-counters` can return `nil` instead of `0` (which I think it should not do).

src/clj/game/cards/events.clj
```diff
@@ -215,11 +215,14 @@
...
+                    :effect (req (add-counter state :runner target :virus 3))}
```
Saintis: Should be able to use `effect` here instead of `req` if you so fancy. I personally do not mind.

PR #3355, opened 2018-04-20
WiN: Viral Weaponization

jwarwick:


src/clj/game/cards/agendas.clj
```diff
@@ -1193,6 +1193,16 @@
     :msg "lose 2 bad publicity"
     :effect (effect (lose :bad-publicity 2))}
 
+   "Viral Weaponization"
+   {:effect (effect (register-events
+                      {:corp-turn-ends
+                       {:msg "to do 1 net damage for each card in the grip"
```
JoelCFC25: As in a previous instance drop the "to" in the `:msg` string--right now you'll get "Player uses Viral Weaponization to to do 1 net..."

jwarwick: Thanks. I read right through the extra `to` every time.

src/clj/game/cards/agendas.clj
```diff
@@ -1193,6 +1193,16 @@
     :msg "lose 2 bad publicity"
     :effect (effect (lose :bad-publicity 2))}
 
+   "Viral Weaponization"
+   {:effect (effect (register-events
+                      {:corp-turn-ends
+                       {:msg "do 1 net damage for each card in the grip"
+                        :effect (req (let [cnt (count (:hand runner))]
```
nealterrell: This needs to be delayed completion.

PR #3358, opened 2018-04-21
update monger with 3.1.0 fork to resolve compiler problem 

### LINKED ISSUE #3328: Compiler warnings
We currently have two compile time warnings due to symbol name conflicts brought in by new functions in clojure core 1.9.  Non impacting however should be dealt with.

`WARNING: any? already refers to: #'clojure.core/any? in namespace: monger.collection, being replaced by: #'monger.collection/any?`
This one is a small problem in Monger with Clojure 1.9.0 onwards. The problem is resolved in the master branch there - but is waiting on another fix to push the 3.2.0 version.    Could be resolved temporarily by having our own private fork until 3.2.0 is pushed.

`WARNING: boolean? already refers to: #'clojure.core/boolean? in namespace: clojure.core.typed.contract-utils, being replaced by: #'clojure.core.typed.contract-utils/boolean?`
I think this is another problem in the the Trello library we use.  Will see if I can resolve and then send an update to @jwarwick

danhut:
Updated to Monger to use a personal fork from the 3.1.0 package - with a fix for Clojure 1.9 compilation.  Excludes any? from collections.clj
`(:refer-clojure :exclude [find remove count drop distinct empty? any? update])`

Have subscribed to Monger for as and when they patch this.

Fix #3328

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -11,7 +11,7 @@
                  [cheshire "5.6.3"]
                  [org.omcljs/om "0.9.0"]
                  [sablono "0.3.4"]
-                 [com.novemberain/monger "3.1.0"]
+                 [danhut/monger "3.1.0"]
                  [org.clojure/core.match "0.3.0-alpha4"]
                  [differ "0.3.1"]
                  [com.taoensso/sente "1.11.0"]


NoahTheDuke:
Truly the hero we need.

PR #3359, opened 2018-04-21
WiN: Overseer Matrix

jwarwick:


src/clj/game/cards/upgrades.clj
```diff
@@ -653,6 +653,31 @@
...
+                                 :player :corp
+                                 :end-effect (effect (clear-wait-prompt :runner)
+                                                     (effect-completed eid))
+                                 :yes-ability {:cost [:credit 1]
+                                               :msg "give the Runner 1 tag"
+                                               :delayed-completion true
+                                               :effect (req (tag-runner state :runner eid 1))}}}
+                               card nil))}]
+     {:trash-effect
+      {:req (req (and (= :servers (first (:previous-zone card))) (:run @state)))
```
nealterrell: I don't believe a run is required for this effect, to account for trashing outside of runs.

src/clj/game/cards/upgrades.clj
```diff
@@ -653,6 +653,31 @@
...
+       :effect (effect (register-events {:runner-trash (assoc om :req (req (or (= (:zone (get-nested-host target)) (:previous-zone card))
```
nealterrell: Why does `in-same-server?` not work here?

src/clj/game/cards/upgrades.clj
jwarwick: I think I do need the run requirement here. This is for when Overseer Matrix itself is trashed and we want the effect to continue for the remainder of the run. If it's trashed outside of a run, then we don't want to register the event handler.

When a `runner-trash` event happens, I don't check for a run.

src/clj/game/cards/upgrades.clj
jwarwick: This code fires after Overseer Matrix has been trashed, so I think we need to look where it was, not where it is (Archives).

src/clj/game/cards/upgrades.clj
nealterrell: I forgot that `:runner-trash` triggers before the card moves zones, so we're good here.

PR #3360, opened 2018-04-21
Reset game state when starting a game

### LINKED ISSUE #3101: Game starts with a client thinking a run has initiated
Sometimes starting a game will have one client display a run has initiated, and the game will not start till that player uses the "/end-run" command. (To be specific, it hangs after players have accepted starting hands)

I'm not sure the exact sequence to reproduce, but I think it happens if you are spectating a game that ends mid-multi access and then you start a new game.

jwarwick:
When a new game was launched, any existing game state was merged into the new game state. This likely was causing the problem where games started with runs happening or win prompts being displayed.

When `Watching` a game, we were relying on the stale game state being set. Moved the state update message until after launching the game.

Fixes #3101


nealterrell:
I see. Good catch!

PR #3364, opened 2018-04-24
WIP: Access code rewrite

nealterrell:
Fix order of events triggered by the access code, and clean it up. Remove the old solution for Film Critic, and turn it into a normal ability handler that triggers on `:access` and, if  used, prevents other access handlers from resolving (thanks to an upgrade to `trigger-event-simult`). 

We no longer show "You must access X" prompts for agendas, which was only necessary for Film Critic to have a window for use.

I have not touched Imp, which has several failing tests. Could use some eyes on the run.clj changes before trying to finish up with Imp.


src/clj/game/core/runs.clj
```diff
@@ -313,41 +313,40 @@
...
-  "Apply game rules for accessing the given list of cards (which generally only contains 1 card.)"
-  ([state side cards] (handle-access state side (make-eid state) cards nil))
-  ([state side eid cards] (handle-access state side eid cards (:title (first cards))))
-  ([state side eid cards title]
-   ;; Indicate that we are in the access step.
+(defn access-card
+  "Apply game rules for accessing the given card."
+  ([state side card] (access-card state side (make-eid state) card nil))
+  ([state side eid card] (access-card state side eid card (:title card)))
+  ([state side eid c title]
```
NoahTheDuke: If we're no longer doing multiple cards, can we just call it `card` instead of `c`?

PR #3366, opened 2018-04-24
Leaking drawn card info to clients

jwarwick:
The `[:register :most-recent-drawn]` list of cards is sent to both Corp and Runner, allowing the other player to see which cards have been drawn by an opponent.

Stripping that info out of `:register` before sending state changes.

### DIFF
--- a/src/clj/game/main.clj
+++ b/src/clj/game/main.clj
@@ -52,7 +52,10 @@
    "close-deck" core/close-deck})
 
 (defn strip [state]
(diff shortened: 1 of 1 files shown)


nealterrell:
Does `:register` need to be sent at all? Does the UI use that key?

jwarwick:
@nealterrell sadly, the UI does use some register fields to determine if cards are playable (for instance). I debated pulling them out to separate fields, but that started to get a bit strange keeping state in two places.

Issue #3367, opened 2018-04-24
Because I Can (21066) shuffle effect should be optional


ben-albon:
"Make a run on a remote server. If successful, instead of accessing cards, you _may_ force the Corp to shuffle all cards in that server into R&D."

The "force the Corp to shuffle all cards installed in that server into R&D" effect should be optional - the runner should have the option, on succesful run, to not fire the effect but instead access cards normally.

The current implementation does not give the option to access - on succesful run, all installed cards in the remote are automatically shuffled into R&D.

NoahTheDuke:
Easy fix for someone: remove `:mandatory true`!

ben-albon:
@NoahTheDuke - Thanks!

If this looks good, and you don't think I've missed anything, I'll submit a pull request:

https://github.com/ben-albon/netrunner/commit/ac6480a5d615f66cfcb08270acc7464a86c437a3
https://github.com/ben-albon/netrunner/commit/afb11c7df6dfbb9d1e156cbab3bf5ad8e4d03ef2

Saintis:
This will be fixed once code is deployed. Good work!

PR #3368, opened 2018-04-24
Issue#3367 - Because I Can (21066) shuffle effect should be optional

ben-albon:
See https://github.com/mtgred/netrunner/issues/3367


NoahTheDuke:
Welcome to the fun!

PR #3369, opened 2018-04-24
Clean up optionals and delayed completion key from cards.

danhut:
Based on the recent Find the Truth wiki entry and a walk through the prompt code - this is a clean up of cards which are slightly wrong though the issue is non-impacting.

Removes :no-ability keys which only call effect-complete since this happens for free.
:delayed completion keys do nothing in this part of the structure so delete them too.

```clojure
{:optional
 :delayed-completion true
 {:prompt "say-something"
  :delayed-completion true}
```

clean up off NEXT WAVE 2

### DIFF
--- a/src/clj/game/cards/hardware.clj
+++ b/src/clj/game/cards/hardware.clj
@@ -437,14 +437,12 @@
                              (pos? (count-virus-programs state))))
               :optional
                    {:prompt "Place a virus counter?"
-                    :yes-ability {:effect (effect (continue-ability
-                                                    {:prompt "Select an installed virus program"
-                                                     :choices {:req #(and (installed? %)
-                                                                          (has-subtype? % "Virus")
-                                                                          (is-type? % "Program"))}
-                                                     :msg (msg "place 1 virus counter on " (:title target))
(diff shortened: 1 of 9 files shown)


src/clj/game/cards/hardware.clj
```diff
@@ -444,17 +441,15 @@
...
-                                  {:prompt "Select an installed virus program"
-                                   :choices {:req #(and (installed? %)
-                                                        (has-subtype? % "Virus")
-                                                        (is-type? % "Program"))}
-                                   :msg (msg "place 1 virus counter on " (:title target))
-                                   :effect (effect (add-counter target :virus 1))}
-                                  card nil))}}}}}
+                   {:prompt "Place a virus counter?"
+                    :yes-ability {:effect (effect (continue-ability
+                                                    {:prompt "Select an installed virus program"
```
nealterrell: Can this be pulled down one level directly into the `yes-ability`? So get rid of the continue-ability?

nealterrell: Did you miss a commit?

PR #3376, opened 2018-04-27
Hivemind updates strength of programs on change

jwarwick:
Made `Hivemind` update the `current-strength` of installed virus programs when it is added to play, leaves play, or gains a counter. This mostly servers to update the UI so the breaker strength of things like `Aumakua` is accurately displayed.


jwarwick:
Added fix for `Darwin`, since it had interaction with the `Hivemind` update (minimally). You will need to re-run `lein fetch` to update the database before the fix takes affect.

PR #3379, opened 2018-04-28
Hand size rewrite

Saintis:
Rewrites how hand size is handled. I intend to apply this approach to memory, tags and bad-pub. Could also be extended to credits as well (but that will be a right faff).

Very few non-code changes (you can now reduce your hand-size to below 0 with the +/- buttons).

Also renames `deduce` to `deduct` and adds two helper functions used with `gain`, `lose` and `deduct`.

Note: Test fails due to Imp + TFP.


Saintis:
One comment that I have in general is that currently all `:in-play` effects are changing the `:hand-size` modifier. So this could maybe be rewritten that if `:hand-size` does _not_ have a map it just changes the modifier (would make the card implementations a bit cleaner). Any preferences?

src/clj/game/utils.clj
```diff
@@ -15,6 +15,16 @@
   [n]
   ((fnil zero? 1) n))
 
+(defn safe-inc-n
+  "Helper function to safely update a value by n. Returns a function to use with `update` / `update-in`"
+  [n]
+  #(+ (or % 0) n))
```
jwarwick: Could do this: `(fnil + 0)` to be more Clojurey. Unless you are worried about things other than `nil`.

Saintis: Well, it's more whether I can still return it as a function then

jwarwick: `defn safe-inc-n [n] #(+ n ((fnil identity 0) %)))`. Much more readable...
I tried to work `partial` in there for bonus points.

Saintis: Yeah, not really more readable.. I tried partial but that broke something with Atlas...

NoahTheDuke:
There’s a lot here, but it overall looks really good.

src/clj/game/cards/hardware.clj
```diff
@@ -636,17 +636,17 @@
...
                            tagold (get-in old [:runner :tag])]
                        (when (> tagnew tagold)
-                         (gain state :runner :hand-size-modification (- tagnew tagold)))
+                         (gain state :runner :hand-size {:mod (- tagnew tagold)}))
                        (when (< tagnew tagold)
-                         (lose state :runner :hand-size-modification (- tagold tagnew)))))))
+                         (lose state :runner :hand-size {:mod (- tagold tagnew)}))))))
     :leave-play (req (remove-watch state :obelus)
-                     (lose state :runner :hand-size-modification (:tag runner)))
+                     (lose state :runner :hand-size {:mode (:tag runner)}))
```
JoelCFC25: Do you want this to be `:mod` instead of `:mode`?

PR #3380, opened 2018-04-28
Reimplements Endless EULA

Saintis:
Tests fail due to Imp + TFP


msbeck:
Can you add a second option for breaking all 6 at once? That was in one of my failed attempts to commit this ICE

Saintis:
Done! Credit to you for making the original code :)

Saintis:
I’ll let Msbeck submit this instead since it’s his work anyway :)

PR #3385, opened 2018-04-28
Embezzle

presheaf:
I think this could be cleaned up a bit, but it seems to work.

### DIFF
--- a/src/clj/game/cards/events.clj
+++ b/src/clj/game/cards/events.clj
@@ -461,7 +461,7 @@
    {:msg "gain 3 [Credits]" :effect (effect (gain :credit 3))}
 
    "Embezzle"
-   (letfn [(name-string [cards] (clojure.string/join " and " (map :title cards)))] ; either 'card' or 'card1 and card2'
+   (letfn [(name-string [cards] (join " and " (map :title cards)))] ; either 'card' or 'card1 and card2'
     {:req (req hq-runnable)
      :effect (effect
               (run :hq {:req (req (= target :hq))
@@ -469,8 +469,7 @@
                         {:mandatory true
                          :msg (msg "reveal 2 cards from HQ and trash all " target "s") ;should maybe lower-case target
                          :prompt "Choose a card type"
-                         ; Identity is included here because it's technically allowed and relevant with e.g. Hostile Infrasructure
-                         :choices ["Agenda" "Asset" "Upgrade" "Operation" "ICE" "Identity"] 
+                         :choices ["Asset" "Upgrade" "Operation" "ICE"]
                          :effect (req (let [chosen-type target
                                             cards-to-reveal (take 2 (shuffle (:hand corp)))
                                             cards-to-trash (filter #(is-type? % chosen-type) cards-to-reveal)]
--- a/src/clj/game/cards/events.clj
+++ b/src/clj/game/cards/events.clj
@@ -460,6 +460,29 @@
    "Easy Mark"
    {:msg "gain 3 [Credits]" :effect (effect (gain :credit 3))}
 
+   "Embezzle"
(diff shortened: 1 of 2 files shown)


presheaf:
Regarding the failed test, the one which fails (at least on my machine) is one with Imp, which also failed before these commits, so I don't think it has anything to do with Embezzle.

src/clj/game/cards/events.clj
```diff
@@ -460,6 +460,29 @@
...
+   (letfn [(name-string [cards] (clojure.string/join " and " (map :title cards)))] ; either 'card' or 'card1 and card2'
+    {:req (req hq-runnable)
+     :effect (effect
+              (run :hq {:req (req (= target :hq))
+                        :replace-access
+                        {:mandatory true
+                         :msg (msg "reveal 2 cards from HQ and trash all " target "s") ;should maybe lower-case target
+                         :prompt "Choose a card type"
+                         ; Identity is included here because it's technically allowed and relevant with e.g. Hostile Infrasructure
+                         :choices ["Agenda" "Asset" "Upgrade" "Operation" "ICE" "Identity"]
```
Saintis: Card only allows you to name "asset, ice, operation, or upgrade".

src/clj/game/cards/events.clj
```diff
@@ -460,6 +460,29 @@
    "Easy Mark"
    {:msg "gain 3 [Credits]" :effect (effect (gain :credit 3))}
 
+   "Embezzle"
+   (letfn [(name-string [cards] (clojure.string/join " and " (map :title cards)))] ; either 'card' or 'card1 and card2'
```
Saintis: Do we not already have a util function like this? Feels like it would be used with cards trashed to damage and other multi-card reveal effects.

test/clj/game_test/cards/events.clj
```diff
@@ -753,6 +753,19 @@
...
 
+(deftest embezzle
+  ;; Check that trashed cards are trashed face-up
+  (do-game
+    (new-game (default-corp [(qty "Project Atlas" 1)])
+              (default-runner [(qty "Embezzle" 1)]))
+    (take-credits state :corp)
+    (is (= 5 (:credit (get-runner))))
+    (play-run-event state (first (:hand (get-runner))) :hq)
+    (prompt-choice :runner "Agenda")
```
Saintis: Can't actually chose agenda.

src/clj/game/cards/events.clj
```diff
@@ -460,6 +460,29 @@
...
+                        {:mandatory true
+                         :msg (msg "reveal 2 cards from HQ and trash all " target "s") ;should maybe lower-case target
+                         :prompt "Choose a card type"
+                         ; Identity is included here because it's technically allowed and relevant with e.g. Hostile Infrasructure
+                         :choices ["Agenda" "Asset" "Upgrade" "Operation" "ICE" "Identity"] 
+                         :effect (req (let [chosen-type target
+                                            cards-to-reveal (take 2 (shuffle (:hand corp)))
+                                            cards-to-trash (filter #(is-type? % chosen-type) cards-to-reveal)]
+                                        (system-msg state side (str " reveals " (name-string cards-to-reveal) " from HQ"))
+                                        (when-not (empty? cards-to-trash)
```
Saintis: Maybe some sort of message if the runner doesn't trash cards / gains credits.

src/clj/game/cards/events.clj
Saintis: Also, do we not import string/join, so you could use it just as `join`?

src/clj/game/cards/events.clj
presheaf: Ah, that's a good point. I mentally just read that as 'a card type', thanks.

PR #3387, opened 2018-04-29
Compile

presheaf:
Could not find a clean way to make encounter effect only trigger when it should, so went with the same implementation as Diana's Hunt (card adds itself to temp. zone and notifies player that they should click it to trigger it.)

### DIFF
--- a/src/clj/game/cards/events.clj
+++ b/src/clj/game/cards/events.clj
@@ -378,6 +378,41 @@
      {:delayed-completion true
       :effect (effect (continue-ability (choice all) card nil))})
 
+   "Compile"
+   {:implementation "Trigger only on first encounter not enforced"
+    :prompt "Choose a server"
+    :msg "make a run and install a program on encounter with the first piece of ICE"
+    :choices (req runnable-servers)
+    :delayed-completion true
+    :abilities [{:label "Install a program using Compile"
+                 :delayed-completion true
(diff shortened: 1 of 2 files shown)


jwarwick:
Should we make the Runner chose `Stack` or `Heap`? Maybe just list all the programs in both locations. I guess you need to differentiate if you have a copy in each location.

presheaf:
The reaaon I made them choose was to avoid installing a card from the wrong place by accident. If you have some clean way of listing the cards along with their locations I could do that as well, but this was the best one I could come up with.

src/clj/game/cards/events.clj
```diff
@@ -378,6 +378,41 @@
...
+                                                                       (shuffle! state :runner :deck)))})
+                                                     card nil))}
+                                   card nil))}]
+    :effect (effect (run target nil card)
+                    (prompt! card (str "Click Compile in the Temporary Zone to install a Program") ["OK"] {})
+                    (resolve-ability
+                      {:effect (req (let [c (move state side (last (:discard runner)) :play-area)]
+                                      (card-init state side c {:resolve-effect false})))} card nil))
+    :events {:run-ends {:effect (req
+                                 (let [compile-installed (first (filter #(get-in % [:special :compile-installed]) (game.core/all-installed state :runner)))]
```
nealterrell: Indentation.

PR #3391, opened 2018-04-29
WIP - first cut - prevention upgrade

danhut:
This is an upgrade to the damage and tag prevention system.  The card-def now includes a :req check for a given type and subtype like so:
```clojure
{:prevent {:damage {:net {:req (req (first-chance? state side))}}
                :tag {:all {:req (req (first-chance? state side))}}}
```

For simple cards this will just be 
```clojure
(req true)
```

Three cards are shown so far in this WIP
- Feedback Filter
- No One Nome
- Recon Drone

Let me know concerns or something I am missing - I will migrate everything shortly otherwise.

### DIFF
--- a/src/clj/game/cards/resources.clj
+++ b/src/clj/game/cards/resources.clj
@@ -1,6 +1,6 @@
 (in-ns 'game.core)
 
-(declare close-access-prompt)
+(declare close-access-prompt event-count)
 
 (defn- genetics-trigger?
   "Returns true if Genetics card should trigger - does not work with Adjusted Chronotype"
@@ -1138,7 +1138,9 @@
     :abilities [{:cost [:credit 2] :msg "avoid 1 tag" :effect (effect (tag-prevent 1))}]}
 
    "No One Home"
-   (letfn [(start-trace [type]
+   (letfn [(first-chance? [state side] (and (< (event-count state side :pre-tag) 2)
+                                            (< (event-count state side :pre-damage) 2)))
+           (start-trace [type]
              (let [message (str "avoid any " (if (= type :net) "amount of net damage" "number of tags"))]
              {:player :corp
               :label (str "Trace 0 - if unsuccessful, " message)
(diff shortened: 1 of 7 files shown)


src/clj/game/cards/resources.clj
```diff
@@ -1138,7 +1138,9 @@
     :abilities [{:cost [:credit 2] :msg "avoid 1 tag" :effect (effect (tag-prevent 1))}]}
 
    "No One Home"
-   (letfn [(start-trace [type]
+   (letfn [(first-chance? [state side] (and (< (event-count state side :pre-tag) 2)
+                                            (< (event-count state side :pre-damage) 2)))
```
nealterrell: Should this be "the sum of pre-tag and pre-damage is less than 2"? Since it's first time for either of those events.

src/clj/game/core/rules.clj
```diff
@@ -228,33 +228,31 @@
...
+   (let [n (damage-count state side type n args)
+         prevent (get-in @state [:prevent :damage type])]
+     (if (and (not unpreventable) (cards-can-prevent? state side prevent :damage type))
+       ;; runner can prevent the damage.
+       (do (system-msg state :runner "has the option to avoid damage")
+           (show-wait-prompt state :corp "Runner to prevent damage" {:priority 10})
+           (swap! state assoc-in [:prevent :current] type)
+           (show-prompt state :runner nil (str "Prevent any of the " n " " (name type) " damage?") ["Done"]
+                        (fn [_]
+             (let [prevent (get-in @state [:damage :damage-prevent type])]
```
nealterrell: Indentation here is a little misleading.

src/clj/game/cards/resources.clj
danhut: Good catch - I think you have that right.  I spent a while banging my head on this one due to the pre-damage firing before the prevent.

* First damage fires and we get :pre-damage event which other cards use. So count of pre-damage is now 1.
* Then the prevention logic fires and we check in the No One Home req that the count of pre-damage events is less than 2 (ie. this is the first time).  Strictly I could check for exactly 1 but want with less than 2.

Now let's assume I get my first tag event after this
* First tag event happens and we get a :pre-tag event fire.  Count of :pre-tag = 1, count of :pre-damage
* No One Home will not pass the req check

OR
* I get my second damage effect
* Count of :pre-damage = 2 so NOH will not pass.

src/clj/game/core/rules.clj
danhut: will fix!

danhut:
Still under discussion - another revisit taking a future card format rework into account.
```clojure
{:interactions {:prevent-net {:req (req (first-chance? state side))}
                        :prevent-tag {:req (req (first-chance? state side))}}}
```

Noting that I did not intend to move the abilities themselves for the cards into this map - as you will see with recon drone we need two distinct :req checks…  If we want to move the abilities inside then I will need to rename the prevention truthy test to something other than :req (maybe :prevent?)

PR #3392, opened 2018-04-29
WiN: Jackpot!

jwarwick:
Attempted implementation of `Jackpot!` using the `card-moved` event.
Not working, possibly because the `move` ability isn't awaitable.


src/clj/game/cards/resources.clj
```diff
@@ -858,6 +859,31 @@
...
+                                   :no-ability {:effect (effect (clear-wait-prompt :corp)
+                                                                (effect-completed eid))}
+                                   :yes-ability
+                                   {:prompt "Choose how many [Credit] to take"
+                                    :choices {:number (req (get-in card [:counter :credit] 0))}
+                                    :delayed-completion true
+                                    :effect (req (gain state :runner :credit target)
+                                                 (system-msg state :runner (str "trashes Jackpot! to gain " target " credits"))
+                                                 (clear-wait-prompt state :corp)
+                                                 (trash state :runner eid card {:cause :ability-cost}))}}}
```
nealterrell: This effect is not the cost of a triggered ability; the trash is a side effect. `:ability-cost` is for "[Trash]: ..." abilities, notably to trigger Geist / Tech Trader.

src/clj/game/cards/resources.clj
```diff
@@ -858,6 +859,31 @@
...
+                                 card nil))}]
+     {:events
+      {:runner-turn-begins {:effect (effect (add-counter :runner card :credit 2))}
```
nealterrell: This is a pretty OP implementation: supposed to be only 1 credit added per turn 😉 .

src/clj/game/core/rules.clj
```diff
@@ -535,9 +535,11 @@
...
 (defn as-agenda
   "Adds the given card to the given side's :scored area as an agenda worth n points."
-  [state side card n]
-  (move state side (assoc (deactivate state side card) :agendapoints n) :scored)
-  (gain-agenda-point state side n))
+  ([state side card n] (as-agenda state side (make-eid state) card n))
+  ([state side eid card n]
+   (move state side (assoc (deactivate state side card) :agendapoints n) :scored)
+   (when-completed (trigger-event-sync state side eid :as-agenda (assoc card :as-agenda-side side :as-agenda-points n))
+                   (gain-agenda-point state side n))))
```
nealterrell: Same comment here. You'll need to call effect-completed by hand.

jwarwick:
@nealterrell good catch - I updated `Jackpot!` to only give you two credits per turn if you are a donor.

PR #3396, opened 2018-04-29
WiN: Gebrselassie

### LINKED ISSUE #3384: System Seizure only keeps the first strength boost on an icebreaker
If you run a server, boost a lady with +1 strength (now str 4), pass that ice, boost lady for +1 again (now 5 str), then lady will be reset to str 4 when passing the second ice.

Edit: It also does not work on the second run of any turn (even if you did not boost any icebreaker the first run).

nicohasa:
As written it works on the same trigger as System Seizure, so if you play both one of them will overwrite the other, depending on the order you play them. Atm there is no way to order events automatically it seems. I might look into that.

Also closes #3384

### DIFF
--- a/src/clj/game/cards/hardware.clj
+++ b/src/clj/game/cards/hardware.clj
@@ -415,13 +415,22 @@
                                                 (system-msg :runner (str "places 1 virus counter on Friday Chip")))}}}}})
 
    "Gebrselassie"
-   {:abilities [{:label "Host Gebrselassie on an installed non-AI icebreaker"
+   {:abilities [{:msg (msg "host it on an installed non-AI icebreaker")
                  :cost [:click 1] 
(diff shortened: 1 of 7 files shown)


src/clj/game/cards/hardware.clj
```diff
@@ -414,6 +414,24 @@
                                {:effect (effect (add-counter :runner card :virus 1)
                                                 (system-msg :runner (str "places 1 virus counter on Friday Chip")))}}}}})
 
+   "Gebrselassie"
+   {:abilities [{:msg (msg "host it on an installed non-AI icebreaker")
+                 :cost [:click 1] 
+                 :choices {:req #(and 
+                                   (installed? %) 
+                                   (has-subtype? % "Icebreaker") (not (has-subtype? % "AI")))}
```
nealterrell: My preferred style here is for the first function to go on the same line as the `and`, then the other two to be on their own lines.

PR #3397, opened 2018-04-29
WiN: Laamb

nicohasa:
The on-encounter is implemented as a paid ability.
Should not be able to combo with Paperclip, so this should not be a problem.

If you end up advancing a Changeling while still encountering it, weird things might happen, so please don't.

### DIFF
--- a/src/clj/game/cards/icebreakers.clj
+++ b/src/clj/game/cards/icebreakers.clj
@@ -614,6 +614,31 @@
                  :req (req (ice? (get-nested-host card)))
                  :msg "break 1 subroutine on the host ICE"}]}
 
+   "Laamb"
+   (auto-icebreaker ["Barrier"]
+                    (let [subtype "Barrier"]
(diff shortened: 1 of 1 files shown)


src/clj/game/cards/icebreakers.clj
```diff
@@ -614,6 +614,31 @@
...
 
+   "Laamb"
+   (auto-icebreaker ["Barrier"]
+                    (let [subtype "Barrier"]
+                      {:abilities [(break-sub 2 0 subtype)
+                                   (strength-pump 3 6)
+                                   {:once :per-turn
+                                    :cost [:credit 2]
+                                    :label (str "Turn currently encountered ice into " subtype)
+                                    :msg (msg "to turn currently encountered ice into " subtype)
```
nealterrell: Get rid of the "to", which is created automatically by the `msg`. You can also use `current-ice` inside this macro to print a better message. (No such luck for the `:label` above it.)

PR #3401, opened 2018-05-01
WiN: Remote Enforcement

presheaf:
Card implementation seems to work fine. However, the test fails on the line `(prompt-choice :corp "New remote")`, and I have no idea why. Testing out the test scenario manually, everything seems to go fine, so I'm fairly confident the card implementation is not at fault. Looking at the stack trace, it seems like at some point a string (one might guess the string "New remote") tries to be converted into a map and causes an exception, so my guess would be that the game expects a card as choice here, somehow, but I don't understand why. Any pointers would be appreciated!

### DIFF
--- a/src/clj/game/cards/agendas.clj
+++ b/src/clj/game/cards/agendas.clj
@@ -1032,9 +1032,8 @@
                                                                :choices (filter #(not (#{"HQ" "Archives" "R&D"} %))
                                                                                 (corp-install-list state chosen-ice))
                                                                :effect (effect
-                                                                        (corp-install chosen-ice target {:install-state :rezzed-no-rez-cost})
                                                                         (shuffle! :deck)
-                                                                        (effect-completed eid))}
+                                                                        (corp-install eid chosen-ice target {:install-state :rezzed-no-rez-cost}))}
                                                               card nil)))}}}
 
    "Research Grant"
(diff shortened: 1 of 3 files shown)


src/clj/game/cards/agendas.clj
```diff
@@ -1019,6 +1019,24 @@
...
+                             :effect (req (let [chosen-ice target]
+                                            (continue-ability state side
+                                                              {:delayed-completion true
+                                                               :prompt (str "Select a server to install " (:title chosen-ice) " on")
+                                                               :choices (filter #(not (#{"HQ" "Archives" "R&D"} %))
+                                                                                (corp-install-list state chosen-ice))
+                                                               :effect (effect
+                                                                        (corp-install chosen-ice target {:install-state :rezzed-no-rez-cost})
+                                                                        (shuffle! :deck)
+                                                                        (effect-completed eid))}
```
jwarwick: You could do the `shuffle!` action first, then pass the `eid` to `corp-install`. Then there's no need for `effect-completed`. Unless there's some reason the shuffle would need to happen after the install (keep tracking of the card? I don't think that should be an issue, since the zone hasn't changed).

presheaf:
I think this is fine now. Thanks a lot for the help fixing this test in Slack.

PR #3402, opened 2018-05-01
WIN: Slipstream card implementation (WIP)

msbeck:
Used Bishop and Rook as an archetype for choosing ice with criteria, and Copycat for moving a runner to that position.

### DIFF
--- a/src/clj/game/cards/resources.clj
+++ b/src/clj/game/cards/resources.clj
@@ -1545,6 +1545,26 @@
                :runner-turn-ends {:effect (effect (update! (dissoc card :server-target)))}}
       :abilities [ability]})
 
+   "Slipstream"
+    {:abilities [{:label "[Trash]: Choose a piece of ice protecting a central server in the same position as the passed ice and approach that ice."
+                     :req (req (and (:run @state)))
+                     :effect (effect (req (let [icepos (ice-index state current-ice)]
+                                            (resolve-ability state side
+                                             {:prompt (msg "Choose a piece of ICE protecting a central server at position " icepos)
+                                              :choices {:req #(and (is-central? (second (:zone %)))
+                                                                   (ice?)
+                                                                   (=(icepos (ice-index state %))))}
+                                              :msg (msg "move to " (card-str state target))
+                                              :effect (req (let [dest (second (:zone target))
+                                                                 tgtndx (ice-index state target)]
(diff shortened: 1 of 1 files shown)


src/clj/game/cards/resources.clj
```diff
@@ -1545,6 +1545,26 @@
...
+                                             {:prompt (msg "Choose a piece of ICE protecting a central server at position " icepos)
+                                              :choices {:req #(and (is-central? (second (:zone %)))
+                                                                   (ice?)
+                                                                   (=(icepos (ice-index state %))))}
+                                              :msg (msg "move to " (card-str state target))
+                                              :effect (req (let [dest (second (:zone target))
+                                                                 tgtndx (ice-index state target)]
+                                                             (swap! state update-in [:run]
+                                                                    #(assoc % :position tgtndx :server [dest]))
+                                                             (trash state side card {:cause :ability-cost})
```
nealterrell: Same comment as the Jackpot PR: this not an ability cost. It should, however, be unpreventable because of the "If you do" clause.

src/clj/game/cards/resources.clj
```diff
@@ -1545,6 +1545,26 @@
...
+                                                             (system-msg :runner
```
nealterrell: Missing `state`.

msbeck: in the `system-msg`?

nealterrell:
Yep.

src/clj/game/cards/resources.clj
nealterrell: Yep!

msbeck:
There's room for me to revisit this and dynamically create a list of prompts of valid servers to move to, but as of now this works and it also literally allows the user to do what the card says: "choose a piece of ice"

screen shot 2018-05-03 at 2 17 03 pm

msbeck:
Hold on this, I'm realizing an error. You have to approach ice at the _same_ position as the last passed ice, not the position of the ice you are in front of after passing.

msbeck:
Updated to allow user to trigger before passing ice to circumvent the "when passed" abilities of ice. Considered complete from my end.

msbeck:
Updated messaging for more usability.

screen shot 2018-05-03 at 11 03 54 pm
screen shot 2018-05-03 at 11 03 26 pm

PR #3405, opened 2018-05-01
Kampala: Zer0

nicohasa:


test/clj/game_test/cards/hardware.clj
```diff
@@ -1177,3 +1177,19 @@
...
+  ;; Zer0 - Once per turn, deal 1 damage to self, to gain 1 credit and 2 cards.
+  (do-game
+    (new-game (default-corp)
+              (default-runner [(qty "Zer0" 1) (qty "Sure Gamble" 2) (qty "Corroder" 1)]))
+    (starting-hand state :runner ["Zer0" "Corroder"])
+    (take-credits state :corp)
+    (play-from-hand state :runner "Zer0")
+    (is (= 4 (:credit (get-runner))) "Runner has 4 credits")
+    (let  [z (get-hardware state 0)]
+      (print z)
```
Saintis: Normally we don't put print functions in the tests

test/clj/game_test/cards/hardware.clj
```diff
@@ -1177,3 +1177,19 @@
    (prompt-choice :runner "Expose a card")
    (prompt-select :runner (get-ice state :archives 0))
    (is (= 3 (:credit (get-runner))) "Gained 1 more credit from exposing")))
+
+(deftest zer0
+  ;; Zer0 - Once per turn, deal 1 damage to self, to gain 1 credit and 2 cards.
+  (do-game
+    (new-game (default-corp)
+              (default-runner [(qty "Zer0" 1) (qty "Sure Gamble" 2) (qty "Corroder" 1)]))
```
Saintis: The single `qty` cards could be replaced with just a string of the card name (e.g. `"Zer0"`)

test/clj/game_test/cards/hardware.clj
nicohasa: wups, that's left after I tried to figure out what went wrong at some point

nealterrell:
We may need to look at making runner-install use `pay-sync`, but for now this should be OK.

PR #3406, opened 2018-05-01
WIP: Change all events to use single map

nicohasa:
Instead of having to remember that for damage, first target is type,
second is card-source, third is amount, it would be better
to have a single target {:type :net :amount 3 :card snare}.

Have changed the following events to use a single map as target (and
fixed errors introduced by this change)

```
:pre-resolve-damage
:damage
:pre-damage
:prevented-damage
:pre-resolve-tag
:runner-gain-tag
:pre-tag
```

More to come!

### DIFF
--- a/src/clj/game/cards/agendas.clj
+++ b/src/clj/game/cards/agendas.clj
@@ -1059,7 +1059,7 @@
       :events {:corp-turn-ends nil :runner-turn-ends nil}}
 
    "Sentinel Defense Program"
-   {:events {:pre-resolve-damage {:req (req (and (= target :brain) (> (last targets) 0)))
+   {:events {:pre-resolve-damage {:req (req (and (= (:type target) :brain) (> (:amount target) 0)))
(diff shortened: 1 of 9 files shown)


danhut:
This is amazing!  I was literally thinking this needed to get done as I walked home last night.

nealterrell:
This is a good and necessary change, but beware relying on tests to discover errors from the conversion. Many cards do not have tests. It sucks, but I would feel much better with a manual review of all uses of each particular handler.

nicohasa:
Ya I naively started using tests, but i did end up grepping for the relevant keywords (If I've missed anything it's on `:damage` as that is used for a bunch of other things as well, but think I got it all).

NoahTheDuke:
This can be closed, right?

nicohasa:
Ya, I haven't had time to work on anything jinteki-related for a long time. When I get back if I wanna fix this I have to start from scratch anyway.

PR #3407, opened 2018-05-02
WiN: Warden Fatuma

jwarwick:
Basic structure. Firing new sub doesn't work. Rez doesn't work.

None of the abilities that populate ice subroutines work correctly for any cards.


jwarwick:
Updated ice subroutine handlers. Working now. All tests pass, though I suspect that means other assets and ice that interact with ice subroutines don't have tests.
screen shot 2018-05-02 at 3 52 07 pm

jwarwick:
Well, if I'm going to fix them, I think that's the first step. Luckily I think only 2 or 3 cards try, the rest just print a console message.

NoahTheDuke:
This is an excellent step forward for other "Adds subs" cards. Would it be possible to extract the main bits to a generalized function we could plug in elsewhere?

jwarwick:
`Wetwork Refit` and `Sub Boost` were the only cards that attempted to modify ICE subroutines in place. Those are now updated. Probably good enough for now. Future work could include `Loki`, `Marker`, `Project Wotan`, `Project Kusanagi`, `TL;DR`, and `Sensei` (among others).

screen shot 2018-05-02 at 10 37 23 pm

src/clj/game/cards/assets.clj
```diff
@@ -1679,6 +1679,35 @@
                    :delayed-completion true
                    :effect (req (as-agenda state :runner eid card 2))}}
 
+   "Warden Fatuma"
+   (let [new-sub {:label "[Warden Fatuma] force the Runner to lose 1 [Click], if able"
```
NoahTheDuke: Super small nitpick: can "force" be capitalized?

PR #3410, opened 2018-05-02
Client Lock Update

### LINKED ISSUE #2930: Improve Client Lock
1. Lock is removed by any incoming command from the server.  Explore if we can remove this only in response to our own action - this should remove "too fast" clicking incidents which I suspect are the result of another action clearing the lock
2. Confirm typing indicator is a system event and as such will not set or clear lock
3. Explore making own "say" commands not set a lock as they don't change game state

danhut:
Improves how we handle client to server locking.  Full brain dump here:
https://github.com/mtgred/netrunner/wiki/Client-Lock-and-Handling-Client-Commands

Build failing due to unrelated TFP and Embezzle checks.  Nothing to do with this.

**Solution:**
*Create and track an action id (:aid) for each player
*This is incremented by 1 with each call for the side to handle-action
*This is sent to the client
*The client checks for each new game state. If its own side :aid has changed it can clear the lock safely.

Also cleans up two now unused typing commands, and re-orders commands to alphabetical.

Fixes #2930

### DIFF
--- a/src/clj/game/main.clj
+++ b/src/clj/game/main.clj
@@ -13,43 +13,41 @@
    "typingstop" core/typingstop})
 
 (def commands
-  {"typing" core/typing
-   "typingstop" core/typingstop
-   "concede" core/concede
-   "system-msg" #(core/system-msg %1 %2 (:msg %3))
+  {"ability" core/play-ability
+   "access" core/successful-run
+   "advance" core/advance
    "change" core/change
-   "move" core/move-card
-   "mulligan" core/mulligan
-   "keep" core/keep-hand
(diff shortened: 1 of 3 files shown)


nealterrell:
Looks simple enough.

I believe there's a race condition though: if two messages arrive simultaneously, but one takes much longer to process than the other, both action IDs will be updated and sent back as the response to the fast message, unlocking both players' UI even though only one player's action was recorded. I believe that moving `set-action-id` AFTER handling the command would fix this?

danhut:
Great catch. This can happen before the swap and diff thingy.

PR #3412, opened 2018-05-02
WiN: Masvingo

NoahTheDuke:
This one's pretty simple.

### DIFF
--- a/src/clj/game/cards/ice.clj
+++ b/src/clj/game/cards/ice.clj
@@ -1344,7 +1344,7 @@
     :advanceable :always
     :abilities [{:label "Gain subroutines"
                  :msg (msg "gain " (:advance-counter card 0) " subroutines")}]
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/ice.clj
```diff
@@ -1339,6 +1339,14 @@
                                 (when (<= 3 (+ (:advance-counter card 0) (:extra-advance-counter card 0)))
                                   (end-run state side)))}]}
 
+   "Masvingo"
+   {:implementation "Number of subs is manual"
+    :advanceable :always
+    :abilities [{:label "Gain subroutines"
+                 :msg (msg "gain " (:advance-counter card 0) " subroutines")}]
+    :events {:rez {:effect (effect (add-prop card :advance-counter 1))}}
```
nealterrell: This triggers on every rez of any card 😛 . This should probably just be in the `:effect` of the card itself, which triggers when the card is rezzed.

PR #3413, opened 2018-05-02
WiN: Logic Bomb

NoahTheDuke:
Couldn't figure out how to pull `(:click (:runner @state))` out, so I just left it all over the place. Also, I can't help myself but prettify other messy cards when I see them.

### DIFF
--- a/src/clj/game/cards/resources.clj
+++ b/src/clj/game/cards/resources.clj
@@ -1057,7 +1057,7 @@
 
    "Logic Bomb"
    {:implementation "Bypass effect is manual"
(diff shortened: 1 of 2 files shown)


presheaf:
I see I'm a bit too late on this one, but IIRC `runner` is defined in `req/effect` as `(:runner @state)`, so it seems like you could at least shorten it to `(:click runner)` unless I'm missing a good reason not to.

NoahTheDuke:
Oh damn, good call. That’s my bad. I’ll add a fix to Freedom Khumalo.

PR #3414, opened 2018-05-02
WiN: Freedom Khumalo

NoahTheDuke:
~Here's the really dumb version, that doesn't cost or check or use virus counters at all, it just auto-trashes the first card encountered. Based off Imp, obvs.~

Updated below


NoahTheDuke:
Now more intelligent: `prompt-select` cards to get virus counters from. Once enough have been selected, card is trashed. No cancel or undo yet, tho. That's next.

msbeck:
Trypano virus's are spendable too! ;-)

PR #3421, opened 2018-05-04
Prevention upgrade -ready for merge

### LINKED ISSUE #3264: No One Home triggers on every instance of net damage
No One Home should only provide an option to prevent damage/tags once per turn. Currently it triggers every time Net Damage happens:

EricF uses Aiki to make the Runner draw 2 cards.
EricF uses Aiki to do 1 net damage.
jarlstig has the option to avoid damage.
jarlstig will not prevent damage.
jarlstig trashes Data Folding due to net damage.
EricF uses Aiki to do 1 net damage.
jarlstig has the option to avoid damage.
jarlstig will not prevent damage.
jarlstig trashes Saker due to net damage.
jarlstig continues the run.
jarlstig
jarlstig
?
EricF has no further action.
jarlstig accesses Project Kusanagi from Server 5.
jarlstig steals Project Kusanagi and gains 0 agenda points.
EricF uses Jinteki: Personal Evolution to do 1 net damage.
jarlstig has the option to avoid damage.
jarlstig will not prevent damage.
jarlstig trashes Underworld Contact due to net damage.
EricF
EricF
sorry, writing bug report on NOH

danhut:
This is an upgrade to the damage and tag prevention system.  The card-def now includes a :req check for a given type like so:
```clojure
 {:interactions {:prevent [{:type [:net :brain]
                                          :req (req true)}]}
```

The set of prevention types and :req checks is actually a list.  This gives us future flexibility to have multiple and different checks for different types.

This check is used to determine if we should offer the user a prompt to fire prevention cards.  The actual prevention ability is outside of the interactions code block so the user will still click those cards to fire them.  There are some things we cannot handle today like recurring credit which would make a menu system NOT work vs. the new way Imp is done

Fixes #3264

Prevention types to handle in core code logic:

- [x] Tag
- [x] Damage
- [x] Expose 
- [x] Jack-Out
- [x] Trash - hardware, program
- [x] Bad pub


src/clj/game/core/flags.clj
```diff
@@ -433,3 +433,37 @@
...
+(defn ab-can-prevent?
+  "Checks if the specified ability definition should prevent.
+  Checks for a :req in the :prevent map of the card-def.
+  Defaults to false if req check not met"
+  ([state side card req-fn]
+   (ab-can-prevent? state side (make-eid state) card req-fn))
+  ([state side eid card req-fn]
+   (cond
+     req-fn (req-fn state side eid card nil)
+     :else false)))
```
jwarwick: Any reason we want this to default to `false`? For abilities, a missing `:req` defaults to `true`. Then we don't have to put `:req (req true)` everywhere.

danhut: There was a rationale.  Honest....

Maybe... I think it was that if default to false means we don't stick warnings on the UI if someone forgets to write a req so we default to OFF... with abilities we default to True/ON as they have been played.  Protection is something to be turned ON not something to default ON?

It seemed right at the time - but fine to change it if folks want.  I think we have other stuff where we always want (req true) like :silent and :interactive - so we are nothing if not inconsistent :D

jwarwick: I didn't think about `:silent` or `:interactive`. Guess it does depend on the situation. I don't know that I have a preference given the different use cases.

NoahTheDuke:
Would Broadcast Square fall under this? Something like:

```clojure
"Broadcast Square"
{:interaction {:prevent [{:type [:bad-publicity]
                         :req (req true)}]}
:events {:pre-bad-publicity {:delayed-completion true
                             :trace {:base 3
                                     :msg "prevents all bad publicity"
                                     :effect (effect (bad-publicity-prevent Integer/MAX_VALUE))}}}}
```

danhut:
@NoahTheDuke  I upgraded the core prevent bad-pub code because it was there - however it is not used at all right now.  

For broadcast square it will fire its trace automatically if the card is rezzed already so I am not sure what hooking it into the prevent logic will add anything - and in fact it is not a "may" trace so we add a click for no reason.

It *could* be used to prompt the corp to rez broadcast square before the BP is taken?  Out of my scope ;)

NoahTheDuke:
A, good call. I was over eager!

Saintis:
Some merge conflicts.

danhut:
Conflicts sorted

nealterrell:
I hate to do this to you my friend... but I think that `:type` should be a set, not a vector, as the order of the prevented types is irrelevant, and we should prefer a set with its constant-time lookup. You can then use `contains?` instead of `(some #{type} ...)` which is cleaner.

PR #3430, opened 2018-05-05
Made lobby notifications respect sound settings

jwarwick:
Don't know if this was intentional or not, but sounds in the game lobby always played. This PR blocks them if `Sounds` are disabled in the `Settings.`

If people think this behavior was a feature, then we could add a second sound setting for notifications.


Saintis:
Also some merge conflicts

PR #3442, opened 2018-05-07
WIP: Changed checking of requirement for trigger-event-sync

nicohasa:
Also made `:corp-install` awaitable, this also fixes the asa+mirrormorph bug.

Ok so the biggest change in this PR is that when we call trigger-event-sync, we used to

- Fetch all cards that should trigger on the given event, put them in a list
- Sort them by current-player (first all current-player events, then all opponent events)
- Go through the list, checking if each event should trigger, by checking if the card still exist,
that its `:req` is fulfilled, and that the event is not suppressed for that card (basically Crisium Grid).
- For every event that should fire, resolve its effect.
- Inform the world that an event of the correct type has happened.

This created some problems when I made `:corp-install` awaitable, as specifically Asa triggers on
a `:corp-install` but also triggers a `:corp-install`. As the last step was to inform the world that an
event of type `:corp-install` had happened, this meant that both cards installed with Asa counted
as the first `:corp-install`.

What I now do:

- Fetch all cards that should trigger on the given event, put them in the list
- Go through the list, only keeping the cards that should trigger, by checking that its
`:req` is fulfilled, and that the event is not suppressed.
- Sort them by current-player
- Inform the world that an event of the correct type has happened.
- Go through the filtered and sorted list, for every event check if the card is still there,
If it is, resolve that events effect, but first remove `:req` from the ability. We have already checked `:req`, and since we informed the world of a `:corp-install` happening in the meantime, keeping the req would stop many abilities from firing.

This is why I believe this is correct:

1. This is (kind of) how Netrunner rules work. When an event happens, Netrunner creates a list of all cards that should trigger off the event (which would include checking its requirements, for cards like Beth), then goes through the list and fires the abilities. If the game state changes in the meantime, so the requirements are no longer fulfilled, the card should still trigger (gaining a credit from Corp Sales Team should still trigger the 'old' Beth ability), as long as the card has not been trashed.
2. This gives us an excellent way of programming if-when vs when-if abilities. The if-when abilities can be programmed with a `:req` checking the if, while a when-if can just have an effect that calls `continue-ability` with a requirement in the new ability.

What this does not solve:
- An ability should be stopped from triggering if the _event itself_ is cancelled (bypassing Tollbooth, hosting agenda on Film Critic, etc). However, if you want an event that could be cancelled, then you
should use `trigger-event-simult` with a fitting `cancel-fn`. In the future, maybe all events are called via this, and that will be great.

Alternative solution:
A quicker way of solving the problem of when the game should inform the world of an event being triggered, is to just do that as the first step, and then change `first-event?` to check if the filtered list of those events have 1 element, instead of being empty. The alternative solution is less radical, but I still think the original solution is a bit more in line with the game rules, and might therefore help us in the future when implementing cards.


src/clj/game/core/installing.clj
```diff
@@ -252,35 +252,38 @@
...
-                                             (update! state side (assoc (get-card state moved-card) :rezzed true :seen true)))
-                                           (when-not (:delayed-completion cdef)
-                                             (effect-completed state side eid)))
-
-                                       ;; All other cards. Trigger effect-completed.
-                                       :else
-                                       (effect-completed state side eid))
-
-                                     (trigger-event state side :corp-install (get-card state moved-card))
+                                 (do (letfn [(event [state side eid _] (trigger-event-sync state side eid :corp-install (get-card state moved-card)))]
```
NoahTheDuke: You don’t need the `do`, as let binding have an implicit `do`.

nicohasa: Right you are!

nicohasa:
Things that should not crash are crashing now. But I have a better idea of what I WANT to do, so I'm probably gonna nuke this branch, start anew from master.

NoahTheDuke:
Close it!

Issue #3454, opened 2018-05-08
Film critic and Jackpot issue


vkl125:
![image](https://user-images.githubusercontent.com/18687557/39734147-22117c5e-5233-11e8-9a6f-f8425b3f130a.png)

msbeck:
Can you give a description of the issue please?

vkl125:
Film critic and jackpot issue.

msbeck:
Can you please describe what the issue with Film critic and Jackpot! is?

vkl125:
Film critic not sync with jackpot.

vkl125:
Screenshot said Click Click to add an agenda to score area from Film critic. Then Screenshot showed Jackpot remained on the board with money.

Saintis:
Bug seems to be that using FC to add an agenda to the runner's score area does not trigger Jackpot.

PR #3463, opened 2018-05-08
Film Critic fixes

### LINKED ISSUE #3462: Film Critic attempts to score hosted MCA Informant
If the Corp hosts MCA Informant on a Film Critic, the Film Critic ability can be triggered to attempt to move it the scored area. This causes a null pointer exception.

jwarwick:
Require a hosted agenda to fire the `Film Critic` ability.
Fixes #3462


jwarwick:
Made `Film Critic` use `as-agenda` so `Jackpot!` will fire.

Fixes #3454

PR #3465, opened 2018-05-08
Fix Trypano infinite recursion.

nealterrell:
`:counter-added` is triggered on ANY card getting a counter. Tryp trashes ice -> Friday Chip gets a counter -> Tryp sees new counter, trashes ice again -> Friday Chip gets a counter -> etc.


jwarwick:
We alphabatize to keep @JoelCFC25 happy.

PR #3477, opened 2018-05-10
Paperclip shouldn't crash if fired on facedown ice

jwarwick:
`current-strength` isn't set for facedown ice. It was being used to set the default strength in the `Paperclip` prompt.


nealterrell:
It's almost like he has access to server crash logs...

jwarwick:
Don't give away all my secrets!

PR #3481, opened 2018-05-10
FIX: Disallow trash abilities on accessing agendas in archives

NoahTheDuke:
Missed this when building the Access prompts.

### DIFF
--- a/src/clj/game/core/runs.clj
+++ b/src/clj/game/core/runs.clj
@@ -221,7 +221,7 @@
         trash-ab-cards (when (not= (:zone c) [:discard])
                          (->> (concat (all-active state :runner)
                                       (get-in @state [:runner :play-area]))
-                              (filter #(can-trigger? state :runner (:trash-ability (:interactions (card-def %))) % [c]))))
+                              (filter #(can-trigger? state :runner (->> % card-def :interactions :trash-ability) % [c]))))
         ability-strs (map #(->> (card-def %) :interactions :trash-ability :label) trash-ab-cards)
         ;; strs
         steal-str (when (and can-steal-this? can-pay-costs?)
--- a/src/clj/game/core/runs.clj
+++ b/src/clj/game/core/runs.clj
@@ -221,7 +221,7 @@
         trash-ab-cards (when (not= (:zone c) [:discard])
                          (->> (concat (all-active state :runner)
                                       (get-in @state [:runner :play-area]))
-                              (filter #(can-trigger? state :runner (->> % card-def :interactions :trash-ability) % [c]))))
+                              (filter #(can-trigger? state :runner (get-in (card-def %) [:interactions :trash-ability]) % [c]))))
         ability-strs (map #(->> (card-def %) :interactions :trash-ability :label) trash-ab-cards)
         ;; strs
         steal-str (when (and can-steal-this? can-pay-costs?)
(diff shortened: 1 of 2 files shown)


test/clj/game_test/cards/programs.clj
```diff
@@ -379,7 +379,16 @@
...
+        (is (= 0 (:agenda-point (get-runner))) "Runner did not steal TFP"))))
+  (testing "vs cards in Archives"
+    (do-game
+      (new-game (default-corp ["Hostile Takeover"])
+                (default-runner ["Imp"]))
+      (core/move state :corp (find-card "Hostile Takeover" (:hand (get-corp))) :discard)
+      (take-credits state :corp)
+      (play-from-hand state :runner "Imp")
+      (run-empty-server state "Archives")
+      (is (= ["Steal"] (->> (get-runner) :prompt first :choices)) "Should only get the option to steal"))))
```
Saintis: I'd add "when accessing Hostile Takeover from archives"

src/clj/game/core/runs.clj
```diff
@@ -218,9 +218,10 @@
         cost-as-symbol (when (= 1 (count cost-strs)) (costs-to-symbol cost))
         ;; any trash abilities
         can-steal-this? (can-steal? state side c)
-        trash-ab-cards (->> (concat (all-active state :runner)
-                                    (get-in @state [:runner :play-area]))
-                            (filter #(can-trigger? state :runner (:trash-ability (:interactions (card-def %))) % [c])))
+        trash-ab-cards (when (not= (:zone c) [:discard])
+                         (->> (concat (all-active state :runner)
+                                      (get-in @state [:runner :play-area]))
+                              (filter #(can-trigger? state :runner (:trash-ability (:interactions (card-def %))) % [c]))))
```
Saintis: I'd change the nested keywords to a `get-in`.

test/clj/game_test/cards/programs.clj
NoahTheDuke: Why?

Saintis: More explicit, and easier to understand what the issue is when the test fails

src/clj/game/core/runs.clj
```diff
@@ -221,7 +221,7 @@
         trash-ab-cards (when (not= (:zone c) [:discard])
                          (->> (concat (all-active state :runner)
                                       (get-in @state [:runner :play-area]))
-                              (filter #(can-trigger? state :runner (:trash-ability (:interactions (card-def %))) % [c]))))
+                              (filter #(can-trigger? state :runner (->> % card-def :interactions :trash-ability) % [c]))))
```
Saintis: This double arrow `->>` could be a single arrow `->` to save a valuable character

src/clj/game/core/runs.clj
NoahTheDuke: Heh yeah, I’ll probably go back to `get-in` next time I’m on the computer. I realized I could pull one more level out without thinking about how it’d read later. I personally like how it looks, but only when it’s... conversational? Like when checking prompts in a test. This isn’t quite that, so I think you’re right.

src/clj/game/core/runs.clj
NoahTheDuke: Honestly, I just forgot that the single threading macro places it immediately after the func. I keep thinking it places it to the left of the function, lol.

test/clj/game_test/cards/programs.clj
NoahTheDuke: Makes sense! I’ll pay attention to that in the future.

PR #3482, opened 2018-05-10
Pass the side of a card as its event handler side. Fix #3458.

### LINKED ISSUE #3458: Gang Sign a Snare!, runner gets prompts
A runner Gang Signed into my Snare! in HQ. After that, he got a "waiting for the corp to resolve trigger", but I did not get a Snare! prompt. When he `/close-prompt`'ed, the prompt to use Snare! was waiting for him. He triggered it, paying 2 credits (and dying).

nealterrell:
I think this should be fine. No tests fail, at least.

### DIFF
--- a/test/clj/game_test/cards/resources.clj
+++ b/test/clj/game_test/cards/resources.clj
@@ -811,7 +811,7 @@
     (take-credits state :corp)
     (play-from-hand state :runner "Gang Sign")
     (take-credits state :runner)
-    (play-and-score state :corp "Hostile Takeover")
+    (play-and-score state "Hostile Takeover")
     (prompt-choice :runner "Card from hand")
     ;; Runner has "wait for Snare, wait for on-access" prompts.
     (is (= 2 (count (:prompt (get-runner)))) "Runner only has the Waiting prompt, not Snare!'s pay-prompt")
--- a/test/clj/game_test/cards/resources.clj
+++ b/test/clj/game_test/cards/resources.clj
@@ -803,6 +803,22 @@
       (prompt-choice :runner "Card from hand")
       (prompt-choice :runner "Steal"))))
 
+(deftest gang-sign-correct-prompts
+         ;; Gang Sign - accessing from HQ, not including root. Issue #2113.
+   (do-game
+     (new-game (default-corp [(qty "Hostile Takeover" 1) (qty "Snare!" 1)])
+               (default-runner [(qty "Gang Sign" 1)]))
+     (take-credits state :corp)
+     (play-from-hand state :runner "Gang Sign")
+     (take-credits state :runner)
+     (play-from-hand state :corp "Hostile Takeover" "New remote")
+     (score-agenda state :corp (get-content state :remote1 0))
+     (prompt-choice :runner "Card from hand")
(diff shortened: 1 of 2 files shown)


test/clj/game_test/cards/resources.clj
```diff
@@ -803,6 +803,22 @@
       (prompt-choice :runner "Card from hand")
       (prompt-choice :runner "Steal"))))
 
+(deftest gang-sign-correct-prompts
+         ;; Gang Sign - accessing from HQ, not including root. Issue #2113.
```
Saintis: This comment is too far indented. Should be aligned with `do-game` below.

Saintis: Also, you should move this to `deftest gang-sign` with `testing "correct prompts"` or similar. See Noah's great tests 😉

test/clj/game_test/cards/resources.clj
```diff
@@ -803,6 +803,22 @@
...
+   (do-game
+     (new-game (default-corp [(qty "Hostile Takeover" 1) (qty "Snare!" 1)])
+               (default-runner [(qty "Gang Sign" 1)]))
```
Saintis: Since I found out that we can replace `(qty X 1)` with just `X` I prefer that for single copies of cards.

test/clj/game_test/cards/resources.clj
```diff
@@ -803,6 +803,22 @@
...
+     (take-credits state :corp)
+     (play-from-hand state :runner "Gang Sign")
+     (take-credits state :runner)
+     (play-from-hand state :corp "Hostile Takeover" "New remote")
+     (score-agenda state :corp (get-content state :remote1 0))
```
Saintis: There is a snazzy `play-and-score` function that condenses these two functions into one.

test/clj/game_test/cards/resources.clj
```diff
@@ -803,6 +803,22 @@
...
+     (prompt-choice :runner "Card from hand")
+     ;; Runner has "wait for Snare, wait for on-access" prompts.
+     (is (= 2 (count (:prompt (get-runner)))) "Runner only has the Waiting prompt, not Snare!'s pay-prompt")
```
Saintis: Would this not be true if Runner had pay prompt and corp had waiting prompt?

test/clj/game_test/cards/resources.clj
```diff
@@ -803,6 +803,22 @@
...
+     ;; Core has "pay for Snare, wait for agenda-scored" prompts.
+     (is (= 2 (count (:prompt (get-corp)))) "Corp has the prompt to use Snare!")))
```
Saintis: Same as comment above, but the converse

test/clj/game_test/cards/resources.clj
nealterrell: No, the runner would have 3 prompts at that point. (Pay for Snare. Waiting for Corp to use Snare [this prompt is always shown to Runner regardless of who sees the Pay prompt]. Waiting for on-access effects to resolve.)

Saintis: Ah, I see

Issue #2469, opened 2017-04-04
Various less important things.


PesiTheWizard:
- [x] **1)** Using Top Hat doesn't tell the corp what # card from the top was accessed.
- [x] **2)** An Exploratory Romp run technically ends before the card to lose counters is chosen. Corp could abuse this unintentionally.
- [x] **3)** Using Information Sifting gives 2 accessing logs for each card accessed.
- [x] **4)** Accessing Mumbad Virtual Tour with Neutralize All Threats installed and not being able to pay for it gives a lot more yellow warnings than one would expect.
- [x] **5)** First Responders doesn't care who did the damage.
- [ ] **6)** Apocalypse'd runner cards stop being active without triggering leave-play. Good for Muertos Gang Member, bad for Origami and cards that define an ability like Raymond Flint. (TODO: Encompasses enough to warrant it's own issue)
~~**7)** Manually increasing Bad Publicity with the control panel seems to trigger Raymond Flint.~~
*Probably not actually a bug*
- [x] **8)** Playing Möbius with Doppelgänger installed can mess things up somehow so that the runner gains 4 credits on every successful R&D run, just like the first bug with it. This can build up when playing other Möbii. Playing Möbius with Same Old Thing sometimes doesn't make the 2nd run happen.
- [x] **9a)** Using Sacrifice to forfeit an agenda worth -1 points causes corp to lose 1 credit and gain 1 Bad Publicity. This happens weather it's a 0-pointer with Political Graffiti or a -1point ambush that has made it's way to the corp's score area.
~~**9b)** Sacrificing an "Political Graffiti"-ed agenda this way refunds the lost agenda points twice over. After sacrificing a 0-pointer with 3 Graffitis, then scoring a 2-pointer, the corp had 5 agenda points. (The 2-pointer was Corporate War. Having a negative credit balance when doing so brings it up to 0.)~~
*Is a PG-only issue*
- [x] **10)** Haas-Bioroid: Stronger Together and Employee Strike interact strangely. Strengths seem to not be updated until a run is made. Can lead to a visual situation where if the current goes away and then a bioroid is rezzed outside of a run, some of them have the strength bonus and some don't.
- [x] **11)** Installing ICE with a scored ABT doesn't ignore the additional install cost of Network Exchange.

JoelCFC25:
2 and 5 are easy, at any rate...7 is expected because Raymond uses a watch state. 

6 just makes me cringe, been over that ground so many times I can't bear to look at it anymore.

[...]

PR #3483, opened 2018-05-10
FIX: Better handling for costs in access prompts

### LINKED ISSUE #3471: Runner forced to trash MVT even when on < 5 credits
Twice today in games the runner has been forced to trash MVT firing the CTM trace even when they had < 5 credits, and no other trash mechanism such as Imp/Freedome &c.

NoahTheDuke:
Fixes bug where Mumbad Virtual Tour would force trash when Runner had < 5 credits. Also fixes bug with not showing trash ability choices on agenda access prompts with multiple steal costs.

Fixes #3471, #2469

### DIFF
--- a/src/clj/game/core/runs.clj
+++ b/src/clj/game/core/runs.clj
@@ -133,7 +133,8 @@
                 trash-cost-str (when can-pay
                                  [(str "Pay " trash-cost "[Credits] to trash")])
                 ;; If the runner is forced to trash this card (Neutralize All Threats)
-                forced-to-trash? (and can-pay
+                forced-to-trash? (and (or can-pay
+                                          (seq trash-ab-cards))
                                       (or (get-in @state [:runner :register :force-trash])
                                           (card-flag-fn? state side card :must-trash true)))
                 trash-msg (when can-pay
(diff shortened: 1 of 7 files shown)


Saintis:
For Imp + MVT, as the runner is force to trash MVT if installed and able I would expect prompts to give Pay and Imp option if >= 5 credits and just Imp otherwise (so no "No action" option). Is this possible with the current access stuff?

NoahTheDuke:
I think you're right, let me fix that up.

src/clj/game/core/runs.clj
```diff
@@ -234,39 +238,41 @@
...
+
+                                      ;; Steal normally
+                                      (= target "Steal")
+                                      (steal-agenda state :runner eid c)
+
+                                      ;; Pay single additiional cost to steal
+                                      (.contains target "Pay")
+                                      (if (> n 1)
+                                        ;; Use the better function for multiple costs
+                                        (resolve-ability state :runner
```
nealterrell: I'm fairly certain this needs to be continue-ability. Otherwise this eid is left hanging.

NoahTheDuke: Damn, good call. Thanks for catching that.

Issue #2829, opened 2017-10-04
Ekomind MU Bug


konals:
I've been playing a couple games with Ekomind and I've noticed that I seem to have more MU than I should.

I just finished a game where I had Ekomind installed with 15 cards in hand, and two LLDS Memory Diamonds.

This should have given me 17 MU. I also had 14 MU of programs installed, which should have brought me to a total of 3 MU, instead I was at 15 MU free.

I'm not sure exactly what causes the issue but I think Ekomind is assigning extra memory for some reason.

I did drag a card back to my hand that I didn't intend to discard at one point.

JoelCFC25:
Yeah, this is actually a longstanding problem--a very little used card which is probably why no one's bothered to say anything. Ekomind's watch state is actually adjusting what your _maximum_ MU is, but is not accounting in any way for what MU is being consumed by installed programs.

NoahTheDuke:
@Saintis, does your MU refactor handle this?

Saintis:
Yeah, this should fix the Ekomind issue

PR #3496, opened 2018-05-12
Major MU rewrite

Saintis:
Changes how MU is handled. It is now handled similar to how hand size was recently rewritten to work.

the `:memory` key of the runner's state now has a map with 3 keys: `:base`, `:mod` and `:used`.
`:base` tracks the base MU which is 4 unless Ekomind is installed. `:mod` tracks +MU from hardware, CT etc. (and -MU from Bad Times). `:used` tracks amount of MU used by installed programs.

There are several new helper functions:

- `available-mu` for checking how much memory is available / unused 

- `free-mu` for freeing up MU when a program is uninstalled etc. This affects `:used`.

- `use-mu` for using up MU when a program is isntalled etc. This affects `:used`.

GUI has been updated to show both unused MU and total MU:
screen shot 2018-05-12 at 19 15 04

In addition, the `gain` and `deduct` functions have been updated to handle numerical values for attributes that expect a map (currently just `:hand-size` and `:memory`), in which case the default sub-attribute is `:mod`. This means all my work changing all `:in-play [:memory 1]` to `:in-play [:memory {:mod 1}]` is no longer necessary (and is even unnecessary).

I would like feedback on how we should display the MU information on the client. Unused MU only (how it was before this PR)? Unused / Total MU? Unused / Used? Used / Total? etc.


Saintis:
Bonus stuff:

- adds `:no-mu` parameter to `runner-install` that ignores MU on install (for installing on hosting cards like Leprechaun)

- Generator function for the Ancient Greek breakers (Adept, Sage and Savant)

- Tidies up some other code, e.g. cloud breakers

NoahTheDuke:
Just to understand, some of your changes are from numbers to `{:mod}` and sometimes you revert those changes. Why?

NoahTheDuke:
Ah, thanks for the explaining.

This looks great and is a great step forward. Will make changes really easy to handle.

Saintis:
Should also fix #2829.

Saintis:
Have now cleaned up all the new `{:mod x}` values found in relation to memory and hand-size since they are no longer strictly necessary. Makes card implementations much cleaner. 

I swear this was not a ploy to increase my number of lines added / removed.

Saintis:
This is ready to merge (once Neal reviews it)

PR #3504, opened 2018-05-14
Remove app.js when running `lein clean`

jwarwick:


Saintis:
Does this mean you need to run `lein cljs build` after every clean? Since cljs updates are more rare than clj updates I feel like this will make more work for cases when you are only working on game-engine code.

NoahTheDuke:
Agreed. `app.js` takes some 30 seconds to compile on my machine, and I rarely need it to be cleared.

jwarwick:
Are you sure you guys are being paranoid enough?
Like @JoelCFC25 level sure?

NoahTheDuke:
True! Maybe I’m not.

nealterrell:
Not a fan in general. I appreciate the ~paranoia~ increased certainty, but I have to clean all the time when switching branches and killing `app.js` in order to recompile the server code seems like overkill. Can we add a separate root command, like `lein clean-cljs`?

PR #3507, opened 2018-05-15
Only allow a user to be in a game once

jwarwick:
To further encourage the use of smurf accounts.

Any of our internal data structures that use the database id to identify a player cannot handle the case of a player as both `Runner` and `Corp` (or `Spectator`). This attempts to block you from joining or watching a game which you are already participating.


nealterrell:
Looks good.

PR #3516, opened 2018-05-17
Don't crash if there is a nil value for a trace

jwarwick:


NoahTheDuke:
Ooo, I should merge this into surveillance sweep

PR #3517, opened 2018-05-17
Better check for user comparisons in the UI

### LINKED ISSUE #3175: All cards in hand are facedown
screen shot 2018-03-11 at 12 53 02 am

Sometimes when the game freezes, I need to leave the game and return or I click refresh, and when I rejoin the game, all my cards in my hand are facedown. How do I fix or prevent this from happening?

### LINKED ISSUE #3502: Cards weren't shown when starting a new game in the mulligan
https://i.imgur.com/ibav9Xl.jpg

A leave game>refresh>rejoin did fix it but it's def a glitch.

jwarwick:
In the UI we were comparing the entire user structure to determine if a player was the Runner/Corp/Spectator. If the structure changed, either by select an alt art card or getting a stats update, or some other method that put the user state trivially out of sync between the client and the server, then we would assume the user was a Spectator. This tended to result in the user's cards all appearing facedown.

Now we just compare the `:_id` field.

Fixes #3502 
Fixes #3175

### DIFF
--- a/src/cljs/netrunner/gameboard.cljs
+++ b/src/cljs/netrunner/gameboard.cljs
@@ -42,6 +42,9 @@
       (= (get-in state [:corp :user :_id]) user-id) :corp
       :else :spectator)))
 
+(defn not-spectator? []
+  (not= :spectator (get-side @game-state)))
+
 (defn init-game [state]
   (let [side (get-side state)]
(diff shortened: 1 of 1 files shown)


danhut:
You are da bomb

PR #3524, opened 2018-05-19
KA: Mlinzi

jwarwick:


src/clj/game/cards/ice.clj
```diff
@@ -1451,6 +1451,33 @@
                                                                    (shuffle! :deck))}
                                                  card nil)))}]}
 
+   "Mlinzi"
+   (letfn [(net-or-trash [net-dmg mill-cnt]
+             {:label (str "Do " net-dmg " net damage unless the Runner trashes the top "
```
nealterrell: I think this should just read "Do X net damage". The "unless..." part is irrelevant to the Corp when they are identifying which subroutine to resolve, which is the point of the ability label.

src/clj/game/cards/ice.clj
```diff
@@ -1451,6 +1451,33 @@
...
+                          mill-cnt " cards of the Stack.")
+              :effect (req (show-wait-prompt state :corp "Runner to choose an option for Mlinzi")
+                           (resolve-ability state :runner
+                                            {:prompt "Take net damage or trash cards from the Stack?"
+                                             :choices [(str "Take " net-dmg " net damage")
+                                                       (str "Trash the top " mill-cnt " cards of the Stack")]
+                                             :effect (req (if (.startsWith target "Take")
+                                                            (do
+                                                              (system-msg state :corp (str "uses Mlinzi to give the Runner "
```
nealterrell: "give the Runner" -> "do"

PR #3525, opened 2018-05-21
KA: Reclaim

jwarwick:


src/clj/game/cards/resources.clj
```diff
@@ -1417,6 +1417,30 @@
                  :delayed-completion true
                  :effect (effect (expose eid target) (trash card {:cause :ability-cost}))}]}
 
+   "Reclaim"
+   {:abilities
+    [{:label "[Click],[Trash],trash a card from your Grip: install a program, piece of hardware, or virtual resource from your Heap"
```
nealterrell: I believe the click cost will already be appended to the label. Also, I think we only put the trash icon into the label if that is the only cost of the ability. (Based on code for Contract Killer, Alix, MCA Austerity.)

src/clj/game/cards/resources.clj
```diff
@@ -1417,6 +1417,30 @@
...
+      :cost [:click 1]
+      :req (req (not-empty (:hand runner)))
+      :prompt "Choose a card to trash"
+      :choices (req (cancellable (:hand runner) :sorted))
+      :delayed-completion true
+      :effect (req (when-completed (trash state :runner card {:cause :ability-cost})
+                                   (when-completed (trash state :runner target {:cause :ability-cost})
```
nealterrell: `:ability-cost` is intended for abilities with a [Trash] cost icon, not for a "trash a card from your grip" cost. (Important for Geist interaction.) This one should just be `:unpreventable true`.

src/clj/game/cards/resources.clj
```diff
@@ -1417,6 +1417,30 @@
...
+                                                   (resolve-ability
```
nealterrell: Make this continue-ability for maximum delayed completion.

PR #3527, opened 2018-05-21
KA: Hippo

jwarwick:


nealterrell:
I don't understand why there's a selection prompt here. Are you reading this card as "trash the furthest-out ice that had all its subroutines broken"? I'm not sure that's the right interpretation.

jwarwick:
Yes, it appears that's how I was interpreting it. Re-reading the card, I have no idea how I came to that conclusion.

PR #3528, opened 2018-05-21
KA: Flame-out

jwarwick:


nealterrell:
This card would really benefit from a rework of runner install to show a list of potential hosts instead of going directly to the rig. Oh well.

PR #3532, opened 2018-05-22
Catch no `:agendapoints` field when stealing

jwarwick:
Another NPE appearing in the production log file.


NoahTheDuke:
What agendas don't have `:agendapoints`?

jwarwick:
Presumably all agendas have that field, but something else is corrupting the `state`. I hope that if we catch the crash, we can get to an actually useful bug report or better stack trace.

PR #3540, opened 2018-05-24
Set default value for Hivemind virus counter lookup

jwarwick:
Fixes crash in production log.

### DIFF
--- a/src/clj/game/cards/programs.clj
+++ b/src/clj/game/cards/programs.clj
@@ -450,7 +450,7 @@
       :trash-effect {:effect update-programs}
       :events {:counter-added {:req (req (= (:cid target) (:cid card)))
                                :effect update-programs}}
-      :abilities [{:req (req (> (get-in card [:counter :virus]) 0))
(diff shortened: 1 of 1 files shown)


src/clj/game/cards/programs.clj
```diff
@@ -450,7 +450,7 @@
       :trash-effect {:effect update-programs}
       :events {:counter-added {:req (req (= (:cid target) (:cid card)))
                                :effect update-programs}}
-      :abilities [{:req (req (> (get-in card [:counter :virus]) 0))
+      :abilities [{:req (req (> (get-in card [:counter :virus] 0) 0))
```
NoahTheDuke: While you’re here, wanna make this use `pos?`?

NoahTheDuke: It'd be easy enough to write our own version of `pos?`:
```clojure
(:refer-clojure :exclude [pos?])
...
(defn pos?
  [x]
  (if (instance? Number x)
    (. clojure.lang.Numbers (isPos x))
    false))
```

PR #3541, opened 2018-05-25
KA: Black Hat

presheaf:
Implements Black Hat. Seems to work fine when testing by hand, except with Shiro, but this seems to be Shiro's fault and will be fixed once Akiko is live, as that has a Shiro fix.

Test seems a bit weird. I would appreciate any advice on better checking that 3 cards are actually accessed.


jwarwick:
I generally check `(empty? (:prompt (get-X)))` after I've accessed things to be sure I've seen the correct number. I think there's also a cards-accessed field in `register`.

PR #3544, opened 2018-05-25
Handle New Construction when :advance-counter is not defined

jwarwick:
Handling NPE crash in the production log.


NoahTheDuke:
Nice. Hopefully I can fix all these with a general `get-counters` I’m working on.

PR #3545, opened 2018-05-25
Handle cloud breaker undefined runner link

jwarwick:
Catching NPE crash in production log.


NoahTheDuke:
Okay, time for a `get-link`!

PR #3547, opened 2018-05-25
Fixed Mutate

### LINKED ISSUE #3488: Mutate just trashes the ice
I tested this twice, firing mutate on a rezzed ice just trashes the ice, reads out in the chat log that it was played, but does nothing.  There was ice left in RnD (just heritaged it to the top).

jwarwick:
Instead of relying on the corp trash event, we provide a prompt to select rezzed ice and handle the result.

Fixes #3488

### DIFF
--- a/src/clj/game/cards/operations.clj
+++ b/src/clj/game/cards/operations.clj
@@ -820,29 +820,32 @@
                          true)))))}
 
    "Mutate"
-   {:additional-cost [:ice 1]
-    :effect (effect (register-events (:events (card-def card)) (assoc card :zone '(:discard))))
-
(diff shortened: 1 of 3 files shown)


nealterrell:
This is a good change. I wonder if we can upgrade corp-install to specify install position, so we don't have to do all the hacky swaps in here. It's not a huge deal, but there's probably a broken interaction with Subcontract and 419 here, since the `:corp-install` event is awaitable, the prompts from Subcontract will probably get intermingled with 419's pay-to-stop-expose. Now that I've mentioned this, I'm sure such a corp deck will become Tier 1 tomorrow.

PR #3555, opened 2018-05-27
Ready to Test - OM -> Reagent Migration

danhut:
This is still a work in progress.  More work than I thought (again).

Implementation nodes (in progress here)
https://github.com/mtgred/netrunner/wiki/Reagent-Implementation

Raising a PR but not really wanting help or anyone to commit code please... mostly want people to know this is still happening and not far away.

This PR will move us away from the dead OM library to Reagent.   This will provide:
- Easier to read front-end code that hopefully we are not scared to touch
- Access to React performance and bug fixes
- Simpler components
- Readable documentation

The approach taken has been to create an entire new set of name-spaces called "nr".  This is to avoid nuking all the old code before I am ready, and to make my life easier.  It will also allow us to easily keep both versions for a while if people do dual commits into both sets of name spaces.

The app is also named app10.js   - this was to avoid some known issues with NGINX caching when we switch.  

Work to be done... Namespace re-writes/ports:
- [x] About
- [x] Account
- [x] Ajax - done
- [x] Appstate
- [x] Auth 
- [x] Cardbrowser.  
- [x] Chat
- [x] Deck builder
- [x] Game board
- [x] Gamelobby
- [x] Help
- [x] News - Not tested.  Not sure how to test
- [x] Stats
- [x] Utils 
- [x] Ws 
- [x] Pull updates from netrunner ns on master when everything else is working


danhut:
Re-opened vs. Dev as @nealterrell seems to want to test this and not merge direct to master

PR #3557, opened 2018-05-28
WIP: Musaazi

Giesch:
* Yusuf copypasta implementation


NoahTheDuke:
You should check out the function [pick-virus-counters-to-spend](https://github.com/mtgred/netrunner/blob/e2241b439f50352283293154e425f36e117f4b59/src/clj/game/cards.clj#L169) and how it's used in [Freedom Khumalo](https://github.com/mtgred/netrunner/blob/e2241b439f50352283293154e425f36e117f4b59/src/clj/game/cards/identities.clj#L342). The goal is actually to move Yusuf to using `pick-virus-counters-to-spend` as well.

nealterrell:
We'll merge this and then iterate.

PR #3563, opened 2018-05-30
KA: PAD Tap

NoahTheDuke:
Placeholder for now. Just want to get my name in the books for it before someone else snags it.


nealterrell:
I'm guessing the ability can't actually be used by the corp in the UI...

PR #3570, opened 2018-05-31
Rosetta double freeing mu

jwarwick:
`Rosetta` should not be attempting to free mu for card being removed from game.

Also caught a few instances of testing `installed?` without an argument.


NoahTheDuke:
Nice catch. I’ve seen that `installed?` bit a bunch. Didn’t know it was wrong. Glad you’re fixing it!

jwarwick:
It's okay to use standalone (eg., `:req installed?`) but not like `#(and installed? (= % target))`.

NoahTheDuke:
Ahhh that’s gotta be it then

test/clj/game_test/cards/resources.clj
```diff
@@ -1685,6 +1685,27 @@
...
+    (play-from-hand state :runner "Corroder")
+    (is (= 3 (core/available-mu state)) "Corrder cost 1 mu")
+    (is (= 2 (:credit (get-runner))) "Starting with 2 credits")
+    (card-ability state :runner (get-resource state 0) 0)
+    (prompt-select :runner (get-program state 0))
+    (prompt-choice :runner (find-card "Gordian Blade" (:deck (get-runner))))
+    (is (= 3 (core/available-mu state)) "Gordian cost 1 mu, Corroder freed")
+    (is (zero? (:credit (get-runner))) "Ending with 0 credits")
+    (is (= 1 (count (:rfg (get-runner)))) "Corroder removed from game")
+    (is (= 1 (count (get-in @state [:runner :rig :program]))) "One program installed")
```
NoahTheDuke: `(get-in @state [:runner :rig :program])` could be `(get-program state)`

PR #3582, opened 2018-06-04
FIX: Product Recall trash bonus miscalculation

### LINKED ISSUE #3560: Product Recall incorrectly calculating trash cost in IG
When using product recall in Industrial Genomics, it is counting both face down and face up cards in Archive to calculate trash cost.  It should only use face down.

In fact it is worse than that.  Trashing a 3 trash cost card, with 6 face down and 4 faceup in archives gave me 15.

NoahTheDuke:
Product Recall was clearing the trash bonus at the end of the effect instead of at the beginning, which caused lingering trash cost bonuses to be rolled into the new calculations when `:pre-trash` was triggered within the card effect.

This fixes that bug and adds a test to handle the case.

Fixes #3560

### DIFF
--- a/src/clj/game/cards/identities.clj
+++ b/src/clj/game/cards/identities.clj
@@ -499,8 +499,10 @@
       :abilities [ability]})
 
    "Industrial Genomics: Growing Solutions"
-   {:events {:pre-trash {:effect (effect (trash-cost-bonus
-                                           (count (filter #(not (:seen %)) (:discard corp)))))}}}
+   {:events
+    {:pre-trash
+     {:effect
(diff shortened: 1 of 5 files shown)


Saintis:
Does this work with encryption protocol?

NoahTheDuke:
Should work fine. It calls `(trigger-event :pre-trash)` to get those updates and then calls `(trash-cost)` to get the most up-to-date trash cost information, given all bonuses.

nealterrell:
I think I'd rather write something equivalent to `modified-install-cost` instead of doing the dissoc and trigger manually in a card definition (which is a bit of a code smell).

NoahTheDuke:
Done!

PR #3583, opened 2018-06-04
FIX: :no-trash-or-steal shouldn't increment in Archives

### LINKED ISSUE #3388: Maw not triggering
Just flagging up a case when Maw didn't trigger when it should have.

Situation:

Alice as the ID with Maw and Archives Interface installed does something on her side of the board for a couple of clicks, then runs archives third click. The Corp trashes an Obokata to her ability and she accesses archives seeing, but not stealing, the Obokata in the bin and removing from game some other card with Archives Interface. Then last click she runs HQ, accessing the sole card in HQ, another Obokata which she declines to steal. As the first card accessed not in Archives for the turn, this should have triggered Maw to send it to the bin, but it didn't trigger.

They're the cards I remember that seem relevant, hope that is helpful!

NoahTheDuke:
We were incrementing `:no-trash-or-steal` when accessing cards in Archives, and while I could hear an argument for it needing to increment in that case because you can technically decline to steal agendas in Archives with additional steal costs, this system only exists for Maw and I'm only interested in fixing it right now.

Fixes #3388

### DIFF
--- a/src/clj/game/core/runs.clj
+++ b/src/clj/game/core/runs.clj
@@ -46,7 +46,7 @@
     (trigger-event state side :no-steal c))
   (when (and (get-card state c)
              ;; Don't increment :no-trash-or-steal if accessing a card in Archives
-             (not (find-cid (:cid c) (get-in @state [:corp :discard]))))
+             (not= (:zone c) [:discard]))
     (swap! state update-in [:runner :register :no-trash-or-steal] (fnil inc 0)))
   (trigger-event-sync state side eid :post-access-card c))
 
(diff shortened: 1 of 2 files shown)


src/clj/game/core/runs.clj
```diff
@@ -44,7 +44,9 @@
   (when (and (is-type? c "Agenda")
              (not (find-cid (:cid c) (get-in @state [:runner :scored]))))
     (trigger-event state side :no-steal c))
-  (when (get-card state c)
+  (when (and (get-card state c)
+             ;; Don't increment :no-trash-or-steal if accessing a card in Archives
+             (not (find-cid (:cid c) (get-in @state [:corp :discard]))))
```
nealterrell: Can this be done a little more directly, like checking the zone of `c`? Also, couldn't we predicate this on `c` not being an agenda?

NoahTheDuke: Yeah, checking the zone is better. Forgot about that.

We still want it increment when you see an agenda, as that's what Maw looks for. Otherwise, declining to steal an agenda with an additional cost won't trigger Maw.

PR #3608, opened 2018-06-07
FIX: PAD Tap missing trash message

### LINKED ISSUE #3600: PAD Tap needs corp trash message
Need to add a `:system-msg` to the `corp-abilities`.

NoahTheDuke:
Adds a message to when the corp trashes PAD Tap.

Fixes #3600


NoahTheDuke:
I feel like that'd look bad: "Corp uses PAD Tap to trash PAD Tap". If you think it's okay, I'll switch it up.

Saintis:
Actually, keep the system message, but add in the cost

PR #3611, opened 2018-06-07
Catch nil tag field in Obelus

jwarwick:
Crash in production log


NoahTheDuke:
:-(((((

Time for `(get-tags state)`!

Saintis:
I'll get on it!

jwarwick:
It's just another case where the state gets corrupted (likely `nil`). Since this uses a `watch` it always fires and just happens to be the place that crashes.

PR #3621, opened 2018-06-09
Make safe-inc-n handle nil values of n

jwarwick:
Fixing crash in production log


nealterrell:
Yeah, I think I'd rather have the log message if `n` is nil. Is there a reason to hide that message?

jwarwick:
The challenge is that when you get the exception, the state has already been put in a bad state, so you can't tell what has happened. The stacktrace doesn't indicate the offending card. I'm hoping to patch this and have it crash somewhere more useful or have users complain about not earning agenda points in a specific case.

PR #3622, opened 2018-06-09
Make start-trace handle nil base and bonus

jwarwick:
Fix crash in production log


NoahTheDuke:
`fnil` handles both base and bonus, tho.

PR #3636, opened 2018-06-12
Reject messages with banned words

NoahTheDuke:
Following up on @jwarwick's excellent chat updates, I thought it might behoove us to have a really stupid chat filter, and put in only the worst words, rot13'd so we don't have to deal with them sitting open on our servers.

### DIFF
--- a/src/clj/web/chat.clj
+++ b/src/clj/web/chat.clj
@@ -38,13 +38,20 @@
   (defn- rot13 [in]
     (apply str (map #(get rot13-map % %) in))))
 
-(let [banned-words (->> (slurp "resources/public/chat/banned-words.txt")
-                        (s/split-lines)
-                        (s/join "|")
-                        (str "(?i)"))]
-  (defn- has-banned-words
-    [msg]
-    (re-find (re-pattern banned-words) (rot13 msg))))
+(def banned-words "")
+
+(defn- get-banned-words []
+  (when (empty? banned-words)
+    (def banned-words
+      (->> (slurp "data/banned-words.txt")
+           (s/split-lines)
+           (s/join "|")
+           (str "(?i)"))))
+  banned-words)
+
+(defn- has-banned-words?
+  [msg]
+  (re-find (re-pattern (get-banned-words)) (rot13 msg)))
 
 (defn- insert-msg
   [{{{:keys [username emailhash]} :user} :ring-req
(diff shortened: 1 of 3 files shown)


Saintis:
Maybe we should add a toast to inform the user their message was rejected due to a bad word? Or do proper old-school and replace the bad word with "****".

src/clj/web/chat.clj
```diff
@@ -30,11 +30,29 @@
         msg-cnt (mc/count db msg-collection {:username username :date {"$gt" start-date}})]
     (< msg-cnt max-cnt)))
 
-(defn- insert-msg [{{{:keys [username emailhash]} :user} :ring-req
-                    {:keys [:channel :msg]} :?data :as event}]
+(let [alpha (into #{} "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ")
```
Saintis: Why is this both upper and lower case? Is it not better to match just on lowercase?

src/clj/web/chat.clj
```diff
@@ -30,11 +30,29 @@
...
+      rot13-map (->> (cycle alpha)
+                     (drop 26)
+                     (take 52)
+                     (zipmap alpha))]
+  (defn- rot13 [in]
+    (apply str (map #(get rot13-map % %) in))))
+
+(let [banned-words (->> (slurp "resources/public/chat/banned-words.txt")
```
Saintis: Do we want to store the file in a public place or can we put it elsewhere?

src/clj/web/chat.clj
```diff
@@ -30,11 +30,29 @@
...
+                        (s/split-lines)
+                        (s/join "|")
+                        (str "(?i)"))]
+  (defn- has-banned-words
```
Saintis: This could have a `?` at the end since it's to be used as a boolean check

src/clj/web/chat.clj
NoahTheDuke: Yeah, that's probably a good point. I just copy-pasted this from a gist I found, so I can definitely make that change.

src/clj/web/chat.clj
NoahTheDuke: We could put it anywhere, this was just my first thought. Placing it in /data was my other thought. The only issue is getting tests to pass if it's not being pulled from somewhere.

src/clj/web/chat.clj
NoahTheDuke: Nice, good call!

src/clj/web/chat.clj
NoahTheDuke: This actually doesn't work, because unless we want to call "lower-case" on all messages before checking them, we'll miss any capital letters. Easier to convert to rot13 and then ignore-case in the regex.

NoahTheDuke:
Shelving this for now as no one was pleased with the solution.

PR #3642, opened 2018-06-13
Reagent Dev Server. - fixes

danhut:
Fixes:
- Corp discount counter
- Block user menu position
- updated with OM changes from dev
- @JohnWarwicks chat delete stuff
- @Saintis tweaks


Saintis:
Yeah, I would have had no clue how to fix that. Reagent is still very magical to me :)

PR #3644, opened 2018-06-13
Rename delayed completion functions

nealterrell:
`:delayed-completion` -> `:async`

`when-completed` -> `wait-for`

Shorter without semantic loss of meaning. Async/await are common terms in other languages like C# and JavaScript. I would have preferred `await` but that is a function in clojure.core, so `wait-for` suffices.

### DIFF
--- a/src/clj/game/core/rules.clj
+++ b/src/clj/game/core/rules.clj
@@ -70,11 +70,11 @@
                 (not (and (has-subtype? card "Priority")
                           (get-in @state [side :register :spent-click])))) ; if priority, have not spent a click
          ;; Wait on pay-sync to finish before triggering instant-effect
-         (when-completed (pay-sync state side card (if ignore-cost 0 total-cost) {:action :play-instant})
-                         (if-let [cost-str async-result]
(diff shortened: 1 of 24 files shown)


src/clj/game/core/rules.clj
```diff
@@ -463,27 +463,27 @@
...
   ([state side card args] (trash state side (make-eid state) card args))
   ([state side eid {:keys [zone type] :as card} {:keys [unpreventable cause suppress-event] :as args}]
-   (when-completed (prevent-trash state side card eid args)
-                   (if-let [c (first (get-in @state [:trash :trash-list eid]))]
-                     (resolve-trash state side eid c args)
-                     (effect-completed state side eid)))))
+   (wait-for (prevent-trash state side card eid args)
+             (if-let [c (first (get-in @state [:trash :trash-list eid]))]
+  (resolve-trash state side eid c args)
+  (effect-completed state side eid)))))
```
NoahTheDuke: Looks like this and the line above it are indented wrong

PR #3654, opened 2018-06-20
WIP: Run Timing Overhaul

NoahTheDuke:
Issue: Currently, the way runs are handled is very loose, which means that both the timings are messed up and the when or how things can be done is unclear. This is both a refactoring and an expansion of the current run timings to better map to the official run timing chart as given in the latest (4.1) FAQ and Rules Reference.

In my limited testing, it works as expected but is very ugly. It needs a lot of refining, in both UI and UX, but once it's done, this will open the doors to allow for greater and more sustainable card automation in the future. For example, automating `on encounter` would be significantly more feasible than previously as it has a clearly defined place in the timing of the run.

Issues: Every single run test is broken, nearly every single card that modifies run timing is broken, cards that replace access or modify run positioning are broken, and uh... I am not looking forward to writing all of the tests necessary to handle this. Shit's gotta get done, tho.

Fuck it, yolo.


src/clj/game/core/runs.clj
```diff
@@ -4,30 +4,224 @@
          gain-run-credits update-ice-in-server update-all-ice
          get-agenda-points gain-agenda-point optional-ability
          get-remote-names card-name can-access-loud can-steal?
-         prevent-jack-out card-flag? can-run?)
+         prevent-jack-out card-flag? can-run? run-approach-ice
+         jack-out successful-run end-run update-step corp-phase-53)
 
 ;;; Steps in the run sequence
+(defn approach-server
+  "[5]: the runner approaches the attacked server"
```
Saintis: Is there a specific reason for these being in reverse order?

src/clj/game/core/runs.clj
```diff
@@ -4,30 +4,224 @@
...
+  [state side eid]
+  ;; "When approached" conditionals meet their trigger condition
+  (prn "approach-server")
```
Saintis: I assume these will be removed when the rewrite is done yeah?

src/clj/game/core/runs.clj
```diff
@@ -4,30 +4,224 @@
...
+  (trigger-event-sync state side (make-eid state) :approach-server (-> @state :run :server))
```
Saintis: I'm not sure how trigger-event-sync works, but should this not be waited for before proceeding?

src/clj/game/core/runs.clj
```diff
@@ -4,30 +4,224 @@
...
+  ;; [5.2] the runner chooses whether to jack out or declare a successful run
+  ; (show-wait-prompt state :corp "Runner actions. Paid abilities")
+  (wait-for (resolve-ability
+              state :runner
+              {:prompt "Complete a successful run, or jack out? Paid abilities"
```
Saintis: Is the "Paid abilities" text to show that you can use paid abilities?

src/clj/game/core/runs.clj
```diff
@@ -4,30 +4,224 @@
...
+                            (if (= "Jack out" target)
+                              (update-step state side eid :step5 :step7)
+                              (effect-completed state side eid)))}
+              nil nil)
+            (when-not (:ended (:run @state))
+              ; (show-wait-prompt state :runner "Corp may rez non-ice. Paid abilities")
+              (wait-for (resolve-ability
+                          state :corp
+                          {:prompt "Act before access?"
+                           :choices ["Action before access" "No more actions"]
```
Saintis: I like the current way of handling the Corp's "act after commit to access" window, those this mean that you will end up with an extra prompt after runner selects "successful run" in the case the corp does not want to act?

Saintis:
One more thing, can we name `:corp-phase-53` something more descriptive and specific-phase-number-agnostic?

src/clj/game/core/runs.clj
NoahTheDuke: I borrowed this code from Neal and he had them in this order as they called each other directly. Could easily be reordered now.

src/clj/game/core/runs.clj
NoahTheDuke: Yeah, it's not super graceful right now. I wanted to fully remove the `corp-phase-53` system, but _a lot_ of stuff rests on it, so this was my stop-gap.

NoahTheDuke:
I _do_ plan on finishing this some day, but right now it's too far behind and is built on an out-dated rules doc. The next iteration will be based on the Comprehensive Rules run timing chart.

PR #3656, opened 2018-06-21
RaR: Cradle

NoahTheDuke:
Implements Cradle, with a test.

### DIFF
--- a/src/clj/game/cards/icebreakers.clj
+++ b/src/clj/game/cards/icebreakers.clj
@@ -470,6 +470,20 @@
                     {:abilities [(break-sub 1 1 "Barrier")
                                  (strength-pump 1 1)]})
 
+   "Cradle"
+   {:abilities [(break-sub 2 0 "Code Gate")]
+    :events {:card-moved {:silent (req true)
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/icebreakers.clj
```diff
@@ -470,6 +470,20 @@
...
                                  (strength-pump 1 1)]})
 
+   "Cradle"
+   {:abilities [(break-sub 2 0 "Code Gate")]
+    :events {:card-moved {:silent (req true)
+                          :req (req (and (= "Runner" (:side target))
+                                         (= [:hand] (or (:zone target)
+                                                        (:previous-zone target)))))
+                          :effect (effect (update-breaker-strength card))}
+             :runner-draw {:silent (req true)
```
Saintis: Does runner draw not trigger `:card-moved`?

NoahTheDuke: It does not.

nealterrell: I don't know why, but `draw` doesn't use `move`; it manually updates the state vectors.

PR #3657, opened 2018-06-21
RaR: Tycoon

NoahTheDuke:
Implements Tycoon, mostly. Includes a half-hearted test, that honestly isn't great. Someone come fix this.

### DIFF
--- a/src/clj/game/cards/icebreakers.clj
+++ b/src/clj/game/cards/icebreakers.clj
@@ -1003,7 +1003,7 @@
                     {:abilities [(break-sub 1 2 "Barrier" (effect (update! (assoc-in card [:special :tycoon-used] true))))
                                  (strength-pump 2 3)]
                      :events {:pass-ice {:req (req (get-in card [:special :tycoon-used]))
-                                         :effect (effect (update! (dissoc-ic card [:special :tycoon-used]))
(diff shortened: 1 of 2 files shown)


Saintis:
You could crib code from faerie and hook up an event to pay the corp when you use the break subs ability.

NoahTheDuke:
Hell yeah, that's the stuff. Thanks, I'll set that up.

src/clj/game/cards/icebreakers.clj
```diff
@@ -998,6 +998,13 @@
                     {:abilities [(break-sub 1 1 "Code Gate")
                                  (strength-pump 1 1)]})
 
+   "Tycoon"
+   (auto-icebreaker ["Barrier"]
+                    {:abilities [(break-sub 1 2 "Barrier" (effect (update! (assoc-in card [:special :tycoon-used] true))))
+                                 (strength-pump 2 3)]
+                     :events {:pass-ice {:req (req (get-in card [:special :tycoon-used]))
+                                         :effect (effect (gain-credits :corp 2))}}})
```
nealterrell: Think you need to dissoc your `:tycoon-used` key here so the Corp doesn't keep gaining on each passed ice.

PR #3658, opened 2018-06-21
RaR: Mache

NoahTheDuke:
Implements Mache with a test. Currently fails because NRDB hasn't updated with the new cards.

### DIFF
--- a/src/clj/game/cards/hardware.clj
+++ b/src/clj/game/cards/hardware.clj
@@ -620,8 +620,8 @@
                  :counter-cost [:power 3]
                  :effect (effect (draw :runner 1))}]
     :events {:runner-trash {:once :per-turn
-                            :async true
                             :req (req (and (card-is? target :side :corp)
+                                           (:access @state)
                                            (:trash target)))
                             :effect (effect (system-msg (str "places " (:trash target) " power counters on Mâché"))
                                             (add-counter card :power (:trash target)))}}}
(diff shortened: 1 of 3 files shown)


src/clj/game/cards/hardware.clj
```diff
@@ -614,6 +614,18 @@
...
     :effect (effect (set-prop card :rec-counter (count (:ices (get-in @state [:corp :servers :hq])))))}
 
+   "Mâché"
+   {:abilities [{:label "Draw 1 card"
+                 :msg "draw 1 card"
+                 :counter-cost [:power 3]
+                 :effect (effect (draw :runner 1))}]
+    :events {:runner-trash {:once :per-turn
+                            :async true
+                            :req (req (and (card-is? target :side :corp)
```
Saintis: How restrictive is the `:runner-trash` event? Does it only fire when card is accessed?

I'm wondering if this will trigger if you use PolOp, Apocalypse or Keyhole to trash a card.

Might want to add that case to the test as well.

NoahTheDuke: Ah yikes, you're probably right. Hmm, I'll have to think about how to handle this.

nealterrell: In theory, `(:access @state)` should only be true while a card is being accessed. In theory.

NoahTheDuke:
Oh duh! I’ll add one.

PR #3660, opened 2018-06-22
RaR: Insight + typo fix in Yusuf/Musaazi

presheaf:
I'm happy to add a test if needed, just let me know. Also sorry about sneaking in a typo fix (missing space).

### DIFF
--- a/src/clj/game/cards/events.clj
+++ b/src/clj/game/cards/events.clj
@@ -1051,6 +1051,19 @@
     :choices (req runnable-servers)
     :effect (effect (run target nil card))}
 
+   "Insight"
+   {:async true
+    :implementation "Rearranging R&D is not optional"
+    :effect (req
+             (let [from (take 4 (:deck corp))]
+               (if (pos? (count from))
+                 (do (show-wait-prompt state :runner "Corp to rearrange the top cards of R&D")
+                     (wait-for
+                      (resolve-ability state :corp (reorder-choice :corp from) card targets)
+                      (do (clear-wait-prompt state :runner)
+                          (system-msg state :runner (str " reveals (top:) " ; here (get-in @state ...) is needed to refresh ref. to deck after reordering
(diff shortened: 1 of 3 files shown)


NoahTheDuke:
Always love a good typo fix. This looks good! I'd much appreciate a test, but I understand how it is.

src/clj/game/cards/events.clj
```diff
@@ -1051,6 +1051,19 @@
     :choices (req runnable-servers)
     :effect (effect (run target nil card))}
 
+   "Insight"
+   {:async true
+    :implementation "Rearranging R&D is not optional"
+    :effect (req
+             (let [from (take 4 (:deck corp))]
+               (if (pos? (count from))
```
Saintis: This could be `when` if you don't have an `else` clause.

src/clj/game/cards/events.clj
```diff
@@ -1051,6 +1051,19 @@
     :choices (req runnable-servers)
     :effect (effect (run target nil card))}
 
+   "Insight"
+   {:async true
+    :implementation "Rearranging R&D is not optional"
```
Saintis: What does this mean? Is the Corp not allowed to re-arrange cards into the same arrangement?

src/clj/game/cards/events.clj
```diff
@@ -1051,6 +1051,19 @@
...
+    :effect (req
+             (let [from (take 4 (:deck corp))]
+               (if (pos? (count from))
+                 (do (show-wait-prompt state :runner "Corp to rearrange the top cards of R&D")
```
Saintis: Might want to say how many cards Corp is rearranging, so "top `(count from)` cards".

src/clj/game/cards/events.clj
presheaf: What I mean is that the card says the Corp 'may' rearrange, but I think asking the corp whether they want to will only cause confusion, so I assume the Corp wants to rearrange and rely on them to keep the order as-is if they don't want to. Maybe this note should just be deleted - seems like it causes confusion too. :)

PR #3668, opened 2018-06-25
RaR: API-S Keeper Isobel

jwarwick:


src/clj/game/cards/assets.clj
```diff
@@ -185,6 +185,22 @@
                                       card nil)))}}
       :abilities [ability]})
 
+   "API-S Keeper Isobel"
+    (letfn [(counters-available? [state] (some #(pos? (get-counters % :advancement)) (all-installed state :corp)))]
+      {:flags {:corp-phase-12 (req (counters-available? state))}
+       :abilities [{:req (req (and (:corp-phase-12 @state)
+                                   (counters-available? state)))
+                    :once :per-turn
+                    :label "Remove an advancement token from an installed card and gain 3 [Credits]"
```
nealterrell: I would prefer this shorter, and with an indication that it's a start-of-turn ability. "Remove an advancement token (start of turn)" would be my suggestion.

src/clj/game/cards/assets.clj
```diff
@@ -185,6 +185,22 @@
...
+                    :prompt "Select a card to remove an advancement token from"
+                    :choices {:req #(and (pos? (get-counters % :advancement))
+                                         (installed? %))}
+                    :effect (req (let [cnt (get-counters target :advancement)]
+                                   (set-prop state side target :advance-counter (dec cnt))
+                                   (gain-credits state :corp 3)
+                                   (system-msg state :corp (str "uses API-S Keeper Isobel to remove an advancement token from "
+                                                                (:title target) " and gains 3 [Credits]"))))}]})
```
nealterrell: This should use `card-str` instead of `:title`.

PR #3669, opened 2018-06-25
RaR: Building Blocks

jwarwick:


NoahTheDuke:
Might want to check that both costs are ignored.

src/clj/game/cards/operations.clj
```diff
@@ -239,6 +239,14 @@
     :msg "do 7 meat damage"
     :effect (effect (damage eid :meat 7 {:card card}))}
 
+   "Building Blocks"
+   {:choices {:req #(and (= (:side %) "Corp")
+                         (ice? %)
```
Saintis: Technically the card only specified it has to be a Barrier, that it is also a piece of ICE is an assumption on our part.

test/clj/game_test/cards/operations.clj
```diff
@@ -180,6 +180,20 @@
...
+    (new-game (default-corp ["Building Blocks" "Ice Wall"])
+              (default-runner))
+    (core/gain state :corp :credit 1)
+    (play-from-hand state :corp "Building Blocks")
+    (is (= 1 (:credit (get-corp))) "Corp starts with 1 credit")
+    (prompt-select :corp (find-card "Ice Wall" (:hand (get-corp))))
+    (prompt-choice :corp "New remote")
+    (let [iw (get-ice state :remote1 0)]
+      (is (= 1 (:credit (get-corp))) "Corp spent no credits")
+      (is (:rezzed (refresh iw)) "Ice Wall is installed and rezzed"))))
```
Saintis: I'd also like a test where you try to select a non-Barrier, and maybe a non-Barrier, non-ice as well.

test/clj/game_test/cards/operations.clj
```diff
@@ -180,6 +180,20 @@
...
     (is (= 7 (count (:hand (get-corp)))) "Drew 2 cards")))
 
+(deftest building-blocks
+  ;; Building Blocks - install and rez a barrier from HQ at no cost
+  (do-game
+    (new-game (default-corp ["Building Blocks" "Ice Wall"])
+              (default-runner))
+    (core/gain state :corp :credit 1)
+    (play-from-hand state :corp "Building Blocks")
+    (is (= 1 (:credit (get-corp))) "Corp starts with 1 credit")
```
Saintis: I'd like this to check that Building Blocks cost 5 to play. Same effect, but different error message.

PR #3671, opened 2018-06-25
RaR: Under the Bus

jwarwick:


src/clj/game/cards/operations.clj
```diff
@@ -1698,6 +1698,16 @@
                                               :effect (effect (corp-install target nil))}
                                              card nil)))}
 
+   "Under the Bus"
+   {:req (req (and (:accessed-cards runner-reg)
```
danhut: This should be `runner-reg-last`?

src/clj/game/cards/operations.clj
```diff
@@ -1698,6 +1698,16 @@
...
+                   (not-empty (filter #(has-subtype? % "Connection") (all-active-installed state :runner)))))
+    :prompt "Choose a connection to trash"
+    :choices {:req #(and (= (:side %) "Runner") (has-subtype? % "Connection") (installed? %))}
```
Saintis: Technically card specifies "connection resource". I'd like to check for both in case a connection something or other is printed.

PR #3672, opened 2018-06-25
RaR: Broad Daylight

jwarwick:


src/clj/game/cards/agendas.clj
```diff
@@ -291,6 +291,23 @@
...
                :msg "make the Runner lose 2 tags"}}
 
+   "Broad Daylight"
+   {:prompt "Take 1 bad publicity?"
+    :choices ["Yes" "No"]
+    :async true
+    :effect (req (when (= target "Yes")
+                   (wait-for (gain-bad-publicity state :corp 1)
+                             (system-msg state :corp "used Broad Daylight to take 1 bad publicity")))
+                 (let [bp (get-in @state [:corp :bad-publicity] 0)]
```
Saintis: Is this going to include the BP from the above wait-for? Even if you have that trace asset and chose to not pay into the trace? Or Raymond Flint etc.

test/clj/game_test/cards/agendas.clj
```diff
@@ -362,6 +362,41 @@
...
+      (is (= 1 (get-counters (get-scored state :corp 0) :agenda)) "Should gain 1 agenda counter")
+      (play-and-score state "Broad Daylight")
+      (prompt-choice :corp "No")
+      (is (= 1 (:bad-publicity (get-corp))) "Corp doesn't gain bad pub")
+      (is (= 1 (get-counters (get-scored state :corp 1) :agenda)) "Should gain 1 agenda counter")
+      (play-and-score state "Broad Daylight")
+      (prompt-choice :corp "Yes")
+      (is (= 2 (:bad-publicity (get-corp))) "Corp gains 1 bad pub")
+      (is (= 2 (get-counters (get-scored state :corp 2) :agenda)) "Should gain 2 agenda counters")
+      ))
```
Saintis: PANIC! TRAILING CLOSING BRACKETS!

src/clj/game/cards/agendas.clj
jwarwick: Yes, I believe it should wait for everything that gets triggered from the bad-pub event to complete. Then the `state` should be up to date.

test/clj/game_test/cards/agendas.clj
jwarwick: /facepalm

src/clj/game/cards/agendas.clj
nealterrell: @Saintis actually has the right idea. Anything that's not inside the `wait-for` will plow ahead if any `gain-bad-publicity` listeners are themselves async. I think there are no such interactions with gaining BP, but for safety the `let` should also be inside the wait-for.

NoahTheDuke: Does Broadcast Square affect this?

jwarwick: Ah, right, I see. It's not actually blocking, we're just keeping track of what to follow up with when it's done.

PR #3513, opened 2018-05-15
Changes order of trigger-event info/req checking AND MORE

nicohasa:
This PR does the following:

- Makes `trigger-event`, `trigger-event-sync` and `trigger-event-simult` swap in the event as happened as it's first order of business
- Changes `first-event?` to handle this change correctly (and fixes all bugs with this I could find, hopefully all)
- Makes `trigger-event-sync` and `trigger-event-simult` check their requirements *only* when the event is triggered, so they do not take into account changes to game state. The cards Activist Support, Underworld Contact and System Outage if-tests moved to inside the effect, to reflect this.
- Makes corp-install use `trigger-event-sync` instead of `trigger-event` to fix problem with Asa. This also fixes some other problems.
- While I was at it, makes `corp-install` check that `:pre-corp-install` and `:server-created` are actually done before continuing. None of these do anything that should be less than instant, but you never know, some of them might do something requiring prompts in the future.
- Made test for the Asa + MirrorMorph thing.

Closes #3495 #3409 #3393 and maybe more???

PR #3682, opened 2018-06-26
RaR: some Anarch cards

Saintis:
Implements some Anarch cards: Gnat and Guinea Pig.

### DIFF
--- a/src/clj/game/cards/events.clj
+++ b/src/clj/game/cards/events.clj
@@ -874,6 +874,12 @@
    "Government Investigations"
    {:flags {:psi-prevent-spend (req 2)}}
 
+   "Guinea Pig"
+   {:msg "trash all cards in the grip and gain 10 [credits]"
+    :effect (req (doseq [c (:hand runner)]
+                   (trash state :runner c {:unpreventable true}))
+                 (gain-credits state :runner 10))}
+
    "Hacktivist Meeting"
    {:implementation "Does not prevent rez if HQ is empty"
     :events {:rez {:req (req (and (not (ice? target))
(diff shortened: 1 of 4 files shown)


src/clj/game/cards/events.clj
```diff
@@ -874,6 +874,12 @@
    "Government Investigations"
    {:flags {:psi-prevent-spend (req 2)}}
 
+   "Guinea Pig"
+   {:msg "trash all cards in the grip and gain 10 [credits]"
+    :effect (req (doseq [c (:hand runner)]
+                   (trash state :runner c {:unpreventable true}))
```
jwarwick: Does `trash` need to be `async`?

jwarwick: Can't think of anything, but I've been seeing crashes in `Compile` and `Street Peddler` when their `trash` calls aren't `async`, so thought it might apply here as well.

Saintis: Not making it async since that will need some recursive trashing to trash cards one-by-one when they should be trashed simultaneously. If it causes issues / crashes I can revisit it.

Should be good to merge now.

src/clj/game/cards/identities.clj
```diff
@@ -738,6 +738,11 @@
                                                               " [Credits] from the rez of " (:title current-ice)))
                                   (swap! state update-in [:bonus] dissoc :cost))))}]}
 
+   "Nathaniel \"Gnat\" Hall: One-of-a-Kind"
+   {:events {:runner-turn-begins {:msg "gain 1 credit"
+                                  :req (req (> 3 (count (:hand runner))))
```
nealterrell: We recently (#3442 evolving into #3513) changed how `:req` works with event handlers to more properly align with game rules. The net effect is that `:req` is for deciding whether the handler exists at all, notably for cards that read "If X, then when Y, do Z". If a card instead reads "When X, then if Y, do Z", we've moved the Y condition into the effect of the card.

PR #3686, opened 2018-06-27
TEST: Alice Merchant

### LINKED ISSUE #3684: Alice Merchant trash ability
Currently, Alice Merchant's ability forces the corp trash the card from HQ face-up.  I'm pretty sure this isn't correct (as far as I can tell it's supposed to be face down), and it's hard to work around.

NoahTheDuke:
~Alice Merchant was forcing the corp to trash from hand, which wouldn't set `:seen`. This led to the card being faceup in Archives when using an access replacement effect. I've fixed it so that the card is discarded face down first and foremost, before access is checked.~

Now it's just a good test for Alice.

Fixes #3684

### DIFF
--- a/src/clj/game/cards/identities.clj
+++ b/src/clj/game/cards/identities.clj
@@ -124,8 +124,8 @@
                                  :player :corp
                                  :choices (req (:hand corp))
                                  :msg "force the Corp to trash 1 card from HQ"
-                                 :effect (effect (clear-wait-prompt :runner)
-                                                 (trash :corp eid (assoc target :seen false) nil))}
+                                 :effect (effect (trash :corp target)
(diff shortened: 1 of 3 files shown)


Saintis:
How come this didn't work without a fix? I would expect any card trashed from HQ to be unseen unless specified to be otherwise.

NoahTheDuke:
I don’t know. I need to do more research and see if this should be generalized.

Saintis:
I'm still not convinced we need to "fix" Alice

NoahTheDuke:
I tested without the "fix", and the cards are facedown. This'll just be a test PR now.

nealterrell:
Thanks for doing the Legwork here.

PR #3690, opened 2018-06-28
RaR: Reboot

jwarwick:
Not positive I'm referencing `state` correctly in the recursive function.


NoahTheDuke:
Doesn't look wrong, but that doesn't mean it isn't, lol.

src/clj/game/cards/events.clj
```diff
@@ -1544,6 +1544,30 @@
                  (doseq [c (-> @state :runner :temp-nvram)]
                    (host state side (get-in @state [:runner :identity]) c {:facedown true}))))}
 
+   "Reboot"
+   (letfn [(install-cards [state side eid card to-install]
+             (if (empty? to-install)
+               (do
+                 (move state side (find-latest state card) :rfg)
+                 (effect-completed state side eid))
+               (let [f (first to-install)]
```
Saintis: You might be able to make this an `if-let` and therefore save a layer of indentation.

src/clj/game/cards/events.clj
```diff
@@ -1544,6 +1544,30 @@
...
+                 (system-msg state :runner (str "uses Reboot to install " (card-str state f) " facedown"))
```
nealterrell: Can we condense this into a single list of all 5 chosen cards? 5 log messages seems excessive.

src/clj/game/cards/events.clj
```diff
@@ -1544,6 +1544,30 @@
...
+                 (wait-for (runner-install state :runner f {:facedown true :no-msg true})
+                           (install-cards state side eid card (rest to-install))))))]
+     {:req (req archives-runnable)
+      :effect (effect
+                (run :archives
+                     {:req (req (= target :archives))
+                      :replace-access
+                      {:prompt "Choose up to five cards to install"
```
nealterrell: This effect should be `:mandatory true`.

PR #3691, opened 2018-06-29
RaR: Saraswati Mnemonics: Endless Exploration

Zeofar:
Imeplements Saraswati Mnemonics: Endless Exploration with test.

Worth noting that this carries the same issue as Mushin No Shin, allowing the Corp to Rez/Score the affected card as soon as they end their turn.

### DIFF
--- a/src/clj/game/cards/identities.clj
+++ b/src/clj/game/cards/identities.clj
@@ -890,6 +890,38 @@
                                 (trigger-event state side :ice-subtype-changed ice)))}]
     :events {:run-ends nil}}
 
+   "Saraswati Mnemonics: Endless Exploration"
+   (letfn [(install-card [chosen]
+                        {:prompt "Select a remote server"
+                         :choices (req (conj (vec (get-remote-names state)) "New remote"))
+                         :async true
+                         :effect (req (corp-install state side (assoc chosen :advance-counter 1) target)
+                                      (effect-completed state side eid)
+                                      (let [tgtcid (:cid chosen)]
+                                        (register-turn-flag! state side
+                                                             card :can-rez
+                                                             (fn [state side card]
+                                                               (if (= (:cid card) tgtcid)
+                                                                 ((constantly false) (toast state :corp "Cannot rez due to Saraswati Mnemonics: Endless Exploration." "warning"))
+                                                                 true)))
+                                        (register-turn-flag! state side
+                                                             card :can-score
+                                                             (fn [state side card]
+                                                               (if (and (= (:cid card) tgtcid)
+                                                                        (>= (get-counters card :advancement) (or (:current-cost card) (:advancementcost card))))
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/identities.clj
```diff
@@ -890,6 +890,38 @@
                                 (trigger-event state side :ice-subtype-changed ice)))}]
     :events {:run-ends nil}}
 
+   "Saraswati Mnemonics: Endless Exploration"
+   (letfn [(install-card [chosen]
+                        {:prompt "Select a remote server"
+                         :choices (req (conj (vec (get-remote-names state)) "New remote"))
+                         :async true
+                         :effect (req (corp-install state side (assoc chosen :advance-counter 1) target)
+                                      (effect-completed state side eid)
```
jwarwick: You should probably move the `corp-install` to the end of the `install-card` function, then you can drop the `effect-completed`. Unless `register-turn-flag!` gets unset during the install.

Something like:
```
 (req (let [tgt...]
          ...
         )
        (corp-install state side eid (assoc ...) target))
```

Note passing the `eid` into `corp-install`, that's what lets you drop the `effect-completed`.

nealterrell: John has the right idea. Currently, you will trigger `corp-install` (which will show a "Choose a server dialog"), and then because you are not `wait-for`-ing that call, you will mark your ability as resolved immediately. It's not a huge deal because this is an `:ability` and not an event handler, but let's do it proper like! :)

src/clj/game/cards/identities.clj
```diff
@@ -890,6 +890,38 @@
...
+                                                                 true)))
+                                        (register-turn-flag! state side
+                                                             card :can-score
+                                                             (fn [state side card]
+                                                               (if (and (= (:cid card) tgtcid)
+                                                                        (>= (get-counters card :advancement) (or (:current-cost card) (:advancementcost card))))
+                                                                 ((constantly false) (toast state :corp "Cannot score due to Saraswati Mnemonics: Endless Exploration." "warning"))
+                                                                 true)))))})]
+   {:abilities [{:async true
+                 :label "[Click], 1 [Credits]: Install a card from HQ in a remote server, then place 1 advancement token on it. You cannot score or rez that card until your next turn begins."
```
nealterrell: The `:label` won't be shown since this card only has 1 ability, but to follow our loose standard: labels should not be a word-for-word copy of the card text (which the player can already see), but a brief outline of what the ability does. Something like "Install a card from HQ" would be fine. (Also note that your `:cost` vector will auto-generate the click and credit icons on the label, if the label were shown.)

src/clj/game/cards/identities.clj
Zeofar: Refactored as described, but the `eid` argument got the compiler to spit errors. I tried just moving the position of `corp-install` without it and everything *seems* to work as intended. Is this alright or is there potential for a hidden issue?

PR #3694, opened 2018-06-29
Rar: Patchwork

nealterrell:
This implementation upgrades the `runner-install` and `play-instant` functions so they await their "`:pre-X`" events before calculating costs to play/install. This allows Patchwork to use a prompt to ask the Runner to trash a card to lower costs. This prompt is only shown if there are at least 2 cards in hand (including the card-to-play). Clicking Done instead of a card will decline the discount.

Compared with @nicohasa's alternative idea -- in which Patchwork gets an ability that you manually click prior to installing/playing your desired card -- this solutions has both pros and cons:

1. Can't "forget" to apply the discount -- you're prompted in the middle of your play/install process.
2. Works with all sources of installs/plays (Same Old Thing, Clone Chip, or straight from grip)
3. Prompts on _every_ play/install until it is used. That requires 1 extra click for each play that you don't want the discount. 
    a. Hard to know if the average player will more frequently want to use or decline this card's discount. My gut says use. In this solution, we require 1 mouse click per decline; in the alternative, it's 1 mouse click per use. 

Open to discussion, and for abandoning this attempt if it's more work for the players.

### DIFF
--- a/src/clj/game/cards/hardware.clj
+++ b/src/clj/game/cards/hardware.clj
@@ -796,6 +796,30 @@
                           :effect (effect (update! (dissoc card :Omnidrive-prog))
                                           (use-mu (:memoryunits target)))}}}
 
+   "Patchwork"
+   (letfn [(patchwork-discount [cost-type bonus-fn]
+             {:async true
+              :req (req (and (not (get-in @state [:per-turn (:cid card)])) ; manual once-per-turn check
+                             (= "Runner" (:side target))
+                             ;; The card being installed/played is still in the hand; we therefore need at least 2
+                             ;; cards for this effect to make sense.
+                             (< 1 (count (:hand runner)))))
+              :effect (req (let [playing target]
+                             (continue-ability
+                               state side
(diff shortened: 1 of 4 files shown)


src/clj/game/cards/hardware.clj
```diff
@@ -796,6 +796,30 @@
...
                                           (use-mu (:memoryunits target)))}}}
 
+   "Patchwork"
+   (letfn [(patchwork-discount [cost-type bonus-fn]
+             {:async true
+              :req (req (and (not (get-in @state [:per-turn (:cid card)])) ; manual once-per-turn check
+                             (= "Runner" (:side target))
+                             ;; The card being installed/played is still in the hand; we therefore need at least 2
+                             ;; cards for this effect to make sense.
+                             (< 1 (count (:hand runner)))))
```
nicohasa: If I install something via SMC or play an event via Same Old Thing with one card in hand, Patchwork *should* still trigger

nicohasa: (Also, Patchwork won't give +1 Memory atm)

PR #3695, opened 2018-06-29
RaR: Patchwork (Improved!)

nicohasa:
Built upon #3694, but fixed so it does not ask you every time, just after you've clicked patchwork.

Still gives you the option to cancel, and will then keep the ability until you use it.

Removes the ability between turns so you can't "save" it for a turn, then use it twice in a row.

### DIFF
--- a/src/clj/game/cards/hardware.clj
+++ b/src/clj/game/cards/hardware.clj
@@ -799,37 +799,34 @@
    "Patchwork"
    (letfn [(patchwork-discount [cost-type bonus-fn]
              {:async true
-              :req (req (and (= "Runner" (:side target))
-                             ;; The card being installed/played is still in the hand; we therefore need at least 2
-                             ;; cards for this effect to make sense.
-                             (< 0 (count (remove (partial same-card? target) (:hand runner))))))
+              :req (req (and (get-in card [:special :patchwork])
+                             (= "Runner" (:side target))
+                             ;; We need at least one card (that is not the card played) in hand
+                             (not-empty (remove (partial same-card? target) (:hand runner)))))
               :effect (req (let [playing target]
                              (continue-ability
                                state side
                                {:prompt (str "Trash a card to lower the " cost-type " cost of " (:title playing) " by 2 [Credits].")
+                                :priority 2
                                 :choices {:req #(and (in-hand? %)
                                                      (= "Runner" (:side %))
                                                      (not (same-card? % playing)))}
                                 :msg (msg "trash " (:title target) " to lower the " cost-type " cost of "
                                           (:title playing) " by 2 [Credits]")
-                                :effect (req (trash state side target {:unpreventable true})
-                                             (bonus-fn state side [:credit -2])
-                                             (unregister-events state side card))
+                                :effect (effect (trash target {:unpreventable true})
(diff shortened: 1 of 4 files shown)


src/clj/game/cards/hardware.clj
```diff
@@ -796,6 +796,41 @@
                           :effect (effect (update! (dissoc card :Omnidrive-prog))
                                           (use-mu (:memoryunits target)))}}}
 
+   "Patchwork"
+   (letfn [(patchwork-discount [cost-type bonus-fn]
+             {:async true
+              :req (req (and (= "Runner" (:side target))
+                             ;; The card being installed/played is still in the hand; we therefore need at least 2
+                             ;; cards for this effect to make sense.
+                             (< 0 (count (remove (partial same-card? target) (:hand runner))))))
```
nealterrell: Let's use `not-empty` instead.

src/clj/game/cards/hardware.clj
```diff
@@ -796,6 +796,41 @@
...
+                                                     (= "Runner" (:side %))
+                                                     (not (same-card? % playing)))}
+                                :msg (msg "trash " (:title target) " to lower the " cost-type " cost of "
+                                          (:title playing) " by 2 [Credits]")
+                                :effect (req (trash state side target {:unpreventable true})
+                                             (bonus-fn state side [:credit -2])
+                                             (unregister-events state side card))
+                                :cancel-effect (effect (effect-completed eid))}
+                               card nil)))})]
+     {:in-play [:memory 1]
```
nealterrell: What, you mean this is a console too? 😁

src/clj/game/cards/hardware.clj
```diff
@@ -796,6 +796,41 @@
...
+      :implementation "Click Patchwork before playing/installing a card."
+      :abilities [{:once :per-turn
+                   :effect (effect (register-events {:pre-play-instant (patchwork-discount "play" play-cost-bonus)
```
nealterrell: So I think this would be better if the events were always "on", and we just set a flag in the card under `[:special :patchwork]` or something. The `:req` for the handler above would then require that flag to be set. I know a lot of cards to the register/unregister trick but it's pretty hacky.

src/clj/game/cards/hardware.clj
Saintis: And update the comment above this line.

src/clj/game/cards/hardware.clj
```diff
@@ -799,37 +799,34 @@
...
-                             (< 0 (count (remove (partial same-card? target) (:hand runner))))))
+              :req (req (and (get-in card [:special :patchwork])
+                             (= "Runner" (:side target))
+                             ;; We need at least one card (that is not the card played) in hand
+                             (not-empty (remove (partial same-card? target) (:hand runner)))))
               :effect (req (let [playing target]
                              (continue-ability
                                state side
                                {:prompt (str "Trash a card to lower the " cost-type " cost of " (:title playing) " by 2 [Credits].")
+                                :priority 2
```
nealterrell: Good call here, to hide prompts from Clone Chip / heap breakers, etc.

Issue #3607, opened 2018-06-07
No wait for Political Dealings


HubGitAccoutn:
When the Political Dealings dialogue comes up, this doesn't lock the board for the Runner as they wait for the Corp to act.

Usually, this wouldn't be too much of an issue as most card draw is in the Corp turn. However, with things like Fisk Investment Seminar, they Runner will on taking clicks unaware that the Corp is thinking or trying to install an agenda somewhere.

When waiting for the Corp choice to install, the Runner should really get a "Waiting for Corp action" type message.

This will have to come up every time the Corp draws while Political Dealings is active though, otherwise the very fact that the dialogue appears will alert the Runner to the fact that the Corp just drew an agenda, even if they don't then choose to reveal and install it. I'd suggest the Corp could maybe get a: "You didn't draw any agendas for Political Dealings" dialogue when no agendas are drawn, with just an "Okay" button or something similar.

NoahTheDuke:
Thanks for the report! We'll have to see the best way to make this happen.

NoahTheDuke:
Note to self: Jinja City Grid has the same issue.

Saintis:
In order to minimise obstructive prompts, I would suggest that Jinja and Political Dealings only do wait-prompts on the runner's turn.

PR #3698, opened 2018-06-30
Show fake prompt from Jinja City Grid when drawing non-ice on runner's turn

### LINKED ISSUE #3577: Laramy Fisk and Jinja City Grid: Fisk ability and access happens twice if ICE is drawn
If Jinja is rezzed, and they draw an ICE from Fisk's ability: I receive the option to use Fisk's ability again (regardless of whether they use Jinja to install the ICE). Also I see the access prompt again immediately after the first access. I observed this behaviour across 4 different matches.

nealterrell:
As recommended by @Saintis  in #3607. Fixes #3577. Should also be applied to Political Dealings.

### DIFF
--- a/src/clj/game/cards/upgrades.clj
+++ b/src/clj/game/cards/upgrades.clj
@@ -469,21 +469,38 @@
                                                               ices grids)
                                                  card nil)))}))]
 
-     {:events {:corp-draw {:req (req (some #(is-type? % "ICE")
-                                           (:most-recent-drawn corp-reg)))
-                           ;; THIS IS A HACK: it prevents multiple Jinja from showing the "choose a server to install into" sequence
+     {:events {:corp-draw {;; THIS IS A HACK: it prevents multiple Jinja from showing the "choose a server to install into" sequence
                            :once :per-turn
                            :once-key :jinja-city-grid-draw
                            :async true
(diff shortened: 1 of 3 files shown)


src/clj/game/cards/upgrades.clj
```diff
@@ -469,21 +469,38 @@
...
+                                            (when (= :runner (:active-player @state))
+                                              (show-wait-prompt state :runner "Corp to resolve Jinja City Grid"))
+                                            (if (not-empty ices)
+                                              (continue-ability state side (choose-ice ices grids) card nil)
+                                              (effect-completed state side eid)))
+                                          ;; else, if it's the runner's turn, show a fake prompt so the runner can't infer that ice weren't drawn
+                                          (= :runner (:active-player @state))
+                                          (continue-ability
+                                            state :corp
+                                            {:prompt "This prompt is being shown so the Runner can't infer that you did not draw any ice"
```
Saintis: This is cute and all, but I think just a prompt saying "You didn't draw any ice to use with Jinja" is just as clear and less informal / silly.

NoahTheDuke: Agreed

nealterrell: Doesn't that wording beg the question of why it isn't shown on the corp's turn?

PR #3702, opened 2018-06-30
RaR: Miss Bones

jwarwick:


src/clj/game/cards/resources.clj
```diff
@@ -1139,6 +1139,17 @@
                  :effect (effect (derez target)
                                  (trash card {:cause :ability-cost}))}]}
 
+   "Miss Bones"
+   {:data {:counter {:credit 12}}
+    :implementation "Credit use restriction not enforced"
+    :abilities [{:counter-cost [:credit 1]
+                 :msg "gain 1 [Credits] for trashing installed cards"
+                 :async true
+                 :effect (req (gain-credits state :runner 1)
```
Saintis: This shouldn't use gain right since it doesn't technically go to runner credit pool right?

src/clj/game/cards/resources.clj
```diff
@@ -1139,6 +1139,17 @@
                  :effect (effect (derez target)
                                  (trash card {:cause :ability-cost}))}]}
 
+   "Miss Bones"
+   {:data {:counter {:credit 12}}
+    :implementation "Credit use restriction not enforced"
+    :abilities [{:counter-cost [:credit 1]
+                 :msg "gain 1 [Credits] for trashing installed cards"
```
Saintis: Might want this to say somethig similar to what Scrubber says, if it doesnt already.

src/clj/game/cards/resources.clj
```diff
@@ -1139,6 +1139,17 @@
...
+                 :async true
+                 :effect (req (gain-credits state :runner 1)
+                              (if (zero? (get-counters (get-card state card) :credit))
+                                (trash state :runner eid card {:unpreventable true})
```
Saintis: If you pass in eid to trash you shouldn't need effect completed right?

src/clj/game/cards/resources.clj
jwarwick: I don't think we have a way to denote that credits are only for a specific usage. `Scrubber` adds them to the credit pool.

src/clj/game/cards/resources.clj
jwarwick: Actually `Scrubber` doesn't mention the `trashing only` part. Probably should update it to add that to the log.

src/clj/game/cards/resources.clj
jwarwick: Correct, that's my understanding. Passing `eid` to `trash` makes it responsible for completing the effect. The other branch of the `if` statement, I'm still responsible for calling `effect-completed`.

src/clj/game/cards/resources.clj
NoahTheDuke: We have `take-credits` for this very use case.

src/clj/game/cards/resources.clj
Saintis: Ah, I missed this was different branches of the `if` statement.

src/clj/game/cards/resources.clj
Saintis: That, or leave it as just "take 1 credit"

src/clj/game/cards/resources.clj
Saintis: Yeah, I was more thinking about the extra `runner-gain-credit` events that `gain-credits` fires off that should be avoided.

PR #3707, opened 2018-07-01
RaR: Hydra

Saintis:
Implements Hydra and adds test.

Label is full text as on the card, but message on sub fire is just the relevant bit of the effect (so _either_ 3 net damage _or_ 1 tag etc.)


nealterrell:
You are a D.R.Y. god.

nealterrell:
A future improvement would be to dynamically change the subroutine labels upon tag gain/removal, but this is good for now :).

PR #3709, opened 2018-07-02
Chronos runner current not rfg

msbeck:
@NoahTheDuke Please review.

Chronos project was RFG'ing runner currents, when the corp actions should trigger first (removal of heap from the game) and _then_ the current is trashed. 

This change keeps the runner's current in their trash by not RFG'ing it. I changed the RFG to happen during after-active-player instead of first-ability.


msbeck:
Scratch this PR. I think rules indicate the way it works already is the way it's supposed to work.

PR #3716, opened 2018-07-04
RaR: Ika

presheaf:
Auto-pump is displayed even when currently encountered ice is not the one Ika is hosted on. This is in part intentional to allow for easier use of Flame-Out, and in part laziness.

### DIFF
--- a/src/clj/game/cards/icebreakers.clj
+++ b/src/clj/game/cards/icebreakers.clj
@@ -709,17 +709,15 @@
    "Ika"
    (auto-icebreaker ["Sentry"]
                     {:abilities [(break-sub 1 2 "Sentry")
-                                 (strength-pump 2 3 :all-run)
+                                 (strength-pump 2 3)
                                  {:label "2 [Credits]: Host Ika on a piece of ICE"
-                                  :effect (req (let [ika (get-card state card)]
-                                                 (resolve-ability state side
-                                                                  {:prompt (msg "Host Ika on a piece of ICE")
-                                                                   :cost [:credit 2]
-                                                                   :choices {:req #(and (ice? %)
-                                                                                        (installed? %)
-                                                                                        (can-host? %))}
-                                                                   :msg (msg "host it on " (card-str state target))
-                                                                   :effect (effect (host target card))} card nil)))}]})
+                                  :prompt (msg "Host Ika on a piece of ICE")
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/icebreakers.clj
```diff
@@ -706,6 +706,21 @@
                                  (break-sub 1 1 "Code Gate")
                                  (strength-pump 1 1)]})
 
+   "Ika"
+   (auto-icebreaker ["Sentry"]
+                    {:abilities [(break-sub 1 2 "Sentry")
+                                 (strength-pump 2 3 :all-run)
```
nealterrell: Why is this pump all-run?

src/clj/game/cards/icebreakers.clj
```diff
@@ -706,6 +706,21 @@
...
+                                 {:label "2 [Credits]: Host Ika on a piece of ICE"
+                                  :effect (req (let [ika (get-card state card)]
+                                                 (resolve-ability state side
```
nealterrell: Can this effect be merged into the main ability? Is the resolve-ability necessary?

src/clj/game/cards/icebreakers.clj
presheaf: No good reason! I'll fix it.

src/clj/game/cards/icebreakers.clj
presheaf: No. The reason I had this was because I started trying to verify that host and target ice were not the same, but this is unneccessary. Removing it.

src/clj/game/cards/icebreakers.clj
```diff
@@ -709,17 +709,15 @@
    "Ika"
    (auto-icebreaker ["Sentry"]
                     {:abilities [(break-sub 1 2 "Sentry")
-                                 (strength-pump 2 3 :all-run)
+                                 (strength-pump 2 3)
                                  {:label "2 [Credits]: Host Ika on a piece of ICE"
```
nealterrell: The cost will now automatically be added to the subroutine label because of the `:cost` vector, so we can remove it from this string.

presheaf: Aha, meaning I can just remove the first part of the `:label`?

PR #3719, opened 2018-07-05
RaR: Gatekeeper

NoahTheDuke:
Based off NEXT Sapphire, along with a test.

Additionally, I made all ice strength update at the end of each turn, so `:current-strength` calculates properly when the req for Gatekeeper checks `:this-turn`.


src/clj/game/cards/ice.clj
```diff
@@ -889,6 +889,39 @@
...
 
+   "Gatekeeper"
+   (let [draw {:async true
+               :prompt "Draw how many cards?"
+               :choices {:number (req 3)
+                         :max (req 3)
+                         :default (req 1)}
+               :msg (msg "draw " target "cards")
+               :effect (effect (draw eid target nil))}
+         reveal-and-shuffle {:async true
```
nealterrell: I don't believe the async is necessary, which would also eliminate all the `effect-completed`s.

src/clj/game/cards/ice.clj
```diff
@@ -889,6 +889,39 @@
...
+                             :cancel-effect (effect (shuffle! :deck)
+                                                    (effect-completed eid))
+                             :msg (msg "add "
+                                       (str (join ", " (map :title targets)))
+                                       " to R&D")}
+         draw-reveal-shuffle {:async true
+                              :label "Draw cards, reveal and shuffle agendas"
+                              :effect (req (wait-for (resolve-ability state side draw card nil)
+                                                     (continue-ability state side reveal-and-shuffle card nil)))}]
+    {:strength-bonus (req (if (= :this-turn (:rezzed card)) 6 0))
```
nealterrell: Is this already part of the engine? Dayum.

NoahTheDuke: I know! I was also pleasantly surprised!

src/clj/game/cards/ice.clj
NoahTheDuke: Even with the prompts?

src/clj/game/cards/ice.clj
```diff
@@ -889,6 +889,35 @@
    "Galahad"
    (grail-ice end-the-run)
 
+   "Gatekeeper"
+   (let [draw {:prompt "Draw how many cards?"
```
nealterrell: I think I was unclear; this `draw` ability definitely needs to be async so Jinja City Grid fully resolves before you move on to the shuffle phase of the subroutine. It's the `reveal-and-shuffle` ability that does not need to be, since moving cards and shuffling are both non-async functions.

NoahTheDuke: Ah okay! Thanks.

PR #3720, opened 2018-07-05
RaR: Giordano Memorial Field

NoahTheDuke:
Implements Giordano with a test.

Also of note is a change to the run timing: previously, there was no check for if a run had been ended between the marking of a successful run and access. However, this card now allows that, so I've added a check in `successful-run-trigger` for if the `(:run @state)` has been marked as ended. If it has, clean up the run. If it hasn't, continue!

### DIFF
--- a/src/clj/game/cards/upgrades.clj
+++ b/src/clj/game/cards/upgrades.clj
@@ -352,7 +352,8 @@
                  :effect (effect (mill :corp :runner 2))}]}
 
    "Georgia Emelyov"
-   {:events {:unsuccessful-run {:req (req (= (first (:server target)) (second (:zone card))))
+   {:events {:unsuccessful-run {:req (req (= (first (:server target))
+                                             (second (:zone card))))
                                 :async true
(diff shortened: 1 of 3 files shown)


src/clj/game/cards/upgrades.clj
```diff
@@ -369,6 +370,31 @@
...
+      :req (req this-server)
+      :effect (req (let [credits (:credit runner)
+                         cost (* 2 (count (:scored runner)))
+                         pay-str (str "Pay " cost " [Credits]")]
+                     (show-wait-prompt state :corp (str "Runner to pay " cost " [Credits]"))
+                     (continue-ability
+                       state :runner
+                       {:player :runner
+                        :async true
+                        :prompt (str pay-str " or end the run?")
```
nealterrell: This should probably indicate that Memorial Field is the source of this choice.

NoahTheDuke: Good call.

PR #3723, opened 2018-07-06
RaR: Thunder Art Gallery

nicohasa:
Renamed tag-runner to gain-tags, added lose-tags,
implemented Thunder Art Gallery.


src/clj/game/cards/resources.clj
```diff
@@ -1973,6 +1974,19 @@
...
+                  :optional {:prompt (msg "Use Thunder Art Gallery to install a card?")
+                             :yes-ability {:choices (req (cancellable (remove #(is-type? % "Event") (:hand runner))))
+                                           :prompt (msg "Which card to install?")
+                                           :effect (req (install-cost-bonus state side [:credit -1])
+                                                        (when (runner-can-install? state side target)
+                                                              (system-msg state side "uses Thunder Art Gallery to install a card.")
+                                                              (runner-install state side target)))}}}]
+     {:events {:runner-lose-tag (assoc ability :req (req (= side :runner)))
+               :runner-prevent (assoc ability :req (req (seq (filter #(some #{:tag} %) targets))))
+               }})
```
NoahTheDuke: Trailing parens.

Saintis:
Does this work correctly when the _Corp_ removes the tag?

Saintis:
Maybe you could to change `lose-tag` to `remove-tag` (mirrors the term used in the game) and make sure that the side flag passed in triggers the correct event `:runner-remove-tag` vs. `:corp-remove-tag`

NoahTheDuke:
We had a discussion about the name on slack with me arguing for `lose-tags`. The reason I argued is that it's developer-facing, so consistency is more important than exactly mirroring the game wording. We have `gain-credits` and `lose-credits`, I feel it should be `gain-tags` and `lose-tags`.

I don't know if that's exactly a good reason, tho. As long as the arguments are the same, it's probably not a problem.

nicohasa:
It *should* not trigger when corp removes tags, but I did not make a test for this. ATM it also triggers once per turn instead of first time per turn (so overinstalling with a new one will let you fire it again etc). Don't have time to fix this now though.

src/clj/game/cards/resources.clj
```diff
@@ -1973,6 +1974,18 @@
     :leave-play (req (remove-watch state :theophilius-bagbiter)
                      (swap! state assoc-in [:runner :hand-size :base] 5))}
 
+   "Thunder Art Gallery"
+   (let [ability {:once :per-turn
```
nealterrell: This card is not once-per-turn, it's first-each-turn. You'll need to use `first-event?` for this.

src/clj/game/cards/resources.clj
```diff
@@ -1973,6 +1974,18 @@
...
+                  :optional {:prompt (msg "Use Thunder Art Gallery to install a card?")
```
nealterrell: I prefer cards like this (optional into card selection) to be a single selection prompt, with the `:cancel-effect` being used as the "No" ability. Players know that clicking Done without having selected a card is the same as "no/cancel". The prompt can then be worded "Install a card with Thunder Art Gallery?".

src/clj/game/cards/resources.clj
```diff
@@ -1973,6 +1974,18 @@
...
+                             :yes-ability {:choices (req (cancellable (remove #(is-type? % "Event") (:hand runner))))
+                                           :prompt (msg "Which card to install?")
+                                           :effect (req (install-cost-bonus state side [:credit -1])
+                                                        (when (runner-can-install? state side target)
+                                                              (system-msg state side "uses Thunder Art Gallery to install a card.")
+                                                              (runner-install state side target)))}}}]
```
nealterrell: `runner-install` is an async operation, so the ability itself should be `:async true` and pass its `eid` to `runner-install`. `effect-completed` will then need to be called both in the false branch of the conditional above (converting to `if`) and in the `:cancel-effect` that we'll have to add.

PR #3724, opened 2018-07-06
Fixes Implementation notes on Rebirthed IDs

### LINKED ISSUE #3722: An ID that came into play with Rebirth is displayed as "Unimplemented"
I think it's because the implemention status is set to `:full` on game start of all cards in @allcards, but that does not affect cards that come into play dynamically during the game.

Saintis:
Also fixes Liza not being an option for DJ Fenris.

Fixes #3722.


src/clj/game/core/flags.clj
```diff
@@ -253,11 +253,13 @@
...
 (defn has-subtype?
-  "Checks if the specified subtype is present in the card."
+  "Checks if the specified subtype is present in the card, ignoring case."
   [card subtype]
-  (or (has? card :subtype subtype)
-      (when-let [persistent-subs (-> card :persistent :subtype)]
-        (includes? persistent-subs subtype))))
+  (letfn [(contains-sub? [card]
+            (when-let [subs (:subtype card)]
+              (includes? (lower-case subs) (lower-case subtype))))]
```
nealterrell: Let us pray that this is the intended meaning of subtypes.

Saintis: Considering that's how icebreakers interact with ice (all subtypes are lower-case on icebreakers and capitalised on ice), I sure hope so 😛

PR #3730, opened 2018-07-08
RaR: Fast Break

NoahTheDuke:
Implemented with a fairly robust test. Adds install reqs to the "Root only" upgrades, as our process for handling them is hacky and UI reliant. This at least future-proofs us while not affecting current functionality.

### DIFF
--- a/src/clj/game/cards/operations.clj
+++ b/src/clj/game/cards/operations.clj
@@ -478,6 +478,43 @@
                                card nil)))}
                 card nil))}
 
+   "Fast Break"
+   {:async true
+    :req (req (-> runner :scored count pos?))
+    :effect
+    (req (let [X (-> runner :scored count)
+               draw {:async true
+                     :prompt "Draw how many cards?"
+                     :choices {:number (req X)
+                               :max (req X)
+                               :default (req 1)}
+                     :msg (msg "draw " target " cards")
+                     :effect (effect (draw eid target nil))}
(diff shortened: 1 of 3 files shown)


Saintis:
Maybe go the full distance and get rid of the `:init` stuff on the root-only upgrades. The new system is much better anyway!

NoahTheDuke:
I don't want to change that stuff until we switch over to the Regeant branch, cuz all cljs work we do adds work to poor @danhut. I made a note to myself about this, I'll definitely revisit it once we switch over.

src/clj/game/cards/operations.clj
```diff
@@ -478,6 +478,43 @@
...
+                               {:prompt "Select a card to install"
+                                :choices {:req #(and (= (:side %) "Corp")
+                                                     (not (is-type? % "Operation"))
+                                                     (in-hand? %)
+                                                     (seq (filter (fn [c] (= server c)) (corp-install-list state %))))}
+                                :effect (req (wait-for
+                                               (corp-install state side target server nil)
+                                               (let [server (if (= "New remote" server)
+                                                              (-> (turn-events state side :corp-install)
+                                                                  ffirst :zone second zone->name)
```
nealterrell: I'm having CADDAADAR flashbacks.

NoahTheDuke: Hahaha I had the same feeling.

PR #3731, opened 2018-07-08
RaR: Formicary

presheaf:
Current implementation is that when rezzed, it checks whether its ability is triggerable (runner is in innermost position of server but has not yet committed to access), and if so, asks the corp whether they want to trigger it. It might be less confusing to make this ability manually triggerable. The net damage taken to not ETR is made unpreventable. This may allow for weird interactions (I think if Guru Davinder is on the table, the Runner should not be able to select '2 net damage' on Formicary at all, whereas the current implementation would allow them to) - please let me know if you can think of anything extra weird here.

### DIFF
--- a/src/clj/game/cards/ice.clj
+++ b/src/clj/game/cards/ice.clj
@@ -1456,6 +1456,33 @@
                    :prompt "Choose an ICE to install from HQ"
                    :effect (req (corp-install state side target (zone->name (first (:server run))) {:no-install-cost true}))}]}
 
+   "Formicary"
+   {:req (req (and (:run @state)
+                   (zero? (:position run))
+                   (not (contains? run :corp-phase-43))
+                   (not (contains? run :successful))))
+    :effect (effect (resolve-ability
+                     {:optional
+                      {:prompt "Move Formicary?"
+                       :yes-ability {:msg "rez and move Formicary. The Runner is now approaching Formicary."
+                                     :effect (req
+                                              (move state side card
+                                                    [:servers (first (:server run)) :ices]
+                                                    {:front true})
+                                              (swap! state assoc-in [:run :position] 1))}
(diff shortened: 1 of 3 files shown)


nealterrell:
Agree w/ your reading on Guru Davinder -- runner should not be able to select the option to take 2 net, but that is such a minor case we can punt on it.

I think your implementation idea is fine.

src/clj/game/cards/ice.clj
```diff
@@ -1456,6 +1456,33 @@
                    :prompt "Choose an ICE to install from HQ"
                    :effect (req (corp-install state side target (zone->name (first (:server run))) {:no-install-cost true}))}]}
 
+   "Formicary"
+   {:req (req (and (:run @state)
+                   (zero? (:position run))
+                   (not (contains? run :corp-phase-43))
+                   (not (contains? run :successful))))
+    :effect (effect (resolve-ability
```
nealterrell: Let's move the `:optional` into the root ability of the card instead of the resolve-ability.

src/clj/game/cards/ice.clj
```diff
@@ -1456,6 +1456,33 @@
...
+                       :yes-ability {:msg "rez and move Formicary. The Runner is now approaching Formicary."
+                                     :effect (req
+                                              (move state side card
+                                                    [:servers (first (:server run)) :ices]
+                                                    {:front true})
+                                              (swap! state assoc-in [:run :position] 1))}
+                       :no-ability {:msg "rez Formicary without moving it"}}}
+                     card nil))
+    :subroutines [{:label "End the run unless the Runner suffers 2 net damage"
+                   :effect (req (wait-for (resolve-ability
```
nealterrell: Subroutines currently don't plug into the async system, but to future proof, the sub should be `:async true` in order to use `wait-for`.

PR #3735, opened 2018-07-09
Make tests explicit + assorted test bugfixes - don't merge (yet)

presheaf:
This is the start of an attempt to make the helper functions in `do-game` friendlier. Currently, a lot of things fail silently, meaning it's easy to make mistakes when writing tests. So if you accidentally mistype a card name, causing an expression to return `nil` and not the card, you may just try to select `nil` in a prompt, doing nothing, and leaving you puzzled as to why nothing is happening. Or maybe you try to call `prompt-choice` when `prompt-card` is correct.

This PR makes `prompt-choice`, `prompt-card`, `prompt-select` do a couple of checks to make sure you're not passing in `nil`s or trying to use them for the wrong prompts. In making this happen, I found a bunch of tests which previously passed without testing what they were intending to and fixed (some of) them, so I feel confident in saying that this actually does prevent some common mistakes.

However, in doing so I've basically doubled the length of `do-game`, and I think it could probably use some rethinking - there's currently quite a few helper methods defined, and I doubt they're all necessary. So I'd love some suggestions on what to discard and what to keep.


Saintis:
On more thing you could do is move all the `prompt` helper functions out of the macro definition and into the util file. That does mean all such functions will need `state` passed to them but it should make error stack traces clearer

presheaf:
Good point. Maybe

NoahTheDuke:
I thumbs up‘d it, but let me voice my desire to see the prompt handlers moved out of the macro and into utils where they can be more easily modified and generalized.

Additionally, as we talked about in Slack, I’d love to see the prompt interactions reduce to something like two functions, `click-prompt` and `click-card`. This would require a lot more work, but even with `prompt-card` taking a card map because it’s actually displaying the titles of card maps behind the scenes, it should be possible to convert that and number prompts to intelligently take strings and turn them into the necessary prompt selections.

presheaf:
Your `click-card` would be what's currently called `prompt-select`, right? If so, combining `prompt-choice` and `prompt-card` into something which takes a string or an int should be very feasible.

NoahTheDuke:
Yes, but I'd like to make it take a string too, for when there aren't multiple copies of a given card available to be clicked.

presheaf:
That makes sense. How would it determine what cards are clickable, though, if a string is passed? Maybe just have it throw an exception if multiple cards with that name exists in the game? Provided the error message is informative, I think that would be fine.

presheaf:
I've added a first pass at `click-card` now and tested it in a couple of random tests. Its current behavior is equivalent to `prompt-select` with more informative error messages if a card map is passed, and tries to find the right card if a string is passed. It can probably be extended to look at a couple more cards (for example, stuff on Supplier probably isn't counted at the moment), but provided it fails sensibly and the performance is not atrocious, I don't think this is crucial.

PR #3733, opened 2018-07-08
RaR: Hangeki, Otoroshi, Mind's Eye, Divide and Conquer, Psych Mike, do-access and The Turning Wheel fixes, and more!

NoahTheDuke:
At the time of opening this PR, it's nothing more than a merging of the #3703 and #3713 PRs with a small bigfix thrown in. I plan on adding the other two cards soon. Both should be relatively simple given the wonderful work @presheaf did to de-run accesses.

Once those two cards are finalized and it's all approved, the two linked PRs can be closed and this one will be the one to merge.

EDIT: The two cards are complete and all systems are a-go!

Closes #3703 
Closes #3713

PR #3736, opened 2018-07-09
Mwanza City Grid, handle accessing 0 cards

jwarwick:


NoahTheDuke:
This is actually handled by #3733! But if we can merge this first that’ll be a good stop gap

PR #3738, opened 2018-07-10
RaR: Thimblerig

presheaf:
Does not restrict usage of swap ability at all, so it can be triggered outside start of turn, more than once per turn, or outside ice being passed. I did this because swapping ice manually is notoriously annoying, so it seemed best to make misclicks or missed triggers easily fixable, but can add restrictions if desired. 

Does not prompt for use after pass, which is good to keep runs easy but bad if people rush ahead and do irreversible stuff without noticing that Corp can trigger, so who knows. Does flag phase 1.2 if it rezzed and has a legal target.

### DIFF
--- a/src/clj/game/cards/ice.clj
+++ b/src/clj/game/cards/ice.clj
@@ -2060,6 +2060,16 @@
                                     :msg (msg "force the Runner to lose " (:tag runner) " [Credits]")
                                     :effect (effect (lose-credits :runner (:tag runner)))})]}
 
+   "Thimblerig"
+   {:flags {:corp-phase-12 (req true)}
+    :implementation "Does not restrict usage of swap ability to start of turn or after pass"
+    :abilities [{:label "Swap Thimblerig with a piece of ice"
+                 :prompt "Swap Thimblerig with a piece of ice"
+                 :choices {:req ice?
+                           :not-self (req (:cid card))}
+                 :effect (effect (swap-ice card target))}]
+    :subroutines [end-the-run]}
+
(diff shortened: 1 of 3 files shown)


src/clj/game/cards/ice.clj
```diff
@@ -2060,6 +2060,16 @@
                                     :msg (msg "force the Runner to lose " (:tag runner) " [Credits]")
                                     :effect (effect (lose-credits :runner (:tag runner)))})]}
 
+   "Thimblerig"
+   {:flags {:corp-phase-12 (req (>= (count (filter ice? (all-installed state :corp))) 2))}
+    :implementation "Does not restrict usage of swap ability to start of turn or after pass"
+    :abilities [{:label "Swap Thimblerig with a piece of ice"
+                 :prompt "Swap Thimblerig with a piece of ice"
```
Saintis: Maybe have this say "Choose a piece of ice to swap Thimblerig with?"

src/clj/game/cards/ice.clj
```diff
@@ -2060,6 +2060,16 @@
...
+                 :choices {:req ice?
+                           :not-self (req (:cid card))}
```
Saintis: This is a thing? Awesome!

NoahTheDuke: I had the same reaction!

src/clj/game/cards/ice.clj
presheaf: Makes sense. I'll fix it soonish.

Issue #3701, opened 2018-06-30
Bloodletter's lasting prompt


PesiTheWizard:
When the subroutine on Bloodletter is fired, the corp's prompt to wait for the runner to trash something doesn't go away after the runner does.

Saintis:
The case when runner choses to trash a program needs a `wait-for` on the trash resolve and then close the corp's wait prompt.

PR #3740, opened 2018-07-11
FIX: #3701 Bloodletter corp prompt

Danurai:
clear-wait-prompt regardless of which option the runner chooses.
Also fixed issue with Bloodletter which allowed the runner to trash 2 from the stack with only 1 card in the stack, will now only allow Trash 2 from stack if there are more than 1 cards in the stack.


nealterrell:
Well done!

PR #3742, opened 2018-07-11
RaR: SIU

presheaf:
Ability does not enforce being in 1.2 to make it slightly more forgiving for mistaken clicks on start of turn or forgotten rezzes.

### DIFF
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -1599,6 +1599,20 @@
              :async true
              :effect (effect (damage eid :net 1 {:card card}))}}
 
+   "SIU"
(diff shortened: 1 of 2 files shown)


nealterrell:
Every other optional start-of-turn corp card restricts its use to the phase-12 window, so I would like this card to as well. (Whether they should all be changed is another discussion.) It's especially important corp side IMO because of the mandatory draw.

presheaf:
Points taken. I do think there are quite a few Corp cards which don't restrict themselves to start of turn, though - Rashida and Tech Startup can be click 1 installed, rezzed and triggered, I think.

PR #3743, opened 2018-07-11
RaR: District 99 (feat. @nicohasa)

presheaf:
Took this one over after @nicohasa left it half finished. Card is almost fully functional, but some trashes must be manually triggered. This interacts well with automatic triggers.

Specifically, programs trashed by installing over another program are customarily done by dragging them to heap, so it's hard for the game engine to know that this was a real trash. One option is to hook that up to trashes, but it's also pretty common for mistakes to be corrected like that ("actually, i don't want to install that black orchestra after all"), and I suspect having this trigger trash events causes more problems than it solves, so I did the lazy thing and added a "add power counter" ability. This 'eats' the D99 trigger for that turn, meaning users who trigger D99 manually, then later on pop a spycam won't have to remember they triggered it already.

### DIFF
--- a/src/clj/game/cards/resources.clj
+++ b/src/clj/game/cards/resources.clj
@@ -552,7 +552,7 @@
                                  true)))
                            (trash card {:cause :ability-cost}))}]}
 
-   "Dean Lister"
+   "Den Lister"
    {:abilities [{:req (req (:run @state))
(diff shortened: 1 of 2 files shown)


nealterrell:
I agree with your design decisions.

PR #3745, opened 2018-07-12
Dean fix + test

presheaf:
Fixed my accidental rename of Dean and added a minor test to keep it from happening again.


nealterrell:
You have been punished appropriately!

PR #3760, opened 2018-07-14
WIP: Various Issues and minor fixes / improvements.

Saintis:
Fixes for some smaller issues.

Generally work-in-progress but should be able to be merged whenever.


NoahTheDuke:
Some good fixes here! I like removing the excess keys where unused.

Saintis:
Have tested The Foundary manually as well since the current test we had didn't fail...

PR #3768, opened 2018-07-16
RaR: Acme Consulting and Daruma

nealterrell:
Adds `:additional-tag` counter to Runner's state, giving us three tag counters: `:tag` (physical tags), `:tagged` ("is tagged", Papparazi), and `:additional-tag` (Acme). Updated `is-tagged?` and the `req` macro to account for this.

Acme grants `:additional-tag`  in the following situations:

* Starting a run on a server with a rezzed outermost ice.
* Rezzing the outermost ice during a run.

Gaining a tag thus sets a flag on the identity that is used to determine when to remove the additional tag:

Loses a tag (all of these require the flag being set):

* Run ends.
* Outermost ice is derezzed.
* Passing the outermost ice.

I think that should cover it.

Known issues that we can fix later:

* Does not cause Rachel Beckman to trash herself.
* Does not account for bypassing on approach with Charlatan.

### Daruma

Added a `:approach-server` event. Known issue: can swap a second agenda into a server that already has one. Did not write a test, short on time.

### DIFF
--- a/src/clj/game/cards/upgrades.clj
+++ b/src/clj/game/cards/upgrades.clj
@@ -282,6 +282,37 @@
     :abilities [{:label "[Trash]: Purge virus counters"
                  :msg "purge virus counters" :effect (effect (trash card) (purge))}]}
 
+   "Daruma"
+   (letfn [(choose-swap [to-swap]
+             {:prompt (str "Select a card to swap with " (:title to-swap))
+              :choices {:not-self true
+                        :req #(and (= "Corp" (:side %))
+                                   (#{"Asset" "Agenda" "Upgrade"} (:type %))
+                                   (or (in-hand? %) ; agenda, asset or upgrade from HQ
+                                       (and (installed? %) ; card installed in a server
+                                            ;; central upgrades are not in a server
+                                            (not (#{:hq :rd :archives} (first (:zone %)))))))}
+              :effect (req (wait-for (trash state :corp card nil )
+                                     (move state :corp to-swap (:zone target) {:keep-server-alive true})
+                                     (move state :corp target (:zone to-swap) {:keep-server-alive true})
+                                     (system-msg state :corp
+                                                 (str "uses Daruma to swap " (card-str state to-swap)
+                                                      " with " (card-str state target)))
+                                     (clear-wait-prompt state :runner)))
+              :cancel-effect (effect (clear-wait-prompt :runner))})
+           (ability [card]
+             {:optional {:prompt "Trash Daruma to swap a card in this server?"
+                         :yes-ability {:async true
+                                       :prompt "Select a card in this server to swap"
+                                       :choices {:req #(and (installed? %)
+                                                            (in-same-server? card %))
+                                                 :not-self true}
+                                       :effect (effect (continue-ability (choose-swap target) card nil))}
+                         :no-ability {:effect (effect (clear-wait-prompt :runner))}}})]
+   {:events {:approach-server {:async true
+                               :effect (effect (show-wait-prompt :runner "Corp to use Daruma")
+                                               (continue-ability :corp (ability card) card nil))}}})
+
    "Dedicated Technician Team"
    {:recurring 2}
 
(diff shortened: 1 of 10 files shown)


src/clj/game/core/flags.clj
```diff
@@ -36,7 +36,8 @@
   "Returns true if the runner is tagged."
   [state]
   (or (pos? (get-in state [:runner :tag]))
-      (pos? (get-in state [:runner :tagged]))))
+      (pos? (get-in state [:runner :tagged]))
```
Saintis: Shouldn't `:tagged` be a boolean instead of a number?

src/clj/game/macros.clj
```diff
@@ -47,7 +47,9 @@
             'hq-runnable '(not (:hq (get-in runner [:register :cannot-run-on-server])))
             'rd-runnable '(not (:rd (get-in runner [:register :cannot-run-on-server])))
             'archives-runnable '(not (:archives (get-in runner [:register :cannot-run-on-server])))
-            'tagged '(or (pos? (:tagged runner)) (pos? (:tag runner)))
+            'tagged '(or (pos? (:tagged runner))
```
Saintis: Could change this to call the function, or just remove it as it's just repeated code.

test/clj/game_test/cards/identities.clj
```diff
@@ -76,6 +76,83 @@
...
+      (play-from-hand state :corp "Vanilla" "Archives")
+      (core/rez state :corp (get-ice state :archives 0))
+      (core/rez state :corp (get-ice state :archives 1))
+      (take-credits state :corp)
+      (run-on state :archives)
+      (is (core/is-tagged? @state) "Runner is tagged when encountering outermost ice")
+      (run-continue state)
+      (is (not (core/is-tagged? @state)) "Runner is not tagged when encountering second ice")))
+
+  )
```
NoahTheDuke: sup

Saintis:
Does this work with The Twins? Twins trigger after passing ice and force encounter again (which should have tag if it's the outermost).

NoahTheDuke:
That test is really thorough. I appreciate you noting the known issues. Should you write a pending test to note that it currently doesn't work?

src/clj/game/core/flags.clj
nealterrell: This is not my change, it's been this way ever since Paparazzi.

Saintis: I shall rectify this at some point

nealterrell:
@Saintis no, if someone uses The Twins to re-encounter Data Ward, the runner will not automatically gain a tag. We could make that happen if we refactored all cards that modify run position to go through a function instead of fiddling with `state` directly.

test/clj/game_test/cards/identities.clj
nealterrell: Not my fault I swear.

Saintis:
Seems like a hassle. I'm happy to have it not work with The Twins for now.

src/clj/game/macros.clj
NoahTheDuke: Which function?

nealterrell:
Yeah, if Acme is splashing The Twins and trashing their best ice to force an ETR on another Data Ward, they deserve to resolve things manually.

src/clj/game/cards/upgrades.clj
```diff
@@ -282,6 +282,37 @@
...
+   (letfn [(choose-swap [to-swap]
+             {:prompt (str "Select a card to swap with " (:title to-swap))
+              :choices {:not-self true
+                        :req #(and (= "Corp" (:side %))
+                                   (#{"Asset" "Agenda" "Upgrade"} (:type %))
+                                   (or (in-hand? %) ; agenda, asset or upgrade from HQ
+                                       (and (installed? %) ; card installed in a server
+                                            ;; central upgrades are not in a server
+                                            (not (#{:hq :rd :archives} (first (:zone %)))))))}
+              :effect (req (wait-for (trash state :corp card nil )
```
nealterrell: DON'T DO IT NOAH.

NoahTheDuke: Lmao how dare you

PR #3774, opened 2018-07-17
docker update

rezwits:
Came up with a working docker setup, if this helps...

### DIFF
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -28,6 +28,7 @@ with the order that services start in mattering. Here are some specific issues:
     correctly imported into the database. Hence the kludgey `sleep 30` in
     `compose.sh`.
   - The lein* services download there dependencies every time they're run.
-  - For reasons I don't understand, `docker up lein-netrunner lein-cljs
-    stylus-css coffee-server` causes the lein* services to throw an exception,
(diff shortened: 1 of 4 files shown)


nealterrell:
Awesome job. Can anyone try this out?

rezwits:
Yeah, anyone can just check out YOUR project, and then have docker installed, change localhost to mongodb, in the dev.edn and then good ole, ./compose.sh -b

(was at the dentist)

Thomas Aldershof II
rezwits@me.com

rezwits:
I have two versions in my fork, one is the old crusty, server.coffee/fetch.coffee version.  But I added my cards.json, sets.json, and mwl.json, files just now, so you can fetch from your own webserver, just change the url in the fetch.coffee, to where those 3 json are put.  Then you can do either version.  The master branch is good using docker.  And the "mext" branch (meccg->next), which is the pure clojure server version which is going to be my ßeta for the first year starting September.  Currently my Alpha is the coffee.server, still testing.  I am going to transition, to pure clojure... for that sweet speed!

nealterrell:
Thank you for the comments... but I actually meant if any of the other devs were willing to try this out, since I don't use docker 😃.

PR #3775, opened 2018-07-17
WIP: Various fixes Mk 2

Saintis:
More fixes. 

Fixed Kate and Reina rebirth interaction, so now Rebirth should be fully functional. Removed the rebirth system message that notifies this as well.

Added `rebirth-choice` helper function to make it easier to construct tests with rebirth.


Saintis:
It could have a prompt when the run ends asking if you did trash a piece of ice or not, but I'm happy with this implementation for now. Only the reprisal cards (and Helion Beta Test) checks the `:trashed-card` flag currently, so only issue is if you _don't_ trash an ice and the Corp still plays a Reprisal.

PR #3776, opened 2018-07-17
 Make tests explicit + assorted test bugfixes - feat. @NoahTheDuke

presheaf:
Basically the same as https://github.com/mtgred/netrunner/pull/3735 . We had some trouble with auto-merge leading to a merge commit which was messy to undo (probably not for better gitters, but oh well), so I redid the merge commit in a separate branch.


src/clj/game/cards/identities.clj
```diff
@@ -639,7 +639,9 @@
                                      :rfg))}}]}
 
    "Kate \"Mac\" McCaffrey: Digital Tinker"
-   {:events {:pre-install {:req (req (and (#{"Hardware" "Program"} (:type target))
+   {:effect (req (when (pos? (event-count state side :runner-install))
+                   (swap! state assoc-in [:per-turn (:cid card)] true)))
+    :events {:pre-install {:req (req (and (#{"Hardware" "Program"} (:type target))
```
Saintis: This will clash with my fixes PR, and this effect is too restrictive. It will prevent Kate discount if you have installed anything before you Rebirth.

NoahTheDuke: Yeah, my bad for not pulling this out after talking with you yesterday. Removing this change and the pending change should revert that bit.

PR #3782, opened 2018-07-18
Change Literally Everything about Tests

NoahTheDuke:
UPDATE: I went fucking wild, y'all:
* Moved the json data to individual edn files in the `data/cards` folder.
* Moved all card definitions to their own files in folders by type.
* Moved all card tests to their own files in folder by type.
* Rewrote all card-loading logic in the engine, testing, and card-coverage systems to read from these files.
* Rewrote `card-def` and the card definition logic to act like a cache: lazy load the card definitions as the cards are used, don't unnecessarily load other card defs. This honestly only effects the first time a given card is used, but significantly speeds up how long it takes to load the engine.
* Rewrote test creation logic: now you can start the game with cards already in hand or in the discard, you can set starting credits to be different, you can leave out a side if they're gonna be defaulted, and you can set a given side's identity without using a new function.
* Changed all card-watching and reloading logic to use `hawk`, which is legit straight magic.
* Removed all fixtures as they didn't actually do what I wanted them to.
* Briefly considered/attempted to move tests to using `speclj`, but I realized that it's not been touched in two years so I'm not gonna put our test suite on it only to have it never get fixed when I find the broken bits.
* Added [eftest](https://github.com/weavejester/eftest) as a dependency and plugin. It's a _stupid fast_ test runner, running tests in parallel. At the time of writing, it handles all 871 tests in 20 seconds.
* Added a `card-definitions` test, which currently only looks at labels. I plan on adding to it, expanding it to check things like legal reqs, flags, and other "correctly writing a card" stuff.

ORIGINAL: Makes tests not require mongo to run. To get this to work, you have to run `lein fetch` to refresh the card data. Once you've done that, tho, you can just run `lein test` with no second console open for mongo and it'll work!


test/clj/game_test/cards/events.clj
```diff
@@ -2244,7 +2244,6 @@
       (do-game
         (new-game (default-corp ["Ice Wall" "Mark Yale"])
                   (make-deck whizzard ["Rebirth"]))
-        (println "Reina Rebirth twice test")
```
Saintis: ooops

test/clj/game_test/utils.clj
```diff
@@ -3,24 +3,10 @@
...
-    (mg/disconnect conn)
-    ret))
-
 (defn load-cards []
-  (let [conn (mg/connect {:host "127.0.0.1" :port 27017})
-        db (mg/get-db conn "netrunner")
-        cards (doall (mc/find-maps db "cards"))]
-    (mg/disconnect conn)
-    cards))
+  (read-string (slurp "data/cards.edn")))
```
Saintis: Neat!

NoahTheDuke:
I've decided that to follow through on this, there are ways to make it better, so I've put WIP back in the title.

NoahTheDuke:
After discussion on slack, decided to split into multiple PRs for ease.

PR #3786, opened 2018-07-19
Reagent merge to master

nealterrell:
You da man, @danhut.


project.clj
```diff
@@ -69,19 +72,23 @@
   :cljsbuild {
     :builds [
       {:id "dev"
-       :source-paths ["src/cljs" "src/dev" "src/cljc"]
+       :source-paths ["src/cljs/nr" "src/cljs/dev" "src/cljc"]
        :figwheel true
-       :compiler {:output-to "resources/public/cljs/app.js"
+       :compiler {:output-to "resources/public/cljs/app10.js"
```
NoahTheDuke: Is this to force a cache invalidation for clients?

danhut: That was the idea. There seemed to be some discussion on Nginx issues back when I started this.

PR #3787, opened 2018-07-19
Refactors deck validity checks and adds SOCR7

### LINKED ISSUE #3780: Feature request: Stimhack Online Cache Refresh 7 Legality checker
The legality checker has been hugely useful in previous Stimhack Online Cache Refresh tournaments, both to make sure your own deck is legal, and to make sure your opponent's deck is (particularly important since  you can change your deck every round).

For the next Stimhack Online Cache Refresh (SOCR7) tournament, we are using a different card pool in order to add Reign and Reverie. We are rotating out Red Sands and replacing it with Reign and Reverie, as if R&R was a new cycle. Like the official Cache Refresh pool, we are still using:

- 1 copy of Revised Core
- 1 big box (either Creation & Control, Honor & Profit, Order & Chaos, or Data & Destiny)
- Terminal Directive

Unlike the official cache refresh pool, instead of Red Sands and Kitara, we add:

- Kitara
- Reign and Reverie

So the total card pool is:

- 1 copy of Revised Core
- 1 big box (either Creation & Control, Honor & Profit, Order & Chaos, or Data & Destiny)
- Terminal Directive
- Kitara
- Reign and Reverie

MWL is in effect.

**The request**: It would be very helpful to have a new legality checker to check for SOCR7 decks. This would need to be in addition to the existing Cache Refresh legality checker, since it is not as if any of us has the authority to change the official card pool.

For more details, see here: https://forum.stimhack.com/t/stimhack-online-cache-refresh-7-starts-august-5/10024

Saintis:
Makes it easier to add new deck validity checks. Adds rules for the new Stimhack Online Cache Refresh. Fixes #3780.

### DIFF
--- a/src/cljc/jinteki/decks.cljc
+++ b/src/cljc/jinteki/decks.cljc
@@ -197,8 +197,12 @@
   (let [restricted-cards (remove (fn [card] (some #(= (:setname (:card card)) %) allowed-sets)) (:cards deck))
         restricted-sets (group-by (fn [card] (:setname (:card card))) restricted-cards)
         sorted-restricted-sets (reverse (sort-by #(count (second %)) restricted-sets))
-        [restricted-bigboxes restricted-datapacks] (split-with (fn [[setname cards]] (some #(when (= (:name %) setname) (:bigbox %)) sets)) sorted-restricted-sets)]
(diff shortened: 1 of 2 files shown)


tblackwe:
Asking in chat as well, but can we get this feature added soon?
The tournament goes live next week

PR #3869, opened 2018-09-18
Have Kakugo perform selfcheck on :cid

### LINKED ISSUE #3826: Kakugo not doing auto net damage?
Played a game tonight where I was using Engolo to break Kakugo and the 1 net damage wasn't firing automatically. I believe it would fire automatically previously?

cwgreene:
If Kakugo was passed, but it had been modified, game logic would say
it's not the same card, and not recognize that it had just been passed.
By checking against the :cid property, which is invariant under
temporary transformations, Kakugo sucessfully performs damage even under
type transformation.

Tested with new unit test.

Fixes #3826

### DIFF
--- a/src/clj/game/cards/ice.clj
+++ b/src/clj/game/cards/ice.clj
@@ -1242,7 +1242,7 @@
 
    "Kakugo"
    {:events {:pass-ice {:async true
-                        :req (req (= (:cid target) (:cid card)))
+                        :req (req (same-card? target card))
                         :msg "do 1 net damage"
                         :effect (effect (damage eid :net 1 {:card card}))}}
     :subroutines [end-the-run]}
--- a/src/clj/game/cards/ice.clj
+++ b/src/clj/game/cards/ice.clj
@@ -1242,7 +1242,7 @@
 
    "Kakugo"
    {:events {:pass-ice {:async true
-                        :req (req (= target card))
+                        :req (req (= (:cid target) (:cid card)))
                         :msg "do 1 net damage"
                         :effect (effect (damage eid :net 1 {:card card}))}}
     :subroutines [end-the-run]}
--- a/test/clj/game_test/cards/ice.clj
+++ b/test/clj/game_test/cards/ice.clj
@@ -739,6 +739,26 @@
         (run-continue state)
         (is (= 1 (-> (get-runner) :discard count)) "Runner should take 1 net damage from Kakugo")))))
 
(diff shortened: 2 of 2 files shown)


src/clj/game/cards/ice.clj
```diff
@@ -1242,7 +1242,7 @@
 
    "Kakugo"
    {:events {:pass-ice {:async true
-                        :req (req (= target card))
+                        :req (req (= (:cid target) (:cid card)))
```
Saintis: We have a `same-card?` helper function for this as well :)

cwgreene: updated.

cwgreene: thanks!

test/clj/game_test/cards/ice.clj
```diff
@@ -739,6 +739,26 @@
...
+      (is (= 4 (:credit (get-runner))) "Runner has 4 credits")
+      (is (= 1 (count (:hand (get-runner)))) "Runner has 1 card before run")
+      (let [kakugo (get-ice state :rd 0)
+            engolo (get-program state 0)]
+        (run-on state "R&D")
+        (core/rez state :corp kakugo)
+        (card-ability state :runner engolo 2)
+        (is (core/has-subtype? (refresh kakugo) "Code Gate") "Kakugo was made into a code gate")
+        (run-continue state)
+        (is (= 0 (count (:hand (get-runner)))) "Runner took damage passing kakugo"))))
```
Saintis: Could use `zero?` here instead of `= 0`. Or just `(is-empty (:hand (get-runner)))` (or is it `empty?`...)

NoahTheDuke: `zero` is what we’ve used elsewhere.

NoahTheDuke:
Can you in-line your new test in the existing Kakugo test, by wrapping it in a `testing` call instead of as a new standalone test?

test/clj/game_test/cards/ice.clj
cwgreene: `empty?` makes sense here since it tests the size of a collection. `zero?` tests if the number is zero. Going with `empty?`

cwgreene:
> Can you in-line your new test in the existing Kakugo test, by wrapping it in a `testing` call instead of as a new standalone test?

done.

nealterrell:
Great job!!

PR #3882, opened 2018-10-04
Updated Docker configuration

jwarwick:
Made a new `docker-compose` configuration that pulls the latest `master` release from GitHub and runs it in a container; MongoDB is launched in a separate, linked container.


docker/local/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+WORKDIR /netrunner
+COPY . .
+RUN lein deps
+RUN bower --allow-root install
+RUN stylus src/css -o resources/public/css/
+RUN lein cljsbuild once prod
+RUN lein uberjar
+
+# Final image
+FROM clojure:lein-alpine
```
justinliew: Does this work? I've never done it this way but it seems much better if indeed it cuts dependency checking time down (once the "build" stuff has run once and doesn't change)

jwarwick: Not clear if this speeds up the dependency checking. I'm doing it to get rid of all the build artifacts; it reduces the final image size by ~200MB.

Docker does seem to cache the intermediate layers, so I don't think it's rebuilding unnecessarily.

justinliew: Okay cool that's a win then. I can grab this at some point and see if splitting up the dockerfiles helps some.

jwarwick:
Got the `nginx` reverse proxy working. Had to rollback changes for smaller image size; unable to get `lein fetch` to work without `.git`, `.m2`, and `target` directories.

PR #3887, opened 2018-10-10
Made Blockchain only count face up transactions

### LINKED ISSUE #3859: Blockchain counts facedown transactions for its "on encounter" effect
When the "Gain Subroutines" option is clicked on Blockchain, the resulting number of subroutines counts facedown transactions as well as faceup ones - for example, if there are five faceup transactions and one facedown transaction in Archives, Blockchain will gain three subroutines when it should only gain two.

As an aside, it would also be useful to change the message to say "x uses Blockchain to gain y **additional** subroutines", as that makes the distinction between the printed and gained subroutines clear. (This may be useful wording to change on all ice that gain subroutines, even those that have no printed subroutines like Komainu just in case someone decides to pair it and Sub Boost or whatever, but it's a very fringe case!)

jwarwick:
Fixes #3859 

Also added a `+` to indicate additional subs present on the ice.


Saintis:
Might be worth changing the `gain subs` ability so it prints total subs present, e.g. "Corp uses Blockchain to gain 3 subs (5 in total)".

src/clj/game/cards/ice.clj
```diff
@@ -400,12 +400,18 @@
...
                                                     (:discard corp)))
                                      2)))]
-     {:abilities [{:label "Gain subroutines"
-                   :msg (msg (let [c (sub-count corp)]
-                               (str "gain " c (pluralize " subroutine" c))))}]
+     {:implementation "Number of subs is manual"
+      :abilities [{:label "Gain subroutines"
+                   :effect (req (let [c (sub-count corp)]
+                                  (update! state :corp
+                                           (assoc-in card [:special :extra-subs] (pos? c)))
```
NoahTheDuke: Will calling `pos?` here return a boolean? If so, is that what we want?

test/clj/game_test/cards/ice.clj
```diff
@@ -91,30 +91,53 @@
...
+        (card-subroutine state :corp bc 0)
+        (is (= 13 (:credit (get-corp))) "Corp gained 1 credit from Blockchain")
+        (is (= 4 (:credit (get-runner))) "Runner lost 1 credit from Blockchain"))))
+  (testing "Face down transactions"
+    (do-game
+      (new-game {:corp {:deck ["Blockchain" (qty "Beanstalk Royalties" 5)]}})
+      (core/gain state :corp :credit 2 :click 5)
+      (core/move state :corp (find-card "Beanstalk Royalties" (:hand (get-corp))) :discard)
+      (core/move state :corp (find-card "Beanstalk Royalties" (:hand (get-corp))) :discard)
+      (core/move state :corp (find-card "Beanstalk Royalties" (:hand (get-corp))) :discard)
```
NoahTheDuke: You can initialize with `(new-game {:corp {:hand ["Blockchain" (qty "Beanstalk Royalties" 2)] :discard [(qty "Beanstalk Royalties" 3)] :credit 7}})`, and it'll automatically set this all up for you. Won't set your starting clicks tho.

src/clj/game/cards/ice.clj
jwarwick: Yes, should return a boolean. `:extra-subs` is just an indicator of extra subs, not a count. Extra subs just shows up as `+` icon on the card. Modified ice strength shows up as a number icon.

test/clj/game_test/cards/ice.clj
NoahTheDuke: Crap, really? I'll have to see why. It should just "work". :-(

PR #3890, opened 2018-10-12
Close runner prompt for Remote Enforcement if no ice in deck

### LINKED ISSUE #3748: Scoring a Remote Enforcement with no ice left in the deck caused the game state to stall
I scored a Remote Enforcement with only 3 cards left in deck.  There was no prompt to install ice (I assume there were none left), but my opponent was stuck on a window saying to wait till the corp had installed an ice.  We ended up getting out of it by the runner using /close-prompt .

This was on https://jinteki.zaroth.net

jwarwick:
Fixes #3748


NoahTheDuke:
Spacing/formatting is all messed up, but this is otherwise a great fix.

PR #3903, opened 2018-10-22
Magnum Opus: Crowdfunding

jwarwick:


jwarwick:
`Assimilator` doesn't trigger `on install` events, so it would not gain the credits. I guess it should probably trash itself though. Not sure what `Daily Casts` does in that case. 

I do want to avoid having people accidentally clicking it and having it fire.

PR #3905, opened 2018-10-23
Magnum Opus: Border Control

NoahTheDuke:
what it says on the tin

### DIFF
--- a/src/clj/game/cards/ice.clj
+++ b/src/clj/game/cards/ice.clj
@@ -472,7 +472,7 @@
    {:abilities [{:label "End the run"
                  :effect (effect (trash card {:cause :ability-cost})
                                  (end-run))}]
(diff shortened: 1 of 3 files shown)


src/clj/game/cards/ice.clj
```diff
@@ -468,14 +468,28 @@
                                                              :init-data true})
                                (trigger-event state side :corp-install newice)))}]})
 
+  "Border Control"
+   {:abilities [{:label "End the run"
+                 :effect (effect (trash card {:cause :ability-cost})
+                                 (end-run))}]
+    :subroutines [{:label "Gain credits for ice on this server"
```
Saintis: I'd make this say the same as the card.
```suggestion
    :subroutines [{:label "Gain 1 [credits] for each ice protecting this server"
```

PR #3906, opened 2018-10-23
Magnum Opus: Slot Machine

NoahTheDuke:
Done did.

### DIFF
--- a/src/clj/game/cards/ice.clj
+++ b/src/clj/game/cards/ice.clj
@@ -2105,38 +2105,35 @@
                    :effect (req (do-access state :runner eid [:rd] {:no-root true}))}]}
 
    "Slot Machine"
-   {:implementation "Encounter effect is manual"
-    :abilities [{:label "Roll them bones"
-                 :effect (effect (move :runner (first (:deck runner)) :deck)
-                                 (system-msg (str "uses Slot Machine to put the top card of the stack to the bottom,"
-                                                  " then reveal the top 3 cards in the stack: "
-                                                  (join ", " (map :title (take 3 (:deck runner)))))))}]
-    :subroutines [{:label "Runner loses 3 [credits]"
-                   :msg "Runner loses 3 [credits]"
-                   :effect (effect (lose-credits :runner 3))}
-                  {:label "Gain 3 [credits]"
-                   :effect (req (let [unique-types (->> (:deck runner)
-                                                        (take 3)
-                                                        (map :type)
-                                                        (into #{})
-                                                        count)]
-                                 (when (>= 2 unique-types)
-                                   (system-msg state :corp (str "gains 3 [credits]"))
-                                   (gain-credits state :corp 3))))}
-                  {:label "Place 3 advancement tokens"
-                   :effect (req (let [unique-types (->> (:deck runner)
-                                                        (take 3)
-                                                        (map :type)
-                                                        (into #{})
-                                                        count)]
-                                 (when (= 1 unique-types)
-                                   (continue-ability
-                                     state side
-                                     {:choices {:req installed?}
-                                      :prompt "Choose an installed card"
-                                      :effect (req (system-msg state :corp (str "places 3 advancement tokens"))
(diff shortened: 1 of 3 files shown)


src/clj/game/cards/ice.clj
```diff
@@ -2104,6 +2104,38 @@
...
+    :abilities [{:label "Roll them bones"
+                 :effect (effect (move :runner (first (:deck runner)) :deck)
+                                 (system-msg (str "uses Slot Machine to put the top card of the stack to the bottom,"
+                                                  " then reveal the top 3 cards in the stack: "
+                                                  (join ", " (map :title (take 3 (:deck runner)))))))}]
+    :subroutines [{:label "Runner loses 3 [credits]"
+                   :msg "Runner loses 3 [credits]"
+                   :effect (effect (lose-credits :runner 3))}
+                  {:label "Corp gains 3 [credits]"
+                   :effect (req (let [f (:type (nth (:deck runner) 0))
```
Saintis: I feel there is a nicer way of doing this involving list expansion and `take`.
E.g. `[unique-types (count (into #{} (map :type (take 3 (:deck runner)))))]`
Or as a threading macro: `[unique-types (-> (:deck runner) (take 3) (map :type) (into #{}) count)]`

And then checking if unique-subtypes <= 2 and == 1.

src/clj/game/cards/ice.clj
```diff
@@ -2104,6 +2104,38 @@
                    :async true
                    :effect (req (do-access state :runner eid [:rd] {:no-root true}))}]}
 
+   "Slot Machine"
+   {:implementation "Encounter effect is manual"
+    :abilities [{:label "Roll them bones"
```
Saintis: This is perfect as it.

src/clj/game/cards/ice.clj
```diff
@@ -2104,6 +2104,38 @@
...
+                 :effect (effect (move :runner (first (:deck runner)) :deck)
+                                 (system-msg (str "uses Slot Machine to put the top card of the stack to the bottom,"
+                                                  " then reveal the top 3 cards in the stack: "
+                                                  (join ", " (map :title (take 3 (:deck runner)))))))}]
```
Saintis: I wonder if we want to help the mental math and print the types of each card here as well. E.g. "Sure Gamble (event), Career Fair (event), Bankroll (program)".

src/clj/game/cards/ice.clj
```diff
@@ -2104,6 +2104,38 @@
...
+    :subroutines [{:label "Runner loses 3 [credits]"
+                   :msg "Runner loses 3 [credits]"
+                   :effect (effect (lose-credits :runner 3))}
+                  {:label "Corp gains 3 [credits]"
```
Saintis: This should just be Gain 3 credits I think
```suggestion
                  {:label "Gain 3 [credits]"
```

src/clj/game/cards/ice.clj
```diff
@@ -2104,6 +2104,38 @@
...
+                   :effect (req (let [f (:type (nth (:deck runner) 0))
+                                      s (:type (nth (:deck runner) 1))
+                                      t (:type (nth (:deck runner) 2))]
+                                 (when (or (= f s)
+                                           (= f t)
+                                           (= s t))
+                                   (system-msg state :corp (str "gains 3 [credits]"))
+                                   (gain-credits state :corp 3))))}
+                  {:label "Corp places 3 advancement tokens"
```
Saintis: Similar to other label comment
```suggestion
                  {:label "Place 3 advancement tokens"
```

src/clj/game/cards/ice.clj
NoahTheDuke: Very true! I'll do that.

src/clj/game/cards/ice.clj
NoahTheDuke: Oh that's kind of cool. I like that. I'll add it in!

PR #3908, opened 2018-10-23
Magnum Opus: Embolus

presheaf:
The only design choice I think is questionable is whether Embolus should prompt you if you have 0 creds. I elected to have it do so in part so I didn't have to think about whether there were any weird edge cases in which players might need the prompt in order to manually trigger card abilities to ensure they had the credit to add the counter. However, if anyone has strong opinions on this I'll happily change it, although I'll then want to add a test ensuring it works as well as one can expect with Launch Campaign.

### DIFF
--- a/src/clj/game/cards/upgrades.clj
+++ b/src/clj/game/cards/upgrades.clj
@@ -378,20 +378,20 @@
     :trash-effect {:effect (req (update-all-ice state side))}}
 
    "Embolus"
-   (let [maybe-gain-counter
-         {:once :per-turn
-          :label "Place a power counter on Embolus"
-          :effect (effect (continue-ability {:optional
-                                             {:prompt "Pay 1 [Credit] to place a power counter on Embolus?"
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/upgrades.clj
```diff
@@ -377,11 +377,30 @@
...
+          :effect (effect (end-run))}]
+     {:derezzed-events {:runner-turn-ends corp-rez-toast}
+      :events {:corp-turn-begins maybe-gain-counter
+               :successful-run {:req (req (pos? (get-counters card :power)))
+                                :msg "remove 1 power counter from Embolus"
+                                :effect (effect (add-counter card :power -1))}}
+      :abilities [maybe-gain-counter etr]})
+
    "Expo Grid"
-   (let [ability {:req (req (some #(and (is-type? % "Asset")
```
NoahTheDuke: Why this change?

presheaf: Oops, good question. It was certainly not intentional.

PR #3914, opened 2018-10-30
Magnum Opus: Hired Help

presheaf:
Adding in a new kind of additional cost to a run on one server only was kinda annoying. I opted to have the card instead read "When a run is initiated on this server, the Runner must either trash an agenda or end the run". 

This will not result in any rule violations, I think, as the Runner is actually allowed to run a server, decline to pay costs and then jack out, but it would perhaps be better not to allow them to e.g. dirty laundry a server they can't pay the cost to run on. As it should be fairly easy to undo, I ignored that potential issue and chose the easy way out.

### DIFF
--- a/src/clj/game/cards/upgrades.clj
+++ b/src/clj/game/cards/upgrades.clj
@@ -492,6 +492,31 @@
                  :msg "gain 2 [Credits]"
                  :effect (effect (gain-credits 2))}]}
 
+   "Hired Help"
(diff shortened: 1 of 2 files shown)


presheaf:
Thanks for the kind words! It's not the ideal solution, but the effort to functionality ratio is pretty good.

NoahTheDuke:
Yeah, without overhauling large portions of the codebase to handle a card that will be banned in all formats, this is the best way to go.

PR #3920, opened 2018-11-04
Fix Kyuban/AoT interaction

### LINKED ISSUE #3917: Kyuban pays out twice against Architects of Tomorrow
The Shaper card Kyuban from Reign and Reverie pays out twice if used on a Bioroid Ice against the HB Identity Haas-Bioroid: Architects of Tomorrow. Most likely this is because both Kyuban and AoT check on passing an ice. Kyuban pays out once before AoT checks (as runner you see "wait on Corp using ability" prompt) and once after. Possibly this affects other "on pass" effects too.

kyubanaot

In the screenshot you can see a Kyuban on a Fairchild 2.0. In the log on the right you can see the message "chaosof99 uses Kyuban to gain 2 credits" as Kyuban paid out twice while passing it only once.

jwarwick:
Fixes #3917

### DIFF
--- a/src/clj/game/cards/identities.clj
+++ b/src/clj/game/cards/identities.clj
@@ -464,10 +464,11 @@
                                 {:prompt "Select a Bioroid to rez" :player :corp
                                  :choices {:req #(and (has-subtype? % "Bioroid") (not (rezzed? %)))}
                                  :msg (msg "rez " (:title target))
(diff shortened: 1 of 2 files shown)


jwarwick:
Not real clear on why this fixed it either. It seemed like we got the pass event twice. It worked fine if the corp did rez a piece of ice.

PR #3940, opened 2018-11-13
Rewrite Deck Validation for Nisei formats

NoahTheDuke:
[Nisei](http://nisei.net/) is set to release the specifications for their new set, System Core 2019, along with their new MWLs. This won't play nice with the current system, especially given that there exists the possibility for multiple MWLs in effect at a given time, so the time has come to rewrite some things.

Having polled the internet, I have decided that the 1.1.1.1 and Modded formats are out of favor and shouldn't be supported anymore. We will only support the 5 following formats:

* Snapshot (The current state of "Tournament legal", without the Magnum Opus and Orgcrime cards)
* Standard
* Eternal
* Core Experience
* [Stimhack Online Cache Refresh](https://forum.stimhack.com/t/stimhack-online-cache-refresh-8-round-1/10194) (Currently 8, but will be updated as needed)


src/cljs/nr/deckbuilder.cljs
```diff
@@ -203,7 +205,7 @@
         (->> @all-cards
              (filter #(and (= (:side %) side)
                            (= (:type %) "Identity")))
-             (filter #(not (contains? %1 :replaced_by))))
+             (remove :rotated))
```
jwarwick: I think this should stays as `:replaced_by`, which is used to map alt arts to multiple cards of the same name (eg. in Core 2.0). I don't think you want to remove rotated cards from the lookup search.

NoahTheDuke: That's a good point. How do we want to handle cards with more than two entries, then? What does `:replaced_by` mean when it's in core 1, core 2, and core 3?

jwarwick: This structure makes sense when things are linear, but less clear to me how to handle branching. The goal is not to have multiple copies of a given card name in the list (and treat the different versions of it as alternate art options). We could just leave them all in, though that might be confusing to see `Stimhack (Core Set)`, `Stimhack (Revised Core Set)`, etc.

src/cljc/jinteki/decks.cljc
```diff
@@ -273,97 +228,104 @@
...
-       (>= (card-count cards) (min-deck-size identity))
-       (<= (influence-count deck) (id-inf-limit identity))
-       (every? #(and (allowed? (:card %) identity)
-                     (legal-num-copies? identity %)) cards)
-       (or (= (:side identity) "Runner")
-           (let [min (min-agenda-points deck)]
-             (<= min (agenda-points deck) (inc min))))))
+(defn check-deck-status
+  "Checks the valid and standard keys of a deck-status map to check if the deck is legal, casual or invalid."
+  [{:keys [valid standard eternal core-experience snapshot]}]
```
Saintis: We could make this less hard-coded I think. Just loop over keys and return the string version of the key if legal.

Saintis: Although that would lose internal order, so maybe this is the best way.

src/cljs/nr/deckbuilder.cljs
```diff
@@ -372,7 +374,10 @@
   [deck-status tooltip? violation-details?]
   (let [status (decks/check-deck-status deck-status)
         message (case status
-                  "legal" "Tournament legal"
+                  "standard" "Standard legal"
+                  "eternal" "Eternal legal"
+                  "core-experience" "Core legal"
```
Saintis: Should this not be "Core Experience" for extra technicality points?

src/cljs/nr/deckbuilder.cljs
```diff
@@ -419,7 +424,7 @@
         best-card (lookup (:side card) card)]
     (if (js/isNaN qty)
       (swap! s assoc :quantity 3)
-      (let [max-qty (or (:limited best-card) 3)
+      (let [max-qty (or (:deck-limit best-card) 3)
```
Saintis: Could probably change this to `(:deck-limit best-card 3)` unless that breaks for some reason.

src/cljs/nr/deckbuilder.cljs
```diff
@@ -586,18 +591,23 @@
            (go (while true
                  (let [edit (<! edit-channel)
                        card (:card edit)
-                       max-qty (or (:limited card) 3)
+                       max-qty (or (:deck-limit card) 3)
```
Saintis: Same comment as above.

src/cljc/jinteki/decks.cljc
NoahTheDuke: Yeah, I want to return the smallest version first. Until we let players choose a deck's format and then only check it, we have to work from smallest to biggest available card pool because by definition all Core Experience decks are Standard legal too.

src/cljs/nr/deckbuilder.cljs
NoahTheDuke: I think it's written this way to protect against `{:deck-limit nil}`, but I agree that it's ugly. Seeing as we control the data now, it's probably safe to change those back to the standard way.

src/cljc/jinteki/decks.cljc
NoahTheDuke: Well, shortly players _will_ be able to choose a deck's format.

PR #3949, opened 2018-11-20
Fix alt art display

jwarwick:
Since we overwrite the `cards` collection on startup, we need to re-insert the alternate art info.

Also removed the individual alt art selector from the `Settings`. Users should specify individual alt arts in the `Card Browser`. Mass setting/resetting of alts is still in `Settings.`

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -63,6 +63,7 @@
   :aot [#"game\.*"
         #"web\.*"
         #"tasks.fetch"
+        #"tasks.altart"
         #"jinteki\.*"]
   :jar-name "netrunner.jar"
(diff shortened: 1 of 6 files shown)


NoahTheDuke:
Can users set the alt arts they want to use in the Card Browser right now? Or did my recent update mess that up?

jwarwick:
I think alt selection from the card browser is still working correctly.

NoahTheDuke:
I mean, can they pick which alt art they want to use in game in the card browser?

jwarwick:
Yes, there is a 'Select Art' button with the card info.

PR #3968, opened 2018-12-11
[WIP] Deck API

jwarwick:
**Still a work in progress, do not merge**
**TODO: adding new decks, editing existing decks, deleting decks**

Implements a new UI page (`Integrations`) which allows a user to generate up to 5 API keys. User can also delete existing keys. Keys have an associated expiration date, which is highlighted in the UI if the current date is greater than the expiration date.

screen shot 2018-12-11 at 4 51 46 pm

The API keys are JWTs, with the user's `emailhash` encoded in them. The standard fields of `iat` and `exp` are used to specify when the token was issued and when it will expire.

The JWT is signed with am asymmetric key. The private portion is discarded after signing. The public portion is stored in the `api_keys` Mongo collection. Additionally, the signed token the email hash, and the associated dates are stored in the same Mongo document.

The backed exposes the endpoints `/api/v1/decks` for listing all decks and `/api/v1/deck/<deck_id>` for listing an individual deck. Current format is json, very similar to what is stored in the `decks` collection today. **THIS MAY CHANGE IN LATER COMMITS**

The backed checks for a token in the `Authorization` HTTP headers, expecting the form: `Authorization: Bearer <Token>`. If the token is invalid or expired, the request is denied. If the token is not found in the `api_keys` collection, the request is denied. The `emailhash` from the signed token is used to look up a user in the `users` collection. Only decks for that user may be accessed.

This JWT-based verification scheme is overkill as currently used. But, I hope to reuse the majority of it for future OAuth integration work.

### DIFF
--- a/src/clj/web/api.clj
+++ b/src/clj/web/api.clj
@@ -11,6 +11,7 @@
             [web.admin :as admin]
             [web.news :as news]
             [web.decks :as decks]
+            [web.deck-api :as deck-api]
             [compojure.route :as route]
             [ring.middleware.params :refer [wrap-params]]
             [ring.middleware.keyword-params :refer [wrap-keyword-params]]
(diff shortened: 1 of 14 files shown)


NoahTheDuke:
Any updates, John?

jwarwick:
Haven't moved forward with this. Was hoping to find someone working on a mobile app or other client that could comment on what kind of API would be useful.

NoahTheDuke:
I'll ping some folks, see if anyone has input/ideas!

NoahTheDuke:
Closing until we can work on it more.

Issue #3962, opened 2018-12-05
Suggestion for Hernando Cortez


tolaasin:
I understand that Hernando is hard to implement, but in the interests of supporting game play, could there be some kind of prompt to the players (growl?) that if hernando is installed, and the corp has 10 credits, that the credit totals need adjusting appropriately? It's very easy to forget this otherwise.

PR #3972, opened 2018-12-21
Implement Hernando Cortez

nathanj:
I tested with a few of the ICE listed in the FAQ for this card. If you think there are any cards that this does not work with let me know, and I will add a test for it.

### DIFF
--- a/src/clj/game/cards/resources.clj
+++ b/src/clj/game/cards/resources.clj
@@ -48,6 +48,15 @@
     ;; Add an effect only if there is none in the map
     (merge trash-effect definition)))
 
+(defn count-num-subroutines
+  "Count the number of subroutines for Hernando Cortez. Note that this only
+  counts subroutines that exist as part of the card text. It does not count
+  subroutines that the card will gain once rezzed."
+  [card]
+  (->> (-> card :title server-card :text split-lines)
+       (filter #(starts-with? % "[subroutine]"))
+       count))
+
 ;;; Card definitions
 (def card-definitions
   {"Aaron Marrón"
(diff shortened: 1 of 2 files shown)


nathanj:
Yeah that makes sense. I was worried about the runtime performance since it is dealing with strings, but now that I think about it this is not going to hit that often.

nathanj:
Made the change. I ran into a small problem because `:text` was removed from the card when it is placed in the hand due to `make-card`. I'm not sure why it was removed. I added it back.

Another option would be to get the card from `@all-cards` when Hernando fires if you want to keep the removal of `:text`.

NoahTheDuke:
I think the goal for `dissoc`ing the card text was to make the card objects less weighty, as we don't ever modify the card text and it can be easily retrieved by calling `(-> card :title server-card :text)`.

src/clj/game/cards/resources.clj
```diff
@@ -48,6 +48,15 @@
     ;; Add an effect only if there is none in the map
     (merge trash-effect definition)))
 
+(defn count-num-subroutines
+  "Count the number of subroutines for Hernando Cortez. Note that this only
+  counts subroutines that exist as part of the card text. It does not count
+  subroutines that the card will gain once rezzed."
+  [card]
+  (->> (split-lines (:text card))
```
NoahTheDuke:
```suggestion
  (->> (-> card :title server-card :text split-lines)
```

src/clj/game/core/turns.clj
```diff
@@ -128,7 +128,7 @@
   ([card cid]
   (-> card
       (assoc :cid cid :implementation (card-implemented card))
-      (dissoc :setname :text :_id :influence :number :influencelimit :factioncost))))
+      (dissoc :setname :_id :influence :number :influencelimit :factioncost))))
```
NoahTheDuke: Revert this change.

NoahTheDuke:
Closes #3962

PR #3977, opened 2018-12-27
Implement Sundew

nathanj:
The :runner-spent-click event happens before the run is initiated, so we
do not know which server the run is going to happen on. This prevents us
from fully handling this ability in :runner-spent-click.

To fix, add a new field in the register, :click-type, which we can use
to determine if the click was spent on a run event. If so, we then wait
for the :begin-run event to handle the ability.

--

The :click-type feels a bit hacky to me but it was the best I could think of. Any ideas for a better way?

### DIFF
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -1736,11 +1736,22 @@
                               (play-cost-bonus [:credit 2]))}}}
 
    "Sundew"
-   {:implementation "it's all broken just don't even try man"}
-    ; :events {:runner-spent-click {:once :per-turn
-    ;                               :msg (req (when (not this-server) "gain 2 [Credits]"))
-    ;                               :effect (req (when (not this-server)
-    ;                                              (gain-credits state :corp 2)))}}
+   ; If this a run event then handle in :begin-run as we do not know the server
+   ; being run on in :runner-spent-click.
+   {:events {:runner-spent-click {:req (req (first-event? state side :runner-spent-click))
+                                  :msg (req (if (not= :run (get-in @state [:runner :register :click-type]))
+                                              "gain 2 [Credits]"))
+                                  :effect (req (if (not= :run (get-in @state [:runner :register :click-type]))
+                                                 (gain-credits state :corp 2)))}
+             :begin-run {:once :per-turn
+                         :req (req (first-event? state side :runner-spent-click))
+                         :msg (req (if (and (= :run (get-in @state [:runner :register :click-type]))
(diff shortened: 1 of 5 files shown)


src/clj/game/cards/assets.clj
```diff
@@ -1736,11 +1736,20 @@
...
+   ; being run on in :runner-spent-click.
+   {:events {:runner-spent-click {:once :per-turn
+                                  :req (req (and (first-event? state side :runner-spent-click)
+                                                 (not= :run (get-in @state [:runner :register :click-type]))))
+                                  :msg "gain 2 [Credits]"
+                                  :effect (req (gain-credits state :corp 2))}
+
+             :begin-run {:once :per-turn
+                         :req (req (and (first-event? state side :runner-spent-click)
+                                        (not this-server)))
```
NoahTheDuke: This needs to check that the `:click-type` is `:run`, because only Run subtype events trigger Sundew:
* [If the Runner plays a run event to initiate a run on the server Sundew is installed in, the Corp does not gain 2 credits. [Official FAQ]](https://netrunnerdb.com/en/card/25092)

nathanj: I don't know why I need to check it. It seems like I should know that this is a run because this is in the `:begin-run` event. Can you explain further?

NoahTheDuke: If you use Deuces Wild to make a run on the Sundew server click 1, Sundew still fires because technically you're not spending your first click making a run on the Sundew server, you're playing an event. Only events that have the "Run" subtype count as "run events" for Sundew.

You're only setting the `:click-type` if the click-run ability is used or if an event with a "Run" subtype is played, which is necessary to know if it should or should not block Sundew. Otherwise, literally any kind of run (Deuces Wild for instance) in the first click will block Sundew.

nathanj: Thanks for the info. I was not aware of the difference between a "run event" and an "event" that makes a run.

NoahTheDuke: No worries, I had to ask folks on Stimslack about it to make sure.

nathanj:
Pushed new version. I added several more tests to make sure all the different interactions worked properly. I ended up keeping the `first-event?` in `req` as it seemed necessary for Ashes to work in all scenarios. Based on my reading of Sundew that seems right: the ability always triggers on the first click but then conditionally does not gain the credits.

NoahTheDuke:
This looks phenomenal. Thanks so much for all of the tests and hard work!

PR #3996, opened 2019-01-13
Fix miscapitalised credit-related messages

ouroboros8:


ouroboros8:
@NoahTheDuke I should've specified, but I raised this after noticing that messages for Border Control were not displaying credit symbols in chat, and I verified it for Blockchain as well:
creditbug

Shall I raise an issue to fix the symbol parser?

NoahTheDuke:
Sorry, life got away from me. Yes, please fix the symbol parser if you can. I'm sorry for not believing you, that's super strange! Would be very helpful to have it handle both upper and lower case versions of the symbols.

PR #3999, opened 2019-01-16
Fix crash in Khumalo breakers

jwarwick:


NoahTheDuke:
Damn it, how’d I miss this! Thanks for the catch

jwarwick:
These are easy to fix when I can see them in the logs :)

PR #4003, opened 2019-01-21
Implemented legality check for TBB's Classic format

presheaf:
As I've seen the Stimhack Slack channel for it is pretty active, I've implemented a legality checker for the Classic format (https://runthenet.wordpress.com/2019/01/01/the-classic-format-netrunners-final-form/). 

~~However, it currently seems to crash instantly when a legal deck is input. I'll be the first to admit I don't really know clojurescript and really just made changes by analogy, so I've probably made some silly mistake. If anyone could point me in the right direction I'll happily fix it.~~

Nevermind the above, it now works locally.


presheaf:
No worries, this was a single line of sed. :) Or, well, four words of google and then a copy/paste, if I'm being honest. Anyway, not your fault I use tools the wrong way. Thanks for cleaning up the formats in the first place, made the job super managable.

NoahTheDuke:
You’re welcome! I really need to document the process better cuz it’s manual and opaque. Thanks so much for doing all this.

PR #4005, opened 2019-01-25
Fixed Masanori/Paragon interaction and added regression test.

presheaf:
Also minor change to existing test which would break the day Corp cards can't go in Runner decks.


NoahTheDuke:
Nicely done! Looks like you left some prints in, but otherwise I’d say :shipit:

PR #4008, opened 2019-01-28
Fix Rashida Jaheem's trigger event

### LINKED ISSUE #3855: Rashida Jaheem automatic trigger
I just discovered that Rashida Jaheem triggers automatically when I click the Mandatory Draw button! Unfortunately, the mandatorily drawn card is shown before Rashida's prompt opens, so it's illegal to use Rashida this way.  Is there anyway for the prompt to occur before the mandatory draw, or are we stuck clicking Rashida, rezzing Rashida, clicking Start Turn, clicking Rashida, clicking her use ability, clicking Yes on the prompt, then clicking Mandatory Draw?

At worst, could we bypass the prompt if we click her "gain 3, draw 3" ability? It's pretty obvious that anyone who is going to click on that button will want to use that ability.

ChooJeremy:
Fixes #3855 

Simple change, changing the trigger from start of turn to before mandatory draw. Now the prompt appears after clicking "Start turn" but before taking the mandatory draw.

Further potential changes:
Remove `:flags {:corp-phase-12 (req true)}` - there's no more point for this because the prompt appears?

I noticed a couple of other cards with start of turn abilities have this problem as well, although it is significantly less prominent and seeing the card first doesn't matter as much. For example, Bio-Ethics Association happens after the mandatory draw step too, even though it's technically supposed to happen before. Perhaps those cards could be changed in the same way too? It would make the code somewhat less understandable though (`corp-turn-begins` is more clear than `corp-phase-12`)

Also, I hear there's a slack channel, could I get an invite? I know mtgred has retired, do I email someone else?

### DIFF
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -1462,7 +1462,7 @@
                                     card nil))}]
      {:derezzed-events {:runner-turn-ends corp-rez-toast}
       :flags {:corp-phase-12 (req true)}
-      :events {:corp-turn-begins ability}
+      :events {:corp-phase-12 ability}
       :abilities [ability]})
 
    "Reality Threedee"


NoahTheDuke:
I have a bunch of thoughts on this PR and associate issue because it's not exactly the right fix, but I typed them into slack, lol. You should be added, so when you join up we'll chat more about this.

NoahTheDuke:
Closing as I merged #4010.

PR #4011, opened 2019-01-29
Refactor start of turn and events

### LINKED ISSUE #3855: Rashida Jaheem automatic trigger

NoahTheDuke:
@ChooJeremy tried to fix this in #4008, but the problem is deeper than changing which trigger to fire on. And @nealterrell's PR added async to Rashida, which is obvious in hindsight, but without Marilyn's `:interactive (req true)`, it'd still fire without asking first.

Along with the two card fixes, I moved click-gain to before `:X-phase-12` (which inadvertently fixes/refactors/cleans up Jak Sinclair), and changed the `:X-turn-begins` to be a `trigger-event-simult` call, allowing the player to choose the order of the effects.

That last change, of course, required adding `:interactive (req true)` or `:silent (req true)` to every single `:corp-turn-begins` and `:runner-turn-begins` event, which I used as as opportunity to refactor a number of cards as well. I plan on writing tests for all rewritten/refactored cards (as the rest are presumably fine with a single line added).

### Existing issues:
For some reason, using `core/set-prop` in a test breaks the test. I tried to use it instead of calling `take-credits` multiple times in the test to remove all of the counters from Marilyn Campaign, but it didn't work and was throwing weird errors.

~I've also not looked into refactoring all of the other `:X-turn-begins` effects, as that's a much bigger issue and should be done in different/separate PRs, I think. Moving away from `:X-phase-12` functionality to requiring handling all existing cards is a much bigger change and probably needs involvement from the greater community.~ lmao jk I did it.

Fixes #3855 
Closes #4008 
Closes #4010


nealterrell:
So the "click cards to choose order" window is still there, we just now run the "BTW you forgot to trigger this manual/important effect before your turn started!" events in simult?

src/clj/game/cards/hardware.clj
```diff
@@ -938,19 +944,25 @@
...
+                  :silent (req true)
+                  :req (req (:runner-phase-12 @state))
+                  :effect (effect (update! (assoc card :qianju-active true))
+                                  (lose :click 1))
+                  :msg "lose [Click] and avoid the first tag received until their next turn"}]
+     {:flags {:runner-phase-12 (req true)}
+      :abilities [ability]
+      :events {:corp-turn-ends {:effect (effect (update! (dissoc card :qianju-active)))}
+               :runner-turn-begins
+               {:interactive (req true)
```
nealterrell: So you feel this card is important enough to force the Runner to make a decision every turn?

src/clj/game/cards/agendas.clj
```diff
@@ -341,30 +341,33 @@
...
    "Corporate Sales Team"
-   (let [e {:effect (req (when (pos? (get-counters card :credit))
-                           (gain-credits state :corp 1)
-                           (system-msg state :corp (str "uses Corporate Sales Team to gain 1 [Credits]"))
-                           (add-counter state side card :credit -1)))}]
+   (let [ability {:silent (req true)
+                  :counter-cost [:credit 1]
+                  :msg "gain 1 [Credit]"
+                  :effect (effect (gain-credits :corp 1))}]
      {:effect (effect (add-counter card :credit 10))
```
nealterrell: This card should probably have an ability like PAD Campaign so it can be ordered before cards like Executive Boot Camp.

src/clj/game/cards/agendas.clj
```diff
@@ -1174,7 +1177,8 @@
 
    "SSL Endorsement"
    (let [add-credits (effect (add-counter card :credit 9))
-         remove-credits {:optional {:req (req (pos? (get-counters card :credit)))
+         remove-credits {:interactive (req true)
+                         :optional {:req (req (pos? (get-counters card :credit)))
```
nealterrell: Same comment as for Corp Sales Team, except the manual ability probably should just take 3 credits instead of triggering an Optional.

NoahTheDuke:
That makes a ton of sense and I’m disappointed in myself that I didn’t pick up on it, haha. Looks like I have some changes to undo!

src/clj/game/cards/agendas.clj
NoahTheDuke: Good call, I'll do that.

src/clj/game/cards/hardware.clj
NoahTheDuke: There's a trade-off to be made, but I think that a card like this is being played to gain its' effect, so we should ask. If you don't feel that way, then I can easily remove this part.

src/clj/game/cards/agendas.clj
NoahTheDuke: Done did

NoahTheDuke:
Closing this in favor of splitting into many smaller PRs.

PR #4013, opened 2019-01-30
Add log entry when Thimblrig is moved 

### LINKED ISSUE #4009: No log entry when Thimblrig is moved
When the player uses Thimblrig to move it around, either by start of the turn or during a run, no log entry is created.

ChooJeremy:
Fixes #4009


5N00P1:
Thank you for taking this serious! Some time I only realize it was moved by looking on the board.
I think it should include the position, to make clear where it was moved especially when more then one Thimblrigs are active.

ChooJeremy:
Thanks for the help! I've updated the message to be more appropriate, especially with unrezzed ice.

PR #4016, opened 2019-02-03
Fix research grant to work with team sponsorship

### LINKED ISSUE #3390: Research Grant on-score effect does not trigger if no other copies are installed
Simply put, the corp cannot trigger the on-score effect of Research Grant at all if no other copies are installed at the time the agenda is scored. This is an issue because of Team Sponsorship, which resolves at the same time. A legal play would be to score Research Grant, resolve Team Sponsorship to install a copy of Research Grant, then resolve the scored Research Grant to score the newly installed one. However, Jinteki.net does not check if Research Grant is installed after Team Sponsorship resolves, giving no opportunity to immediately score it.

If a second copy of Research Grant _is_ installed when the first is scored, then a third can be installed and immediately scored using Team Sponsorship, as the option to resolve simultaneous effects in the order of the corp's choice will correctly appear.

Additionally, when scoring Bifrost Array while Research Grant is scored, options for simultaneous effects will correctly appear and allow the corp to install a copy of Research Grant using Team Sponsorship before resolving Bifrost Array and in turn the previously scored Research Grant.

This issue is extremely trivial to reproduce: simply score Research Grant while Team Sponsorship is rezzed and a Research Grant is present in HQ or Archives, but no Research Grants other than the one being scored exist in remote servers. No option to resolve Research Grant's effect will be given even if one is installed during the agenda-scored triggers phase.

ChooJeremy:
Fixes #3390

Fixes as suggested by @nealterrell in the original issue.

Basically moves the requirement to when the effect triggers (So the effect will always fire, but only ends up resolving if the trigger works) and adds interactive so that the user gets to order them manually, instead of the game ordering it manually.

The way the simultaneous triggers work become obvious if there are 2 copies of team sponsorship rezzed, in which case the first trigger adds Research's grants trigger to the simultaneous trigger selection but allows the user to pick between the 2nd team sponsorship's trigger and the research grant's trigger.

### DIFF
--- a/src/clj/game/cards/agendas.clj
+++ b/src/clj/game/cards/agendas.clj
@@ -1118,12 +1118,13 @@
    "Research Grant"
    {:interactive (req true)
     :silent (req (empty? (filter #(= (:title %) "Research Grant") (all-installed state :corp))))
-    :req (req (not (empty? (filter #(= (:title %) "Research Grant") (all-installed state :corp)))))
     :async true
     :effect (effect (continue-ability
                       {:prompt "Select another installed copy of Research Grant to score"
                        :choices {:req #(= (:title %) "Research Grant")}
+                       :interactive (req true)
                        :async true
+                       :req (req (not (empty? (filter #(= (:title %) "Research Grant") (all-installed state :corp)))))
                        :effect (effect (set-prop target :advance-counter (:advancementcost target))
                                        (score eid (get-card state target)))
                        :msg "score another installed copy of Research Grant"}


NoahTheDuke:
Thanks for doing this! Glad you're tackling these forgotten bugs!

NoahTheDuke:
I can get to the test another time.

ChooJeremy:
> 
> 
> Thanks for doing this! Glad you're tackling these forgotten bugs!
> 
> Would you mind writing a test making sure that you can score just one Research Grant without having a second installed? Seems that when I wrote the original test, I didn't account for that.

@NoahTheDuke Oops, I somehow missed this message. I'll try writing one up!

PR #4018, opened 2019-02-05
Look up credit icons case insensitively

### LINKED ISSUE #4017: Border control chat message shows "[credits]" instead of credit symbol
If the first sub on Border Control fires, a message like the following is shown:

> PlayerName uses Border Control to gain 2 [credits].

image

The text `[credits]` is shown rather than the credit symbol like on other cards.

Not 100% sure why and don't have a dev environment stood up right now, but could is be because `[credits]` in this line should be `[Credits]` like it is on other cards?

https://github.com/mtgred/netrunner/blob/f7f6d8fb6bd25ce622716c01ddb6d79e2d559a47/src/clj/game/cards/ice.clj#L479

ouroboros8:
This lookup only seems to be used in the log pane, which uses a somewhat specialised function that returns differing spans depending on whether the input is a reference to a symbol. Elsewhere `add-symbols` from `nr.cardbrowser` is used to do a standard search/replace case insensitively across a string.

I don't understand exactly why log-pane uses a different method, but on the bright side it's probably faster as it's not iterating over regex substitutions? 🤷‍♂

Fixes  #4017

### DIFF
--- a/src/cljs/nr/gameboard.cljs
+++ b/src/cljs/nr/gameboard.cljs
@@ -98,24 +98,24 @@
 (ws/register-ws-handler! :netrunner/diff #(handle-diff (parse-state %)))
 (ws/register-ws-handler! :netrunner/timeout #(handle-timeout (parse-state %)))
 
-(def anr-icons {"[Credits]" "credit"
+(def anr-icons {"[credits]" "credit"
                 "[$]" "credit"
                 "[c]" "credit"
-                "[Credit]" "credit"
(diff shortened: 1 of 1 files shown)


NoahTheDuke:
I didn't know they were doing different things! If you feel up to it, I'd love to see them unified somehow. If not, I think this change will work just fine.

ouroboros8:
I'll have a go at it :)

ouroboros8:
Superseded by  #4027

PR #4022, opened 2019-02-10
Add tests for research grant

ChooJeremy:
See #4016 

Wrote one up for single research grant and one for it's synergy with team sponsorship. I'm not sure about the latter - feels like any UI changes are going to break the test.


NoahTheDuke:
Yeah, our tests are not unit tests at all, so any UI change is liable to break them. Those tests look good tho.

PR #4023, opened 2019-02-10
Support for autoresolving :optional prompts

presheaf:
This PR implements support for `:optional` prompts to additionally take an `:autoresolve` parameter. If not present, things proceed as usual, but if present, it is treated as a 5-function, called when the prompt would be displayed, and if it returns "Yes" or "No", the prompt is not displayed, and instead resolves as if the user had clicked "Yes" or "No".

Further, it implements togglable autoresolve similar to what Consume and Friday Chip has for the following cards:

* Amani Senai
* Marilyn Campaign
* Knobkierie
* Paragon
* 419
* CtM
* SSO
* Find the Truth
* RNG Key
* Laramy Fisk
* Aeneas Informant

(Note that for stuff like CtM/Paragon, where you first click 'Yes', then make another choice, autoresolve only clicks 'Yes' for you - it doesn't compel you to always trace 4 or always bottom the card or whatever.) 

For all of these cards, I think that provided you voluntarily put them in your deck (looking at you, Fisk), the player is likely to essentially always want to click 'Yes' if they can. In case this is wrong, players are always free to leave autoresolve to its default setting of not doing anything. I am tempted to make the default 'Yes' for essentially everything on that list except Fisk, but have refrained from doing so for now because that breaks players' expectations, so I'd like to hear what you think first.


presheaf:
I originally planned on porting a bunch of stuff currently coded as phase 1.2 manually triggered abilities into this system, similar to what I think Noah is doing in his PR, but there's just so much that wants changing that I elected to leave it be for now. 

Also, as a comment on a possible extension, it's tempting to go further and extend this to all kinds of prompts. Porting Consume/Friday Chip actually changed its functionality when multiple cards were simultaneously trashed (rare but possible), as the resulting number prompt isn't :optional, so the current behavior of automatically placing all the counters now becomes prompting the user for how many counters to place. I could definitely add a hack to have it auto-place the maximum, but the proper fix if this is deemed a priority seems to be adding support for `:autoresolve` to a broader class of prompts. 

I think in principle there is no problem with calling a 5-fn and seeing if it outputs one of the prompt choices. There is a bit of casework involved to keep it fully general as there are so many types of prompts, but I think it can be done cleanly. The main reason I didn't bother is it's not clear to me how useful it is. It might be good to support the normal convention of "I'm going to always choose X server unless I say otherwise" for Security Testing, which we could do by adding support for :autoresolve and coding an ability which selects a server and automatically selects that at start of turn. I don't mind doing this if we want more of this, but my gut says implementing this for `:optional` prompts is 80 for 20.

src/clj/game/core/abilities.clj
```diff
@@ -278,20 +278,30 @@
...
+               (if (and (= prompt-choice "Yes")
+                        yes-ability
+                        (can-pay? state side (:title card) (:cost yes-ability)))
+                 (resolve-ability state side (assoc yes-ability :eid eid) card targets)
+                 (if-let [no-ability (:no-ability ability)]
+                   (resolve-ability state side (assoc no-ability :eid eid) card targets)
+                   (effect-completed state side eid)))))]
+     (let [autoresolve-as (when-let [autoresolve-fn (:autoresolve ability)]
+                            (autoresolve-fn state side eid card targets))]
+       (case (when-let [autoresolve-fn (:autoresolve ability)]
```
NoahTheDuke: This is merely a readability/style thing, but would you mind changing this to a `when-let*` and assigning `(autoresolve-fn state side eid card targets)` to a variable, and then saying `case variable-name`? It's slightly wordier, but it helps understanding this line a lot.

presheaf: Yeah, I'll do that. To be honest, I don't actually understand this line when I reread it. Why am I binding the name `autoresolve-as` without using it? So some cleanup is definitely in order.

test/clj/game_test/engine/rules.clj
```diff
@@ -660,3 +660,136 @@
...
+     (new-game {:corp {:deck ["Jackson Howard"]}
+                :runner {:deck [(qty "Aeneas Informant" 2)]}})
+     (play-from-hand state :corp "Jackson Howard" "New remote")
+     (take-credits state :corp)
+     (core/gain state :runner :click 50)
+     (play-from-hand state :runner "Aeneas Informant")
+     (letfn [(run-jackson []
+               (run-empty-server state "Server 1")
+               (click-prompt state :runner "No action"))
+             (get-aeneas1 [] (get-resource state 0))]
```
NoahTheDuke: Just so you know, this is effectively the same as `(let [aeneas (get-resource state 0)]` and then using `(refresh aeneas)` everywhere you wrote `(get-aeneas1)`. I don't think it's worth changing your tests over, but it is good to be aware of.

src/clj/game/cards.clj
```diff
@@ -207,6 +207,24 @@
                                                (vals selected-cards)))]
                            (effect-completed state side (make-result eid {:number counter-count :msg msg})))))}))
 
+(defn autoresolve-toggler [toggle-kw ability-name]
+  "Makes a card ability which lets the user toggle auto-resolve on an ability. Setting is stored under [:special toggle-kw]."
+  {:label (str "Toggle auto-resolve on " ability-name)
+   :prompt (str "Set auto-resolve on " ability-name " to:")
+   :choices ["Always" "Never" "Ask"]
+   :effect (effect (update! (update-in card [:special toggle-kw] (fn [x] (keyword (clojure.string/lower-case target)))))
```
NoahTheDuke:
```suggestion
   :effect (effect (update! (update-in card [:special toggle-kw] (fn [x] (slugify target))))
```
This only works if you require slugify from `jinteki.utils`, but it's there to be used!

src/clj/game/cards/programs.clj
```diff
@@ -223,11 +221,7 @@
                              (str "gain " (* 2 global-virus) " [Credits], removing " (quantify local-virus "virus counter") " from Consume"
                              (when (pos? hivemind-virus)
                                    (str " (and " hivemind-virus " from Hivemind)")))))}
-                {:effect (effect (update! (update-in card [:special :auto-accept] #(not %)))
-                                 (toast (str "Consume will now "
-                                             (if (get-in card [:special :auto-accept]) "no longer " "")
-                                             "automatically add counters.") "info"))
-                 :label "Toggle automatically adding virus counters"}]}
+                {:effect (autoresolve-toggler :auto-accept "adding virus counters")}]}
```
NoahTheDuke:
```suggestion
                (autoresolve-toggler :auto-accept "adding virus counters")]}
```

src/clj/game/cards/programs.clj
```diff
@@ -195,18 +195,16 @@
...
                                                mult-ab {:prompt "Place virus counters on Consume?"
                                                         :choices {:number (req amt-trashed)
                                                                   :default (req amt-trashed)}
                                                         :msg (msg "place " (quantify target "virus counter") " on Consume")
                                                         :effect (effect (add-counter :runner card :virus target))}
-                                               ab (if (> amt-trashed 1) mult-ab sing-ab)
-                                               ab (if (get-in card [:special :auto-accept]) auto-ab ab)]
+                                               ab (if (= 1 amt-trashed) sing-ab mult-ab)]
                                            (continue-ability state side ab card targets)))}}
     :effect (effect (toast "Tip: You can toggle automatically adding virus counters by clicking Consume."))
```
NoahTheDuke: Is this needed with the new toasts?

src/clj/game/cards/programs.clj
```diff
@@ -895,10 +891,12 @@
              :successful-run {:req (req (and (#{:hq :rd} target)
                                              (first-event? state :runner :successful-run #{[:hq] [:rd]})))
                               :optional {:prompt "Fire RNG Key?"
+                                         :autoresolve (autoresolve-lookup :auto-fire)
                                          :yes-ability {:prompt "Guess a number"
                                                        :choices {:number (req 20)}
                                                        :msg (msg "guess " target)
-                                                       :effect (effect (update! (assoc-in card [:special :rng-guess] target)))}}}}}
+                                                       :effect (effect (update! (assoc-in card [:special :rng-guess] target)))}}}}
+    :abilities (autoresolve-toggler :auto-fire "RNG Key")}
```
NoahTheDuke:
```suggestion
    :abilities [(autoresolve-toggler :auto-fire "RNG Key")]}
```

test/clj/game_test/engine/rules.clj
presheaf: Thanks for the reminder. This was on purpose, but only done out of laziness - I keep getting annoyed at forgetting to refresh cards, so saving the, uh, 4 characters seemed worthwhile. I won't really defend it, though. :P

src/clj/game/cards/programs.clj
presheaf: Good point, it shouldn't be.

src/clj/game/core/abilities.clj
```diff
@@ -278,20 +278,30 @@
...
+   (letfn [(prompt-fn [prompt-choice]
+             (let [yes-ability (:yes-ability ability)]
+               (if (and (= prompt-choice "Yes")
+                        yes-ability
+                        (can-pay? state side (:title card) (:cost yes-ability)))
+                 (resolve-ability state side (assoc yes-ability :eid eid) card targets)
+                 (if-let [no-ability (:no-ability ability)]
+                   (resolve-ability state side (assoc no-ability :eid eid) card targets)
+                   (effect-completed state side eid)))))]
+     (let [autoresolve-fn     (:autoresolve ability)
```
NoahTheDuke:
```suggestion
     (let [autoresolve-fn (:autoresolve ability)
```
There are arguments to be made both ways, so insert all applicable arguments here, lol.

src/clj/game/core/abilities.clj
```diff
@@ -278,20 +278,30 @@
...
+           autoresolve-answer (when autoresolve-fn
+                                (autoresolve-fn state side eid card targets))]
+       (case autoresolve-answer
+         "Yes" (prompt-fn "Yes")
+         "No"  (prompt-fn "No")
+               (do (toast state side (str "This prompt can be skipped by clicking "
```
NoahTheDuke: see above about spacing, lol

src/clj/game/cards/programs.clj
```diff
@@ -139,7 +139,38 @@
...
+                              (when-let [hiveminds (filter #(= "Hivemind" (:title %)) (all-active-installed state :runner))]
+                                        (doseq [h hiveminds]
+                                               (update! state side (assoc-in h [:counter :virus] 0)))))
+                 :msg (msg (let [local-virus (get-counters card :virus)
+                                 global-virus (get-virus-counters state card)
+                                 hivemind-virus (- global-virus local-virus)]
+                             (str "gain " (* 2 global-virus) " [Credits], removing " (quantify local-virus "virus counter") " from Consume"
+                             (when (pos? hivemind-virus)
+                                   (str " (and " hivemind-virus " from Hivemind)")))))}
+                {:effect (autoresolve-toggler :auto-accept "adding virus counters")}]}
```
NoahTheDuke:
```suggestion
                (autoresolve-toggler :auto-accept "adding virus counters")]}
```

NoahTheDuke: Just in case it gets lost in the reorder. :wink:

src/clj/game/cards/programs.clj
```diff
@@ -139,7 +139,38 @@
...
+                              (when-let [hiveminds (filter #(= "Hivemind" (:title %)) (all-active-installed state :runner))]
+                                        (doseq [h hiveminds]
+                                               (update! state side (assoc-in h [:counter :virus] 0)))))
+                 :msg (msg (let [local-virus (get-counters card :virus)
+                                 global-virus (get-virus-counters state card)
+                                 hivemind-virus (- global-virus local-virus)]
+                             (str "gain " (* 2 global-virus) " [Credits], removing " (quantify local-virus "virus counter") " from Consume"
+                             (when (pos? hivemind-virus)
+                                   (str " (and " hivemind-virus " from Hivemind)")))))}
+                {:effect (set-autoresolve :auto-accept "adding virus counters")}]}
```
NoahTheDuke:
```suggestion
                (set-autoresolve :auto-accept "adding virus counters")]}
```

src/clj/game/cards/programs.clj
```diff
@@ -707,12 +701,14 @@
...
                                              {:prompt "Spend a power counter on Nyashia to access 1 additional card?"
+                                              :autoresolve (get-autoresolve :auto-nyashia)
                                               :yes-ability {:msg "access 1 additional card from R&D"
                                                             :effect (effect (access-bonus :rd 1)
                                                                             (add-counter card :power -1)
                                                                             (clear-wait-prompt :corp))}
                                               :no-ability {:effect (effect (clear-wait-prompt :corp))}}}
-                                            card nil))}}}
+                                            card nil))}}
+    :abilities [(get-autoresolve :auto-nyashia "Nyashia")]}
```
NoahTheDuke:
```suggestion
    :abilities [(set-autoresolve :auto-nyashia "Nyashia")]}
```

presheaf:
Thanks, nice catches!

PR #4024, opened 2019-02-10
Add toast prompt for Mti Mwekundu: Life Improved 

### LINKED ISSUE #3618: Mti Mwekundu: Life Improved has no usage prompt
The last two days I've been testing this ID, and got confused as it is marked as fully automated on the automation page, but didn't figure out how to use it. After some experimentation I found it involves clicking the ID when getting the "action before access" prompt. I might have missed it but would it be possible to have a pop-up or message indicating when it's possible to use Mti Mwekundu? Thank you!

ChooJeremy:
Fixes #3618

Added an info toast that appears when the runner has approached the server and Mti is available for use.
image

Not sure if this is the proper way of implementing it, I basically created a new event for it and checked if it has fired so that the toast would only appear if Mti is usable.

Also, I found a bug where after using the ability it would print something like:

> j installs ICE protecting [:rd].

So I fixed it to:
> j installs ICE protecting R&D.

### DIFF
--- a/src/clj/game/cards/identities.clj
+++ b/src/clj/game/cards/identities.clj
@@ -762,23 +762,24 @@
       :abilities [ability]})
 
    "Mti Mwekundu: Life Improved"
-   {:abilities [{:effect (req (trigger-event state :corp :mti-ability))}]
-    :events {:approach-server {:req (req (not (first-event? state side :mti-ability)))
-                               :effect (req (toast state :corp "You may use Mti Mwekundu: Life Improved to install ice from HQ." "info"))}
-             :mti-ability {:once :per-turn
-                           :label "Install a piece of ice from HQ at the innermost position"
-                           :req (req (and (:run @state)
-                                          (zero? (:position run))
-                                          (not (contains? run :corp-phase-43))
-                                          (not (contains? run :successful))))
-                           :prompt "Choose ICE to install from HQ"
-                           :msg "install ice at the innermost position of this server. Runner is now approaching that ice"
-                           :choices {:req #(and (ice? %)
-                                                (in-hand? %))}
-                           :effect (req (corp-install state side target (zone->name (first (:server run)))
-                                                      {:ignore-all-cost true
-                                                       :front true})
-                                        (swap! state assoc-in [:run :position] 1))}}}
+   (let [ability {:once :per-turn
+                  :label "Install a piece of ice from HQ at the innermost position"
+                  :req (req (and (:run @state)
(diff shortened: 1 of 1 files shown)


presheaf:
Nice! When playing Mti I took quite some time to tell when the ability was supposed to be clicked, so this is definitely worthwhile. 

As for the way you've done it, I think we can find a way to tell whether an ability has triggered without making a special event. I'm not 100% sure on what the idiomatic way to check if an ability has triggered this turn is, but I think `can-trigger?` in `abilities.clj` is what you want. Failing that, at least its source code should tell you where to check in the state to find out. If this proves tricky let me know and I can help you puzzle it out.

NoahTheDuke:
Agreed with presheaf! This is great and should probably not using a unique `trigger-event`.

ChooJeremy:
Thanks! I'll take a look into `can-trigger?`

src/clj/game/cards/identities.clj
```diff
@@ -762,19 +762,24 @@
...
-                 :prompt "Choose ICE to install from HQ"
-                 :msg "install ice at the innermost position of this server. Runner is now approaching that ice"
-                 :choices {:req #(and (ice? %)
-                                      (in-hand? %))}
-                 :effect (req (corp-install state side target (:server run) {:ignore-all-cost true
-                                                                             :front true})
-                              (swap! state assoc-in [:run :position] 1))}]}
+   (let [ability {:once :per-turn
+                  :label "Install a piece of ice from HQ at the innermost position"
+                  :req (req (and (:run @state)
```
NoahTheDuke:
```suggestion
                  :req (req (and run
```

src/clj/game/cards/identities.clj
```diff
@@ -762,23 +762,24 @@
...
+                  :choices {:req #(and (ice? %)
+                                       (in-hand? %))}
+                  :effect (req (corp-install state side target (zone->name (first (:server run)))
+                                             {:ignore-all-cost true
+                                              :front true})
+                               (swap! state assoc-in [:run :position] 1))}]
+     {:abilities [ability]
+      :events {:approach-server {:req (req (can-trigger? state side ability card nil))
+                                 :effect (req (toast state :corp "You may use Mti Mwekundu: Life Improved to install ice from HQ." "info"))
+                                 :msg "test message 2" }}})
```
NoahTheDuke: :eyes:

ChooJeremy: Oh whoops!

Issue #4012, opened 2019-01-30
Dadiana Chacon doesn't trash itself if runner drops to 0c after trace


ChooJeremy:
Reproduction steps
* Runner has Dadiana Chacon installed
* Runner makes a run
* Runner hits an ice with a trace subroutine (i.e. surveyor)
* Corp fires the trace
* Runner spends all his credits on the trace

Runner is now at 0 credits. Dadiana Chacon isn't trashed. Even after the run finishes, Dadiana Chacon is still installed and runner doesn't take any damage.

Note:
The message `XX trashes Dadiana Chacon and suffers 3 meat damage.` appears, but nothing happens.

NoahTheDuke:
Thanks for the report! I'll take a look.

presheaf:
Just at a guess, I would think the problem is partially that Dadiana prints that it trashes itself when _registering_ the event which is actually going to trash it, not when the event resolves - this is probably why you see the log message being printed but nothing happening. It can be fixed by moving the `system-msg` to a `:msg` in the registered event. 

However, I'm not sure why Chacon doesn't actually trash himself. Maybe the event never actually resolves. I notice that Order of Sol accomplishes its effect by resolving an ability instead. This feels like a more correct way to do it, but looking at the blame it seems it was changed to an event to fix #3004. 

Luckily for us, the new rules sequence actually fixes #3004 for us, as playing cards should now 1) move them to play area, 2) pay the play cost. So I think the way to fix this bug _should_ actually be reverting Chacon back to resolving an ability, then modifying whichever of `play-instant` or `complete-play-instant` in `rules.clj` to pay the cost only after moving the event to the play area. I use "should" here because I think it would be perfectly acceptable to figure out why the trash Chacon event isn't resolving and making sure it does, but (famous last words) I don't think my proposed fix is too hard, and being compliant with the rules seems like a good thing.

PR #4026, opened 2019-02-12
Changed event play rules to be compliant with Nisei rules. Fixed Chacon+trace bug

### LINKED ISSUE #3004: Dadiana Cachon can trash the card being played
With Dadiana Cachon installed, if you try to install a card that costs all of your credits, the meat damage resolves before the card leaves your hand, and can be discarded by the damage.

presheaf:
The event is now moved to play area *before* paying the cost, not after. This means that if the cost is not successfully paid (can happen with Price of Freedom, thank you tests), the card must be moved back, but aside from that, I don't think there are any new complications. This change is kinda deep in the system though, so I'd appreciate a second pair of eyes on this.

Doing this allows us to effectively undo ae4c08a, which fixed #3004, so I did, fixing #4012 . Was there any other reason you changed Chacon's watch from resolving an ability to triggering an event, @danhut ?

### DIFF
--- a/src/clj/game/cards/resources.clj
+++ b/src/clj/game/cards/resources.clj
@@ -550,11 +550,10 @@
                                  (trash card {:cause :ability-cost}))}]}
 
    "Dadiana Chacon"
-   (let [trashme {:effect (effect (system-msg "trashes Dadiana Chacon and suffers 3 meat damage")
-                                  (register-events {:play {:req (req (= "Runner" (:side target)))
-                                                           :effect (effect (unregister-events card)
-                                                                           (damage eid :meat 3 {:unboostable true :card card})
-                                                                           (trash card {:cause :ability-cost}))}} card))}
+   (let [trashme {:effect (effect (unregister-events card)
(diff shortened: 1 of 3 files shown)


presheaf:
Also, I was thinking about making a test verifying that you can't play Sure Gamble and have Chacon trash it out of hand, but I didn't see how. I could just try 10 times and verify that it didn't happen any of the times, but that seemed bad.

NoahTheDuke:
Do you want to move the card back to the hand unconditionally? Or do you want to move it to wherever the card came from? I don't remember how to check that, but probably in `:zone` or something.

presheaf:
Good point - if the event was played using e.g. Same Old Thing it should go into its old zone.

NoahTheDuke:
If we felt spicy, we could use this opportunity to fix a big hole in our handling of card effects: we move the card directly to the heap once the `:effect` has fired, instead of waiting until some equivalent of `effect-completed` to indicate that the entire card's effect has finished, meaning that we have to bind a lot of card effects to the card being in the discard pile.

PR #4027, opened 2019-02-12
Refactor card and icon rendering

### LINKED ISSUE #4017: Border control chat message shows "[credits]" instead of credit symbol

ouroboros8:
Implement a new common functions engine for interspersing card and icon HTML into text strings containing card titles or icon short codes.

Previously most icon interpolation was handled by the `add-symbols` function, which produced stringified HTML which was then added to relevant DOM elements using `:dangerouslySetInnerHTML`. Because this method is unsafe for user-supplied input (as it would allow HTML injection) a different mechanism (largely accessed through the `create-span` function) was used by the log pane and global chat components This latter mechanism also inserts card previews.

The new functions use a common mechanism for both kinds of replacement based on regex -> replacement hashmaps (`card-smap` and `icon-smap`). Functionality provided by `add-symbols` is replaced with `render-icons`, and `create-span` is replaced by `render-icons-and-cards`. Both of these functions take a string or HTML fragment (in reagent's preferred vector form) and return another HTML fragment containing a mix of plain text and icon/card spans.

Fixes #4017

### DIFF
--- a/src/cljs/nr/cardbrowser.cljs
+++ b/src/cljs/nr/cardbrowser.cljs
@@ -8,7 +8,7 @@
             [nr.account :refer [alt-art-name]]
             [nr.ajax :refer [GET]]
             [nr.utils :refer [toastr-options banned-span restricted-span rotated-span
-                              influence-dots slug->format format->slug]]
+                              influence-dots slug->format format->slug render-icons]]
             [reagent.core :as r]))
 
 (def cards-channel (chan))
@@ -99,37 +99,6 @@
   [only-version cards]
   (reduce (partial expand-alts only-version) () (reverse cards)))
 
-(defn add-symbols [card-text]
-  (-> (if (nil? card-text) "" card-text)
-      (make-span "\\[Credits\\]" "credit")
-      (make-span "\\[Credit\\]" "credit")
-      (make-span "\\[Click\\]" "click")
-      (make-span "\\[Subroutine\\]" "subroutine")
-      (make-span "\\[Recurring Credits\\]" "recurring-credit")
-      (make-span "\\[recurring-credit\\]" "recurring-credit")
-      (make-span "1\\[Memory Unit\\]" "mu1")
-      (make-span "2\\[Memory Unit\\]" "mu2")
-      (make-span "3\\[Memory Unit\\]" "mu3")
-      (make-span "\\[Memory Unit\\]" "mu")
-      (make-span "1\\[mu\\]" "mu1")
-      (make-span "2\\[mu\\]" "mu2")
-      (make-span "3\\[mu\\]" "mu3")
-      (make-span "\\[mu\\]" "mu")
-      (make-span "\\[Link\\]" "link")
-      (make-span "\\[Trash\\]" "trash")
-      (make-span "\\[adam\\]" "adam")
-      (make-span "\\[anarch\\]" "anarch")
-      (make-span "\\[apex\\]" "apex")
-      (make-span "\\[criminal\\]" "criminal")
-      (make-span "\\[hb\\]" "haas-bioroid")
-      (make-span "\\[haas-bioroid\\]" "haas-bioroid")
-      (make-span "\\[jinteki\\]" "jinteki")
-      (make-span "\\[nbn\\]" "nbn")
-      (make-span "\\[shaper\\]" "shaper")
-      (make-span "\\[sunny\\]" "sunny")
-      (make-span "\\[weyland\\]" "weyland-consortium")
-      (make-span "\\[weyland-consortium\\]" "weyland-consortium")))
-
 (defn non-game-toast
   "Display a toast warning with the specified message."
   [msg type options]
@@ -213,7 +182,7 @@
    [:div.text
     [:p [:span.type (str (:type card))] (if (empty? (:subtype card))
                                           "" (str ": " (:subtype card)))]
-    [:pre {:dangerouslySetInnerHTML #js {:__html (add-symbols (:text card))}}]
+    [:pre (render-icons (:text (first (filter #(= (:title %) (:title card)) @all-cards))))]
 
     [:div.formats
      (doall (for [[k name] (-> slug->format butlast)]
--- a/src/cljs/nr/gameboard.cljs
+++ b/src/cljs/nr/gameboard.cljs
@@ -304,72 +304,6 @@
   (and (get-in @game-state [:options :spectatorhands])
        (not (not-spectator?))))
 
-(def ci-open "\u2664")
-(def ci-seperator "\u2665")
-(def ci-close "\u2666")
-
-(defn is-card-item [item]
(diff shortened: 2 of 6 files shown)


src/cljs/nr/gameboard.cljs
```diff
@@ -665,7 +595,8 @@
        (fn [i ab]
          [:div {:on-click #(do (send-command "corp-ability" {:card card
                                                              :ability i}))
-                :dangerouslySetInnerHTML #js {:__html (add-symbols (str (ability-costs ab) (:label ab)))}}])
+                }
```
NoahTheDuke: 👀

src/cljs/nr/utils.cljs
```diff
@@ -1,5 +1,6 @@
 (ns nr.utils
-  (:require [clojure.string :refer [join] :as s]))
+  (:require [clojure.string :refer [join, lower-case, split] :as s]
```
NoahTheDuke: Style thing: commas are considered white-space by Clojure, so we typically leave them out.

src/cljs/nr/utils.cljs
```diff
@@ -91,3 +92,138 @@
...
+  (into [] (concat [head (merge attr {:key n})] tail))))
+
+(defn render-fragment-impl [fragment replacement-smap]
+  "Given a fragment, shallowly replaces text in the fragment with icon and,
+  optionally, card preview HTML"
+  (let [counter (atom 0)
+        set-next-key (fn [elem] (set-react-key (do (swap! counter inc) @counter) elem))]
+    (->> (reduce replace-in-fragment fragment (ordered-keys replacement-smap))
+         (replace replacement-smap)
+         (map-if vector? set-next-key)
```
NoahTheDuke: I think `map-if` should either be inlined or defined in the above let-binding. Having it at the top of the file makes discoverability and understanding harder.

src/cljs/nr/gameboard.cljs
```diff
@@ -305,72 +304,6 @@
...
-  (and (> (.indexOf item ci-seperator) -1)
-       (= 0 (.indexOf item ci-open))))
-
-(defn extract-card-info [item]
-  (if (is-card-item item)
-    [(.substring item 1 (.indexOf item ci-seperator))
-     (.substring item (inc (.indexOf item ci-seperator)) (dec (count item)))]))
-
-(defn create-span-impl [item]
-  (if (= "[hr]" item)
```
NoahTheDuke: Looks like this has been left out.

src/cljs/nr/gameboard.cljs
ouroboros8: Oopsie :)

src/cljs/nr/utils.cljs
ouroboros8: Damn, I kept catching myself inserting unnecessary commas, good catch! Need to get a linter :D

src/cljs/nr/utils.cljs
ouroboros8: Sounds good to me :+1:

src/cljs/nr/gameboard.cljs
ouroboros8: huh, yes I must have missed that, will take a look

src/cljs/nr/utils.cljs
```diff
@@ -91,3 +92,137 @@
...
+          "[nbn]" "nbn"
+          "[shaper]" "shaper"
+          "[sunny]" "sunny"
+          "[weyland]" "weyland-consortium"
+          "[weyland-consortium]" "weyland-consortium"}
+      (map (fn [[k v]] [(regex-of k) (span-of v)])))))
+
+(defn card-smap-impl []
+  "A map of case sensitive card regexes to card span fragments"
+  (letfn [(unpack-codes [card] [:title card :code card])
```
NoahTheDuke: Watch out for lack of `()`, as this will evaluate to `[:title {:title "Card Name" ...} :code {:title "Card Name" ...}]`

src/cljs/nr/gameboard.cljs
ouroboros8: Added this in b683f9da157eca86225e6a8cb46cee07f1013b7b

src/cljs/nr/utils.cljs
ouroboros8: Now using `juxt` as you suggested in slack :+1:

src/cljs/nr/cardbrowser.cljs
```diff
@@ -213,7 +182,7 @@
    [:div.text
     [:p [:span.type (str (:type card))] (if (empty? (:subtype card))
                                           "" (str ": " (:subtype card)))]
-    [:pre {:dangerouslySetInnerHTML #js {:__html (add-symbols (:text card))}}]
+    [:pre (render-icons (:text (first (filter #(= (:title %) (:title card)) @all-cards))))]
```
NoahTheDuke:
```suggestion
    [:pre (render-icons (:text (get @all-cards (:title card))))]
```
Does the original work? `@all-cards` should be a map with the card title as key. You could also do `(get-in @all-cards [(:title card) :text])` if you prefer the look of that.

NoahTheDuke:
This is looking great! Love the doc strings, love the refactoring you're doing. Very much appreciated.

ouroboros8:
> This is looking great! Love the doc strings, love the refactoring you're doing. Very much appreciated.

I've just updated the ones on the core functions to go into the implementation in a bit more detail, hope that's helpful!

src/cljs/nr/cardbrowser.cljs
ouroboros8: Nope the original always renders and empty pre-tag (`add-symbols` was permissive and would spit out empty string if given `nil`)

Sadly `@all-cards` is a sequence of maps:
```=> (take 2 @all-cards)
({:faction "Neutral",
  ...
  :title "The Shadow: Pulling the Strings",
  ...
  :text
  "Draft format only.\nYou can use agendas from all factions in this deck."}
 {:faction "Neutral",
  ...
  :title "The Masque: Cyber General",
  ...
  :text "Draft format only."})
```
and I couldn't figure out a more elegant way to query that structure.

NoahTheDuke: Yikes. Why is that happening? Let me look into this, cuz that shouldn't be the way it is...

PR #4028, opened 2019-02-12
Fixed Ash prematurely ending the run. Added regression test.

### LINKED ISSUE #3995: Stimhack run ends at the wrong time with Ash
Stimhack run on a remote with Ash and an agenda. 3 real credits available and 3 stimhack credits at access. Ash trace fires, but then the stimhack damage and end of run triggered before the option to trash Ash was available.

presheaf:
Fixes #3995. This one I'm actually puzzled hasn't become an issue before, to be honest, because as far as I can tell the run seems to instantly end when the Ash trash prompt comes up.

### DIFF
--- a/src/clj/game/cards/upgrades.clj
+++ b/src/clj/game/cards/upgrades.clj
@@ -57,7 +57,9 @@
                                                         (swap! state update-in [:run :run-effect]
                                                                #(assoc % :replace-access
                                                                        {:mandatory true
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
It's possible something in our run timing code changed or how awaitable abilities work changed. The old test wasn't checking that the run still existed before clicking the prompt, so it's no wonder it fell through the cracks.

PR #4030, opened 2019-02-13
Refactor Start of Turn 

NoahTheDuke:
This is a redo of #4011, leaving out all of the major card refactoring.

# Changelog

* Change `:X-turn-begins` to use `trigger-event-simult`. Functionally, this changes very little as the only start of turn effect marked `:interactive` is Marilyn Campaign. However, this lays the groundwork for other cards to also be marked `:interactive` and finer handling of complex start of turn effects.
* Move click-gaining logic to before we trigger `:X-phase-12`, so as to more closely align with current turn timing rules. Functionally, this changes very little as no current implementations expect otherwise.
* Include check in "Action Phase" button logic to be be disabled if during `:X-phase-12`, as click gaining happens before this phase now.


Saintis:
As an added bonus, this should make Jak Sinclair work better :)

NoahTheDuke:
Yes indeed!

PR #4034, opened 2019-02-14
Some quick fixes (Liquidation, RNG Key, Psychographics)

kevkcc:
Psychographics rewritten so that you can't pay 5 credits to place 3 advancement tokens because the runner has 3 tags.

### DIFF
--- a/src/clj/game/cards/operations.clj
+++ b/src/clj/game/cards/operations.clj
@@ -896,7 +896,7 @@
 
    "Liquidation"
    {:async true
-    :req (req (some #(rezzed? %) (all-installed state :corp)))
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
Some agendas are installed face-up, so you shouldn’t be checking against that.

kevkcc:
I thought faceup agendas weren't technically rezzed, that's why I put that in

NoahTheDuke:
Hmmm lemme check on that

NoahTheDuke:
> Faceup agendas are neither rezzed nor unrezzed.

3.2.3.a of Comp Rules v1.1. You’re right, the non-agenda req stays in!

PR #4036, opened 2019-02-14
Card Definition Macro Refactor and DRY (req, effect, msg)

NoahTheDuke:
This is a big refactor of our current card definition macros: `req`, `effect`, and `msg`. The goal is to increase performance and DRY primary places for errors.

## Why
Right now, our card def macros share only some of the helper bindings. `effect` has the fewest, `req` and `msg` shadow the game.core function `run`, some bindings are only used in a single card. It's kind of a mess, it's hard to remember what does or doesn't exist, and most of the time, less than half of the bindings are used in a given card def which means we're creating those bindings every time for nothing.

This is especially visible when writing `(req true)`, which is simply a way of taking in the 5-function arguments and returning `true` unconditionally, but right now let-binds 24 variables of varying complexity. Those situations could be rewritten as `(constantly true)`, but it's cleaner to be consistent with `req`.

This leads me to the conclusion that we shouldn't be binding those variables when it's unnecessary and we should be consistent across all macros.

## How
In short, we parse the AST of the functions the macro wraps, building a set of all free locals in it, aka all symbols that don't resolve to a namespace. Instead of listing all our helper vars in a let, we put them in a quoted list, pair them up, take their symbol and use it as the key. With our set of free locals, we build a list of all the used forms and then unwrap that list into our `let`.

With the power of the `~@` macro operator, we can reduce `effect` and `msg` down to variations on `req`: `effect` applies a pre-processor to each function call in the macro, and `msg` wraps the output in a post-processor.

## Notes:
* Unified `req`, `effect`, and `msg` locals into a single set. This required changing the `game.core` function `run` to be `make-run`, as changing the `effect` local `run` made less sense and was overall less clear: the action you're taking is to "make a run", so naming the function after that makes the most sense.
* Applied minor clean-up to `wait-for`, to make it hygienic. (Doesn't affect any current code, but it's much safer now to heavily nest `wait-for`s.)
*  Because this is done during compile, it improves performance in certain use-cases and has zero performance decreases. On some machines it might increase compile times, but I've run this a bunch and have seen no real differences.
* Because only the specifically used locals are being included in the `let`, care must be taken when creating new ones: They cannot reference each other and can only reference functions that will exist wherever they're used. This impacts locals such as `run-ices` and `current-ice`, which relied on `run-server` and `run-position` in their definitions. The affected locals have been changed to not rely on any others, defining all their functionality inline. This makes them more wordy, but overall changes nothing. (Honestly, I'd like to pull a lot of this functionality out into `game.core` functions that we can use in here, but c'est la vie)

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -35,6 +35,8 @@
                  [cljsjs/react "16.5.2-0"]
                  [cljsjs/react-dom "16.5.2-0"]
                  [zprint "0.4.10"]
+                 [org.clojure/tools.analyzer "0.7.0"]
+                 [org.clojure/tools.analyzer.jvm "0.7.2"]
                  [org.clojars.frozenlock/reagent-modals "0.2.8"]
                  [hawk "0.2.11"]]
 
--- a/src/clj/game/cards/agendas.clj
+++ b/src/clj/game/cards/agendas.clj
@@ -1289,8 +1289,8 @@
                                            :choices (vec (reverse (map str (range (inc num-ice)))))
                                            :effect (req (corp-install state side chosen-ice chosen-server
                                                                          {:ignore-all-cost true :index (Integer/parseInt target)})
-                                                        (if (and current-run
-                                                                 (= (zone->name (first (:server current-run)))
+                                                        (if (and run
+                                                                 (= (zone->name (first (:server run)))
                                                                     chosen-server))
                                                           (let [curr-pos (get-in @state [:run :position])] 
                                                             (if (>= curr-pos (Integer/parseInt target))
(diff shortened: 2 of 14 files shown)


Saintis:
Nice stuff! Looks great!

Annother approach would be to rename the `run` function to `make-run` and keep the macro binding.

NoahTheDuke:
Oh yeah, that’s a good idea!

NoahTheDuke:
@Saintis It's much more readable and clear this way, so I've switched to it! Thanks for the idea.

presheaf:
I think you're right that locals like `run-position` should be abstracted out. Would there be a drawback to doing this?

NoahTheDuke:
Not that I can think of. I think a refactor like that should be a different PR once this is merged tho. Trying to change as little as possible, lol.

NoahTheDuke:
Yep! That’s what the `find-undefined-locals` function is doing. It’s wild this is possible!

Issue #4020, opened 2019-02-08
enhancement request: end of turn discard for runner should name cards discarded


triorph-emacs:
When you drag a card from grip to heap for your end of turn discard, it names the card for the corp to see what is gone, however if the new "end of turn" discard feature is used, the name is not shown, and its quite easy for the corp to miss something important.

This is a request for a small feature enhancement where the cards are named as they're discarded.

PR #4038, opened 2019-02-17
Added list of discarded runner cards end of turn.

lostgeek:
Fix for issue [#4020](https://github.com/mtgred/netrunner/issues/4020)


NoahTheDuke:
Welcome to jnet development! This looks great. Let me offer one suggestion: instead of placing the list after a `:`, wrap the list in parentheses. That’s currently the style for other similar effects and consistency is good.

lostgeek:
Redid it to fit other similar texts. Also realized, that this step is not "trashing" but "discarding" cards 😁 .

image

NoahTheDuke:
Ooo this is a much better fix! Thank you so much!

PR #4040, opened 2019-02-19
Corrected misplaced bracket in The Artist test

presheaf:
Seems a bracket accidentally barfed its contents.


NoahTheDuke:
Ha! That’s annoying. Thanks for the catch.

PR #4042, opened 2019-02-19
The Nihilist

presheaf:
Straightforward implementation. Was unsure about how it should fit into phase 1.2, but I think there aren't that many cases where you care about the ordering anyway, so I elected to leave it for the big 1.2. cleanup. Let me know if you object.

### DIFF
--- a/src/clj/game/cards/resources.clj
+++ b/src/clj/game/cards/resources.clj
@@ -2028,12 +2028,11 @@
                                                        :effect (effect (breaker-strength-bonus 2))}}) card))}}
 
    "The Nihilist"
-   (let [has-2-virus-tokens? (req (> 2 (reduce + (map #(get-counters % :virus) (get-all-installed state)))))
+   (let [has-2-virus-tokens? (req (<= 2 (reduce + (map #(get-counters % :virus) (get-all-installed state)))))
          corp-choice {:optional {:player :corp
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/resources.clj
```diff
@@ -2027,6 +2027,35 @@
                                 :pre-breaker-strength {:req (req (= (:cid target)(:cid (:hai-target card))))
                                                        :effect (effect (breaker-strength-bonus 2))}}) card))}}
 
+   "The Nihilist"
+   (let [has-2-virus-tokens? (req (<= 2 (reduce + (map #(get-counters % :virus) (get-all-installed state)))))
```
NoahTheDuke: Not that it matters much, but is there any difference between `reduce +` and `+`?

presheaf: Oh, I forgot + can take multiple arguments. Then no!

presheaf: Oh, wait, I also forgot + doesn't take a list. Oops. So `reduce +` could become `apply +`.

Saintis: Well you already have a map, so maybe you want something like `(reduce %(+ %1 (get-counters %2 :virus)) 0 (get-all-installed state))`.

Saintis: Ah, yes, using an existing function seems much better :)

PR #4054, opened 2019-03-04
Expand Optional Prompts to handle Choices

NoahTheDuke:
Look at Rex Campaign to understand what's going on.

I'll write more later.


presheaf:
Thanks for fixing the autoresolve toast!

PR #4056, opened 2019-03-04
Update http-kit to 2.4.0-alpha3 to support Java 11

kevkcc:
As far as I can tell, our dependency on Java 8 goes away if we upgrade to this alpha version of http-kit. Don't know if we want to hold off till the official 2.4.0 though.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -23,7 +23,7 @@
                  [crypto-password "0.2.0"]
                  [binaryage/devtools "0.9.7"]
                  [digest "1.4.6"]
-                 [http-kit "2.3.0"]
+                 [http-kit "2.4.0-alpha3"]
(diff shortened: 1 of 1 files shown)


NoahTheDuke:
We should probably wait. We're not missing anything important from Java 11 right now.

jwarwick:
Last time I tried one of the alpha releases (don't think it was 3), I saw a number of warnings at runtime; we should verify those aren't anything important.

It would be nice to get Java11 support, since it's the default now. IIRC, `lein fetch` was the thing that currently breaks without this upgrade.

kevkcc:
Yea lein fetch is the only breaking change. Anyways, I submitted this PR because I set up Java 11 on my machine and this is what I had to change. Feel free to try it out and whether or not we merge (or when) will be left up to you guys.

PR #4070, opened 2019-03-13
Fix simultaneous ability reqs, Handle Asa/419

### LINKED ISSUE #3590: 419 + Asa: 419 not triggering
When Asa uses its ability to install two card in one click, 419 should trigger on the first install but does not trigger at all. This has happened in two separate games, with combinations of two ice, an ice and an agenda, and an agenda and an upgrade being installed at once.

NoahTheDuke:
In trying to fix the Asa/419 bug, I found out that when we generate event-handlers in `trigger-event-sync` and `trigger-event-simult`, because they use `filter`, the LazySeq is built but not realized. This means the event-handler `:req`s weren't being checked until they were being consumed. Thus, 419 wouldn't trigger because by the time we'd gotten to resolving the `opponent-events`, Asa's ability had executed twice and `first-event?` would return false.

Solution 1) Wrap the filter in a `doall` call, which forces evaluation.

HOWEVER, this breaks how we handle currents! Because currents are trashed in the `:first-ability` called when an agenda is scored or stolen, and New Angeles Sol can play the current that was just trashed by the runner stealing an agenda, when we would check Sol's `:req` to see if a current existed to be played, it wasn't in the trash yet.

Solution 2) Move generating (and thus `:req` checking) the active and opponent event handlers to happen _after_ we've resolved the `:first-ability`.

Closes #3590.

### DIFF
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -71,16 +71,16 @@
    {:effect (effect (gain-credits 3))
     :msg (msg "gain 3 [Credits]")
     :abilities [{:label "[Trash]: Install a non-agenda card from HQ"
-                 :effect (effect (trash card {:cause :ability-cost})
-                                 (corp-install target nil))
-                 :msg (msg (corp-install-msg target))
+                 :async true
                  :prompt "Select a non-agenda card to install from HQ"
-                 :priority true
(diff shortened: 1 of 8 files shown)


presheaf:
Nice read and catches!

PR #4079, opened 2019-03-18
Downfall implementation

NoahTheDuke:
All but MirrorMorph are implemented in this, along with _a lot_ of positive engine and card bug fixes and refactors.

I'll have to see about generating some complete list of changes.


NoahTheDuke:
@nealterrell This is good to go.

PR #4087, opened 2019-03-18
Remove data files from repo

NoahTheDuke:
With the release of Downfall, my experiment with relying on the data files to handle all card information has proven to be both unwieldy and wholly unnecessary. In removing them we go back to relying completely on `lein fetch` to populate our database and can achieve the same db-less testing system by placing all card data into a single untracked file.

Sorry for the hassle over the last 6 months, y'all. lol Just tryna up my contributor stats. :smirk:


Saintis:
This does keep this repo a bit lighter and separates the data a bit :)

PR #4090, opened 2019-03-18
Fixed gain-credit in Exclusive Party

### LINKED ISSUE #4089: Interface Locks up when Exclusive Party is Played
Problem:  When Exclusive Party is played, interface locks up (as if disconnected or lagging).  If you refresh and rejoin, exclusive party will have executed and be in the temporary zone.

To replicate:  Play Exclusive Party (so far as I can tell anytime).

jwarwick:
Fixes #4089


NoahTheDuke:
Thanks for the test!

PR #4095, opened 2019-03-18
Fixed bad card-def calls in tests

jwarwick:


NoahTheDuke:
lol I'm so annoyed this had to be done. I did this in the Downfall branch a while back, but I think a bad merge overwrote it. Thanks for catching these.

PR #4115, opened 2019-03-19
Fixed Earthrise Hotel effect being uncompleted

### LINKED ISSUE #4113: Lots of "Waiting for X to..." prompts never closing
With the release of Downfall, I find myself /close-prompting a lot because the game seems to wait for stuff which is finished. I suspect multiple cards are to blame, so keeping the issue name general. Please comment if you notice this happening with a card not in the list below:

* Earthrise Hotel (as seen in [the code,](https://github.com/mtgred/netrunner/blob/master/src/clj/game/cards/resources.clj#L856) does not register its effect as completed unless on the final tick)

presheaf:
Should fix #4113 .

### DIFF
--- a/src/clj/game/cards/resources.clj
+++ b/src/clj/game/cards/resources.clj
@@ -854,8 +854,9 @@
                   :req (req (:runner-phase-12 @state))
                   :async true
                   :effect (req (wait-for (draw state :runner 2 nil)
-                                         (when (zero? (get-counters (get-card state card) :power))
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
lol I _just_ finished fixing this myself, but you got it first!

presheaf:
Haha, sorry about that. :P Want to commit to the branch directly so you get that sweet credit, or do you mind if i just copy paste it?

PR #4122, opened 2019-03-19
WIP: Make end-turn effects fire on both turns

### LINKED ISSUE #4106: Class Act does not draw cards when installed off a Street Peddler on Corp turn
Here's a copied and pasted game log for a game I was spectating:

> CORP started their turn 4 with 8  and 5 cards in HQ.
> CORP spends  and 1  to advance a card in Server 1.
> CORP spends  and 1  to advance a card in Server 1.
> CORP spends  and 1  to advance a card in Server 1.
> CORP scores The Future Perfect and gains 3 agenda points.
> RUNNER uses Street Peddler to install The Class Act lowering its install cost by 1 .
> RUNNER spends 2  to install The Class Act.
> RUNNER uses Az McCaffrey: Mechanical Prodigy to reduce the install cost of The Class Act by 1 .
> CORP discards 1 card from HQ at end of turn.
> CORP is ending their turn 4 with 5 and 5 cards in HQ.

Class Act did not cause the runner to draw any cards, either on this turn or their following turn.

NoahTheDuke:
This is a small change but potentially large: Now `:end-turn` effects in card defs will fire on whichever turn it's registered, not on the card side's next turn. Additionally, all of those effects are removed immediately after firing.

Fixes #4106

### DIFF
--- a/src/clj/game/cards/resources.clj
+++ b/src/clj/game/cards/resources.clj
@@ -2015,7 +2015,7 @@
                                                           (fn [coll]
                                                             (remove-once #(= (:cid %) (:cid target)) coll)))
                                     {:cause :ability-cost})
-                             (runner-install state side (dissoc target :facedown))))}]}
+                             (runner-install state side eid (dissoc target :facedown) nil)))}]}
 
    "Symmetrical Visage"
    {:events {:runner-click-draw {:req (req (genetics-trigger? state side :runner-click-draw))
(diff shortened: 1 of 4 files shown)


NoahTheDuke:
So this isn't as simple as "move all end-turn effects to happen on both turns", cuz that means something like Viral Weaponization + The Class Act won't fire on the same trigger. What needs to happen is all `:end-turn` effects need to be moved to `:effect`s like Viral Weaponization, where they register an event the turn they're installed, and then unregister that event when it triggers.

PR #4130, opened 2019-03-20
Move Chisel to an ability

NoahTheDuke:
No issue for this yet, but it should have been done in pull 146 in the future fork. Very disappointed with myself that I missed it.


src/clj/game/cards/programs.clj
```diff
@@ -598,20 +598,20 @@
     :abilities [{:cost [:credit 1] :msg (msg "break 1 " (:subtype-target card) " subroutine")}]}
 
    "Chisel"
-   {:effect (effect (system-say "Corp, please press No Action to trigger Chisel"))
+   {:implementation "Encounter effect is manual."
     :hosting {:req (every-pred ice? can-host?)}
+    :abilities [{:req (req (= (:cid current-ice) (:cid (:host card))))
+                 :msg (msg (if (zero? (get-strength current-ice))
+                             (str "trash " (card-str state current-ice))
+                             (str "place 1 virus counter on " (card-str state card))))
```
Saintis: Might want to change to using a `system-msg` for the counter placement.

src/clj/game/cards/programs.clj
```diff
@@ -598,20 +598,20 @@
     :abilities [{:cost [:credit 1] :msg (msg "break 1 " (:subtype-target card) " subroutine")}]}
 
    "Chisel"
-   {:effect (effect (system-say "Corp, please press No Action to trigger Chisel"))
+   {:implementation "Encounter effect is manual."
     :hosting {:req (every-pred ice? can-host?)}
+    :abilities [{:req (req (= (:cid current-ice) (:cid (:host card))))
```
Saintis: There is a shortcut for this - `same-card?` or something

src/clj/game/cards/programs.clj
```diff
@@ -598,20 +598,20 @@
...
+                 :msg (msg (if (zero? (get-strength current-ice))
```
Saintis: What is the strength of an un-rezzed ice? Will this cause ice to be trashed if used erroneously on un-rezzed ice?

src/clj/game/cards/programs.clj
NoahTheDuke: Why? The effect is a firm OR: either ice strength is > 0 so you place a counter OR ice strength is <= 0 so you trash it.

src/clj/game/cards/programs.clj
NoahTheDuke: Good call. Probably should change all the other cards like this too then, lol.

src/clj/game/cards/programs.clj
Saintis: So it doesn't print the clunky "Runner uses Chisel to place 1 virus counter on Chisel."
In general we print "Runner places 1 virus counter on Chisel." instead for these kind of effects (I think)

src/clj/game/cards/programs.clj
NoahTheDuke: It calls `(or (:current-strength card) (:strength card))`, so it'll always populate with the "correct" number.

Saintis: Hmm, maybe the best way to handle unrezzed ice is to add `(rezzed? current-ice)` to the req (if that isn't already handled somehow with how current-ice is bound)

NoahTheDuke: Oh I see what you mean. Good call, I'll add a check for the ice being rezzed.

PR #4139, opened 2019-03-21
Better errors

jwarwick:


src/clj/game/core/cards.clj
```diff
@@ -13,7 +13,10 @@
   [card]
   (if-let [title (:title card)]
     (get cards title)
-    (.println *err* "Tried to select card def for non-existent card")))
+     (.println *err* (with-out-str
```
NoahTheDuke: Seems to be indented too much.

NoahTheDuke:
This looks _awesome_, John. Would it be possible to add the same logic to the "Error in a card-title prompt" and "Error in an integer prompt" as well, to cover all of our bases?

PR #4149, opened 2019-03-23
Fix SDS program selection (steal and score)

### LINKED ISSUE #4140: Spec Work does not work on hosted programs
Sahasara hosted on Net Chip. Runner plays Spec Work, can't select Sahasara for trashing and had to make the effect happen manually.

### LINKED ISSUE #4147: SDS Drone Deployment did not trigger on Corp Score
Scored the agenda which should prompt me to trash a program but did not receive a prompt and my opponent did not receive a prompt either. Could be because all the program cards are hosted on NetChips?

SDS not firing

lostgeek:
Closes #4147 (which was the bug on score). Also fixes a bug on steal, where trash would not happen, if all installed programs were hosted on other cards.

Closes #4140

### DIFF
--- a/src/clj/game/cards/agendas.clj
+++ b/src/clj/game/cards/agendas.clj
@@ -1239,7 +1239,7 @@
    "SDS Drone Deployment"
    {:steal-cost-bonus (req [:program 1])
     :effect (req (show-wait-prompt state :runner "Corp to use SDS Drone Deployment")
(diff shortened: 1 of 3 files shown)


lostgeek:
Nice. This seems to close #4140 as well.

NoahTheDuke:
Love the utility function.

PR #4158, opened 2019-03-26
Upgrade dependencies

NoahTheDuke:
Our dependencies are out of date across the board. I updated all of them, rebuilt the cljs, and played a couple games against myself to double check. Everything seems to be good.


NoahTheDuke:
Good call, I’ll try this out on the playtesters first lol

jwarwick:
Going to bump up to a modern Java version while you are at it? I think the `http-kit` dependency was the last hold up.

PR #4171, opened 2019-03-31
Card Bug Fixes: Lethia, Bug Out Bag, Architect Deployment Test, Flip Switch, CtM, Secure and Protect, Red Level Clearance, Khusyuk

### LINKED ISSUE #4049: Using hippo is not triggering Threat Assessment
trashing with hippo should enable the trigger.
Similar to #3628 it seems

### LINKED ISSUE #4061: Make SSL Endorsement autoresolvable
On Slack, someone suggested adding autoresolve to SSL Endorsement. I thought this was a good idea, but am postponing doing it myself because I thought it might be a good issue to get started with for people interested in development. So here are some pointers on where to begin:

The ability which needs `:autoresolve` added is a line or so away from here: https://github.com/mtgred/netrunner/blob/master/src/clj/game/cards/agendas.clj#L1180

You also need to add an ability allowing the user to set autoresolve settings (examples can be found in e.g. CtM or Aeneas Informant), which needs to be usable even while SSL Endorsement is in the Runner's score area (see 15 Minutes).

### LINKED ISSUE #4074: Erm... Amani Senai, Arabyata, Freedom Through Equality and Legwork?
I had Freedom Through Equality in play. I Legworked. On access the corp rezzed Amani and Arabyata. I accessed Traffic Accident, so nothing happened. Then I accessed AR Enhanced, so stole it and added FTE to my score area. Then it all went wrong.

Amani fired with a base strength zero, which we thought might be an FTE glitch? So we tried to resolve it but it seemed to keep refiring. We were able to close-prompt out of it. But the only options it gave me were for Successful Run and Jack Out, and neither of them were responding. Nothing responding on Corp side either, and close-prompt didn't work either. We had to abandon ship.

I was running as Kit, which I doubt is relevant. But the corp was Sol, which may have thrown another trigger into the mix.

### LINKED ISSUE #4077: Timely Public Release messages
Would it be possible to inform the runner of which piece of ice is installed using Timely Public Release, if that card was face up in Archives?

### LINKED ISSUE #4078: Data Loop on-encounter effect
Is it possible the corp could (also) activate the on-encounter effect for the runner?  This seems to be the only piece of ice where the runner has to do it, and so i'm often having to remind my opponents of this effect.

If there is a technical reason that prevents this, an option on Data Loop for a pre-written message asking the runner to use the effect would be really nice too.

### LINKED ISSUE #4108: Muertos Gang Member uninstall trigger when blanked by Dr. Lovegood
Muertos Gang Member fires the uninstall trigger when blanked by Dr. Lovegood, despite not being uninstalled.

Might also want to test its interaction with Apocalypse, and trashed while blanked.

### LINKED ISSUE #4125: Khusyuk run automatically procs The Turning Wheel
Got Khusyuk'd, and TTW got a counter before the Khusyuk reveal happened.

### LINKED ISSUE #4150: Red Level Clearance
Current implementation allows the corp to choose the same action twice.
In this case, ASA chose to draw 2 cards twice. Pretty sure this isn't the card's design/intention.

I think Deuces Wild doesn't allow this (drawing cards twice) etc.

### LINKED ISSUE #4163: Secure and Protect does not reveal the ICE installed
Secure and Protect requires the corp to reveal the ICE. The name of the ICE installed with Secure and Protect is not displayed in the chat log

image

### LINKED ISSUE #4164: Bug Out Bag trigger doesn't resolve fully
When triggered at end of turn Bug Out Bag produces a similar lag effect as Levy AR Lab Access as described in #4089.
When the lag clears the cards are drawn but the bag hasn't trashed, the end of turn button is still active and the game is locked.
The bag cannot be manually trashed and the turn cannot be ended.

I am using Chrome 73.0.3683.86 on windows 10.

### LINKED ISSUE #4165: Prompts for  Letheia Nisei are squashed
The button to trigger Letheia Nisei is squashed vertically, making it non obvious how to do this.

### LINKED ISSUE #4166: MCA Informant leads to game crash?
A maybe-a-bug report: 
In my last match, the last action before the game crashed (opponent runner had blanked cards, I tried to fix by leaving and re-joining and I could not join) was me installing an MCA informant on runner's Beth Kilrain-Chang.
We did not do anything else, so I suppose the error must have been caused by that MCA informant.

### LINKED ISSUE #4168: CTM is able to fire on the second trash after a Direct Access trash.
See title.

Thanks.

### LINKED ISSUE #4172: Flip Switch prompts during a trace in corps turn
Hi,

image

Flip Switch states that it cannot be used in corps turn, but it prompts the runner to use it, if the corp initiates a trace, for example with a Punitive Counterstrike.

None of the effects should be able to be triggered during corps turn

### LINKED ISSUE #4174: Architect Deployment Test Works Only On ICE
Architect Deployment Test only works if an ICE is revealed.  But it should work with non-ice cards as well.  This has happened to me twice, but I don't have a screenshot, sorry.

### LINKED ISSUE #4178: Flip Switch in the Log
Using Flip Switch to remove a tag is not logged in the console.

### LINKED ISSUE #4193: Daily Quests doesnt behave properly when hosted on Rec Studio
Daily quests fails to see successful runs against the server when hosted on full immersion rec studio.

### LINKED ISSUE #4196: Aggressive Secretary does not record cost to pay in the chat log
Not sure whether this is also the case for other traps such as Junebug ... corp 'uses' agg sec, but it would be more consistent to say 'pays 2c to use Agg Sec'

jamesgrey accesses Aggressive Secretary from Server 8.
campino uses Aggressive Secretary to trash Yusuf.
jamesgrey pays 0  to trash Aggressive Secretary from Server 8.
jamesgrey spends  and 1  to install Clone Chip.
jamesgrey spends 1  to install Yusuf.

NoahTheDuke:
Small one-offs that don't need to be split up.

Closes #4049 
Closes #4061 
Closes #4074 
Closes #4077 
Closes #4078 
Closes #4108 
Closes #4125 
Closes #4150 
Closes #4163 
Closes #4164 
Closes #4165 
Closes #4166 
Closes #4168 
Closes #4172 
Closes #4174 
Closes #4178
Closes #4193 
Closes #4196


src/clj/game/core/flags.clj
```diff
@@ -251,18 +251,37 @@
   [card]
   (or (not (rezzed? card)) (not (:cannot-host (card-def card)))))
 
+(defn agenda? [card]
```
Saintis: Great additions!

NoahTheDuke: Thank you! I want to go through and "convert" all of the existing type checks like this to the same, but that's much bigger than this batch of fixes.

NoahTheDuke:
Passes locally, CircleCI is currently shitting itself.

PR #4191, opened 2019-04-08
Add default ids to init-game

NoahTheDuke:
An alternative to #4181, this adds a default ID to each side. This only applies during tests and "single-player" games (aka games someone starts without an opponent). Will quiet down the logs, and will gracefully allow for expansion in the future if we decide to follow up on this with "real" solo play.


NoahTheDuke:
Whoops, the default ids have lower agenda point requirements! I did not know this, lol.

PR #4194, opened 2019-04-09
Fix Hacktivist Meeting log, fix mill and discard as cost, clean up cost logging

### LINKED ISSUE #4176: Hacktivist message on ice rez
The log shows a message about discarding a card from hand from Hacktivist when ice is rezzed with Hacktivist active.

NoahTheDuke:
Right now, our logging and prompt button creation use two different systems for handling costs, and even more, they're discordant with a lot of the `cost-handler` functionality as well. This attempts to unify them by defaulting to `cost-names`. (This has a nice side effect of changing cards like SDS Drone Deployment to now prompt and log `trash 1 installed program` instead of `trash 1:program`.)

To make this work with milling and trashing from hand as a cost, I've created `pay-mill` and `pay-discard` which async pay the cost and then generate the necessary string.

To make things more consistent, I changed all references to `:net-damage` etc to be `:net`, as we have no need for the specification and everything else uses the non-suffix version.

Closes #4176


src/clj/game/cards/events.clj
```diff
@@ -2023,8 +2022,7 @@
      nil
      nil
      (effect (register-events {:pre-rez-cost {:req (req (ice? target))
-                                              :msg (msg "add an additional cost of " (:cost target)
-                                                        " [Credits] to rez " (card-str state target))
+                                              :msg (msg "add an additional cost (currently unknown) to rez " (card-str state target))
```
Saintis: Why is this "currently unknown"?

NoahTheDuke: If the ice is facedown, the runner doesn't know how much it costs. X is only determined when the corp goes to rez the ice. The runner might make different decisions if they know the card will cost 2 to rez vs 10. The phrasing might could be changed, but we shouldn't be writing the actual cost of the facedown ice in the log either way.

Saintis: Maybe "to double the rez cost of X"?

NoahTheDuke: I like that!

PR #4254, opened 2019-05-27
Refactor cards to use card property utilities

NoahTheDuke:
This is a good step towards enforcing Separation of Concerns and very loosely/lightly Law of Demeter. and Loose Coupling.

Directly touching objects for their properties makes it hard for use to change how we structure our objects in the future, so this affords us a lot of flexibility.


Saintis:
Good stuff!

PR #4260, opened 2019-05-29
Generalize Showing Off

### LINKED ISSUE #4258: Bacterial Programming + Showing Off
It seems Showing Off is turning R&D upside down for the duration of accessing. If the Runner steals Bacterial Programming during this run, this could be very bad.

NoahTheDuke:
Instead of reversing the deck in-place, only reverse the deck when we're selecting cards to access! I'll whip up more tests tomorrow, but this should handle all outstanding bugs with Bacterial Programming, etc.

Closes #4258


lostgeek:
Closes #4258

Issue #362, opened 2015-07-13
Recurring Credit Cards are only partially implemented


justinliew:
Most if not all of the recurring credit cards are only partially implemented. I believe this is because there is no mechanism to filter those recurring credits to specific uses (ie. for using fracters, for rezzing ice).

justinliew:
I can try to take this on as my first task unless there are any objections. I think it would be a good starter task to get me acquainted with the code and how the card definitions work. I'll definitely ask questions as they come up.

nealterrell:
That is indeed the reason! This sounds like a challenging task but definitely doable, and will very much help you learn the code :).

justinliew:
My thought is to break the recurring credit usage into cases, and then deal with each case separately. From all the recurring credit cards, the following categories surfaced:

Runner:
Credits for installing (programs, hardware, virus only)
Credits for using cards (icebreakers, specific icebreaker types, hosted program)
Credits for using during run (anywhere, on HQ)
Credits for playing events (any, run only)
Credits for using traces
Credits for removing tags
Credits for trashing (anything, assets only)

Corp:
Credits for using traces
Credits for advancing cards (anything, ice, in specific server)
Credits for rezzing ice
Credits for installing (anything, in a specific server)

**Question**: Can anyone think of a situation where a player would NOT want to use their recurring credits first before regular credits? I feel like we could automate this by always using recurring credits for a specific category if they are available, and then deduct from regular pool after that. If we think players might want to micromanage which credits get used, then we'll likely need UI entries for every case. I could see the install case being challenging since it's a click to auto install/deduct, so there'd be new UI flow around that.

Thoughts?

nealterrell:
I can think of few times that recurring credits should not be used automatically:

1. Spending stealth credits for icebreaker abilities that don't require stealth credits. I would not necessarily want to use Cloak to pay for Refractor's "break 1 code gate subroutine" ability if I knew there was another code gate coming up that I needed to boost Refractor for.
2. I might not use recurring credits to rez cards during an Account Siphon run, as that might leave me with more credits for the Runner to siphon and double up on. (I would rather spend 3 of my 5 real credits to rez some ice rather than the 3 credits on The Root.)
3. I would certainly want to prioritize which card provides the credit I'm spending in many situations... prioritize Lockpick over Cloak for using Refractor, prioritize Because We Built It over The Root for advancing ice, Paricia over Whizzard for trashing assets, etc.

[...]

PR #4262, opened 2019-05-30
Implement Recurring and Placed Credits

lostgeek:
# Overview
Currently, paying costs is awaitable through `pay-sync`, but we delegate individual cost types to their own functions which may or may not individually be awaited, and right now paying credits is not. This PR introduces a new function that will check for any installed and active cards that have usable credits, and if so, will open a select prompt that's based on the `pick-virus-counters-to-spend` function (used for Khumalo and the Khumalo breakers) to pay or partially pay the appropriate cost.

### Costs
* Create new awaited cost handler for credits: `pay-credits`
* Wrote a bunch of helper functions to be used in that service
* Expand `can-pay?` to accept an `eid` to check installed cards for the appropriate credits
* Add `:pay-credits` to `:interactions` map, which connects the credits on the card (or a custom function) with the `pay-credits` function. It currently has 3 usable fields:
  * `:req`: Works as other reqs, checking to see whether the current action is the right kind to use this card's credits/ability
  * `:type`: Only 3 allowed values: `:credit`, `:recurring`, and `:custom`. `:credit` is for credits that are placed on the card (eg Miss Bones or Cold Read). `:recurring` is for recurring credits (eg Smoke or Dedicated Server). `:custom` is for cards that have payment-related abilities but aren't placed/recurring credits (eg Flame-out or Patchwork)
  * `:custom`: Only used when `:type` is `:custom`. Effectively an `:effect` map, it's an async 5-fn that should use `make-result` to return the amount of credits that will be applied to the effect.

### Engine
* Update `make-eid` calls to use existing `eid` for passing along `:source` and `:source-type` where appropriate
* Update `make-eid` calls at base levels (actions) to provide the necessary `:source-type`
* Make advancing cards payment awaited
* Make removing tag payment awaited
* Make runner install payment awaited
* Add `:play-area` to `all-active` for each side

### Runs
* Make paying for runs awaited
* Clean up `can-pay` logic so we know definitively if the runner can pay to trash a given card or not
* Make paying to trash an accessed card awaited, which cleans up the issue of repeatedly showing the prompt if you haven't taken enough credits from cards
* Fix bug where `handle-run-end` would dissoc `:access` even tho it was still happening

### Traces
* Refactor/clean up to be properly awaited
* Pull out traces as their own process in prompt-handling logic, so as to display the true number of available credits
* Pull out traces in the UI logic as well to handle the above
* Remove paying for traces in prompt-handling, use `pay-sync` within trace prompts where appropriate

### Cards
See below for full list of implemented cards

### Tests
See below for full list of implemented cards. Every single one got a small test for this functionality


src/clj/game/cards.clj
```diff
@@ -207,6 +207,42 @@
                                                (vals selected-cards)))]
                            (effect-completed state side (make-result eid {:number counter-count :msg msg})))))}))
 
+(defn pick-credit-providing-cards
+  "Similar to pick-virus-counters-to-spend. Works on :recurring and normal credits."
+  ([provider-func] (pick-credit-providing-cards (hash-map) 0 nil provider-func))
+  ([target-count provider-func] (pick-credit-providing-cards (hash-map) 0 target-count provider-func))
+  ([selected-cards counter-count target-count provider-func]
```
NoahTheDuke: Can you order these so you're building them one at a time left to right? So `[provider-func]` then `[provider-func target-count]` then `[provider-func target-count selected-cards counter-count]`? It's a small thing, but it helps readability to have a given argument in the same position across multiple arities.

NoahTheDuke: Looks like the virus picker does the same thing. Feel like fixing that up too? :-P

src/clj/game/cards.clj
```diff
@@ -207,6 +207,42 @@
...
+   {:async true
+    :prompt (str "Select a card with recurring credits ("
+                 counter-count (when (and target-count (pos? target-count))
+                                 (str " of " target-count))
+                 " credits)")
+    :choices {:req #(and (< 0 (get-counters % :recurring)) (.contains (provider-func) %))}
```
NoahTheDuke: Could use `pos?` here. And I feel like `.contains` is not quite what we want to use, but I'll have to play with it to see if there's a better method.

src/clj/game/cards.clj
```diff
@@ -207,6 +207,42 @@
...
+    :effect (req (add-counter state :runner target :recurring -1)
```
NoahTheDuke: You're hardcoding `:runner` here, but this should be side-neutral. No need to have two `pick-credit-providing-cards` functions.

src/clj/game/core/costs.clj
```diff
@@ -266,8 +266,22 @@
 
 (defn pay-credits
   [state side eid card amount]
-  (swap! state update-in [:stats side :spent :credit] (fnil + 0) amount)
-  (complete-with-result state side eid (deduct state side [:credit amount])))
+  (case (:source-type eid)
+    :runner-install
+    (letfn [(provider-func []
+              (filter #((or (-> % card-def :interactions :pay-credits) (req false)) state side eid % [card]) (all-active-installed state :runner)))]
```
NoahTheDuke: This feels inefficient. Maybe generate the list of `all-installed` once, and then `(filter #(when-let [pay-credits (-> % card-def :interactions :pay-credits)] (pay-credits state side eid % [card])) the-installed-cards)` The game-state shouldn't be changing when you're paying for something, so you only need to generate the list of available cards once, and then making the anonymous function use `when-let` means we won't even execute the full function for any card that lacks `:pay-credits`.

NoahTheDuke: Wait a minute, actually, you should only call this whole thing once anyways! Since you've moved the `(< 0 (:recurring card))` check into the prompt, you can remove it from the card def. If we need to check things other than `:recurring`, we could pass that in as an argument to the select prompt, but as it is right now, you're checking each card twice to see if it has enough `:recurring` credits.

src/clj/game/cards.clj
lostgeek: Yeah I've just mirrored what I've seen in the virus picker, I guess. I'll change the credit picker, but I'm still hoping to be able to merge the two pickers in the end...

src/clj/game/cards.clj
lostgeek: `pos?` seems to be a function to check for positive numbers.. Sure you mean that one?

I wanted to use `contains?` but it seems to only work for keys of maps.

src/clj/game/cards.clj
lostgeek: Good catch. This was only the first pass though. I want to refactor the whole function anyways and will fix this in that pass!

src/clj/game/cards.clj
lostgeek: All hope is lost. This will never merge 😂 

I'll be refactoring the virus picker.

src/clj/game/cards.clj
lostgeek: I have added `in-coll?` in `utils.clj` to do this.

src/clj/game/core/costs.clj
lostgeek: We need to check providers every time `pick-credit-providing-cards` is called, since cards like Net Mercur exist. Imagine you're using a credit off of Cloak. Now Net Mercur got a credit, but wasn't on the initial list of providers.
I was calling it more often though and have changed that.

lostgeek: Thinking more about it, the problem started with me comparing whole cards instead of `:cid`. Fixed that as well.

lostgeek:
Cards that benefit from this implementation
- Installing cards (runner)
  - [ ]  Cyberfeeder (virus)
  - [ ]  Cybsoft MacroDrive
  - [ ]  Ice Analyzer
  - [ ]  Inside Man
  - [X]  Sahasrara
- Stealth
  - [X]  Cloak
  - [ ]  Cold Read
  - [ ]  Dyson Fractal Generator
  - [ ]  Smoke
  - [ ]  Ghost Runner 
  - [ ]  Lockpick
  - [ ]  Net Mercur
  - [ ]  Silencer
- Paid abilities on icebreakers/programs
  - [ ]  Cyberfeeder
  - [ ]  Flame-out
  - [ ]  Multithreader
  - [ ]  Omni-drive
  - [ ]  Spinal Modem
  - [ ]  The Toolbox
- Credits during the run
  - [ ]  Maui
  - [ ]  Net Celebrity
  - [ ]  Pheromones
- Trashing cards
  - [ ]  Miss Bones
  - [ ]  Paricia
  - [ ]  Scrubber
  - [ ]  Whizzard
- Play events
  - [ ]  Prepaid VoicePAD
  - [ ]  Public Terminal
- Removing Tags
  - [ ]  Crash Space
- Rez cards
  - [ ]  Dedicated Server
  - [ ]  Khondi Plaza
  - [ ]  Mumba Temple
  - [ ]  The Root
- Installing cards (Corp)
  - [ ]  Dedicated Technician Team
  - [ ]  The Root
- Advance cards
  - [ ]  Simone Diego
  - [ ]  Weyland Because We Built It
- Traces
  - [ ]  Compromised Employee
  - [ ]  NBN:MN
  - [ ]  Net Police
  - [ ]  Primary Transmission Dish

NoahTheDuke:
I think Net Mercur needs a better `req`, but I’m loving this.

lostgeek:
**ToDo List**
- [ ] Make `play-auto-pump` awaitable and group together costs
- [ ] Build a `break-ice-full` mechanism to group together all breaking related costs
- [ ] Write tests

NoahTheDuke:
With removing tags, make sure you effect-complete payment and non-payment.

src/clj/game/core/actions.clj
```diff
@@ -574,10 +574,11 @@
   "Click to remove a tag."
   [state side args]
   (let [remove-cost (max 0 (- 2 (get-in @state [:runner :tag-remove-bonus] 0)))]
-    (when-let [cost-str (pay state side nil :click 1 :credit remove-cost)]
-      (lose-tags state :runner 1)
-      (system-msg state side (build-spend-msg cost-str "remove 1 tag" "removes 1 tag"))
-      (play-sfx state side "click-remove-tag"))))
+    (wait-for (pay-sync state side (make-eid state {:source :action :source-type :remove-tag}) nil :click 1 :credit remove-cost)
```
lostgeek: Since removing tags is now awaitable, I should test a case of Sahasrara + Thunder Art Gallery to check if all the prompts are handled correctly.

src/clj/game/core/costs.clj
NoahTheDuke: Net Mercur doesn’t proc until the whole cost is paid, so you can’t spend on cloak and then also spend off Net Mercur for the same cost.

src/clj/game/cards.clj
```diff
@@ -207,6 +207,69 @@
...
+                                            (or (add-counter state side target :credit -1) 1)
+
+                                            :custom ;custom functions should return the number of credits provided
+                                            (or ((-> target card-def :interactions :pay-credits :custom) state side eid card nil) 1))
+                           selected-cards (update selected-cards (:cid target)
+                                                  ;; Store card reference and number of counters picked
+                                                  ;; Overwrite card reference each time
+                                                  #(assoc % :card target :number (inc (:number % 0))))
+                           counter-count (+ counter-count gained-credits)]
+                       (if (has-subtype? target "Stealth")
```
NoahTheDuke: As I said in a different comment, this will have to be moved out of this function and into pay-rest, cuz paying costs is a simultaneous thing: you either pay everything at the same time or pay nothing. Thus, you can't spend a credit from Cloak and a credit from Net Mercur to boost Houdini if Net Mercur doesn't already have a credit on it.

src/clj/game/cards/hardware.clj
```diff
@@ -912,7 +931,11 @@
                                  (update! (assoc (get-card state card) :Omnidrive-prog (:cid target))))}]
     :events {:card-moved {:req (req (= (:cid target) (:Omnidrive-prog (get-card state card))))
                           :effect (effect (update! (dissoc card :Omnidrive-prog))
-                                          (use-mu (:memoryunits target)))}}}
+                                          (use-mu (:memoryunits target)))}}
+    :interactions {:pay-credits {:req (req (and (= :ability (:source-type eid))
+                                                (program? target)
+                                                (= (:title (:host target)) "Omni-drive")))
```
NoahTheDuke: This could be something like `(same-card? card (:host target))`, cuz `card` in this case will be the Omni-drive, right?

src/clj/game/cards/resources.clj
```diff
@@ -1021,7 +1023,12 @@
                  :effect (req (gain-credits state side 1)
                               (trigger-event state side :spent-stealth-credit card)
                               (when (zero? (get-counters (get-card state card) :credit))
-                                (trash state :runner card {:unpreventable true})))}]}
+                                (trash state :runner card {:unpreventable true})))}]
+    ; See Net Mercur for why this implementation was chosen
+    :interactions {:pay-credits {:req (req (and (:run @state)
+                                                (= :ability (:source-type eid))
+                                                (has-subtype? target "Icebreaker")))
```
NoahTheDuke: This needs to be just `(:run @state)`, cuz folks like me spend Ghost Runner credits for SMC installs and Hayley triggers.

src/clj/game/cards/resources.clj
```diff
@@ -1455,7 +1468,14 @@
                                                     (system-msg state :runner (str "places 1 [Credits] on Net Mercur"))
                                                     (add-counter state :runner card :credit 1)
                                                     (effect-completed state side eid))))}
-                                card nil))}}}
+                                card nil))}}
+    ; Normally this should be (req true), but having pay-credits prompts on
```
NoahTheDuke: I disagree strongly. It's an extra click, which is annoying, but just yesterday at my Regionals, I lost because the runner beat my Punitive Trace by spending all of their Net Mercur credits. (This isn't the best example because Trace prompts won't work with this system at the moment, lol, but you get my meaning.)

The card says "anything", so I think it should be for anything. Maybe let this be toggle-able somehow? Give Net Mercur an ability that changes whether you want it to ask or not?

src/clj/game/core/abilities.clj
```diff
@@ -106,16 +106,16 @@
   :makes-run -- indicates if the ability makes a run."
   ;; perhaps the most important function in the game logic
   ([state side {:keys [eid] :as ability} card targets]
-   (resolve-ability state side (or eid (make-eid state)) ability card targets))
+   (resolve-ability state side (or eid (make-eid state {:source card :source-type :ability})) ability card targets))
   ([state side eid ability card targets]
-   (resolve-ability-eid state side (assoc ability :eid eid) card targets)))
+   (resolve-ability-eid state side (assoc ability :eid eid :source card :source-type :ability) card targets)))
```
NoahTheDuke: This is being `assoc`'d to `ability`, not `eid`. Is that on purpose?

src/clj/game/core/events.clj
```diff
@@ -187,19 +187,19 @@
...
+     (wait-for (resolve-ability state side (make-eid state eid) first-ability nil nil)
                (show-wait-prompt state opponent
                                  (str (side-str active-player) " to resolve " (event-title event) " triggers")
                                  {:priority -1})
                ; let active player activate their events first
-               (wait-for (trigger-event-simult-player state side event active-player-events cancel-fn targets)
+               (wait-for (trigger-event-simult-player state side (make-eid state eid) event active-player-events cancel-fn targets)
                          (when after-active-player
-                           (resolve-ability state side after-active-player nil nil))
+                           (resolve-ability state side eid after-active-player nil nil))
```
NoahTheDuke: Is this going to `effect-complete` the eid?

src/clj/game/cards/resources.clj
lostgeek: Oh you're right. While Net Mercur is triggering all over the place (and you want to use it for something other than pumping breakers like 1 in 5 times), Ghost Runner is only during the run and it should be fine having it open the prompt for trashing / traces / SMC as well in those cases.

src/clj/game/cards.clj
lostgeek: That case is working right now though. When you use Houdini, `pick-credit-providing-cards` gets called. Now at this point here (line 264), we're waiting for the `:spent-stealth-credit` trigger, which causes the Net Mercur prompt to happen. You can now place a credit on Net Mercur and only afterwards go back to `pick-credit-providing-cards`, which now sees the Net Mercur credit. 

The downside of my implementation though is that paying for stuff is not simultaneous anymore. It is happening one credit at a time (until you press "Done" and `pay-rest` gets called).

src/clj/game/cards/hardware.clj
lostgeek: I think so. I will test this.

src/clj/game/cards/resources.clj
lostgeek: For now you can simply revert to the old system of clicking on Net Mercur to get credits. That ability hasn't changed.

But I like the idea of having it toggle-able!

src/clj/game/core/abilities.clj
lostgeek: This was on purpose. If I'm reading `resolve-ability-eid` correctly, it expects the `eid` to be wrapped inside the `ability`. I found this weird too, but didn't want to change it, since I don't know what might have been the motivation behind this.

src/clj/game/core/events.clj
lostgeek: Since we're using the same `eid`, this should be `effect-complete`'d later in line 204.

src/clj/game/cards.clj
```diff
@@ -207,6 +207,69 @@
...
+
+                               :credit
+                               (pos? (get-counters % :credit))
+
+                               :custom
+                               true))}
+        :effect (req (let [pay-credits-type (-> target card-def :interactions :pay-credits :type)
+                           gained-credits (case pay-credits-type
+                                            :recurring
+                                            (or (add-prop state side target :rec-counter -1) 1)
```
NoahTheDuke: Is the goal to return a `1`? If so, I'd recommend using `do`, cuz `or` makes it look like you expect `add-prop` to fail or return incorrectly.

src/clj/game/cards.clj
NoahTheDuke: I just loaded it up, and it works incorrectly. I initiate a run on R&D, I have Net Mercur, Cloak, and Black Orchestra installed. I click Blorch, it gives me the credit selection prompt, so I click Cloak, and Net Mercur's prompt appears. I click "Gain 1 credit", and a credit appears on Net Mercur. I click on Net Mercur, and the credit comes off, and the rest of the cost is paid automatically.

> noah pays 1 from Cloak, 1 from Net Mercur and 1 to use Black Orchestra to add 2 strength and break up to 2 subroutines.
image

src/clj/game/cards.clj
lostgeek: Thanks! That was exactly what I wanted to do.

src/clj/game/cards.clj
lostgeek: Oh, you're right of course. I was in too deep implementing stuff, that I forgot how the rules actually work :joy:

NoahTheDuke:
I think we should wait to implement this for Traces. Traces create the dropdown before you spend it, so you won't be able to spend more credits than you currently have in your pool even with available recurring/placed credits.

This also makes me think: what happens when you don't have enough credits in your pool for playing/installing cards normally but you have recurring/placed credits to make up the difference? Does `can-pay?` allow for that? If not, we'll have to adapt it to handle that first.

src/clj/game/core/turns.clj
```diff
@@ -150,9 +150,8 @@
   ([state] (make-eid state nil))
   ([state {:keys [source source-type]}]
    (merge {:eid (:eid (swap! state update-in [:eid] inc))}
-          (when source
-            {:source source
-             :source-type source-type}))))
+          (when source {:source source})
+          (when source-type {:source-type source-type}))))
```
NoahTheDuke: 👏great👏change👏

lostgeek: Yeah. I stumbled on that (and in the end didn't even need it) 😄

lostgeek:
I think we're done here! :confetti_ball:

NoahTheDuke:
Thinking about Net Mercur more, I think we could get 80% by requiring `:icebreaker`, `:run`, or `:trace`. By my count, that will cover nearly all uses. If someone wants to use it to play or install, they can manually take credits off. Everything else I can think of falls under those uses.

What do you think?

NoahTheDuke:
Closes #362

PR #4265, opened 2019-06-07
Implement Subroutine Tracking

### LINKED ISSUE #3638: Sub tracking and quick run feature proposal
The motivation for this is to allow for runs to be made with less Corp interaction, and icebreakers to have auto-break convenience methods. Below, I'll go more in depth on what I want, and give a mockup of the UI I'm imagining.

Currently, the Corp needs to say "yes, I accept that you have broken all subs" for each piece of ice, and
if not, manually select the ones to fire. Ideally, the Corp would only be needed to rez/trigger paid abilities, so a run on a server where all the ice is rezzed should be resolvable entirely by the Runner, with no user interaction needed from the Corp except to rez non-ice cards or trigger paid abilities if necessary.

This breaks down into three sub-goals:

1. Have the game track the number and types of sub on pieces of ice
2. Give the Runner an "Fire unbroken subs" button and corresponding dialog
3. Give the Corp a "Always 'No further action' on rezzed ice" button

## Necessary modifications to game engine
1. The game needs a method get-subroutines taking the state and a piece of ice as argument and returning a list of subs. Once existing ice have their sub lists updated, this will generally be a simple lookup at the :subroutines key, although some ice (e.g. NEXT Silver) needs to be intelligent about this.

2. The Corp should get a togglable ability "Always 'No further action' on rezzed ice" which would effectively instantly press 'No further action' if the ice the run arrow is pointing to is rezzed. It would assume the corp takes other rez/paid abilities when appropriate, which should be fine unless the Runner plays too fast. 

3. The Runner should get a button 'Fire unbroken subs' as an alternative to 'Continue'. This would 1) check how many subs were broken and how many exist, spawn a 'which subs were broken' prompt if needed, and then resolve all unbroken subs without needing the Corp to take any action unless the sub requires it (e.g. Miraju). To see what this prompt might look like, see "Unbroken subs prompt UI". 

4. Icebreakers should get 'break all' convenience methods added. These would trigger the icebreaker's break ability as many times as there are subs on the piece of ice. Some breakers (Endless Hunger) can't break all, of course.

## Unbroken subs prompt UI example
Let's consider a runner who is encountering a rezzed Archer. The Archer was rezzed before the run, and the Corp has enabled auto-pass, so the Runner does not need to wait for the Corp to press 'no further action' before they break/pass Archer. 

The Runner wants to break all subs except the first, but has mistakenly only pressed 'break ice subroutine' twice before pressing "allow unbroken subs to fire". The game defaults to assuming the Runner wants to break the subs in order, but allows the Runner to correct this, and even to mark too many/few subs for firing.

If the runner presses "Done marking broken/unbroken subs" without having marked what the game thinks is the correct number of subs, the game still only fires the ones the Runner says, but grumbles about it to console.

**Prompt once the Runner clicks "Allow unbroken subs to fire"**

_Archer has 4 subs (you broke 2)_
1. _(Broken) The Corp gains 2 credits._
2. _(Broken) Trash a program._
3. _(Unbroken) Trash a program._
4. _(Unbroken) End the run._
5. _Done marking broken/unbroken subs_

**Prompt after button 1 is pressed:** 

_Archer has 4 subs (you broke 2, but have selected 1)_
1. _(Unbroken) The Corp gains 2 credits._
2. _(Broken) Trash a program._
3. _(Unbroken) Trash a program._
4. _(Unbroken) End the run._
5. _Done marking broken/unbroken subs_

**Prompt after button 3 is pressed:** 

_Archer has 4 subs (you broke 2)_
1. _(Unbroken) The Corp gains 2 credits._
2. _(Broken) Trash a program._
3. _(Broken) Trash a program._
4. _(Unbroken) End the run._
5. _Done marking broken/unbroken subs_

**Prompt after button 4 is pressed:** 

_Archer has 4 subs (you broke 2, but have selected 3)_
1. _(Unbroken) The Corp gains 2 credits._
2. _(Broken) Trash a program._
3. _(Broken) Trash a program._
4. _(Broken) End the run._
5. _Done marking broken/unbroken subs_

Once button 5 is pressed, the game would print "The 'Gain 2 credits' subroutine on Archer resolves.", resolve Archer's first sub, then print "Note: the Runner only broke 2 subs, but marked 3 as broken." to console.  

# Clarifications
## Backwards compatibility
If the runner presses 'Continue', the break/fire stepis assumed to have been handled by the players, so they could play the run 100% old-style if they prefer. Subroutines will still be manually firable by the Corp to easily correct mistakes.
## "Fire all" fails gently
If the Runner clicks "Fire all", a list of subs is presented to them with a "Is this correct?" dialog before actually firing anything, which will always allow the Runner to go "whoops" if needed. 
## Encounter effects are still manual
Not 100% sure how this works out with ice like grail/Komainu. 
## Special cases:
Some effects break a variable number of subs (switchblade, black orchestra, ++), so the game cannot always infer how many subs the player meant to break. For simplicity, I suggest having  the game assume the Runner always breaks as many as possible and rely on the Runner to press 'Allow unbroken subs to fire' if they want to. 

Gbhali/Kongamato/Endless Hunger all break special types of subs. I suggest ignoring these and having the game treat them as if their effects said "Break ice subroutine".

NoahTheDuke:
Hot on the heels of @lostgeek's success with #4262, and nearly one full year after @presheaf wrote up #3638, I feel emboldened to implement another facet of the game we've so far left players to handle themselves!

# SUBROUTINE TRACKING

Goals:
- [x] Implement every card that manipulates subroutines
- [x] Correctly number all printed subroutines
- [x] Build UI for choosing which subroutines to break
- [x] Build system to actually “break” subroutines (tracking, logging, paying, triggering, etc)
  - [x] Logging
  - [x] Paying
  - [x] Tracking
  - [ ] ~Triggering events~
- [x] Build UI for marking/disabling broken subroutines
- [x] Create button that fires all unbroken subroutines in sync
- [x] Build UI for Icebreakers and cards that break subroutines (d4v1d, etc)
- [x] Rewrite all card abilities that currently break subroutines
- [x] Expand engine to limit icebreaker “interactions” to rules
- [ ] ~Implement all cards that listen with subroutine-related events~

Triggering events and implementing non-breaking cards that interact with subroutines will come in another PR. This is purely for numbering and "breaking" them in as dumb and obvious a way as possible. It's already massive, so I want to limit the amount of change happening.

Closes #3638


src/clj/game/cards/ice.clj
```diff
@@ -360,9 +370,11 @@
                    :msg (msg (corp-install-msg target))}]}
 
    "Ashigaru"
-   {:abilities [{:label "Gain subroutines"
-                 :msg (msg "gain " (count (:hand corp)) " subroutines")}]
-    :subroutines [end-the-run]}
+   {:events {:card-moved {:req (req (or (= :hand (first (:zone target)))
+                                        (= :hand (first (:previous-zone target)))))
```
lostgeek: Doesn't `:hand` also apply to the Grip?

NoahTheDuke: Whoops, yes it does! Lmao thanks

src/clj/game/cards/agendas.clj
```diff
@@ -1025,8 +1022,24 @@
...
+                 :prompt "Choose a piece of ICE"
+                 :choices {:req #(and (ice? %)
+                                      (rezzed? %))}
+                 :counter-cost [:agenda 1]
+                 :msg (str "make a piece of ICE gain \"[Subroutine] Do 1 net damage\" "
+                           "after all its other subroutines for the remainder of the run")
+                 :effect  (effect (add-extra-sub! (get-card state target)
+                                                  (do-net-damage 1)
+                                                  (:cid card) {:back true})
+                                  (update! (update-in card [:special :kusanagi] #(conj % target))))}]}
```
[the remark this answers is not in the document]
NoahTheDuke: Ice can get bounced or trashed too, lol. There's really no good way to handle effects like this, right now. Once again, we come across a situation where we need floating effects unrelated to cards. :rip:

test/clj/game_test/cards/programs.clj
```diff
@@ -511,6 +511,24 @@
         (is (zero? (get-counters (refresh c) :virus)) "Consume loses counters")
         (is (zero? (get-counters (refresh h) :virus)) "Hivemind loses counters")))))
 
+(deftest ^:test-refresh/focus corroder
```
lostgeek: :eyes: 
Btw. I've got these mappings to deal with that quickly (`<leader>no` for adding and `<leader>ni` for removing the test-refresh thingy). Maybe you'll find it useful as well!
```
map <leader>no mZ?deftest<cr>wi^:test-refresh/focus <esc>'ZmZ
map <leader>ni mZ?deftest<cr>wdf <esc>'ZmZ
```

PR #4273, opened 2019-06-12
Clean up prevention functions

NoahTheDuke:
The prevention functions (specifically the prompts and resolve functions) are inconsistent and kind of messy, so I cleaned them up a bit. I'm still not happy with how this turned out, especially how `expose` works, but like so many lil refactors, it's a good first step towards greater clarity and understanding.


NoahTheDuke:
On second thought, this is too much for a single PR. I'm gonna split it off into individual PRs for every single prevention type.

PR #4290, opened 2019-06-26
Bugfixes around pay-credits prompts

### LINKED ISSUE #4295: Auto-pumping Icebreaker with pay-credits prompt leads to wrong strength
Situation:
* Smoke with Corroder and Net Mercur encounters an IP Block
* Auto-pump prompt opened
* Used Smoke and paid one credit out of credit pool.
  * Credits are correctly paid
* Corroder strength is at 3 (should be 4)

lostgeek:
This PR will contain the bugfixes for the new pay-credits prompts.

### Changes
* Feature: Credit costs are printed explicitly in most cases now, even if zero. 
  * Exception: Installing cards in a remote / root of a central. 
* Fix: Auto-pumping breakers now merges total costs. Closes #4295.
* Fix: All runner cards now add `:source-type` to the `runner-install` eid.
  * This has to be done for every action that is done by card abilities and induces a `pay-credits` prompt


NoahTheDuke:
This looks sweet! I love the changes you made. One note about the tests: we have a `last-log-contains?` that does p much what your comparisons do. Might help make your tests more readable!

NoahTheDuke:
I don't wanna make you undo all of the work you just did, but could we just default to the correct `:source-type` in `runner-install`? That way we can leave it unspecified and it'll choose correctly, and override where we need to.

lostgeek:
Oh you're right. That was a case of having too much fun fiddling with vim macros and too little thinking :joy: 

That being said, we should probably still keep the changes because we're specifying `:source card` in there.

NoahTheDuke:
Maybe just remove the `:source-type` part then? Seems like a lot of repetition in the code.

lostgeek:
Yeah I'll certainly do that and look at all the other cases again (corp-install and so on). This should be the default behavior of all the `pay-credits` relevant actions.

PR #4298, opened 2019-07-02
Fix the scroll bug?

### LINKED ISSUE #4289: scroll in console not working after latest update
Looks like scroll in console have stopped working after latest update
The text stays where u scroll to, and doesnt follow the new events

image

NoahTheDuke:
Closes #4289. Instead of checking equality, check a range of 5 less to 5 more, to be absolutely sure.


aenimosity:
Had this issue intermittently for the last week or so. I think it's a problem with the zoom level of the browser. 100% zoom has no issues. Will post in the other thread as I have no idea how the github system works.

PR #4305, opened 2019-07-03
Miss Bones Fix for trashing on empty

### LINKED ISSUE #4303: MIss Bones does not trash when empty
Nuff said

NoahTheDuke:
This is the start of a larger refactor, as other cards should operate under this functionality as well.

Closes #4303

### DIFF
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -38,17 +38,17 @@
   "Creates a Campaign with X counters draining Y per-turn.
   Trashes itself when out of counters"
   [counters per-turn]
-  (let [ability {:msg (str "gain " per-turn " [Credits]")
-                 :counter-cost [:credit per-turn]
+  (let [num-credits (fn [card] (min per-turn (get-counters card :credit)))
+        ability {:msg (msg "gain " (num-credits card) " [Credits]")
                  :once :per-turn
                  :req (req (:corp-phase-12 @state))
                  :label (str "Gain " per-turn " [Credits] (start of turn)")
-                 :effect (req (gain-credits state :corp per-turn)
-                              (when (zero? (get-counters card :credit))
-                                (trash state :corp card)))}]
+                 :effect (effect (gain-credits (num-credits card))
+                                 (add-counter card :credit (- (num-credits card))))}]
     {:effect (effect (add-counter card :credit counters))
      :derezzed-events {:runner-turn-ends corp-rez-toast}
-     :events {:corp-turn-begins ability}
+     :events (merge (trash-on-empty :credit)
+                    {:corp-turn-begins ability})
      :abilities [ability]}))
 
 (defn as-trashed-agenda
@@ -1523,12 +1523,14 @@
    {:effect (effect (add-counter card :power 3))
     :derezzed-events {:runner-turn-ends corp-rez-toast}
     :events {:corp-turn-begins
-             {:async true
-              :effect (req (add-counter state side card :power -1)
-                           (if (zero? (get-counters (get-card state card) :power))
-                             (do (system-msg state :corp "uses Public Support to add it to their score area as an agenda worth 1 agenda point")
-                                 (as-agenda state :corp eid (dissoc card :counter) 1))
-                             (effect-completed state side eid)))}}}
+             {:req (req (pos? (get-counters card :power)))
+              :effect (effect (add-counter card :power -1))}
+             :counter-added
+             {:req (req (same-card? card target)
+                        (not (pos? (get-counters card :power))))
+              :async true
+              :effect (effect (system-msg "uses Public Support to add it to their score area as an agenda worth 1 agenda point")
+                              (as-agenda eid (dissoc card :counter) 1))}}}
 
    "Quarantine System"
    (letfn [(rez-ice [cnt] {:prompt "Select an ICE to rez"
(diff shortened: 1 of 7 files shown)


src/clj/game/cards/assets.clj
```diff
@@ -38,16 +38,15 @@
...
   [counters per-turn]
   (let [ability {:msg (str "gain " per-turn " [Credits]")
-                 :counter-cost [:credit per-turn]
                  :once :per-turn
                  :req (req (:corp-phase-12 @state))
                  :label (str "Gain " per-turn " [Credits] (start of turn)")
-                 :effect (req (gain-credits state :corp per-turn)
-                              (when (zero? (get-counters card :credit))
-                                (trash state :corp card)))}]
+                 :effect (effect (gain-credits (min per-turn (get-counters card :credit)))
```
lostgeek: Future-proofing. I love it :+1:

src/clj/game/cards/assets.clj
```diff
@@ -585,7 +584,6 @@
    "Drudge Work"
    {:effect (effect (add-counter card :power 3))
     :abilities [{:cost [:click 1]
-                 :counter-cost [:power 1]
```
lostgeek: I think, this can stay. Imho, the ruling would be that Drudge Work is already in the Archives, when it is doing the reshuffling of the last agenda..

You first pay the costs of the paid ability. Then there is a checkpoint, in which Drudge Work gets trashed because there are no power counters left. And then you're doing the ability, which exists even if the card got trashed.

Maybe we should ask the rules team on this.

src/clj/game/cards/assets.clj
```diff
@@ -1477,11 +1475,11 @@
    "Private Contracts"
    {:effect (effect (add-counter card :credit 14))
     :abilities [{:cost [:click 1]
-                 :counter-cost [:credit 2]
```
lostgeek: (re: the comment on Drudge Work) This one should be ok.

src/clj/game/cards/hardware.clj
```diff
@@ -1048,11 +1048,10 @@
    {:data [:counter {:power 4}]
     :interactions {:prevent [{:type #{:meat}
                               :req (req true)}]}
-    :abilities [{:counter-cost [:power 1]
```
lostgeek: See Drudge Work.

src/clj/game/cards/resources.clj
```diff
@@ -614,7 +606,13 @@
       :abilities [ability]
       :move-zone heap-event
       :events {:runner-turn-begins ability
-               :runner-turn-ends nil}})
+               :runner-turn-ends nil
+               :counter-added {:req (req (not (pos? (get-counters card :credit))))
```
lostgeek: Doesn't this need a `(req (same-card? card target)`?

src/clj/game/cards/resources.clj
```diff
@@ -1186,10 +1184,8 @@
     :interactions {:prevent [{:type #{:meat}
                               :req (req true)}]}
     :abilities [{:label "Prevent 1 meat damage"
-                 :counter-cost [:power 1]
```
lostgeek: See Drudge Work.

src/clj/game/cards/resources.clj
```diff
@@ -1308,11 +1304,10 @@
    "Liberated Account"
    {:data {:counter {:credit 16}}
     :abilities [{:cost [:click 1]
-                 :counter-cost [:credit 4]
```
lostgeek: This is fine.

src/clj/game/core/cards.clj
```diff
@@ -227,7 +227,7 @@
            (if-not placed
              (trigger-event state side :advance (get-card state updated-card))
              (trigger-event state side :advancement-placed (get-card state updated-card))))
-       (trigger-event state side :counter-added (get-card state updated-card))))))
+       (trigger-event-sync state side (make-eid state) :counter-added (get-card state updated-card))))))
```
lostgeek: Maybe as future-proofing add `(make-eid state {:source card :source-type :add-prop})`

src/clj/game/cards/resources.clj
```diff
@@ -1391,13 +1391,11 @@
     :interactions {:pay-credits {:req (req (and (= :runner-trash-corp-cards (:source-type eid))
                                                 (installed? target)))
                                  :type :credit}}
-    :abilities [{:counter-cost [:credit 1]
```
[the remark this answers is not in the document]
NoahTheDuke: Yeah, I might have been too excited here. Fully switching over should probably happen another time.

src/clj/game/cards/assets.clj
NoahTheDuke: Yeah, I’ve heard whispers of an Uprising card that places credits on cards, and technically these abilities aren’t costs, they’re effects, so you always do as much as possible.

src/clj/game/cards/assets.clj
NoahTheDuke: No, you’re totally right! I got over-zealous and forgot what the heck I was doing haha. This is the perfect place for `:counter-cost`.

src/clj/game/cards/resources.clj
NoahTheDuke: Yep! Good catch.

src/clj/game/core/cards.clj
NoahTheDuke: Smart.

PR #4309, opened 2019-07-05
Add better damage prevention messages

### LINKED ISSUE #4307: Heartbeat logs every trashed card as facedown
From Andrejs stream: https://youtu.be/2VBr43WCrD4
20:26

### LINKED ISSUE #4308: Prevent damage should show the amount that was already prevented
Preventing multiple damage is not communicated well to the player. We should adjust the prompt to show the amount that was already prevented.

lostgeek:
### Changelog
- Bugfix: Heartbeat was showing every card as facedown. Closes #4307.
- Feature: Damage prevention now shows how much damage was prevented already. Closes #4308.


lostgeek:
Here's the new prevention in action (Heartbeat preventing one net damage):

Prevention

lostgeek:
A lot of broken tests, because the last commit removed the `[Done]` button on a lot of damage prevention prompts.

NoahTheDuke:
Why change the damage prevention function to private? Seems like it doesn’t make things any clearer

lostgeek:
I thought it might make it a bit cleaner, if I don't dump all that code into the `damage-prevent` function.

NoahTheDuke:
Oh I see, I only looked at the commit diff, forgot to check the full diff. Yeah, you made the right call.

PR #4313, opened 2019-07-10
Added (local) timestamp to game board

### LINKED ISSUE #2227: Add timestamp when a game start
to check how much a game is lasting it would be nice to have a timestamp when the game start (GMT 0) can be a standard ;)

bkentropy:
The issue [2227](https://github.com/mtgred/netrunner/issues/2227) mentioned using GMT 0, but I opted for just providing a local time. I'd be glad to hear feedback on placement, GMT vs local, etc. 
Screen Shot 2019-07-09 at 10 13 18 PM

Closes #2227

### DIFF
--- a/src/cljs/nr/gameboard.cljs
+++ b/src/cljs/nr/gameboard.cljs
@@ -1483,6 +1483,11 @@
                   (pos? (:click @me)))
              #(send-command "credit")]]))])})))
 
+(defn starting-timestamp []
+  [:div
+   {:class ["panel" "blue-shade"]}
+   (str "Game start: " (.toLocaleTimeString (js/Date.)))])
+
 (defn gameboard []
   (let [run (r/cursor game-state [:run])
         side (r/cursor game-state [:side])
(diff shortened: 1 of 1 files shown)


NoahTheDuke:
Looks like you’ve put it below the other panels so it’ll get pushed down when you have cards in the play area or remove cards from the game. Try putting it above those! Another idea is to print it in the log when the game begins, to keep it from being too obtrusive. How often do you need to see when you started a game during a game?

bkentropy:
Thanks for being so responsive! I love the game, but haven't played through Jinteki much. I don't have a good feel where things (removed from game cards) end up. 
I like the log idea, you mean the chat log right? I think the user request in [2227](https://github.com/mtgred/netrunner/issues/2227) would find scrolling to the top of the chat pretty reasonable.

NoahTheDuke:
You're so welcome! It's great to see new folks contributing.

So for my first suggestion, where you've added the `starting-timestamp` vector, you've placed it below the `rfg-view`, the `play-area-view`, and the `rfg-view` (for Currents). This means that if you have cards in the "Removed from the game" zone, the div you've created will move below it. Additionally, if you play a run event like Cold Read, it'll be placed before that as well.

If you moved it from line 1595 to 1589 (as the first entry in the `[:div`), it'll keep the timestamp at the top of that pane, regardless of whatever else is placed in there.

The log idea will have to be done in the Clojure side. Look in `src/clj/game/core/turns.clj`, in the `init-game` function. After the `let` bindings, before the first `init-identity` call, you'll want to add a `system-say` call, which prints the local time. I'm not sure how that's done, so you'll have to explore that for yourself, but it shouldn't be too hard!

Good luck and thanks for the help!

PR #4319, opened 2019-07-15
Fencer Fueno now integrated in pay-credits prompt

### LINKED ISSUE #4318: Fencer Fueno - No prompts to use credits
Hey guys,

I'm playing Fencer Fueno in Jesminder with Thunder Art Gallery and there are a number of situations where I should be able to take Fencer Fueno credits but other prompts override my ability to click on Fencer Fueno.

When using Thunder Art Gallery +  DJ Fenris (Liza) or Hot Pursuit (Install prompt overrides my ability to use Fencer)
When accessing cards against Gagarin Deep Space (Access prompt overrides my ability to use fencer)

Also if I access a card I don't have credits to trash (but do have enough credits on Fencer) it won't give me the prompt to trash the card, only No Action even if I can boost my credits after. (which I can't do before because the run isn't successful yet)

I would personally like to see an implementation similar to Net Mercur because when that's installed it asks if I'd like to use any of those credits (although there's the current issue with that and using icebreakers, but obviously fencer fueno is a much more specific window so it's less)

I've tried to be as clear as possible but if you have any questions please don't hesitate.

Thanks a bunch in advance.

lostgeek:
### Changelog
- Integrated Fencer Fueno into `pay-credits` prompt. Closes #4318

### DIFF
--- a/src/clj/game/cards/resources.clj
+++ b/src/clj/game/cards/resources.clj
@@ -920,28 +920,32 @@
                              :effect (req (as-agenda state :runner eid card 0))}}}
 
    "Fencer Fueno"
-   (companion-builder
-     (req (and (pos? (get-counters (get-card state card) :credit))
-               (:successful run)))
-     (effect (show-wait-prompt :corp "Runner to take decision on Fencer Fueno")
-             (continue-ability
-               {:prompt "Pay 1 [Credits] or trash Fencer Fueno?"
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
Haha I think it’s time to toss the companion builder and long form them all

lostgeek:
I don't know. Implementing M`redacted` worked quite well. I think we just need another argument for the `:pay-credits` routine.

PR #4326, opened 2019-07-20
Stacking of similar servers

lostgeek:
image

### Goal
This change will allow a more compact display of asset servers. When there are multiple servers containing one copy of one asset each, these will be displayed stacked as if they were hosted on each other, with a label referring to the correct server numbers.

### Changelog
- Added option to stack similar servers. This is an opt-out feature, that you can disable in the settings.

### DIFF
--- a/src/cljs/nr/gameboard.cljs
+++ b/src/cljs/nr/gameboard.cljs
@@ -133,6 +133,9 @@
   (ws/ws-send! [:netrunner/mute-spectators {:gameid-str (:gameid @game-state)
                                             :mute-state mute-state}]))
 
+(defn stack-servers [ss-state]
+  (swap! app-state assoc-in [:options :stacked-servers] ss-state))
+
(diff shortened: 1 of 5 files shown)


src/cljs/nr/gameboard.cljs
```diff
@@ -748,9 +748,11 @@
   (.stopPropagation event))
 
 (defn label [cursor opts]
-  (let [fn (or (get-in opts [:opts :fn]) count)]
-    [:div.header {:class (when (> (count cursor) 0) "darkbg")}
-     (str (get-in opts [:opts :name]) " (" (fn cursor) ")")]))
+  (let [fn (or (get-in opts [:opts :fn]) count)
+        classes (conj (when (> (count cursor) 0) '("darkbg")) (get-in opts [:opts :classes]))]
```
NoahTheDuke: This could use `pos?`

PR #4330, opened 2019-07-25
Major refactor of costs, fix cost strings to only print "Pay" when necessary

### LINKED ISSUE #4316: build-cost-str strings are broken
image

NoahTheDuke:
Closes #4316


src/clj/game/core/costs.clj
```diff
@@ -145,38 +149,151 @@
...
+      (str (build-cost-label cost) ": " label)
+      label)))
+
+(let [cost-types #{:trash
+                   :hardware :program :resource :connection :ice
+                   :net :meat :brain
+                   :trash-from-deck :trash-from-hand :randomly-trash-from-hand
+                   :agenda :power :virus :advancement
+                   :forfeit :shuffle-installed-to-stack}]
+  (defn cost->string
```
lostgeek: Maybe swap `defn` and `let`?

NoahTheDuke: I wrote it like this because I don't want to recreate the `cost-types` set every call to `cost->string`, as it's static, but I also don't want to `def` it into a namespace-global variable, because that's messy. Doing the `let` binding around the `defn` means that it'll only be created once, and only this function will have access to it.

NoahTheDuke: Decided to reverse this logic, so it's not even needed anymore!

PR #4335, opened 2019-07-30
Small fixes

lostgeek:
### Changelog
- Setting default values for `:source-type` on several actions
- Bugfix: Cards like Executive Boot Camp now work with recurring credits for rezzing
- Bugfix: CtM no longer shows hanging wait prompt on autoresolve


NoahTheDuke:
Looks like we need a “carry-forward-eid” function for all those let blocks!

PR #4368, opened 2019-08-10
Constellation Protocol: use :msg

lostgeek:
Small little change for a rarely used card. 10/10 would fix again.


NoahTheDuke:
I can't believe this card has been so fundamentally broken since last year.

PR #4371, opened 2019-08-11
runner-install: check for misaligned arguments

lostgeek:
This should fix the problems discussed in slack.


NoahTheDuke:
I’d prefer to not do this, because there are a lot of places where we could perform checks like this and we don’t, so making this a special case doesn’t sit well.

PR #4434, opened 2019-08-19
Implement Constant Abilities, Floating Events

### LINKED ISSUE #3341: By Any Means turns off after playing Levy AR Lab Access
Has happened twice, 

Click 1 By Any Means, https://netrunnerdb.com/en/card/21001 run normally or with Multiaccess and it triggers properly.  Hit levy, https://netrunnerdb.com/en/card/03035 Run again, Stops triggering.  This has caused me to steal agendas that were supposed to be trashed each time.

### LINKED ISSUE #3377: By any means not working with counter surveilance
image
Log of game: Below show after playing by any means, the cards accessed by counter surveilance did not work. 
vincentkbl takes 1 tag to place 2 virus counters on God of War.
vincentkbl started their turn 21 with 28  and 4 cards in their Grip.
vincentkbl spends  and 2  to play By Any Means.
vincentkbl spends  and 1  to install Counter Surveillance.
titonosfe uses Housekeeping to force the Runner to trash I've Had Worse from their Grip.
vincentkbl uses Trope to shuffle Exclusive Party, By Any Means, Exclusive Party, I've Had Worse into their Stack.
vincentkbl spends  to use Counter Surveillance to run R&D and trashes Counter Surveillance.
titonosfe spends 4  to rez Hortum.
vincentkbl spends 3  to install Black Orchestra.
vincentkbl spends 3  to use Black Orchestra to add 2 strength and break up to 2 subroutines.
titonosfe has no further action.
vincentkbl continues the run.
vincentkbl indicates to fire all subroutines on Colossus.
titonosfe uses Colossus to give the Runner 1 tag.
titonosfe uses Colossus to trash 1 program.
titonosfe uses Colossus to trash Black Orchestra.
titonosfe uses Skorpios Defense Systems: Persuasive Power to remove Black Orchestra from the game.
titonosfe has no further action.
vincentkbl continues the run.
titonosfe spends 3  to rez Cyberdex Virus Suite.
titonosfe uses Cyberdex Virus Suite to purge virus counters.
vincentkbl uses Acacia and gains 14.
vincentkbl indicates to fire all subroutines on Mausolus.
titonosfe uses Mausolus to gain 1 .
titonosfe uses Mausolus to do 1 net damage.
vincentkbl trashes Mars for Martians due to net damage.
titonosfe uses Mausolus to give the Runner 1 tag.
titonosfe has no further action.
vincentkbl continues the run.
titonosfe has no further action.
vincentkbl uses Knobkierie to place 1 virus counter on God of War.
vincentkbl uses Counter Surveillance to access up to 16 cards by paying 16 .
vincentkbl accesses an unseen card from R&D.
vincentkbl accesses an unseen card from R&D.
vincentkbl
vincentkbl
huh?
vincentkbl
vincentkbl
i tot trashing
vincentkbl accesses an unseen card from R&D.
vincentkbl accesses an unseen card from R&D.
vincentkbl accesses an unseen card from R&D.
!vincentkbl uses a command: /close-prompt
vincentkbl accesses an unseen card from R&D.
vincentkbl accesses an unseen card from R&D.
vincentkbl accesses an unseen card from R&D.
vincentkbl accesses an unseen card from R&D.
vincentkbl accesses an unseen card from R&D.
vincentkbl accesses an unseen card from R&D.
vincentkbl accesses an unseen card from R&D.
vincentkbl accesses an unseen card from R&D.
vincentkbl accesses an unseen card from R&D.
vincentkbl accesses an unseen card from R&D.
!vincentkbl uses a command: /close-prompt
vincentkbl accesses an unseen card from R&D.

### LINKED ISSUE #4039: Wrestling breakers (Laamb / Engolo) additional subtype stays on ICE if breaker is trashed during encounter
If a wrestling breaker is trashed before passing the ICE, the new subtype stays on the ICE forever.

>Bob spends 2 to use Laamb to make Rototurret gain Barrier.
>Alice uses Rototurret to trash Laamb.
>Alice uses Rototurret to end the run.

image

This will not go away even if passing the ICE later on.

NoahTheDuke:
# Overview
Jinteki.net has two kinds of abilities, paid abilities (`:abilities`, along with `:runner-abilities` and `:corp-abilities`) and triggered abilities (`:events`). This does not quite match up with how Netrunner's rules actually handle abilities, which has led to cumbersome work-arounds and what I consider "hacks" to achieve proper handling. This PR started as a simple implementation of Netrunner's Constant Abilities, and has grown to encompass triggered abilities that no longer rely on the triggering card to still be around (so-called floating effects), and abilities that have a duration.

# Changes
I'll be honest, this is a 5k line PR. Detailing every change would be the equivalent of reading the changes themselves. I'll try to cover each major change here.

## Effect System
Currently, numeric modifiers are determined by calling a `:pre-X` event, letting the abilities tied to that event modify a global value on `state` and then adding the global value with the base value afterwards. This is how `get-breaker-strength` and similar functions work, it's how `rez` works, it's the basis for all numeric modifiers.

Going forward, with `:constant-effects`, the `:effect` of a given handler is either a number (`:effect 3`) or a 5-fn that returns a number (`:effect (req (get-counters card :power))`). When `get-effects` is called, all of the effects are gathered and filtered by their `:req` returning a truthy value, and then the value of their `:effect` is put into a list and returned. This allows us to `sum` the values or print them or filter based on external criteria (`:cannnot-lower-strength`).

This is the whole reason for the PR, lol.

## Events
I've changed how `:events` are written in card implementations from a single map of event keywords to ability maps to instead to be a vector of ability maps with `:type` to event keyword added. The primary motivation is to make it consistent with the new effect system. This might seem like a big change and to a point it is, but it's also a necessary one as this allows us to write cards that accurately handle having multiple handlers for the same event. This also allows for writing wrapper functions that don't override the card-def event handlers (`auto-icebreaker` is the big offender).

This also fixes the "issue" of calling `swap!` multiple times whenever registering or unregistering events, as we now build up the list of events and then `swap!` them all in at once.

## Duration
Right now, all effects and events are bound to the played/active card. Once that card stops being active, the effect or event stops firing, even tho in some cases they should be firing regardless. To solve this, I've added `:duration`s to both events and effects. This allows us to register an effect that has a duration of `:end-of-turn` and trust that it will continue to affect the game state regardless of where the card has been moved (By Any Means, for example). Events default to `:while-installed` and effects default to `:constant`.

I've added the necessary `unregister-floating-effects` and `unregister-floating-events` calls to remove both events and effects after a timing structure has changed. This can be seen in `continue`, where after `trigger-event-sync :pass-ice`, `(unregister-floating-effects state side :end-of-encounter)` is called.

Alongside the duration changes, I've added `card-for-ability` to `core/events.clj`, which will only call `get-card` if the ability has a `:while-installed` duration. This handles the cases like By Any Means, where a card has their events registered and then stopped firing once the card is moved. Now we only care about getting the current version of a card when the card is installed and not when we're resolving an operation.

## get-card
I moved `get-card` and `get-card-hosted` into card.cljc because I want them to be generally available to all namespaces, and having them inside of `game.core` was detrimental to that goal. I'm not happy with their implementation, cuz they feel kind of hacky, but that's for another day.

## Costs
See *Modifiers* above to read about how cost modifiers are handled now.

* Rewrite `rez` to better handle alternative payments (Tithonium), as we no longer have to worry about resetting bonuses or calling `no-get-card`.
* Fix `pay-credits` to properly handle negative amounts. Previously, it would blindly add a negative amount (which caused credit gain).
* Rewrite `play-cost`, `rez-cost`, `install-cost`, `trash-cost`, and `run-cost` to have very similar structures: take the base value (from the card in most cases), add in the `cost-bonus` (which comes from cards like Career Fair directly modifying a cost when performing an action), add in any `:X-cost-bonus` from the card definition (which comes from the card's abilities, like Office Supplies), add any effects that would affect it (`:install-cost`, etc), sum all of those together, and then set the minimum (`(max 0)` in most cases). This produces a final number which is used as the credit total needed to perform the action. Because of the effect system, there's no need to call `trigger-event` or clear any bonuses, and no additional costs are touched or generated.
* Rewrite/write `play-additional-cost-bonus`, and similar functions to take any additional costs on the card object, the card definition, and from any association effects

## Ice and Icebreakers
All ice and icebreaker strength modification effects have been moved to the new effect system. This means that a given card doesn't know what kind of pump effects have been placed on it until `ice-strength` is called on it. On the other hand, this allows for directly modifying a given pump effect after it's been created.

As I said above about *Events*, the changes to the event system means that when we create an `auto-icebreaker`, we don't have to worry about overwriting the pre-wired events inside of a card def, we can just write them and they will happily co-exist next to the `breaker-auto-pump` events.

## Runs
* `get-runnable-zones` now properly takes into account run costs, including additional costs.
* Added `total-run-cost` to properly handle all possible costs for a given run.

## Cards rewritten/refactored/fixed:
* Eden Fragment
* IT Department
* Sandburg
* Watchdog
* Bribery
* Brute-Force Hack
* By Any Means
* Executive Wiretaps
* Frantic Coding
* Running Interference
* Scavenge
* Social Engineering
* System Seizure
* Cortez Chip
* Gebrselassie
* LLDS Processor
* Patchwork
* Chum
* Blue Sun
* Kabonesa Wu
* Khan
* Nasir
* Null
* Product Recall
* Dhegdheer
* Wyrm
* Councilman
* Counter Surveillance
* Dean Lister
* Political Operative
* Rosetta 2.0
* The Helpful AI
* The Supplier
* Awakening Center
* Cold Site Server
* Corporate Troubleshooter
* Helheim Servers
* Reduced Service

## Tests
Fixed the bug of some of Rebirth's tests being outside of the `deftest`, so they got run first. Cleaned up a bunch of tests, added some for bug fixes. Added an `effects` file to unit test the effect system.

# Issues
Closes #3341 
Closes #3377 
Closes #4039


test/clj/game_test/cards/assets.clj
```diff
@@ -3434,10 +3434,11 @@
       (click-prompt state :corp "Yes")
       (is (zero? (count-bad-pub state)) "Removed 1 bad pub"))))
 
-(deftest sandburg
+(deftest ^:test-refresh/focus sandburg
```
lostgeek: :eyes:

NoahTheDuke: Whoops!

ouroboros8:
>This might seem like a big change and to a point it is, but it's also a necessary one as this allows us to write cards that accurately handle having multiple handlers for the same event.

ouroboros8:
> See Modifiers above
there is no Modifiers section :) I guess you mean "See Effect System"?

NoahTheDuke:
Whoops, seems like some edits didn’t get saved. Yes, that’s correct

Issue #4496, opened 2019-09-17
Citadel Sanctuary not logging cards correctly


5N00P1:
[username] trashes Citadel Sanctuary and trashes all (5) cards in their grip (clojure.lang.LazySeq@fc688fbd) to use Citadel Sanctuary to prevent all meat damage.

NoahTheDuke:
Thanks for the report!

PR #4497, opened 2019-09-17
Fix message when runner trashes entire hand

kalotay:
This fix adds the cards names trashed when the runner trashes their hand instead of the name of a clojure datastructure.

Fixes https://github.com/mtgred/netrunner/issues/4496

Before fix:
Screenshot 2019-09-17 at 17 57 07

After fix:
Screenshot 2019-09-17 at 18 02 54


NoahTheDuke:
This looks great! Welcome to the team!

PR #4508, opened 2019-09-24
Combine trash and trash-cards

### LINKED ISSUE #4472: Consolidate trash and trash-cards
Can easily change the `trigger-event :trash-card trashed-cards` to be a recursive function to call with `{:cause cause}`.

NoahTheDuke:
* Adds `cause` to `trash-cards` call, so `(second targets)` still refers to the cause.
* Changes `trash` to call `trash-cards` with its inputs, wrapping the single card in a list.
* Removes the restriction on triggering `:X-trash` on currents.
* Renames `resolve-trash-end` to just `resolve-trash`, for clarity.

Closes #4472

### DIFF
--- a/src/clj/game/core/rules.clj
+++ b/src/clj/game/core/rules.clj
@@ -440,8 +440,8 @@
       (swap! state update-in [:trash :trash-list] dissoc eid)
       (swap! state update-in [:trash :trash-list eid] #(remove-once (fn [x] (same-card? x card)) %)))))
 
-(defn- resolve-trash-end
-  ([state side eid card args] (resolve-trash-end state side eid card eid args))
+(defn- resolve-trash
+  ([state side eid card args] (resolve-trash state side eid card eid args))
   ([state side eid {:keys [disabled] :as card} oid
     {:keys [cause keep-server-alive host-trashed] :as args}]
    (remove-from-trash-list state oid card)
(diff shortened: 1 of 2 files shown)


src/clj/game/core/rules.clj
```diff
@@ -460,18 +460,6 @@
          (continue-ability state side trash-effect moved-card (list cause))))
      (effect-completed state side eid))))
 
-(defn- resolve-trash
-  ([state side eid card args] (resolve-trash state side eid card eid args))
-  ([state side eid {:keys [zone] :as card} oid
-   {:keys [cause suppress-event] :as args}]
-  (if (and (not suppress-event)
-           (not= (last zone) :current)) ; Trashing a current does not trigger a trash event.
```
ouroboros8: You don't replicate this check for currents in your new code, is it still needed?

NoahTheDuke: I asked in the nisei rules chat, and they said that right now, no it's not, but that might change in the future. Given that all of the tests pass, I don't think anything is currently changed because of this, but we'll see once this hits live. Thankfully, it's a _very_ easy fix to re-add.

ouroboros8:
LGTM!

src/clj/game/core/rules.clj
NoahTheDuke: As seen by #4529, I am proven wrong, lol. I'll have to figure out a fix without reverting this PR.

PR #4561, opened 2019-10-13
Partially rewrite `fetch` to allow fetching without a database up

ouroboros8:


NoahTheDuke:
Damn, this looks great

README.md
```diff
@@ -13,82 +14,80 @@ http://www.jinteki.net
...
+
+This project runs on Java 8. If you're on OSX or Linux, we recommend using
+[jenv](https://github.com/jenv/jenv/blob/master/README.md) to manage your java environment.
+
+You can check your setup by running
+
+    $ lein version # Your exact version numbers below may vary, but we expect Java 1.8.X
+    Leiningen 2.9.1 on Java 1.8.0_222 OpenJDK 64-Bit Server VM
+
+A quick way to get a MongoDB database set up is using [Docker](https://www.docker.com/):
```
NoahTheDuke: I don't feel comfortable jumping directly to Docker in this case. Seems like an oversized solution for our needs. Until we can Docker-ize the whole app, I think it's best to keep the recommendation for "just install mongo and run it in a tab".

README.md
```diff
@@ -13,82 +14,80 @@ http://www.jinteki.net
...
+    $ lein fetch
 
+You can optionally pass `--no-card-images` if you don't want to download images from
+[NetrunnerDB](https://netrunnerdb.com/), as this takes a while. See `lein fetch help` for further options.
 
-## Installation
+To install frontend dependencies, run:
 
-Install frontend dependencies:
+    $ npm install -g bower
```
NoahTheDuke: What's your reasoning for changing the code formatting from using backticks to indentation?

src/clj/tasks/fetch.clj
```diff
@@ -1,36 +1,49 @@
...
+   ["-n" "--no-db" "Do not load edn data into the database"
+    :id :db
+    :parse-fn not]
+   ["-i" "--card-images" "Fetch card images from NetrunnerDB (default)"
+    :id :card-images
+    :default true]
+   ["-j" "--no-card-images" "Do not fetch card images from NetrunnerDB"
+    :id :card-images
+    :parse-fn not]
+   ])
```
NoahTheDuke: :eyes:

NoahTheDuke:
You'll also want to change the fetch line in this file: https://github.com/mtgred/netrunner/blob/master/.circleci/config.yml

README.md
ouroboros8: I think it makes the unrendered markdown a bit easier to read personally.

src/clj/tasks/fetch.clj
ouroboros8: I think this is ok? Unless I'm missing something.

README.md
ouroboros8: No worries, I prefer docker for local dev with databases most of the time, since it's a bit easier to manage multiple versions if you need to etc. I'll update this to link back out to mongodb install instructions instead.

ouroboros8:
> You'll also want to change the fetch line in this file: https://github.com/mtgred/netrunner/blob/master/.circleci/config.yml

Don't think that's necessary! `lein do fetch --no-card-images, eftest` should work as normal.

NoahTheDuke:
Are you sure? Fetching with the db is the default, right? We can skip setting up mongo and fetching to the db to run the tests

src/clj/tasks/fetch.clj
NoahTheDuke: Just that normally we don't leave hanging parens. This isn't that big of a deal tho, so don't stress.

ouroboros8:
Oh gotcha! Yeah good shout :)

src/clj/tasks/fetch.clj
ouroboros8: ooh yes, no that's a good point! I am a fan of consistency

PR #4576, opened 2019-10-19
Bugfix Top Hat interaction with run based cards

### LINKED ISSUE #4542: Tophat and Mad Dash do not work
It seems that the interaction between Top Hat and Mad Dash is broken. When choosing an agenda card using Top Hat, the meat damage fires and the Mad Dash is not added to the runner's score area. This may be that the run is treated as over upon the firing of Top Hat?

lostgeek:
Top Hat wasn't waiting correctly. Closes #4542

### DIFF
--- a/src/clj/game/cards/hardware.clj
+++ b/src/clj/game/cards/hardware.clj
@@ -1556,8 +1556,9 @@
                                     :replace-access
                                     {:mandatory true
                                      :async true
-                                     :effect (req (when (empty? (get-cards-to-access state))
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/hardware.clj
```diff
@@ -1555,6 +1555,7 @@
                                    {:card card
                                     :replace-access
                                     {:mandatory true
+                                     :async true
                                      :effect (req (when (empty? (get-cards-to-access state))
```
NoahTheDuke: Seeing as we're making the ability async, this should be an `if` and the else branch calls `effect-completed`

PR #4578, opened 2019-10-19
Stop breakers from revealing information

### LINKED ISSUE #4545: Kongamato Reveals Information About Unrezzed ICE
Discovered in a game when my opponent accidentally used Kongamato to break a subroutine after clicking "Continue" so that Kongamato fired on the next (unrezzed) ICE. Tested again and was able to reproduce on any unrezzed ICE.

Reproduction steps:
1. Runner initiates a run with Kongamato installed
2. Runner uses Kongamato while approaching unrezzed ICE
3. Kongamato is trashed and the following text is displayed in the log:
[Runner] trashes Kongamato to use Kongamato to break the first subroutine on Eli 1.0 (" End the run").

This has the potential to reveal information about unrezzed ICE if the runner mis-times the use of Kongamato during a run. It is also possible that this bug affects Gbahali, but I have not had a chance to test that.

lostgeek:
Some breakers didn't check whether ice is rezzed and leaked information about the subroutines. Closes #4545

### DIFF
--- a/src/clj/game/cards/programs.clj
+++ b/src/clj/game/cards/programs.clj
@@ -112,10 +112,11 @@
   (Greek/Philosopher suite: Adept, Sage, Savant)"
   [first-qty first-type second-qty second-type]
   {:cost [:credit 2]
-   :req (req (or (and (has-subtype? current-ice first-type)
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
If you wanna rebase on master, I can merge this!

PR #4588, opened 2019-10-26
Fix tests

lostgeek:
Trick of Lights `:req` was not regarding all cases and is not really necessary anyways.

### DIFF
--- a/src/clj/game/cards/operations.clj
+++ b/src/clj/game/cards/operations.clj
@@ -2076,12 +2076,11 @@
 
    "Trick of Light"
    {:async true
-    :req (req (let [can-be-advanced (some #(when (can-be-advanced? %) %)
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
Do you need the installed check? How do you get counters on a not installed card?

lostgeek:
I was informed yesterday on Slack that hosting a card on Film Critic uninstalls it. It should also lose its counters but currently does not (see #4589). And in principle, this should apply pretty much everywhere we select a card:

> NISEI Comprehensive Rules Document
> 1.14.1. If a card ability requires a player to choose cards, the player can only choose installed
cards unless the ability specifically works on cards in other zones.

NoahTheDuke:
Okay, cool.

PR #4603, opened 2019-11-01
Fully Automated Run Timing!

NoahTheDuke:
This has been a _long time_ coming. With this, runs will be more cumbersome and more accurate than ever thought possible. Every step of the process will be handled by prompts instead of assumed by the front end.

Changing from the old system, where we smooshed approach and encounter together, and poorly handled passing, we now have prompts for each Paid Ability Window, and subroutines fire automatically, so there's no need for pressing "No action" as Corp to "let" the runner pass the ice. Unless they have an effect, they can't pass!

This is going to take a while, because I'm gonna have to first figure out a good usable UX for each of these steps in the run, and then I'm going to have to adapt every test we have that touches runs (it's over 900) to make sure they work as well. :yikes:


lostgeek:
Is there a reason, you switched from the `:encounter-ice` event to an `:on-encounter` key in the ability for ice? I'm not necessarily against it, just wondering why it should be different for ice than for any other card type.

NoahTheDuke:
Well, it's not quite different for other card types. We have `:access` for on access events, `:trash-effect` for trashed cards, `:stolen` for when an agenda is stolen, and we bundle the card effect from installing a runner card as a `card-ability` as well. This was the cleanest way to reduce the boilerplate of writing `:req (req (and (same-card? card target) ...))` in every single `:encounter-ice` event. By using `:on-encounter`, I can guarantee in a single place that the necessary checks are being performed, whereas when doing it "manually" in `:event :encounter-ice`, it's easier to forget a check or be sloppy.

So many of our effects should work like this, but they don't, lol. This seemed like a good place to introduce another.

Issue #4463, opened 2019-08-25
Corroder breaks Pachinko completely with 3 credits instead of 4


demoy:
I have not had a chance to see if other breakers break correctly (beside paperclip which does)

PR #4616, opened 2019-11-12
added Pachinko tests

Kubik161:
I was looking at issue #4463 (marked as good first issue) and found out, that it's already fixed.  Then I checked google docs spreadsheet, found out that Pachinko has no tests, so I at least tried to write some tests. Hope it helps.

Issue #4463 should be closed.


NoahTheDuke:
Thank you for the contribution! This is a great way to get started on development. My first and only note so far is to remove the blank lines between the tests. I use an auto-formatter that messes up otherwise.

PR #4634, opened 2019-12-01
Fix for Gene Splicer ability, new test case

Kubik161:
I have been playing Gene Splicer yesterday and noticed, that ability have incorrect cost (2 adv tokens vs correct 3). I have fixed that and wrote new test for this situation.

### DIFF
--- a/test/clj/game_test/cards/assets.clj
+++ b/test/clj/game_test/cards/assets.clj
@@ -1696,7 +1696,7 @@
         (core/rez state :corp (refresh gs))
         (card-ability state :corp (refresh gs) 0)
         (is (refresh gs) "Gene Splicer is still in remote")
-        (is (zero? (count (get-scored state :corp))) "Score area is still empty")))))
+        (is (empty (get-scored state :corp)) "Score area is still empty")))))
 
 (deftest genetics-pavilion
   ;; Genetics Pavilion - Limit Runner to 2 draws per turn, but only during Runner's turn
--- a/test/clj/game_test/cards/assets.clj
+++ b/test/clj/game_test/cards/assets.clj
@@ -1686,12 +1686,11 @@
         (is (= 1 (:agendapoints (get-scored state :corp 0))) "Gene Splicer added to Corp score area"))))
   (testing "Corp double-advances a Gene Splicer and fails to use its ability to add to their score area as a 1 point agenda"
     (do-game
-      (new-game {:corp {:deck [(qty "Gene Splicer" 2) (qty "Ice Wall" 3) (qty "Vanilla" 2)]}
+      (new-game {:corp {:hand "Gene Splicer"}
                  :runner {:deck [(qty "Sure Gamble" 3)]}})
       (play-from-hand state :corp "Gene Splicer" "New remote")
       (let [gs (get-content state :remote1 0)]
-        (core/advance state :corp {:card (refresh gs)})
-        (core/advance state :corp {:card (refresh gs)})
+        (dotimes [_ 2] (core/advance state :corp {:card (refresh gs)}))
         (take-credits state :runner)
         (core/rez state :corp (refresh gs))
         (card-ability state :corp (refresh gs) 0)
--- a/test/clj/game_test/cards/assets.clj
+++ b/test/clj/game_test/cards/assets.clj
@@ -1690,12 +1690,13 @@
                  :runner {:deck [(qty "Sure Gamble" 3)]}})
       (play-from-hand state :corp "Gene Splicer" "New remote")
       (let [gs (get-content state :remote1 0)]
-        (core/add-counter state :corp gs :advancement 2)
+        (core/advance state :corp {:card (refresh gs)})
+        (core/advance state :corp {:card (refresh gs)})
         (take-credits state :runner)
         (core/rez state :corp (refresh gs))
         (card-ability state :corp (refresh gs) 0)
-        (is (= "Gene Splicer" (:title (get-content state :remote1 0))) "Gene Splicer is still in remote")
-        (is (= nil (:agendapoints (get-scored state :corp 0))) "Score area is still empty")))))
+        (is (refresh gs) "Gene Splicer is still in remote")
+        (is (zero? (count (get-scored state :corp))) "Score area is still empty")))))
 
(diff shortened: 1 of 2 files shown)


test/clj/game_test/cards/assets.clj
```diff
@@ -1683,7 +1683,19 @@
...
         (is (nil? (get-content state :remote1 0)) "Gene Splicer is no longer in remote")
-        (is (= 1 (:agendapoints (get-scored state :corp 0))) "Gene Splicer added to Corp score area")))))
+        (is (= 1 (:agendapoints (get-scored state :corp 0))) "Gene Splicer added to Corp score area"))))
+  (testing "Corp double-advances a Gene Splicer and fails to use its ability to add to their score area as a 1 point agenda"
+    (do-game
+      (new-game {:corp {:deck [(qty "Gene Splicer" 2) (qty "Ice Wall" 3) (qty "Vanilla" 2)]}
+                 :runner {:deck [(qty "Sure Gamble" 3)]}})
+      (play-from-hand state :corp "Gene Splicer" "New remote")
+      (let [gs (get-content state :remote1 0)]
+        (core/add-counter state :corp gs :advancement 2)
```
NoahTheDuke: We have a helper function `advance` which is best suited for this.

test/clj/game_test/cards/assets.clj
```diff
@@ -1683,7 +1683,19 @@
...
+        (take-credits state :runner)
+        (core/rez state :corp (refresh gs))
+        (card-ability state :corp (refresh gs) 0)
+        (is (= "Gene Splicer" (:title (get-content state :remote1 0))) "Gene Splicer is still in remote")
```
NoahTheDuke: Probably better as `(is (refresh gs) ...)` as `refresh` returns either a card object or `nil`. We don't need to be more explicit because we already have a reference to the card object and `refresh` will return `nil` if it changes locations (as it calls `get-card` under the hood).

test/clj/game_test/cards/assets.clj
```diff
@@ -1683,7 +1683,19 @@
...
+        (is (= nil (:agendapoints (get-scored state :corp 0))) "Score area is still empty")))))
```
NoahTheDuke: This doesn't check what you think it does. Because of nil punning, `get-scored` already returns `nil` so calling `:agendapoints` on it returns `nil` as well. Better to call something like `(nil? (get-scored state :corp 0))` as that checks if the card object itself exists or not. Maybe even better to call `(zero? (count (get-scored state :corp)))` as that explicitly checks that there are 0 cards in the corp's score area.

NoahTheDuke:
Thank you so much for the PR! I'm so glad to see new folks join into the fun. Please let me know if you need more explicit instructions, and once we get this merged I can add you to the dev slack.

Kubik161:
Thanks for review, I have modified that test by your instructions. Let me know, if there are any more problems :)

test/clj/game_test/cards/assets.clj
```diff
@@ -1690,12 +1690,13 @@
...
-        (core/add-counter state :corp gs :advancement 2)
+        (core/advance state :corp {:card (refresh gs)})
+        (core/advance state :corp {:card (refresh gs)})
         (take-credits state :runner)
         (core/rez state :corp (refresh gs))
         (card-ability state :corp (refresh gs) 0)
-        (is (= "Gene Splicer" (:title (get-content state :remote1 0))) "Gene Splicer is still in remote")
-        (is (= nil (:agendapoints (get-scored state :corp 0))) "Score area is still empty")))))
+        (is (refresh gs) "Gene Splicer is still in remote")
+        (is (zero? (count (get-scored state :corp))) "Score area is still empty")))))
```
lostgeek: Not wrong (and was suggested by Noah in this way), but isn't `empty?` better than `zero? count` ?

NoahTheDuke: It is! Good looking out.

test/clj/game_test/cards/assets.clj
```diff
@@ -1683,7 +1683,20 @@
         (core/rez state :corp (refresh gs))
         (card-ability state :corp (refresh gs) 0)
         (is (nil? (get-content state :remote1 0)) "Gene Splicer is no longer in remote")
-        (is (= 1 (:agendapoints (get-scored state :corp 0))) "Gene Splicer added to Corp score area")))))
+        (is (= 1 (:agendapoints (get-scored state :corp 0))) "Gene Splicer added to Corp score area"))))
+  (testing "Corp double-advances a Gene Splicer and fails to use its ability to add to their score area as a 1 point agenda"
+    (do-game
+      (new-game {:corp {:deck [(qty "Gene Splicer" 2) (qty "Ice Wall" 3) (qty "Vanilla" 2)]}
```
lostgeek: The other cards (Ice Wall, Vanilla) are not needed. You could simply use `:hand ["Gene Splicer"]` instead.

test/clj/game_test/cards/assets.clj
```diff
@@ -1683,7 +1683,20 @@
...
+                 :runner {:deck [(qty "Sure Gamble" 3)]}})
+      (play-from-hand state :corp "Gene Splicer" "New remote")
+      (let [gs (get-content state :remote1 0)]
+        (core/advance state :corp {:card (refresh gs)})
+        (core/advance state :corp {:card (refresh gs)})
```
lostgeek: You could also use `(dotimes [_ 2] (core/advance ...))` to have it a bit more concise

Kubik161:
Thanks @lostgeek :)

test/clj/game_test/cards/assets.clj
```diff
@@ -1696,7 +1696,7 @@
         (core/rez state :corp (refresh gs))
         (card-ability state :corp (refresh gs) 0)
         (is (refresh gs) "Gene Splicer is still in remote")
-        (is (zero? (count (get-scored state :corp))) "Score area is still empty")))))
+        (is (empty (get-scored state :corp)) "Score area is still empty")))))
```
lostgeek: Watch out. `(empty coll)` returns an empty collection of the same type. You're looking for `(empty? coll)` :wink:

PR #4638, opened 2019-12-02
Increase the size of the card zoom

nathanj:
The majority of card images are 300x418, so display them at that
resolution to prevent downscaling and blurring the text. This means the
image will be larger and crisper. The resulting image is significantly
easier to read in my opinion.

Current zoom:
old-zoom

New zoom:
new-zoom

### DIFF
--- a/src/cljs/nr/account.cljs
+++ b/src/cljs/nr/account.cljs
@@ -63,6 +63,8 @@
   (swap! app-state assoc-in [:options :show-alt-art] (:show-alt-art @s))
   (swap! app-state assoc-in [:options :stacked-servers] (:stacked-servers @s))
   (swap! app-state assoc-in [:options :runner-board-order] (:runner-board-order @s))
+  (swap! app-state assoc-in [:options :log-width] (:log-width @s))
+  (swap! app-state assoc-in [:options :log-top] (:log-top @s))
   (swap! app-state assoc-in [:options :blocked-users] (:blocked-users @s))
   (swap! app-state assoc-in [:options :alt-arts] (:alt-arts @s))
   (swap! app-state assoc-in [:options :gamestats] (:gamestats @s))
@@ -70,6 +72,8 @@
   (.setItem js/localStorage "sounds" (:sounds @s))
   (.setItem js/localStorage "lobby_sounds" (:lobby-sounds @s))
   (.setItem js/localStorage "sounds_volume" (:volume @s))
+  (.setItem js/localStorage "log-width" (:log-width @s))
+  (.setItem js/localStorage "log-top" (:log-top @s))
   (.setItem js/localStorage "stacked-servers" (:stacked-servers @s))
   (.setItem js/localStorage "runner-board-order" (:runner-board-order @s))
   (post-options url (partial post-response s)))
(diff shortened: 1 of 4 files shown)


NoahTheDuke:
This is a pretty dramatic change in the shape and size of the sidebar. I'm not necessarily against it but _a lot_ of players already zoom out to fit more of the board onto the screen at one time. This will only make things worse.

nathanj:
I understand not wanting to change the current layout. Would you be open to having a setting to toggle between the two?

Speaking as a casual player, I do not have many of the cards memorized and I spend a lot of the game looking at the zoomed card image. Trying to read the text is one of my biggest frustrations with jinteki. I also have terrible eyesight though so maybe I'm in the minority.

FarCryFromHuman:
It might be better to have toggleable hover functionality that pulls the text from NRDB. That way it always scales correctly, and includes the latest errata. That's probably quite a bit more work though.

lostgeek:
Could you link the size of the preview panel to the size of the resizable textbox? Then players could freely choose a size they're comfortable with (preferably saved in the user settings).

nathanj:
Oh wow. I never realized you could resize the log. I've pushed the change to resize the card along with the log. Let me know what you think.

The width doesn't persist in settings since I'm not sure the easiest way to do that yet.

lostgeek:
2019-12-03-102706_1920x1080_escrotum

I like these changes a lot! I only see one problem with it: people (myself included) like to take screenshots of the log though to post on Slack or social media. And for this you often drag out the log to make everything fit better into a single picture. When doing this with your implementation, the log gets very short, very fast (see screenshot above).

Can you add another resizing handle to the "north" side of the chat log, where the chat log stays constant in width but the card preview resizes to make room for the log?

Another thing: I saw that you took out the transitions and at least on my machine it changed nothing regarding the card preview. It seems that part was broken anyways... :sweat_smile:
Can you get it to work again? It may just be because I'm now focused on the card preview, but I feel fading it in and out with a very short duration like 200ms would make it feel a lot smoother.

Oh and I'll take a look on how to save the width to the user settings. I don't know how exactly to do it either, but I think I know where to look. I'll get back to you on that.

lostgeek:
I've sent you a PR with an entry for the user settings: https://github.com/nathanj/netrunner/pull/1

I couldn't figure out how to easily initiate the log panel with the current width, but maybe you can find a solution.

nathanj:
Thanks for the patch. From some quick googling it looks like doing an opacity transition should do the trick. I'll play around with it tonight.

I agree with the north handle. I'm thinking changing the width of the log should not auto increase the height of the card. Instead the user would set the width and height of the log they want and the card will fill the rest of the space in the correct ratio. That way the user can decide whether to have a large log or a large card.

nathanj:
New version up. The result is pretty nice I think. I've ran out of time tonight to integrate your patch. I'll check it out tomorrow.

NoahTheDuke:
Don't forget about the patch!

nathanj:
I worked on it one night but wasn't able to get the initialization to work correctly. Been busy with other stuff lately so I haven't revisited it.

NoahTheDuke:
No worries! I'll see what I can do, cuz I don't want to leave this behind.

nathanj:
Finally got back around to this. I ended up storing the values in local storage immediately upon resize which should remove the need to modify them on the settings page. The settings felt kind of clunky to me. Take a look and see what you think.

src/cljs/nr/gameboard.cljs
```diff
@@ -1725,6 +1738,7 @@
                 [log-pane]
                 [log-typing]
                 [log-input]]]
+              (do (resize-card-zoom) nil)
```
nathanj: Is there a better way to do this? I'm not familiar with reagant.

NoahTheDuke: What is the purpose of this line?

nathanj: (resize-card-zoom) is to perform the resize on load. The (do .. nil) is to return a nil so reagant ignores it. Basically I'm trying to run some javascript inline with the html generation.

NoahTheDuke: yeah, okay. I feel like it's not my favorite way to write that, but I can't think of a better way. I can always clean this up later if needed. (FAMOUS LAST WORDS)

src/cljs/nr/account.cljs
```diff
@@ -127,6 +131,33 @@
...
+                :min 100 :max 2000
+                :on-change #(do (swap! s assoc-in [:log-width] (.. % -target -value))
+                                (reset! log-width (.. % -target -value)))
+                :value @log-width}]
+       [:button.update-log-width {:type "button"
+                                  :on-click #(do (swap! s assoc-in [:log-width] (get-in @app-state [:options :log-width]))
+                                                 (reset! log-width (get-in @app-state [:options :log-width])))} "Get current log width" ]])))
+(defn log-top-option [s]
+  (let [log-top (r/atom (:log-top @s))]
+    (println @log-top)
```
lostgeek: You missed a println.

src/cljs/nr/account.cljs
```diff
@@ -127,6 +131,33 @@
    (doseq [card (vals (:alt-arts @app-state))]
      (update-card-art card art s))))
 
+(defn log-width-option [s]
+  (let [log-width (r/atom (:log-width @s))]
+    (println @log-width)
```
lostgeek: Here as well.

nathanj:
Oops I hadn't expected that last commit to be merged in. Guess I should have labeled this WIP. I wasn't sure if you wanted to keep those settings. Since I changed it to store the values in local storage immediately upon resize it didn't seem necessary to have the setting. Changing the values manually just seems kind of clunky.

NoahTheDuke:
Lmao okay

PR #4643, opened 2019-12-03
Settings for the log width

lostgeek:
I've added a settings option called `:log-width` that lives in the `app-state`. You can get the current width with `(get-in @app-state [:options :log-width])`.

What is still missing is setting the initial width of the log panel... There's probably an easy way of doing that but I can't see it right now.


lostgeek:
This is not where this PR was supposed to go...

PR #4659, opened 2019-12-04
Implement Bass CH1R180G4

kalotay:


NoahTheDuke:
Thanks for this! Can you add a check to the test that Bass is properly trashed? Something like `(is (refresh bass))` to the end of the test after checking the number of clicks.

PR #4665, opened 2019-12-05
server name added to cayambe prompt, new tests

### LINKED ISSUE #4625: Multiple cayambe grid triggers should specify server
As it stands, if you have multiple servers with cayambe grid installed at the start of the turn, you just receive multiple prompts with the text 'cayambe grid'
The pop-up window states "place 1 advancement token on an ice protecting this server", without a specification as to which server it is.
I believe either the prompt should specify the server, or the popup after clicking it should specify the server.

Kubik161:
Closes #4625 

Added server name to prompt title. And some tests.


NoahTheDuke:
Thanks for this!

PR #4673, opened 2019-12-05
Fix Rebirth and DJ Fenris, and simplify prompt handling

### LINKED ISSUE #4672: Broken Rebirth and DJ Fenris prompts
Rebirth and DJ Fenris prompts no longer work. This issue was introduced by #4645

ouroboros8:
The front-end now only passes card titles to the backend when a prompt
with a choice of cards is clicked. The backend then compares this title
against the titles of the cards in the prompt hash-map.

Resolves #4672

### DIFF
--- a/src/clj/game/core/actions.clj
+++ b/src/clj/game/core/actions.clj
@@ -195,12 +195,9 @@
 (defn resolve-prompt
   "Resolves a prompt by invoking its effect function with the selected target of the prompt.
   Triggered by a selection of a prompt choice button in the UI."
-  [state side {:keys [choice card] :as args}]
-  (let [servercard (get-card state card)
-        card (if (not= (:title card) (:title servercard))
-               (server-card (:title card))
-               servercard)
-        prompt (first (get-in @state [side :prompt]))
+  [state side {:keys [choice] :as args}]
+  (let [prompt (first (get-in @state [side :prompt]))
+        card (get-card state (:card prompt))
         choices (:choices prompt)]
     (cond
       ;; Shortcut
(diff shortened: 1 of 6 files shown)


src/clj/game/core/actions.clj
```diff
@@ -250,20 +247,14 @@
...
-                            choices)
-            button (first buttons)]
-        (if button
-          (do (when-let [effect-prompt (:effect prompt)]
-                (effect-prompt button))
+      (let [match (first (filter #(or (= choice %)
+                                       (= choice (:title % "")))
+                                  choices))]
+        (if match
+          (do ((:effect prompt) match)
```
NoahTheDuke: Not all prompts will have effects, so you'll want to leave the `when-let` in.

ouroboros8: Huh. If there's no effect in the default case does the prompt just do nothing?

NoahTheDuke: You're right, I was thinking of a different part of the engine. Please ignore me.

src/clj/game/core/actions.clj
```diff
@@ -250,20 +247,14 @@
...
-                                                    (lower-case (:title choice "do-not-match")))]
-                                   (or (= choice-str (lower-case %))
-                                       (= choice-str (lower-case (:title % ""))))))
-                            choices)
-            button (first buttons)]
-        (if button
-          (do (when-let [effect-prompt (:effect prompt)]
-                (effect-prompt button))
+      (let [match (first (filter #(or (= choice %)
+                                       (= choice (:title % "")))
```
NoahTheDuke:
```suggestion
                                      (= choice (:title % "")))
```

Issue #4662, opened 2019-12-04
Mobius gives 4cr on the first run as well


manveruppd:
![moebius](https://user-images.githubusercontent.com/44246400/70177628-a6774900-16d2-11ea-93fa-c176b882a5d1.png)

Mobius gives the runner 4cr on the first successful run on R&D as well as the second one. This happens only if Mobius is recurred from the heap in any way. I've tested this myself and can confirm: it doesn't happen on the first time you play it, but it did happen if I recurred it with Deja Vu, Same Old, or Steve Cambridge.

NoahTheDuke:
Thanks for the report!

PR #4689, opened 2019-12-08
refactor moebius tests

kalotay:
Möbius tests were refactored and test for issue #4662 was added. Currently the recursion tests fail


NoahTheDuke:
I found the issue, and in "fixing" it, it opened a whole can of worms ha. I'm taking a look to see how to help you out here.

PR #4700, opened 2019-12-10
Fix for Jinja and 419 blocking each other

### LINKED ISSUE #3806: Game stuck with Jinja waiting for 419 and 419 waiting for Jinja
The game entered a state that couldn't be continued with both sides waiting for the opponent's trigger to resolve - a 419 trigger on the runner side and a Jinja City Grid trigger on the corp side.

The screenshot below should show the flow of events that led to the bug, but I've added it in text as well:

The corp (Sportsmetal) has a remote server with a Jinja City Grid installed and rezzed in it, and Arella Salvatore and Ash installed in it and unrezzed.
The runner (419) ran HQ and used 4 Turning Wheel counters to access 2 additional cards
419 accessed and stole a Corporate Sales Team. Sportsmetal chose to gain 2 credits.
419 then accessed another Corporate Sales Team. Sportsmetal then chose to draw 2 cards.
Sportsmetal drew Turing and IP Block (I believe in that order). Jinja City Grid fired, and Sportsmetal chose to reveal and install Turing. Before Sportsmetal could resolve the Jinja trigger for IP Block, Sportsmetal had to wait for 419 to choose to resolve its trigger on the installed-via-Jinja Turing. However, the 419 player stated that on their end they had a message saying "Waiting for the corp to resolve Jinja City Grid".
The game could not be continued as both players were waiting for the opponent's trigger to resolve before they could resolve their trigger (that the opponent was waiting on to be resolved before they could resolve their's). A deadlock, of sorts.

jinja419bug

Kubik161:
Closes #3806 

I have fixed this by increasing priority of 419 prompt. Waiting prompt has priority 1, so I had to use :priority 2 instead of :priority true

### DIFF
--- a/src/clj/game/cards/identities.clj
+++ b/src/clj/game/cards/identities.clj
@@ -52,6 +52,7 @@
                        state side
                        {:optional
                         {:prompt "Expose installed card unless Corp pays 1 [Credits]?"
+                         :priority 2
                          :player :runner
                          :autoresolve (get-autoresolve :auto-419)
                          :no-ability {:effect (req (clear-wait-prompt state :corp))}
--- a/test/clj/game_test/cards/identities.clj
+++ b/test/clj/game_test/cards/identities.clj
@@ -114,7 +114,28 @@
       (play-from-hand state :corp "Building Blocks")
       (click-card state :corp "Ice Wall")
       (click-prompt state :corp "HQ")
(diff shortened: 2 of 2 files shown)


NoahTheDuke:
God, the priority system sucks haha. Thank you for the fix!

PR #4718, opened 2019-12-11
Paperwall fix

### LINKED ISSUE #4712: Paper Wall has the wrong timing
Should be `:event :subroutine-broken`.

Kubik161:
Closes #4712 

Fix and new test.


NoahTheDuke:
You’re knocking these out! Thank you! Can you move the trashed check to before the `run-continue` call? Your test would pass before your fix right now.

PR #4722, opened 2019-12-12
Tranqulity Home Grid interaction with ICE fix

### LINKED ISSUE #4721: Installing ice protecting the server turns off Tranquility Home Grid
Tranquility Home Grid  is rezzed and click one I installed an ice protecting that remote server. Click two i installed an agenda in the remote with tranquility. It should fire but it didn't. I think it might be a bug.

Kubik161:
Closes #4721

### DIFF
--- a/src/clj/game/cards/upgrades.clj
+++ b/src/clj/game/cards/upgrades.clj
@@ -1407,8 +1407,7 @@
                                  (agenda? target)
                                  (upgrade? target))
                              (= (second (:zone card)) (second (:zone target)))
-                             (first-event? state :corp :corp-install #(and (= (-> card :zone second) (-> % first :zone second))
-                                                                           (= (nth (-> card :zone) 2) (nth (-> % first :zone) 2))))))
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/upgrades.clj
```diff
@@ -1407,7 +1407,8 @@
                                  (agenda? target)
                                  (upgrade? target))
                              (= (second (:zone card)) (second (:zone target)))
-                             (first-event? state :corp :corp-install #(= (-> card :zone second) (-> % first :zone second)))))
+                             (first-event? state :corp :corp-install #(and (= (-> card :zone second) (-> % first :zone second))
```
NoahTheDuke: This seems like a good place to use `in-server?` instead of manually checking the `:zone`.

Kubik161: Feel free to correct me :) I was trying to use in-same-server? but was unable to make it work :)

NoahTheDuke: Huh! If it's not working, then something has gone wrong. I'll take a look.

PR #4757, opened 2019-12-15
Null ability fix

Kubik161:
While fixing Nero ability, I have discovered that Null ability is broken too. 

Previous implementation allowed to use ability only on first encounter each turn. It should allow to use ability on any encounter, once per turn.


NoahTheDuke:
Good catch, thank you!

NoahTheDuke:
Merged with b8824af432c4d9b51465d38746a4d5cbb7ae4bff

PR #4786, opened 2019-12-18
Turning Wheel bounce

### LINKED ISSUE #4775: Add bouncing ability to Turning Wheel
Add an ability to Turning Wheel to represent "bouncing" on some ice. This should set the corresponding flags for unsuccessful runs.

This can also be added to Au Revoir

Kubik161:
Will close #4775 sometimes in the future :)

Added new ability on Turning Wheel which will allow runner to bounce HQ/R&D and add power token. As @lostgeek suggested, it's not done as a run, so we have to set some run flags manually? I kind of need help with this, could you please check my changes? Thanks. :)

EDIT: Second question: Is there any way to show Toast or print chat message, if :req evaluates as false? I would like to add some message like "There is no suitable ICE for automatic bounce" for better UX.

### DIFF
--- a/src/clj/game/cards/resources.clj
+++ b/src/clj/game/cards/resources.clj
@@ -2644,20 +2644,8 @@
                 :silent (req true)}]
       :abilities [(ttw-ab "R&D" :rd)
                   (ttw-ab "HQ" :hq)
-                  {:label "Bounce HQ"
-                   :cost [:click 1]
-                   :req (req true)
-                   :effect (req (add-counter state side card :power 1)
-                                (swap! state update-in [:runner :register :unsuccessful-run] #(conj % :hq))
-                                (swap! state assoc-in [:run :unsuccessful] true)
-                                (system-msg state :runner (str "places a power counter on " (:title card))))}
-                  {:label "Bounce R&D"
-                   :cost [:click 1]
-                   :req (req true)
-                   :effect (req (add-counter state side card :power 1)
-                                (swap! state update-in [:runner :register :unsuccessful-run] #(conj % :rd))
-                                (swap! state assoc-in [:run :unsuccessful] true)
-                                (system-msg state :runner (str "places a power counter on " (:title card))))}]})
+                  (ttw-bounce "R&D" :rd)
+                  (ttw-bounce "HQ" :hq)]})
 
    "Theophilius Bagbiter"
    {:effect (req (lose-credits state :runner :all)
--- a/src/clj/game/cards/resources.clj
+++ b/src/clj/game/cards/resources.clj
@@ -2625,7 +2625,13 @@
             :req (req run)
             :msg (msg "access 1 additional card from " name " for the remainder of the run")
             :effect (req (access-bonus state side server 1))})
-          ]
+                  (ttw-bounce [name server]
+                   :label (str "Bounce " name)
+                   :cost [:click 1]
+                   :effect (req (add-counter state side card :power 1)
+                                (swap! state update-in [:runner :register :unsuccessful-run] #(conj % server))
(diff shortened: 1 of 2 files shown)


lostgeek:
While I love the fact, that you check ETR subroutines, I think for easy use we should remove any checks from that bounce button. There are cases like having enough Bad Publicity to break the outermost ice, not having installed programs and therefore being able to bounce on a Rototurret, etc..

For setting the correct flags, you can look at `core/runs.clj:1175`. I think the following two lines should be enough:
```
(swap! state update-in [:runner :register :unsuccessful-run] #(conj % server))
(swap! state assoc-in [:run :unsuccessful] true)
```

Kubik161:
My idea was to allow bounce only on ICE, that has simple ETR subs. Anything more complicated (bad pub to break, no punishment by subs etc) would still need to use Run action. 

But no problem, I can remove whole :req part and just let them add power counter by click. And let players handle all that stuff. I just wanted to prevent some new player or anyone else misunderstanding this as legal ability of Turning wheel.

lostgeek:
Yeah. And in general I love the idea. But in this case it is a shortcut and should only be used when players know and agree on what game actions it actually stands for.

Kubik161:
@lostgeek So, I have removed those restrictions :)

FarCryFromHuman:
I wonder if there's a simple but declarative way to mark "user override" type controls, maybe using purple text or something? That way users know they are opting into a potentially invalid game state as per the rules?

src/clj/game/cards/resources.clj
```diff
@@ -2619,12 +2619,13 @@
...
-                   :req (req run)
-                   :msg (msg "access 1 additional card from " m " for the remainder of the run")
-                   :effect (req (access-bonus state side s 1))})]
+   (letfn [(ttw-ab [name server]
+            {:label (str "Access an additional card in " name)
+            :cost [:power 2]
+            :req (req run)
+            :msg (msg "access 1 additional card from " name " for the remainder of the run")
+            :effect (req (access-bonus state side server 1))})
+          ]
```
lostgeek:
```suggestion
                  (ttw-bounce [name server]
                   :label (str "Bounce " name)
                   :cost [:click 1]
                   :effect (req (add-counter state side card :power 1)
                                (swap! state update-in [:runner :register :unsuccessful-run] #(conj % server))
                                (swap! state assoc-in [:run :unsuccessful] true)
                                (system-msg state :runner (str "places a power counter on " (:title card))))}]
```

You can define the entire thing as a function to avoid writing the same code twice below.

src/clj/game/cards/resources.clj
```diff
@@ -2636,7 +2637,21 @@
...
+                                (swap! state update-in [:runner :register :unsuccessful-run] #(conj % :hq))
+                                (swap! state assoc-in [:run :unsuccessful] true)
+                                (system-msg state :runner (str "places a power counter on " (:title card))))}
+                  {:label "Bounce R&D"
+                   :cost [:click 1]
+                   :req (req true)
+                   :effect (req (add-counter state side card :power 1)
+                                (swap! state update-in [:runner :register :unsuccessful-run] #(conj % :rd))
+                                (swap! state assoc-in [:run :unsuccessful] true)
+                                (system-msg state :runner (str "places a power counter on " (:title card))))}]})
```
lostgeek: You can replace this by `(ttw-bounce "R&D" :rd)` and `(ttw-bounce "HQ" :hq)`

NoahTheDuke:
I committed the final version using `--author="kubik <jakub.juroska@centrum.cz>"` so as to give you authorship and put it in another PR as merging this one was too difficult.

PR #4821, opened 2019-12-21
Implement Simulchip

NoahTheDuke:


src/clj/game/cards/hardware.clj
```diff
@@ -1678,6 +1678,31 @@
...
+                       :req (req (and (same-card? card target)
+                                      (let [pred #(and (runner? (first %))
+                                                       (installed? (first %))
+                                                       (program? (first %)))]
+                                        (zero? (+ (event-count state nil :runner-trash pred)
+                                                  (event-count state nil :corp-trash pred))))))
+                       :value [:program]}]
+   :abilities [{:async true
+                :label "Install a program from the heap"
+                :req (req (and (not (install-locked? state side))
```
lostgeek: Love how clean Simulchip is now. Will be even cleaner once you implement the bit you spoke about, that the ability only fires if there are any valid `:choices`.

I'd think then you'd also move `install-locked?` into the `:choices` req, right?

src/clj/game/core/ice.clj
```diff
@@ -388,8 +388,14 @@
 (defn- break-subroutines-pay
   [ice cost broken-subs args]
   (when (seq broken-subs)
-    {:msg (break-subroutines-msg ice broken-subs args)
-     :cost cost}))
+    {:effect
+     (effect
+       (continue-ability
+         (let [cost (icebreaker-ability-cost state side {:cost cost} card ice)]
+           (println cost)
```
lostgeek: :eyes:

src/clj/game/cards/hardware.clj
NoahTheDuke: Uh, maybe? Will probably want to change how we handle `install-locked?` in that case.

PR #4866, opened 2020-01-03
fixed doppelganger interaction with makers eye

### LINKED ISSUE #4864: Doppelgänger doubling The Maker's Eye effect
When the runner plays [The Maker's Eye](https://netrunnerdb.com/en/card/01036) while [Doppelgänger](https://netrunnerdb.com/en/card/20025) is on the field, if the run is successful they can choose to make another run with Doppelganger's effect, if they choose to make this free run on R&D the effect from The Maker's Eye triggers again if the run is successful, allowing them to access three cards from R&D twice.

Kubik161:
Closes #4864 

I have tested this fix in local browser. I wasn't able to write automatic test, seems like Makers eye not opening prompt in test case. Don't know why.

### DIFF
--- a/src/clj/game/cards/hardware.clj
+++ b/src/clj/game/cards/hardware.clj
@@ -488,6 +488,7 @@
                             :makes-run true
                             :effect (effect (update! (dissoc card :dopp-active))
                                             (clear-wait-prompt :corp)
+                                            (effect-completed (make-result eid {:successful true}))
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
This is not the right way to do this. I'm not sure what is, but this isn't it. In calling `effect-completed` early with the same `eid`, you're moving things out of order. I'll do some research and see what the underlying cause is.

PR #4870, opened 2020-01-03
Mind Control fix

### LINKED ISSUE #4847: Mind Game functions incorrectly
The run timings update did a number on **Mind Game**:

1. the option to manually jack out by clicking on the card itself has been removed, in favour of it occurring in the prompt window at the appropriate time.  This isn't a problem by itself, except for the next two issues.
2. When the subroutine resolves and the runner is deflected to another server, Jnet forces them to be at the encounter step.  This is what **AgInfusion**'s ability does, but **Mind Game** (and **Code Replicator**) are supposed to force the Runner to *approach* the ICE, which gives them a window to jack out before encountering it. I have been getting by this by letting the runner pass the ICE without firing subroutines, except there are major issues with **Anansi** and **Kakugo**, as one can imagine.
3. When the Runner finally has the option to jack out, Jnet forces them to choose a *program* to put on the bottom of the stack, while **Mind Game** should allow any installed card.

Kubik161:
Closes #4847 

Fix and new test case.

### DIFF
--- a/src/clj/game/cards/ice.clj
+++ b/src/clj/game/cards/ice.clj
@@ -2086,11 +2086,12 @@
                                          (swap! state update-in [:run]
                                                 #(assoc % :position (count (get-in corp (conj dest :ices)))
                                                         :server (rest dest)))
+                                         (set-phase state :approach-ice)
                                          (register-floating-effect
                                            state side card
                                            {:type :jack-out-additional-cost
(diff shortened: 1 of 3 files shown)


NoahTheDuke:
Looks good! Only change needed is to remove the unused references to `:add-program-to-bottom-of-deck`.

PR #4875, opened 2020-01-05
Fix The Maker's Eye

### LINKED ISSUE #4864: Doppelgänger doubling The Maker's Eye effect

NoahTheDuke:
A redo of #4866

Closes #4864

### DIFF
--- a/src/clj/game/cards/events.clj
+++ b/src/clj/game/cards/events.clj
@@ -2667,6 +2667,7 @@
    :async true
    :effect (effect (make-run eid :rd nil card))
    :events [{:event :successful-run
+             :unregister-once-resolved true
              :silent (req true)
              :req (req (= target :rd))
              :effect (effect (access-bonus :rd 2))}]})
--- a/test/clj/game/cards/hardware_test.clj
+++ b/test/clj/game/cards/hardware_test.clj
@@ -926,19 +926,49 @@
 
 (deftest doppelganger
   ;; Doppelgänger - run again when successful
-  (do-game
-    (new-game {:runner {:deck ["Doppelgänger"]}})
-    (core/gain state :corp :bad-publicity 1)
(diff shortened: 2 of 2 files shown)


src/clj/game/cards/events.clj
```diff
@@ -2667,6 +2667,7 @@
    :async true
    :effect (effect (make-run eid :rd nil card))
    :events [{:event :successful-run
+             :unregister-once-resolved true
```
lostgeek: How is the event triggering when Maker's Eye is in the discard? Shouldn't those cards be inactive?

NoahTheDuke: It’s active in the play area and trashed the effect finishes.

lostgeek: Sure, but shouldn't it be trashed as soon as the run ends and not be triggered on the second Dopplegänger run?

NoahTheDuke: No, cuz the Doppelganger run is created in 10.3.1 and the rest of the steps wait until Doppelganger's run is finished.

lostgeek: Ah, so Maker's Eye isn't in the trash during the Doppelgänger run, but it's effect is read as if it said "If _that run_ is successful, access 2 additional cards from R&D". Correct?

NoahTheDuke: Yep!

PR #4878, opened 2020-01-05
Fix Dedicated Neural Net when accessing a single card

### LINKED ISSUE #4874: Dedicated Neural Net not working
Psi starts. Win or lose, runner gets normal random access.

NoahTheDuke:
Closes #4874


Kubik161:
thanks!

PR #4883, opened 2020-01-06
Process all normal prompts with uuid

NoahTheDuke:
Follow-up to ideas in #4881


test/clj/game/cards/agendas_test.clj
```diff
@@ -146,6 +146,7 @@
    (play-and-score state "Architect Deployment Test")
    (click-prompt state :corp "OK")
    (click-prompt state :corp "Cancel")
+   (println (map :text (:log @state)))
```
lostgeek: :eyes:

PR #4891, opened 2020-01-09
Gachapon install ability fix

### LINKED ISSUE #4888: Gachapon + Credits for Installs
If you try to install a card from gachapon, but can only afford it with hosted credits (i.e. Paladin Poemu), the card doesn't install and is instead shuffled back in.

Kubik161:
Closes #4888 

Paladin Poemu cradits checks eid for source-type `runner-install`. Gachapon install ability had source-type `ability`, so `all-active-pay-credit-cards` ignored Paladin. I have changed source-type for Gachapon install ability from `ability` to `runner-install` and it seems to work. Did I break anything else with this change?

Fix and  new test case.


NoahTheDuke:
Looks good to me! I think this shows an issue with our `eid` system, but we shouldn’t worry about it for now.

PR #4892, opened 2020-01-09
new SMC filter to show only programs which can be installed

### LINKED ISSUE #4863: pop smc to install afterimage with only 5 credits makes the afterimage disappear.
this isn't a valid move, but the game apparently hid the afterimage card so my opponent couldn't play it.

Board state:

runner has SMC down and 5 credits.

Clicks SMC to go fetch Afterimage.  Can't actually pay for it.  Card ..,. disappears?  We had to unwind.

Kubik161:
Closes #4863 

Couldn't reproduce original issue. Afterimage stays in stack when there is not enough credits to pay. 

I have added new filter for SMC to only show programs which can by installed. 
Also, SMC has both `cancel` and `No install` options in prompt. Is this correct? Shouldn't there be only one?

### DIFF
--- a/src/clj/game/cards/programs.clj
+++ b/src/clj/game/cards/programs.clj
@@ -2231,8 +2231,11 @@
                                                (str "shuffle their Stack")))
                                    :priority true
                                    :choices (req (cancellable
-                                                   (conj (vec (sort-by :title (filter program?
-                                                                                      (:deck runner))))
+                                                   (conj (filter #(can-pay? state side
+                                                                      (assoc eid :source card :source-type :runner-install) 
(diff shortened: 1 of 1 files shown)


NoahTheDuke:
The "Cancel" is for accidentally clicking SMC and deciding you don't want to use it, but I think that's a bad idea, I agree. So I'm fine with you removing `cancellable` in this PR.

Kubik161:
> The "Cancel" is for accidentally clicking SMC and deciding you don't want to use it, but I think that's a bad idea, I agree. So I'm fine with you removing `cancellable` in this PR.

Done :)

PR #4937, opened 2020-01-14
Reverse Infection log fix

### LINKED ISSUE #4861: Reverse Infection print error with two Cyberdex Sandboxes scored
image

There were 2 virus counters in play on a Sandstone at the time. I think it's saying 22 and 00 because of the two scored Cyberdex agendas?

Kubik161:
Closes #4861 

Fix and new assertion fot correct log.

### DIFF
--- a/test/clj/game/cards/operations_test.clj
+++ b/test/clj/game/cards/operations_test.clj
@@ -2829,25 +2829,26 @@
       (click-prompt state :corp "Gain 2 [Credits]")
       (is (= 7 (:credit (get-corp))) "Corp gained 2 credits")))
   (testing "Purge virus counters, counts both Runner and Corp side (Sandstone)")
-  (do-game
-    (new-game {:corp {:deck ["Reverse Infection" "Sandstone"]}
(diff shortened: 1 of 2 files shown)


test/clj/game/cards/operations_test.clj
```diff
@@ -2829,25 +2829,26 @@
...
+      (play-from-hand state :runner "Virus Breeding Ground")
+      (play-from-hand state :runner "Datasucker")
+      (take-credits state :runner)
+      (core/add-counter state :corp (get-ice state :hq 0) :virus 3)
+      (core/add-counter state :runner (get-resource state 0) :virus 3)
+      (core/add-counter state :runner (get-program state 0) :virus 3)
+      (play-from-hand state :corp "Reverse Infection")
+      (click-prompt state :corp "Purge virus counters")
+      (is (last-log-contains? state "Corp uses Reverse Infection to purge 9 virus counters and trash 3 cards from the top of the stack.") "Should write correct log")
+      (is (= 7 (:credit (get-corp))) "Corp did not gain credits")
```
NoahTheDuke: Can you remove this and the later lines? This test is only for checking the log, so we shouldn't recheck all of the previous test's assertions

PR #4938, opened 2020-01-14
Close prompt when user drags card to heap/archives

### LINKED ISSUE #587: Street Peddler can duplicate cards

Kubik161:
Closes #587

Fix is taken from issue #2952 and modified to be used when user drags card to Heap/Archives.

### DIFF
--- a/src/clj/game/core/actions.clj
+++ b/src/clj/game/core/actions.clj
@@ -156,7 +156,7 @@
                (or (= last-zone :play-area)
                    (same-side? side (:side card))))
       (let [move-card-to (partial move state s c)
-            card-prompts (filter #(= (get-in % [:card :title]) (:title c) ) (get-in @state [side :prompt]))
+            card-prompts (filter #(same-card? :title % c) (get-in @state [side :prompt]))
             log-move (fn [verb & text]
                        (system-msg state side (str verb " " label from-str
                                                    (when (seq text)
--- a/src/clj/game/core/actions.clj
+++ b/src/clj/game/core/actions.clj
@@ -156,6 +156,7 @@
                (or (= last-zone :play-area)
                    (same-side? side (:side card))))
       (let [move-card-to (partial move state s c)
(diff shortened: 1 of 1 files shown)


NoahTheDuke:
My question is: why are players dragging a card to the trash when there's a prompt open? What is happening?

Kubik161:
I don't know :) It's super old bug from 2015. We could just close that issue, if it wasn't reported since. But I don't like that dragging with open prompt can duplicate cards.

src/clj/game/core/actions.clj
```diff
@@ -156,18 +156,23 @@
                (or (= last-zone :play-area)
                    (same-side? side (:side card))))
       (let [move-card-to (partial move state s c)
+            card-prompts (filter #(= (get-in % [:card :title]) (:title c) ) (get-in @state [side :prompt]))
```
NoahTheDuke:
```suggestion
            card-prompts (filter #(same-card? :title % c) (get-in @state [side :prompt]))
```

NoahTheDuke:
Yeah, not duplicating cards is very good, I just worry about breaking state in other subtle ways because of this.

Kubik161:
Sure, it's up to you. If you think that there is possibility of introducing new bugs with this commit, then we won't merge it :) Would there be any safer way to fix this? Or we can leave it as it is, it doesn't seem to be frequent problem.

src/clj/game/core/actions.clj
NoahTheDuke: lmao ah shit, i hecked this up. i shouldn't write suggestions before i've had my coffee. don't fix it cuz i'm still undecided on whether the PR is a good idea, but it should be `(same-card? :title (:card %) c)`

NoahTheDuke:
Fuck it, let's find out what breaks.

PR #4944, opened 2020-01-15
Prevent jack out fix

Kubik161:
I have discovered that Whirlpool subs to prevent jack out has wrong implementation. And that An Offer You Can't Refuse doesn't prevent jack out at all. So, there are fixes and tests for that.


NoahTheDuke:
Nicely caught!

Kubik161:
... and fixes for jack out after Inazuma and Susanoo-no-Mikato.

PR #4946, opened 2020-01-15
Guinea pig interaction with buffer drive fixed

### LINKED ISSUE #4942: Guinea pig + Buffer Drive
Tried to use Guinea Pig + Buffer Drive when I had several cards in hand.  It trashes all cards in hand alright, but Buffer Drive only offers me to return the first card from the grip, not whichever I want.

Kubik161:
Closes #4942


NoahTheDuke:
For this, you'll want to `(wait-for (trash-cards ...) (gain 10 credits) (effect-completed state side eid))`

Kubik161:
it doesn't work with wait-for, seems to be caught in some infinite loop.

Issue #4971, opened 2020-01-28
Several runtiming issues


lostgeek:
- [x] Make "Rez ice" automatically hit continue for you
- [x] Split "Pass ice" button into two options for "passing and jacking out" and "passing and continuing"
- [x] Add better log message to TTW along the lines of "Runner spends :click_nisei: to bounce off of R&D for a Turning Wheel token (shortcut)."
- [x] Remove/delay "Runner has no further actions." from the log to prevent giving away information to the Corp
- [x] Add auto-continue for the Corp
- [x] Do something with the run arrow to separately indicate approach and encounter

PR #4979, opened 2020-01-31
Hide continue message in certain cases to prevent giving away information

lostgeek:
new-flow

1. On approaching ice the Runner can fix a misclick by undoing the click. Otherwise only passing priority is possible
2. On encountering ice they can decide to either "pass and jack out" at step 2d of the timing chart or "pass and continue"
3. This decision can be changed as long as the Corp has not pressed the `no-action` button yet
4. On approaching the server, the Corp has to act first
5. Now the runner can jack out or have a successful run

Relates to #4971


NoahTheDuke:
Can you write a test for these?

lostgeek:
I've thought about this more and realized it is more complicated than I initially thought.

This whole change was sparked by paulyg on Slack, who pointed out that in some cases, pressing `continue` as the Runner can give away information. For example when you have a Chisel on Street Peddler and an installed Devil Charm and are approaching an unrezzed piece of ice. In this case, if the Corp rezzes the ice, you will want to use the PAW to install Chisel _before_ you go into the encounter step. So you do not press "No further actions" and thereby reveal that you have a Chisel on the Street Peddler.

So far so good. Now the problem is though that on jnet we're merging together Steps 2b to 2e of the [run-timing chart](https://i.imgur.com/kSVfhhs.png). By clicking `continue` as the Runner, we're both indicating that we are not jacking out in 2d, but also giving up our PAW in 2e.

A solution would be to handle jacking out differently. For example, we could split up the "pass ice" button into "pass ice and jack out" and "pass ice and continue" as kubik161 suggested. In this case, we have the two PAWs in 2b and 2e left to consider.

2b can be relevant for the Stimhack vs unrezzed Border Control interaction, where the Corp can rez and use the BC while they have priority. But this case wasn't covered by our previous implementation either and I can't find a solution where we cover it without introducing further clicks on both sides.

My proposed solution would be:
1. Split up "Pass ice" into "Pass ice and jack out" / "Pass ice and continue"
2. Remove the log message from the Runner pressing `continue` on approaching ice
3. Print the log message from the Corp pressing `no-action` on approaching ice
4. Keep the log message from the Runner pressing `continue` on encountering ice
5. Remove the log message from the Corp pressing `no-action` on encountering ice

NoahTheDuke:
Love your thought process. Thanks for tackling this!

src/clj/game/core/runs.clj
```diff
@@ -257,6 +261,8 @@
               (update-all-icebreakers state side)
               (reset-all-ice state side)
               (cond
+                (get-in @state [:run :jack-out-after-pass])
+                (jack-out state :runner (make-eid state))
```
NoahTheDuke: Don't forget to move this to approach

PR #4981, opened 2020-01-31
Change color of run-arrow depending on run phase

lostgeek:
arrow
From left to right:
- Approaching Wraparound
- Encountering Wraparound
- Approaching Enigma
- Encountering Enigma

Relates to #4971


NoahTheDuke:
Love the idea but I don’t think making the arrow lighter is the most accessible method. I suspect players will find it hard to see, especially those with vision problems. Can we change the color to another primary color?

lostgeek:
image
How is this for readability?

NoahTheDuke:
That looks sick, yeah. Thank you!

PR #4986, opened 2020-02-02
Saisentan fix

### LINKED ISSUE #4982: Saisentan not firing all 3 subs when one of them hits an extra card.
Runner ran into Saisentan with no sentry breaker and 3 cards in hand (2 hardware, 1 resource).
Named hardware and pressed fire all subs
First sub fired hit Hardware and did a second damage (another hardware)
Second sub fired and hit Resource.
Third sub did not fire.

saisentan not firing 3 times

lostgeek:
Closes #4982


test/clj/game/cards/ice_test.clj
```diff
@@ -2883,8 +2883,8 @@
       (let [sai (get-ice state :hq 0)]
         (core/rez state :corp sai)
         (run-continue state)
-        (card-ability state :corp sai 0)
         (click-prompt state :corp "Event")
+        (card-ability state :corp sai 0)
```
NoahTheDuke: These should be removed, right? It's now an `:on-encounter` ability.

Issue #5063, opened 2020-04-28
Subroutines on Yagura are in the wrong order


kalotay:
The order should be peek R&D and optionally move to bottom then net damage. The current implementation has net damage first.

NoahTheDuke:
Thanks for the report!

PR #5069, opened 2020-05-02
Yagura: fix subroutine order

zdmitche:
This fixes the subroutine order of [`Yagura`](http://www.netrunnerdb.com/en/card/04093). It should be:
1. Look at the top card of R&D. You may add that card to the bottom of R&D.
2. Do 1 net damage.

However, 1 and 2 were switched.

Closes https://github.com/mtgred/netrunner/issues/5063


NoahTheDuke:
Oh hell yeah! Thank you so much!

PR #5076, opened 2020-05-03
Auto continue fix

lostgeek:
There was a possible infinite loop in the `continue` + `no-action` logic. This PR removes the `no-action` function and includes its functionality in `continue`.

`continue` now checks, whether both players have pressed it and then progresses to the next phase.


NoahTheDuke:
Hell yeah, this is a great change

PR #5089, opened 2020-05-06
specify icon when creating web app of jinteki.net on iOS

mtwebb:
This specifies the icon used when creating webapp version of jinteki.net (see the old icon on the left and the new icon on the right in the attached image). 

changes are: 
- add icons for different ios devices 
- set metadata to use the one of the icons (the one I could test on device). 

Details can be found here. 
https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/ConfiguringWebApplications/ConfiguringWebApplications.html

jinteki_webapp


NoahTheDuke:
This looks great! Thank you so much!

PR #5100, opened 2020-05-11
Heap-breakers now work on ice with multiple subtypes

### LINKED ISSUE #5099: Orion and conspiracy breakers
Hitting an Orion with multiple conspiracy breakers in the bin, the runner is only prompted to install one of them (not sure what affects which one) rather than all.

lostgeek:
Heap breakers were checking for a single `:conspiracy` key, instead of using their own. This lead to problems with install triggers on multi-type ice.

Closes #5099


NoahTheDuke:
Thank you! You can use :discard in the runner map instead of putting all of the cards in hand and then calling (trash-from-hand)

PR #5108, opened 2020-05-13
Fix for any number of args for /summon command

### LINKED ISSUE #5107: /summon command does not work for three-word-or-more card titles.
The summon command seems unable to summon a few cards, and my best guess as to the common theme is that those cards have titles that are three words or more. 

netrunner summon

Not pictured, I also tested it with The Future is Now (four words) which did not work either.

Incidentally, while /summon isn't a very important command for gameplay, having this command work would be greatly helpful for a tournament I would like to run, and would likely be helpful for playtesting purposes, though of course I have no idea if this is an easy or difficult fix. So if it is fixable, it would be quite appreciated!

Kubik161:
Closes #5107 

I have removed maximum number of args (2) for chat commands. Only commands using args are `/summon` and `/counter` . Both seems to work correctly with any number of args.

### DIFF
--- a/src/clj/game/core/io.clj
+++ b/src/clj/game/core/io.clj
@@ -310,122 +310,121 @@
   (let [[command & args] (split text #" ")
         value (if-let [n (string->num (first args))] n 1)
         num   (if-let [n (-> args first (safe-split #"#") second string->num)] (dec n) 0)]
-    (when (<= (count args) 2)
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
Thank you! Wanna add a test to the commands test file?

Kubik161:
Added some tests :)

PR #5116, opened 2020-05-18
Proposal for panel changes

lostgeek:
image

This is a proposal for adding more information to the ice panel, which has been unused since the run changes added rezzing and firing subroutines to the prompt menu.


lostgeek:
This will get included in #5299 in some form.

Issue #5097, opened 2020-05-09
White Hat doesn't reveal cards in chat log


JessicaEArmstrong:
Hey!

So White Hate 'reveals' cards, and then shows the cards shuffled back in the chat, but doesn't show the cards initially revealed when this should be open information. 

The runner still sees them, and the corp can obviously see HQ all the time, but it means spectators can't, and if the players need to refer back, they are missing!

Thanks :)

NoahTheDuke:
Thanks for the report!

PR #5117, opened 2020-05-19
added revealed cards to message from white hat

bloyot:
Resolves: https://github.com/mtgred/netrunner/issues/5097

Resulting message looks like: 
WhiteHatSnippet

If you prefer a different message format I'm happy to change it.

The `when-let` is just to handle edge case where the corp has no cards.

### DIFF
--- a/src/clj/game/cards/events.clj
+++ b/src/clj/game/cards/events.clj
@@ -2904,7 +2904,7 @@
      :trace {:base 3
              :unsuccessful
              {:async true
-              :msg (msg "reveal all cards in HQ" (when-let [hand (seq (get-in @state [:corp :hand]))]
+              :msg (msg "reveal all cards in HQ" (when-let [hand (seq (:hand corp))]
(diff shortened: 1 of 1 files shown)


NoahTheDuke:
Thank you for this!

src/clj/game/cards/events.clj
```diff
@@ -2904,7 +2904,8 @@
      :trace {:base 3
              :unsuccessful
              {:async true
-              :msg "reveal all cards in HQ"
+              :msg (msg "reveal all cards in HQ" (when-let [hand (seq (get-in @state [:corp :hand]))]
```
NoahTheDuke, line 2907:
```suggestion
              :msg (msg "reveal all cards in HQ" (when-let [hand (seq (:hand corp))]
```

bloyot: Ah didn't realize `msg` also bound `corp` and `runner` for you, perfect.

PR #5120, opened 2020-05-19
Boomerang shuffle fix (multiple boomerangs used in single run)

### LINKED ISSUE #5112: Only one boomerang shuffles after successful run
I have used two boomerangs during run, run was successful, only one got shuffled. 

Expected behavior is both being shuffled back.

I assume, that our current implementation: 
```
   ... after boomerang is used...

(register-events
    ;; Boomerang is trashed at this point
    (find-card "Boomerang" (:discard (:runner @state)))

    ... etc ...
```
is selecting same card from discard every time. We should probably somehow track every copy of boomerang separately?

Kubik161:
Closes #5112 

Fix and new test.

I was thinking for few days. My "hack" solution is to `find-card` in reverse order. This way, we will always find latest trashed boomerang and register event on correct card. Should be good enough until we find a way to track cards entering discard?

Also my test seems to work, but I'm getting complaint about `click-card should only be used with prompts requiring the user to click on cards on table`. Is there any obvious mistake? Thanks for review.

### DIFF
--- a/src/clj/game/cards/hardware.clj
+++ b/src/clj/game/cards/hardware.clj
@@ -183,7 +183,7 @@
                   {:effect (effect
                              (register-events
                                ;; Boomerang is trashed at this point
-                               (reverse-find-card "Boomerang" (:discard (:runner @state)))
+                               (find-card "Boomerang" (reverse (:discard (:runner @state))))
                                (let [server (:server run)]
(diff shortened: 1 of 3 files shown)


Kubik161:
test has been fixed :)

Issue #5095, opened 2020-05-08
Negotiator subroutines can be broken with clicks instead of credits


kalotay:
The runner should be able to use 2 credits to break a sub on negotiator (like F2p): https://netrunnerdb.com/en/card/08019

The current implementation let's the runner use clicks instead.

NoahTheDuke:
Thanks for the report!

PR #5121, opened 2020-05-20
fixing negotiator sub break cost from clicks to credits

bloyot:
Resolves https://github.com/mtgred/netrunner/issues/5095


NoahTheDuke:
Wanna add a test for this card?

bloyot:
Sure, can get to it tomorrow probably, should be straightforward.

bloyot:
Added test cases.

Issue #5090, opened 2020-05-06
Red Planet Couriers doesn`t clear info correctly


Dreadnote:
Red Planet Couriers doesn`t update subs on Masvingo correctly. so when it used, but no new advancement tokens still placed - Masvingo shows incorrect number of subs. (In example 1 token got moved, but Masvingo still have 1 ETR subs. And its relevant to even 11-token-Masvingos)
Maybe this is bug with Masvingo itself.

NoahTheDuke:
Thanks for the report!

PR #5124, opened 2020-05-21
added update-ice-advancement event triggered by red planet couriers

bloyot:
Resolves https://github.com/mtgred/netrunner/issues/5090

I also considered putting this logic inside `update-all-ice` similar to `update-ice-strength` but wasn't sure if that made sense so opted for minimal impact.


NoahTheDuke:
This is a good idea, but it's not quite the way we want to do it. I bet we'll want to add some sort of advancement updating, like we do `update-all-ice`, but that should be properly handled instead of in a one-off like this.

bloyot:
Updated per our discussion, and added test. Thanks!

NoahTheDuke:
This looks great. Nicely done!

PR #5138, opened 2020-05-26
Clean up :zone messiness

### LINKED ISSUE #4571: Daily Quest broken with Full Immersion Rec Studio
DQ hosted on Rec Studio doesn't give runner creds on successful run

NoahTheDuke:
In trying to tackle #4571, I realized that we're not doing a good job of consistently checking for hosting. This is an effort to fix that: Reduce the use of `:zone` to the bare minimum, instead routing all checks through the renamed `get-zone` (`get-nested-zone` previously). There are still a lot of places we could be doing better (all of the `(is-remote? (second (get-zone %)))` and similar lines should be a single function call), but I think this is a good first step.

Closes #4571

### DIFF
--- a/src/clj/game/cards/agendas.clj
+++ b/src/clj/game/cards/agendas.clj
@@ -300,8 +300,7 @@
                             :choices {:card #(and (agenda? %)
                                                   (not= (:title %)
                                                         "Bifrost Array")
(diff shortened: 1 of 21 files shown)


lostgeek:
Oooh. I love this! :+1:

PR #5140, opened 2020-05-26
Corrected spend-msg of Paule's Cafe

### LINKED ISSUE #5126: Paule's Café ability cost is not logged
I saw this issue during a Stimhack run (if that matters), where the runner installed 3 programs, the log said (line after line) the cost of each program being installed (first discounted), but not that it cost them 1 credit to use each time.

But, it does appear that the cost is taken out of the Runner's credit pool correctly.

lostgeek:
Closes #5126

### DIFF
--- a/src/clj/game/cards/resources.clj
+++ b/src/clj/game/cards/resources.clj
@@ -1889,13 +1889,10 @@
                                               (:uniqueness %)))
                                 count
                                 -)
-                           0))
-                       (edit-spend-msg [msg] ; adds 1 to the credit cost message
-                         (let [credit-msg (re-find #"\d+ \[Credits\]" msg)
-                               cost (Integer/parseInt (first (split credit-msg #" ")))]
-                           (clojure.string/replace msg (re-pattern (str cost #" \[Credits\]")) (str (inc cost) " [Credits]"))))]
+                           0))]
                  {:async true
                   :label "Install hosted card"
+                  :msg "install hosted card"
                   :cost [:credit 1]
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
I don't think this is the best way. It's inconsistent with other abilities and isn't accurate, because you're not paying extra to install Desperado, you're paying 1 to use the ability and then the normal (or reduced) amount to install Desperado when resolving the ability. I'd be more inclined to use something like `:msg (msg "install " (:title target))`. This is more verbose, but is also clearer about exactly what's happening.

If done the way you have now, if there's a card ability that cares about how much you've spent to install something, it will be off by one when using Paule's. You'll expect it to fire with X = 4, but it'll fire with X = 3 because the install is actually 3, we've just changed what is printed to say 4.

lostgeek:
Good point. But your suggestion would always print "Runner pays 1 [Credits] to install Desperado", which is even more misleading in my opinion. How would you feel about printing "1+3 [Credits]" instead?

NoahTheDuke:
It would say “Runner pays 1 Credit to use Paule’s Cafe to install Desperado. Runner pays X Credits to install Desperado from Paule’s Cafe, with some discount.”

lostgeek:
That's a good idea. I think this was my aversion against having a wordy log, but splitting this up works great. Now it is:

```Runner pays 1 [Credits] to use Paule's Café to install hosted card.
Runner pays 2 [Credits] to install Corroder using Paule's Café.
```

PR #5142, opened 2020-05-26
/unique command to toggle uniqueness of a card

lostgeek:
This PR adds a `/unique` command, that toggles the uniqueness of a card. It can be used to play games with non-errata'd versions of cards like Wireless Net Pavillion.


NoahTheDuke:
Wanna add a small test for this?

NoahTheDuke:
eh, we can get a test another time

PR #5149, opened 2020-06-02
Restore fix for unseen agenda. New tests.

Kubik161:
Restore should show which card is installed. Asset/Upgrade/etc will be rezzed, so there is no problem. Agenda cannot be rezzed, so we should print card title in chat log.

Added tests for Restore.

### DIFF
--- a/src/clj/game/cards/operations.clj
+++ b/src/clj/game/cards/operations.clj
@@ -1683,7 +1683,7 @@
                                             (in-discard? %))}
                       :effect (req (wait-for
                                      (corp-install state side target nil {:install-state :rezzed})
-                                     (let [seen (assoc-in target [:seen] true)]
+                                     (let [seen (assoc target :seen true)]
                                         (system-msg state side (str "uses Restore to "
                                                                       (corp-install-msg seen)))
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
Thanks for this! A couple things:

1) you don’t need to use `assoc-in` when there’s only one item in he vector. Instead use `assoc`.

2) `:seen` is used by the front end to make cards face up, and we don’t want to do that with agendas, because only Public agendas are face up when installed. There should be an option you can pass into Corp install message that will force print the name, but I don’t remember it.

3) the agenda test should check that the card is still facedown

Kubik161:
I have fixed problems 1 and 3. 

About 2: I'm setting `:seen` only for purpose of printing chat message. Install is done with unmodified version of `target`. I hope that I understand what `let` keyword does, but maybe not? :)

I haven't been able to find any way to force `corp-install-msg` to use title. I see two options now: 
a) I can hardcode chat log print into `Restore` and remove `corp-install-msg` call
b) I can modify `corp-install-msg` to accept boolean to force title print

Which would be better?

NoahTheDuke:
Ah, you’re right! My fault for looking at this on my phone. Looks good to me

Issue #5167, opened 2020-06-19
Personal workshop not automatically installing card with 0 power counters


CarlHawley:
repro:
host a card with >0 cost on Personal Workshop.
start of turn, remove counter from card.
if going from 1 to 0 counters the card doesn't auto install, but if going from 0 to -1 it will.
workaround:
can also use pay X ability to pay 0 and install.

NoahTheDuke:
Thanks for the report!

PR #5176, opened 2020-06-26
fixed issue with personal workshop not installing at zero counters

bloyot:
Resolves https://github.com/mtgred/netrunner/issues/5167 and also adds tests for personal workshop. 

I believe the issue was the code is first checking if counters == 0 to install, then afterwards removing the counter. I simply flipped this logic, since we always want to remove the counter (the abilities actual effect) then conditionally install the hosted card. 

Separately I think the `:once :per-turn` is also a bug, though one that probably doesn't come up much. It won't let you click the spend 1 credit prompt more than once, and also won't let you use personal workshop on two different cards (if you wanted to for some reason). I removed it, but if there's some reason I'm not thinking of to have once per turn on there let me know!

### DIFF
--- a/src/clj/game/cards/resources.clj
+++ b/src/clj/game/cards/resources.clj
@@ -1928,13 +1928,12 @@
   (let [remove-counter
         {:async true
          :req (req (not (empty? (:hosted card))))
-         :once :per-turn
          :msg (msg "remove 1 counter from " (:title target))
          :choices {:card #(:host %)}
-         :effect (req (if (not (pos? (get-counters (get-card state target) :power)))
-                        (runner-install state side eid (dissoc target :counter) {:ignore-all-cost true})
-                        (do (add-counter state side target :power -1)
-                            (effect-completed state side eid))))}]
+         :effect (req (do (add-counter state side target :power -1)
+                          (if (not (pos? (get-counters (get-card state target) :power)))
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/resources.clj
```diff
@@ -1928,13 +1928,12 @@
...
-         :once :per-turn
          :msg (msg "remove 1 counter from " (:title target))
          :choices {:card #(:host %)}
-         :effect (req (if (not (pos? (get-counters (get-card state target) :power)))
-                        (runner-install state side eid (dissoc target :counter) {:ignore-all-cost true})
-                        (do (add-counter state side target :power -1)
-                            (effect-completed state side eid))))}]
+         :effect (req (do (add-counter state side target :power -1)
+                          (if (not (pos? (get-counters (get-card state target) :power)))
+                            (runner-install state side eid (dissoc target :counter) {:ignore-all-cost true}))
```
NoahTheDuke, line 1935: Looks like the final paren here needs to be after the next line.

NoahTheDuke:
This looks great! Would you mind removing the "(start of turn)" part of the ability label, like 1935? Or actually, add another ability that is "remove 1 token (start of turn)" that doesn't cost a credit. Not quite ideal, but it's how we do start of turn stuff rn.

bloyot:
Removed just the start of turn part for now, per our discussion in slack. Thanks!

PR #5182, opened 2020-07-03
Make q loop run ability async to fix masterwork interaction

### LINKED ISSUE #4973: Masterwork + Q-Loop Ordering Doesn't Work
In the event that I want to make a run and trigger Q-Loop before masterwork, I get the sequence of prompts below (order is from top to bottom), which forces me to make a masterwork decision before I see the top of my deck.

image

mikeprosser:
Fixes #4973 

When running with a Masterwork (v37) and a Prognostic Q-Loop the two triggers were previously not ordering properly, selecting Q-Loop to fire first would then offer the Masterwork install, then show you the top two cards of the deck (which were now incorrect if the hardware installed with Masterwork had triggered it's other ability and drawn a card). Now both abilities are async and can be resolved correctly in either order.

Demo: https://www.youtube.com/watch?v=X5YwxTmqMf4

### DIFF
--- a/src/clj/game/cards/hardware.clj
+++ b/src/clj/game/cards/hardware.clj
@@ -1405,10 +1405,12 @@
                         :prompt "Look at top 2 cards of the stack?"
                         :player :runner
                         :autoresolve (get-autoresolve :auto-fire)
-                        :yes-ability {:msg "look at the top 2 cards of the stack"
+                        :yes-ability {:async true
+                                      :msg "look at the top 2 cards of the stack"
                                       :effect (effect (prompt! card (str "The top two cards of your Stack are "
                                                                          (join ", " (map :title (take 2 (:deck runner))))
-                                                                         ".") ["OK"] {}))}}}]
+                                                                         ".") ["OK"] {}
+                                                                         (effect-completed state side eid)))}}}]
    :abilities [(set-autoresolve :auto-fire "Prognostic Q-Loop")
                {:label "Reveal and install top card of stack"
                 :once :per-turn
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
Thank you! Wanna write a small test for this?

mikeprosser:
@NoahTheDuke Sure thing, I think the two I've just added cover both parts of this (q-loop firing first and q-loop firing second and showing the right cards if masterwork draws first)

NoahTheDuke:
Looking at this again, I think it could be better handled by changing the whole `:yes-ability` to be a prompt: `:yes-ability {:msg "blah blah" :prompt (msg "The top two cards are " (join ", " (map :title (take 2 (:deck runner))))) :choices ["OK"]}` Ability maps don't need an `:effect` entry, and we shouldn't be using `prompt!` in abilities as it's cumbersome and prone to misunderstanding. (I should change it to be private to avoid these issues, but that's a lot of work, lol).

To explain my suggestion: the `:msg` will happen on use, `:prompt` is the text that's shown to the player above the buttons, `:choices` is a vector of buttons (they can be strings or card objects), and the macro `msg` is like `req` except that it calls `str` on all expressions inside (which as a reminder converts to a string and concats all sub expressions). `:prompt` and `:choices` are considered "async by default", meaning that you only need to include `:async true` if the `:effect` will be async.

NoahTheDuke:
Hey Mike, wanna clean this up a little? I'd love to merge it.

mikeprosser:
Hey @NoahTheDuke 

Apologies for the delay on this, I think the change I just made matches up with your suggestion and is definitely a lot tidier / more readable for the same effect. Cheers for the help :)

PR #5185, opened 2020-07-05
Fairchild 3 etr async

### LINKED ISSUE #5184: Fairchild 3 end the run sub doesn't always immediately do so
Clicking the "fire all unbroken subs" button on Fairchild 3 presents the corp with the "do 1 brain damage or end the run" subroutine as soon as the runner has selected an option for the second sub. If the runner chooses to trash an installed card, and the corp then chooses to end the run while the runner is deciding which installed card to trash the end the run is logged to chat but doesn't actually do anything until both players then click no more action.

Pretty minor as it does eventually end the run but a bit confusing about what's actually happening, I've seen a few people manually re-fire the sub rather than click no more action because of the ambiguity

https://youtu.be/Vbgq-wigaDw

mikeprosser:
Fixes #5184 

I've set the trash-or-pay subs on Fairchild 1.0, 2.0 and 3.0 to be async so the runner can fully resolve their trash effects before the corp gets prompted to either attempt to end the run or brain damage them if resolving all, and the runner can resolve the first trash before deciding what to do with the second sub.

I've also set up basic tests for each of the 3 pieces of ice which ensure that their main behaviour is working properly but I can't think of / see on another test for a piece of ice a simple way to replicate the exact "Resolve all unbroken subroutines" behaviour as it works in game which would be necessary here (to check, for example, that if you hit the button on F3 then selected brain damage that the brain damage itself doesn't land until the runner has chosen their second card to trash.

### DIFF
--- a/src/clj/game/cards/ice.clj
+++ b/src/clj/game/cards/ice.clj
@@ -1183,9 +1183,11 @@
              :player :runner
              :prompt "Choose one"
              :choices ["Pay 1 [Credits]" "Trash an installed card"]
+             :async true
              :effect (req (if (= target "Pay 1 [Credits]")
                             (do (pay state side card :credit 1)
-                                (system-msg state side "pays 1 [Credits]"))
+                                (system-msg state side "pays 1 [Credits]")
+                                (effect-completed state side eid))
                             (continue-ability state :runner runner-trash-installed-sub card nil)))}]
     {:subroutines [sub
                    sub]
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
Great work! I think a slightly better solution than calling effect-completed manually is to move the system-msg to before the pay call, and the call pay-sync instead, passing the eid in. That way, any async effects from the payment will be awaited properly.

PR #5220, opened 2020-07-28
Wall to wall prompt fix

### LINKED ISSUE #5217: Wall to Wall options don't clear when selected
previously what one of the three options on Wall to Wall available at start of Corp turn is selected it disappeared, Now each option remains even after being selected. Dont have a screen shot available.

Kubik161:
Closes #5217


NoahTheDuke:
Ope, forgot I took care of this in #5216.

PR #5223, opened 2020-07-29
Prevent successful run fix

### LINKED ISSUE #5092: No Alice Merchant trigger on second run on Archives when first was fizzled by Crisium
Steps:
- Have a Crisium on Archives
- Alice runs Archives and trashes the Crisum
- Alice runs Archives again
- Alice ability should fire, but does not

(In the example screenshot here, clicks 1 and 2 were install DDOS and running HQ)
Screenshot 2020-05-06 at 14 45 45

Kubik161:
Closes #5092 

Fixed both cards that prevent successful run (Crisium Grid, Transport Monopoly). 
`The first time you make a successful run` type abilities now works correctly.

Added new effect type `:block-successful-run`

New test for Crisium Grid.

### DIFF
--- a/src/clj/game/core/runs.clj
+++ b/src/clj/game/core/runs.clj
@@ -4,7 +4,7 @@
          update-ice-in-server update-all-ice get-agenda-points get-remote-names
          card-name can-access-loud can-steal?  prevent-jack-out card-flag? can-run?
          update-all-agenda-points reset-all-ice no-action make-run encounter-ends
-         pass-ice do-access successful-run complete-run)
+         pass-ice do-access successful-run)
 
 (defn add-run-effect
   [state side run-effect]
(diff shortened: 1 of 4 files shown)


src/clj/game/core/runs.clj
```diff
@@ -4,7 +4,7 @@
          update-ice-in-server update-all-ice get-agenda-points get-remote-names
          card-name can-access-loud can-steal?  prevent-jack-out card-flag? can-run?
          update-all-agenda-points reset-all-ice no-action make-run encounter-ends
-         pass-ice do-access successful-run)
+         pass-ice do-access successful-run complete-run)
```
NoahTheDuke, line 7: Instead of forward declaring the function here, how about moving `complete-run` above `successful-run`.

NoahTheDuke: Generally yeah. `declare` exists to help out when necessary (mutually recursive functions) but it’s best to keep things “in order”. Helps with hot reloading code too.

PR #5225, opened 2020-07-29
Dr. Lovegood is now automatic

### LINKED ISSUE #5162: Dr. Lovegood should be a start of turn trigger
Since Dr. Lovegood is a mandatory trigger, it should be realized as a start of turn trigger. Also makes it easier to remember to trigger if there are multiple other start of turn triggers.

Kubik161:
Closes #5162 

I'm not sure, if this is acceptable solution. I have added new event/trigger `:pre-runner-turn-begins` to active Dr. Lovegood before anything else. If I try to keep trigger on `:runner-turn-begins`, other cards activate before Dr. Lovegood. I assume that it is because other resources are not marked as `:interactive`.  But we can't mark every card as `:interactive` just in case someone plays Lovegood.

Is there better solution?


Kubik161:
My previous understanding was wrong :) I have moved trigger back to `:runner-turn-begins` and slightly modified `:trigger-event-simult-player` to check for disabled card during same event (e.g. Lovegood disabling card that has also `runner-turn-begins` effect).

src/clj/game/core/events.clj
```diff
@@ -293,7 +293,9 @@
                                     (if (should-continue state handlers)
                                       (continue-ability state side
                                                         (choose-handler
-                                                          (remove-once #(same-card? target (:card %)) handlers))
+                                                          ;;check if any card has been disabled (Dr. Lovegood)
+                                                          (remove-once #( true? (:disabled (get-card state (:card %))))
```
NoahTheDuke, line 297:
```suggestion
                                                          (remove-once #(:disabled (get-card state (:card %)))
```

src/clj/game/core/turns.clj
```diff
@@ -36,16 +36,16 @@
...
-             (unregister-floating-events state side :start-of-turn)
-             (unregister-floating-effects state side (if (= side :corp) :until-corp-turn-begins :until-runner-turn-begins))
-             (unregister-floating-events state side (if (= side :corp) :until-corp-turn-begins :until-runner-turn-begins))
-             (when (= side :corp)
-               (wait-for (draw state side 1 nil)
-                         (trigger-event-simult state side eid :corp-mandatory-draw nil nil)))
-             (swap! state dissoc (if (= side :corp) :corp-phase-12 :runner-phase-12))
-             (when (= side :corp)
-               (update-all-advancement-costs state side)))))
+            (unregister-floating-effects state side :start-of-turn)
```
NoahTheDuke, line 39: These changes to the formatting are unnecessary.

NoahTheDuke:
Merged manually

PR #5231, opened 2020-07-31
Automatically shuffle archives

Kubik161:
Simple implementation. Unseen cards are shuffled every time anything is trashed by any side.


NoahTheDuke:
Thanks for the idea! This won't work if cards are moved to archives without using `trash`, so maybe the login needs to be moved into `move` somehow? What do you think?

Kubik161:
Is there any reason for players to move cards manually into discard? I think that everything can be done via UI. Or are we using `move` in code somewhere instead of trash?

PR #5233, opened 2020-08-01
Ugly fix for Dr. Lovegood and Lewi Guilherme interaction

### LINKED ISSUE #3345: Lewi + Dr. Lovegood
The interaction is supposed to be that on the runner's turn Lewi is blanked so the runner doesn't have to pay, but then it's active on the corp's turn. 

On jnet, the corp hand size keeps +1.

### LINKED ISSUE #5208: Lewi Guilherme with added prompts
The runner has duplicate prompts for Corp timing windows when the corp previously agreed to pay to keep Lewi. These prompts happen most often running a run.

### LINKED ISSUE #5229: Lewi, Part 2
There is no longer any prompts that come up, but when using Dr. Lovegood to blank Lewi Guillherme on my turn, the corp's hand size does not go back down on their turn. In fact, it will continue to go up on their turn if left unchecked.

Kubik161:
Closes #5229
Closes #5208 
Closes #3345

I don't think that this is optimal way to fix this interaction, but I couldn't think of more clean solution. Is there better way to specify which side is targeted by `:in-play` effect?

### DIFF
--- a/src/clj/game/cards/resources.clj
+++ b/src/clj/game/cards/resources.clj
@@ -1432,9 +1432,7 @@
     {:flags {:drip-economy true ;; for Drug Dealer
              :runner-phase-12 (req (< 1 (count (filter #(card-flag? % :drip-economy true)
                                                        (all-active-installed state :runner)))))}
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
This isn't a bad idea but I don't think it's the right way to go about it.

Issue #5237, opened 2020-08-01
Capstone doesn't give enough draws when you trash multiple copies of an installed card


presheaf:
If you have Capstone installed and use it to trash 2 copies of Capstone from your grip, you only get 1 draw, not 2. The reason for this is [the way it counts](https://github.com/mtgred/netrunner/blob/571667450c36f1c80f8b1314330f23c551888090/src/clj/game/cards/hardware.clj#L277) which of your trashed cards are already installed. Because this is done through the use of a set, it won't care that you trashed 2 copies of Capstone. This can be fixed by replacing `(count overlap)` on [this line](url) by `(count (filter overlap trashed-card-names))`.

Issue #5236, opened 2020-08-01
Rex Campaign triggers when trashed by the Runner


presheaf:
When a rezzed Rex Campaign is trashed by the Runner, the Corp gets the choice to remove a bad publicity or gain 5 creds. This happens because the effect of Rex Campaign is implemented by having it auto-trash when there are no counters left, and [trigger its effect](https://github.com/mtgred/netrunner/blob/571667450c36f1c80f8b1314330f23c551888090/src/clj/game/cards/assets.clj#L1718) as an effect of trashing the card. I can't think of a good reason to do this, so a simple fix is just rewriting it to be like e.g. [Urban Renewal.](https://github.com/mtgred/netrunner/blob/571667450c36f1c80f8b1314330f23c551888090/src/clj/game/cards/assets.clj#L2137)

NoahTheDuke:
Thanks for the report!

Issue #5235, opened 2020-08-01
Dirty Laundry triggers twice off Doppelgänger


presheaf:
If the Runner has Doppelgänger installed and plays Dirty Laundry, when the successful run ends, they do not get the option to gain the Dirty Laundry money before triggering Doppelgänger. Additionally, if the Doppelgänger run is also successful, when it ends they gain 5 creds and then 5 more creds, as Dirty Laundry also triggers off the Doppelgänger run.

I think the issue here mainly has to do with [this event in the Dirty Laundry definition.](https://github.com/mtgred/netrunner/blob/571667450c36f1c80f8b1314330f23c551888090/src/clj/game/cards/events.clj#L751) Firstly, (I think) the Runner should have the option to gain the Dirty Laundry money before triggering Doppelgänger. They don't get this because the event is `:silent true`, so removing this will allow them to trigger Dirty Laundry first. However, if they choose to make the Doppelgänger run first, they still shouldn't get two triggers. This seems to be because Dirty Laundry doesn't check that the run it sees finish is the run it started. I'm not totally sure whether the engine has support for determining the ID of a run or whatever, so a proper fix for this might be impossible, but a pretty good fix would be just making this event `:once :per-turn`.

presheaf:
Oh, and High-stakes Job probably has all the same issues.

NoahTheDuke:
Thanks for the report!

presheaf:
It also seems Legwork has the same issue, as do multiple other cards. I'll fix the ones I see, but I think the root cause here is really that the engine doesn't have (?) support for checking whether a successful run occurring during the resolution of a run event was the run the run event initiated, meaning a lot of run events end up using the same trick. I can't think of any other cards this causes issues with except Doppelgänger, but I'll try adding support for giving each run a unique id and registering that id in the card initiating it. I think it won't be that gnarly, but we can see how well it turns out in the PR, and I'll understand if the simple fix is best.

PR #5239, opened 2020-08-03
Run id tracking and various bug fixes

### LINKED ISSUE #5234: Successful run effects trigger on empty Jackson Howard server
This seems to have been introduced in the run timing rework, and might also be an issue stemming from incorrect use of the buttons. To reproduce, have a server with only Jackson Howard, and have the runner run the server with Desperado installed. Have the Corp use Jackson before pressing "No further action" or anything else. When the Corp presses "No further action" and the Runner presses "Access server", Desperado will trigger. This does not happen if the Corp instead presses "Action before access" to go to 4.3 and triggers Jackson in that window. 

I don't actually know whether the Corp is allowed to use Jackson before 4.3 or not (although I think they are), so this might be a non-issue, but I think several players might be used to doing it this way, so it might happen often enough to be annoying. The root problem seems to be that the run doesn't automatically end when Jackson is triggered (meaning that the Runner could for example hit "Jack out" to trigger Au Revoir even if the Corp wanted to deny them the money by using Jackson), but a quickfix might be modifying [this function](https://github.com/mtgred/netrunner/blob/571667450c36f1c80f8b1314330f23c551888090/src/clj/game/core/runs.clj#L362) to check whether the server is empty before calling successful run, and doing `handle-end-run` instead if it is.

presheaf:
Fixes #5234 , #5235 , #5236 , #5237 . I'm happy to split this up if needed, but I figured most changes were small enough that a big PR was ok. The only interesting changes here are the addition of a run-id for each run which is (if applicable) stored in the card which initiated it, and the addition of a helper name `this-card-run` in the `req` macro to make it easier to include a test that the run an event handler is seeing is actually the run its card started. This has caused some issues with Doppelganger which I couldn't see a good fix for otherwise. I considered using durations, but felt this was sort of making the same assumption causing trouble in the first place, so opted not to. 

A simpler fix which fixes all (?) Doppelganger interactions is just making all the events I've used `this-card-run` for be `:once :per-turn` instead, and I'll understand if that's a better solution. Otherwise, it should be noted that run ids are generated as eids for convenience. I don't think this causes any trouble with the event system, but I might be wrong there.

EDIT: This PR grew a bit in scope. It now makes `:successful-run`, and the pre-/post- versions take the run objects as targets, making `this-card-run` less fragile. This necessitated changing a bunch of old card definitions, and I took the opportunity to introduce a convenience function to make them more readable while I was at it and touch up one or two which I struggled to read. It is a bit of a big change just to make a rarely-played console work, so as before I'm okay with reverting it.

### DIFF
--- a/src/clj/game/macros.clj
+++ b/src/clj/game/macros.clj
@@ -26,12 +26,11 @@
       rd-runnable (not (:rd (get-in (:runner @state) [:register :cannot-run-on-server])))
       archives-runnable (not (:archives (get-in (:runner @state) [:register :cannot-run-on-server])))
       tagged (is-tagged? state)
-      ;; true if the current run (or the previous run if no current run) was initiated by this card
+      ;; only intended for use in event listeners on (pre-/post-, un-)successful-run or run-ends
+      ;; true if the run was initiated by this card
       this-card-run (and (get-in card [:special :run-id])
                          (= (get-in card [:special :run-id])
-                            (if (:run @state)
-                              (:run-id (:run @state))
-                              (:run-id (get-in (:runner @state) [:register :last-run])))))
+                            (:run-id (first targets))))
       this-server (let [s (get-zone card)
                         r (:server (:run @state))]
                     (= (second s) (first r)))]
--- a/src/clj/game/macros.clj
+++ b/src/clj/game/macros.clj
@@ -26,6 +26,12 @@
       rd-runnable (not (:rd (get-in (:runner @state) [:register :cannot-run-on-server])))
       archives-runnable (not (:archives (get-in (:runner @state) [:register :cannot-run-on-server])))
       tagged (is-tagged? state)
+      ;; true if the current run (or the previous run if no current run) was initiated by this card
+      this-card-run (and (get-in card [:special :run-id])
+                         (= (get-in card [:special :run-id])
+                            (if (:run @state)
+                              (:run-id (:run @state))
+                              (:run-id (get-in (:runner @state) [:register :last-run])))))
       this-server (let [s (get-zone card)
                         r (:server (:run @state))]
                     (= (second s) (first r)))]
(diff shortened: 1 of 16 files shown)


test/clj/game/cards/events_test.clj
```diff
@@ -1427,17 +1427,35 @@
...
+     (take-credits state :corp)
+     (play-from-hand state :runner "Dirty Laundry")
+     (click-prompt state :runner "Archives")
+     (run-continue state)
+     (is (= 8 (:credit (get-runner))) "Gained 5 credits")
+     (play-from-hand state :runner "Dirty Laundry")
+     (click-prompt state :runner "Archives")
+     (run-jack-out state)
+     (is (= 6 (:credit (get-runner))) "Run unsuccessful; gained no credits")))
+
```
NoahTheDuke, line 1442: Mind removing this extra line?

presheaf: Whoops, not at all.

src/clj/game/macros.clj
```diff
@@ -26,6 +26,12 @@
       rd-runnable (not (:rd (get-in (:runner @state) [:register :cannot-run-on-server])))
       archives-runnable (not (:archives (get-in (:runner @state) [:register :cannot-run-on-server])))
       tagged (is-tagged? state)
+      ;; true if the current run (or the previous run if no current run) was initiated by this card
+      this-card-run (and (get-in card [:special :run-id])
+                         (= (get-in card [:special :run-id])
+                            (if (:run @state)
+                              (:run-id (:run @state))
+                              (:run-id (get-in (:runner @state) [:register :last-run])))))
```
NoahTheDuke, line 34: Is this a good idea? I realize it would be much more annoying to do it another way, but what about checking the `run` passed into `:run-ends`? I see that we're only passing the server into `:successful-run`, which sucks lol. jk This is a fine solution for now. We can fix `:successful-run` and move to something that isn't so stateful another time.

presheaf: I ended up reworking :successful-run and friends to take the run dict as target instead, making them consistent with :run-ends and :unsuccessful-run. It made the PR bigger than it used to, but it should alleviate some of the (justified!) concerns that this would cause issues.

PR #5247, opened 2020-08-04
Hippo double trigger fix

### LINKED ISSUE #4970: Hippo sometimes triggers when it shouldn't
Hippo should only trigger the first time all subroutines are broken on an outermost piece of ice each turn. However, by putting a Hortum on R&D, a Sapper on HQ, I was able to cheat using hippo by running R&D, breaking Hortum fully, jacking out, installing Hippo, running HQ and breaking Sapper again, at which point Hippo triggered erroneously. 

I think the predicate in Hippo's req isn't quite right and is checking whether this specific ice has been broken already this turn as opposed to whether any (at the time) outermost piece of ice has been broken.

Kubik161:
Closes #4970

Fix and new test.

This PR fixes problem described in original issue. However, there are still situations when hippo will/won't trigger as should according to rules. Only permanent fix (that I can think of) would be setting special flag when runner first breaks all subs on outermost ICE and Hippo strictly checking only this flag and nothing else. Otherwise we have situations like previously broken inner ice becoming outermost during runner turn, outermost ice being trashed, etc.

### DIFF
--- a/src/clj/game/cards/hardware.clj
+++ b/src/clj/game/cards/hardware.clj
@@ -905,10 +905,15 @@
                                       (effect-completed state side eid)))}}}]})
 
 (define-card "Hippo"
+  (letfn [(build-hippo-pred [outermost-ices]
+    (fn [events]
+      (or (map #(and (same-card? % (first events))
+                     (every? :broken (:subroutines (first events)))) outermost-ices))))]
   {:events [{:event :subroutines-broken
              :optional
-             {:req (req (let [pred #(and (same-card? (last run-ices) (first %))
-                                         (every? :broken (:subroutines (first %))))]
+              {:req (req (let [servers (->> (:corp @state) :servers seq flatten)
+                               outermost-ices (filter #(some? %) (map #(last (:ices %)) servers))
+                               pred (build-hippo-pred outermost-ices)]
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/hardware.clj
```diff
@@ -905,10 +905,15 @@
                                       (effect-completed state side eid)))}}}]})
 
 (define-card "Hippo"
+  (letfn [(build-hippo-pred [outermost-ices]
+    (fn [events]
+      (not (empty? (filter #(true? %) (map #(and (same-card? % (first events))
```
Kubik161, line 910: Also this `not empty filter` seems bad, but at least it works. I've tried to use `or` here and it didn't work with `map` result.

NoahTheDuke: Whoops, I missed this. For future reference, instead of `(not (empty (filter`, you can write `(seq (filter` or `(not-empty (filter`. Additionally, you don't have to wrap single-argument functions (predicates), you can just use it directly: `(filter true? (map ...`. This still isn't quite ideal because we want to be able to return truth-y things, not just `true`, so a better way would be `(filter identity` to remove false-y values.

PR #5249, opened 2020-08-04
Mind Game additional cost can now be payed only by runner cards

### LINKED ISSUE #5224: When mind game is fired, it allows the runner to add an installed Corp card at the bottom of the Runner’s stack

Kubik161:
Closes #5224

I have changed `add-installed-to-bottom-of-deck` to target only runner cards. It's used only by `Mind Game`.

Same thing is already done for `:shuffle-installed-to-stack`.


NoahTheDuke:
Changing this to be runner only limits us if there's ever a corp card that has the same cost.

PR #5253, opened 2020-08-05
Calva added to .gitignore

Kubik161:


NoahTheDuke:
This is best for your personal global `.gitignore`, but I'll include it just in case.

PR #5255, opened 2020-08-05
Experimental Muertos uninstall fix

### LINKED ISSUE #4532: Muertos uninstall while blanked still triggers free rez
Duplicate of #4108 which was closed by #4171 but the issue is still occurring.
```
secretfisk uses Dr. Lovegood to make the text box of Muertos Gang Member blank for the remainder of the turn.
secretfisk uses Aesop's Pawnshop to trash Muertos Gang Member and gain 3 .
sadstug uses Indian Union Stock Exchange to gain 1 .
Muertos Gang Member allows the Corp to rez Indian Union Stock Exchange at no cost
```

Kubik161:
Closes #4532

Introduces new thing `:uninstall`. 

Current `leave-play` is used to disable constant effects or to reverse gains/loses from cards. 
We trigger `leave-play` whenever card leaves play OR when is being disabled by some other effect (Dr. Lovegood).
We need to trigger `leave-play` ability on Muertos Gang Member only when uninstalled, not when disabled.  
I have added new keyword `:uninstall` which will trigger only when card is really leaving play. (at least I hope so)

I'm very open to other suggestions, because it seems like overkill to add whole new keyword just for one card.

### DIFF
--- a/src/clj/game/cards/resources.clj
+++ b/src/clj/game/cards/resources.clj
@@ -1563,19 +1563,23 @@
                                          card nil)))}]})
 
 (define-card "Muertos Gang Member"
-  {:effect (effect (continue-ability
-                     :corp
+  {:effect (req (show-wait-prompt state :runner "Corp to select a card to derez")
+                   (continue-ability
+                     state :corp
                      {:prompt "Select a card to derez"
                       :choices {:card #(and (corp? %)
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
The card very clearly says uninstalled, I think this is fine.

PR #5264, opened 2020-08-15
Add scripts for local backup and restore

ouroboros8:


ouroboros8:
@NoahTheDuke I said I'd script this up real quick like, a month ago, but here we are! :P

I don't think the failing tests are me.

NoahTheDuke:
Does this need to be run manually or do you set it up with cron or something?

ouroboros8:
Either! :) you could set up a nightly cron along the lines of
```00 03 * * * BACKUPS_KEEP_LAST=5 /path/to/netrunner/backups/backup```
Not sure how big the live db and how much space is on the server, so you might wanna tune the number of backups to retain. It's very basic retention scheme, every time the script runs it deletes all but the newest $BACKUPS_KEEP_LAST backups.

You can also backup manually before deploying changes if needed by just running ```
$ BACKUPS_KEEP_LAST=5 ./backups/backup```

As per the above on retention, this does mean you'll end up deleting the oldest backup, so you'll have 4 days of nightlies + the manual backup you just made. A bit messy but probably good enough.

Issue #5211, opened 2020-07-25
No One Home does not function if meat damage has been taken


NBKelly:
Simple test to confirm:
* Start a game as runner
* `/summon No One Home`
* Install it (add a click)
* `/take-meat 1`
* `/take-net 1`

The above will not get you a prompt, but:

* Start a game as runner
* `/summon No One Home`
* Install it (add a click)
* `/take-net 1`

Will get you a prompt.

Looking at the card and code, I have found:

* `game/cards/resources.clj` - No One Home will trigger on the `pre-tag` and `pre-damage` events, but only on the first instance of either of these events in a turn
```
(first-chance? [state side]
(< (+ (event-count state side :pre-tag)
                  (event-count state side :pre-damage))
               2))
```
* taking meat damage or net damage triggers `pre-damage` event
* `game/core/rules.clj` - pre-damage event
* after meat damage has been taken, the card cannot be used this turn

My suggestion would be separate events for damage types:
`pre-meat-damage`, `pre-net-damage` and `pre-brain-damage`, which then trigger the `pre-damage` events if not prevented. I'm not sure if that's feasible or if there's a better solution.

NoahTheDuke:
Thanks for the report! And good hunting on that bug. Instead of different pre-damages, `event-count` takes a predicate, so we could select only those pre-damage events that are :net damage.

I’d love to see a PR! We can even hop on a slack call and work on it together if you’re interested. Always trying to get folks into jnet development.

PR #5317, opened 2020-10-16
Fix No One Home req to test for first net damage

dsimmer:
To address #5211 

Just getting familiar with the environment and codebase!


test/clj/game/cards/resources_test.clj
```diff
@@ -2983,6 +2983,7 @@
 
 (deftest no-one-home
   ;; Prevent first tag or net damage of the turn if you beat trace0, then trash
+  (testing "Basic test"
```
NoahTheDuke, line 2986: Mind reindenting this whole test?

dsimmer: Whoops, thanks!

Issue #4996, opened 2020-02-10
DBS and TCA should print the position of the card bottomed


tvaduva:
The position of the card that is moved with Daily Business Show (under new errata for CR 1.4) and The Class Act is open information. There should be a log message printed with the position(s) (1st, 2nd, etc.) of card(s) that was/were bottomed.

You shouldn't need to keep track if the card was revealed or accessed earlier (probably making this issue much easier to resolve), because the message that the 2nd card was bottomed wouldn't help an opponent that has not seen the second card.

NoahTheDuke:
Thanks for the report!

Kiv:
I'm interested in working on this one, but I'm having trouble getting started. My environment is up and running, but the "Getting Started with Development" guide seems to be out of date - for example the videos are gone, and the `reset-card-defs` function doesn't exist anymore.

Any tips on what the current workflow is for setting up to test a fix for this?

NoahTheDuke:
Get on [stimslack](https://stimslackinvite.herokuapp.com/), join the #jinteki_net channel, and we'll get you set up!

Kubik161:
@NoahTheDuke this issue should be closed.

PR #5318, opened 2020-10-16
DBS and TCA now print the position of the bottomed card #4996

dsimmer:
To address #4996 

Not sure if the text is OK or if the full text for the position might be preferred, e.g. instead of
"uses Daily Business Show to add card 2 to the bottom of R&D"
perhaps
"uses Daily Business Show to add the second card to the bottom of R&D"
would be better?

Also, does this need a test or is just a text change OK?

### DIFF
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -532,7 +532,7 @@
                                                    :all true}
                                          :effect (req (doseq [c (reverse targets)]
                                                         (system-msg state side (str "uses Daily Business Show to add the "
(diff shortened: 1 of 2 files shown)


dsimmer:
Now it should read:
dolmant2 uses Daily Business Show to add the first card drawn to the bottom of R&D.
and
dolmant uses The Class Act to add the second card drawn to the bottom of the Stack.

NoahTheDuke:
That looks great! One final note is to use the function `inc` instead of `+ 1`.

dsimmer:
Done!

PR #5319, opened 2020-10-16
Swapping, gaining and lose credits are async, who knows what else

NoahTheDuke:
this fucking engine is full of bugs and painful edges. i hate it, tbh. but here we are, gotta do what we gotta do to make it work.

changes:
* all cards that swap are now cleaned up and do what they're supposed to instead of hacking stuff together:
  * Allele Repression and Metamorph are implemented
  * Daruma, Nihongai Grid, Mumbad City Grid, and Midori were rewritten to be more consistent
  * Cleaned up the Deva programs
* Removed ability-resolving from Hivemind.
* Changed `:virus` costs to look for Hiveminds and open a "pick available virus counter" prompt if one exists.
* Change Cordyceps to use `pay`, allowing for selecting counters from Hivemind.
* implement Loki. (our subtype system sucks ass lol.)
* gaining and losing credits are now required to be fully async. don't use `(gain state side :credit number)` or `(lose state side :credit number)`, use `gain-credits` and `lose-credits`.
* `reveal` is required to be async.


crimeminister:
@NoahTheDuke I am a Clojure programmer and recently started looking at this code base as it's pretty sizeable and an interesting application of the language. I'd be curious to know where you consider the rough edges to be, and if anything stands out as being in need of improvement.

NoahTheDuke:
Hey @crimeminister! Join up on the stimslack slack server, get into the #jinteki_net channel, and we can talk about this! https://stimslackinvite.herokuapp.com/

PR #5320, opened 2020-10-19
Reduce websocket chat

dsimmer:
The server doesn't process the typing instruction from spectators so it shouldn't be sent at all.

Also, the oninput handler fires before onchange, causing the typing status sent to the server to always be one character out of date. This is only really noticeable when you delete all the text you were typing and your opponent still sees the typing notification.


NoahTheDuke:
Ayyy this is great. The typing bug has always annoyed me but I couldn’t figure it out.

PR #5331, opened 2020-10-23
Divide and Conquer fix

### LINKED ISSUE #5315: Crisium on Archives does not prevent Divide and Conquer from activating.
The runner made a Divide and Conquer run on archives with a face down Crisium installed on Archives. I rezzed the Crisium Grid at the last window before successful, and the runner accessed archives. It then allowed them to access HQ and R&D, which they should not have.

(I'm sure this scenario happens all the time :doge:.)

Kubik161:
Closes #5315 

Fix and new test.

Divide and Conquer is only card that triggers in `end-access-phase` and requires `successful run`.

### DIFF
--- a/src/clj/game/cards/events.clj
+++ b/src/clj/game/cards/events.clj
@@ -777,7 +777,8 @@
    :effect (effect (make-run eid :archives nil card))
    :events [{:event :end-access-phase
              :async true
-             :req (req (= :archives (:from-server target)))
+             :req (req (and (= :archives (:from-server target))
+                            (get-in @state [:run :successful])))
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/events.clj
```diff
@@ -777,7 +777,8 @@
    :effect (effect (make-run eid :archives nil card))
    :events [{:event :end-access-phase
              :async true
-             :req (req (= :archives (:from-server target)))
+             :req (req (and (= :archives (:from-server target))
+                            (get-in @state [:run :successful])))
```
NoahTheDuke, line 781:
```suggestion
                            (:successful run)))
```

PR #5334, opened 2020-10-23
Cards from play area are moved back to hand when undo-click

### LINKED ISSUE #5030: Undo-click and events
When you undo click, events/operations in the play area stay in the play area, needing manual dragging back to the hand.

Kubik161:
Closes #5030 

Do we need to keep cards in play area after undo-click in any case?


NoahTheDuke:
Lockdowns stay in the corp's play area until the end of their turn. Not ideal.

Kubik161:
Yes, I knew there would be some obvious case why this won't work. Thanks :)

PR #5336, opened 2020-10-24
Performance tests and improvements

dsimmer:
This needs a very thorough review! This is my first serious commit and feedback would be much appreciated.

There is now a script which can generate and start a real lobby (although without decks) with as many spectators as you like. This runs on the current master so you can test it right away and see the issues I am fixing with the other commits. This script has a small issue in that large lobby updates break the clojure websocket client I am using (exceeding the max message size restriction) and can cause clients to be disconnected, but this restriction does not exist for browser clients. Might fix that later.

Reworked the refresh lobby background tick to be more proactive, sending immediately if there have been no sends recently, rate limiting at one per second and only sending an update if something actually definitely changed.

Refresh lobby now distinguishes between data required for those in the lobby and those outside of it, and only sends the extra data to those in the lobby. This private data can get quite large, so this is now sent using differ.

As a result, the client now merges updates with its own state rather than replacing it.

spectator-count was introduced as the public lobby parameter and the full spectator list is now only sent to lobby members as it can get very large.

ws/send! is now private and gated behind a buffer in broadcast-to! to prevent large numbers of updates (e.g. refresh the lobby for 1.1K spectators) from overwhelming the core.async internal take queue.

I tried to separate these into different commits so I can break up this PR if need be.


src/clj/web/game.clj
```diff
@@ -20,41 +19,52 @@
...
+  ([event
+    {:keys [gameid players spectators] :as game}
+    {:keys [type runner-state corp-state spect-state] :as states}
+    ws-id]
+      (def player (some #(= (:ws-id %) ws-id) players))
+      (ws/broadcast-to! [ws-id] event (json/generate-string
+                              {:gameid gameid
+                               :state (if player
+                                 (if (= (:side player) "Corp")
+                                        corp-state
```
NoahTheDuke, line 44: these just need to be indented 2 spaces

src/clj/web/game.clj
```diff
@@ -202,13 +207,14 @@
...
-            ; TODO: this would be better if a full state was only sent to the new spectator, and diffs sent to the existing players.
+            ;; Send existing state to spectator
+            ;; add a chat message, then send diff state to all players.
             (lobby/spectate-game user client-id gameid)
             (main/handle-notification state (str username " joined the game as a spectator."))
-            (ws/send! client-id [:lobby/select {:gameid gameid
-                                                :started started}])
-            (swap-and-send-state! (lobby/game-for-id gameid))
+            (ws/broadcast-to! [client-id] :lobby/select {:gameid gameid
+                                                :started started})
```
NoahTheDuke, line 215: indentation

src/clj/web/lobby.clj
```diff
@@ -227,27 +261,25 @@
...
               :messages        [{:user "__system__"
                                  :text (str username " has created the game.")}]
               :last-update     (t/now)}]
-    (swap! all-games assoc gameid game)
+    (refresh-lobby gameid game)
     (swap! client-gameids assoc client-id gameid)
-    (ws/send! client-id [:lobby/select {:gameid gameid}])
-    (refresh-lobby :create gameid)))
+    (ws/broadcast-to! [client-id] :lobby/select {:gameid gameid})
+    ))
```
NoahTheDuke, line 271: let's put these on the line above

src/cljs/nr/gamelobby.cljs
```diff
@@ -33,22 +34,45 @@
               (:date game)])
            games))
 
+
+(defn process-games-update 
+  [{:keys [diff notification] :as msg}]
+  (swap! app-state update-in [:games]
```
NoahTheDuke, line 40:
```suggestion
  (swap! app-state update :games
```

src/cljs/nr/gamelobby.cljs
```diff
@@ -33,22 +34,45 @@
...
+  (when (and notification (not (:gameid @app-state)))
+    (play-sound notification)))
+
 (ws/register-ws-handler!
   :games/list
-  #(swap! app-state assoc :games (sort-games-list %)))
+  (fn [msg]
+    (let [gamemap (into {} (map #(assoc {} (:gameid %) %) msg))]
+      (process-games-update {:diff {:update gamemap
+                                    :delete (map #(:gameid %) (filter #(not (get gamemap (:gameid %) false)) (:games @app-state)))}}))))
```
NoahTheDuke, line 58: this is an excellent place for 1) putting the bulk of this into a temp variable in the let block above, and 2) using a `->>` macro

src/clj/web/lobby.clj
```diff
@@ -44,26 +45,59 @@
...
+      (when (seq @game-lobby-updates)
+        (let [[old] (reset-vals! game-lobby-updates {})]
+          (doseq [[gameid update] (:update old)]
+              (ws/broadcast-to! (lobby-clients gameid) :games/differ {:diff {:update {gameid update}}})))))
+    (swap! send-queue inc))
+
+  (defn refresh-lobby-update-in
+    [gameid targets func]
+    (let [[old] (swap-vals! all-games update-in (concat [gameid] targets) func)]
+      (def old-key-diff (select-keys (get old gameid) [(first targets)]))
```
NoahTheDuke, line 70: let's use `let` instead of `def`

src/clj/tasks/load_test.clj
```diff
@@ -0,0 +1,121 @@
...
+            [monger.collection :as mc]
+            [monger.result :refer [acknowledged?]]
+            [monger.operators :refer :all]
+            [web.core :refer :all]
+            [game.core-test :refer :all]
+            [game.macros-test :refer :all]
+            [web.lobby :refer :all]))
+
+(defn safe-println [& more]
+  (.write *out* (str (clojure.string/join " " more) "\n")))
```
NoahTheDuke, line 21:
```suggestion
  (.write *out* (str (string/join " " more) "\n")))
```

src/clj/tasks/load_test.clj
```diff
@@ -0,0 +1,121 @@
...
+
+(defn add-test-users [maxUsers]
+  (webdb/connect)
+  (when (not (mc/find-one-as-map db "users" {:username "TestCorp"}))
+    (mc/insert db "users"
+                    {:username         "TestCorp"
+                      :email            "TestCorp@mailinator.com"
```
NoahTheDuke, line 28: Looks like each subsequent pair is indented 1 extra. And you should separate the keys and values by a single space.

src/clj/tasks/load_test.clj
```diff
@@ -0,0 +1,121 @@
...
+                      :email            (str "TestUser" n "@mailinator.com")
+                      :password         (password/encrypt "password")
+                      :isadmin          false
+                      :options          {}})))
+      (range maxUsers))))
+
+(defn login
+  [username password]
+  (let [options {:form-params {:username username :password password}}
+      {:keys [status error headers body]} @(http/post "http://localhost:1042/login" options)]
```
NoahTheDuke, line 56: indentation

src/clj/tasks/load_test.clj
```diff
@@ -0,0 +1,121 @@
...
+        (str "ws://localhost:1042/ws?client-id=" runner-client-id)
+        ; :on-receive #(safe-println 'received %)
+        :on-error #(safe-println "runner error" %)
+        :on-connect (fn [n] (safe-println "Runner Connected"))
+        :on-close (fn [x y] (safe-println "Runner Disconnected"))
+        :headers {"Cookie" (login "TestRunner" "password")}))
+  (safe-println "Create lobby")
+  (handle-lobby-create {:ring-req {:user {:username "TestCorp"}}
+                        :client-id corp-client-id
+                        :?data {:title "Performance Game" :format "standard" :allow-spectator true :spectatorhands false :password "" :room "casual" :side "Corp" :options {}}})
```
NoahTheDuke, line 83: This will be more readable if you put each key-value pair on a new line, like above.

src/clj/tasks/load_test.clj
```diff
@@ -0,0 +1,121 @@
...
+  
+  (def game-id (first (first @all-games)))
+
+  (handle-lobby-join {:ring-req {:user {:username "TestRunner"}}
+                        :client-id runner-client-id
```
NoahTheDuke, line 88: indentation

src/clj/web/core.clj
```diff
@@ -76,7 +77,11 @@
 
     ;; Clear inactive lobbies after 30 minutes
     (web.utils/tick #(lobby/clear-inactive-lobbies 1800) 1000)
-    (web.utils/tick lobby/send-lobby 1000)
+    (web.utils/tick lobby/reset-send-lobby 1000)
+
+    ;; Serve flame graph, used for performance profiling locally
+    ; (prof/serve-files 8080)
+    ; (prof/start)
```
NoahTheDuke, line 84: I'd prefer to not put these kinds of "dev use only" blocks in. Helps keep the uberjar smaller too.

src/clj/web/game.clj
```diff
@@ -20,41 +19,50 @@
...
-                                                corp-diff
-                                                runner-diff)})]))
-  (doseq [{:keys [ws-id] :as pl} spectators]
-    (ws/send! ws-id [:netrunner/diff (json/generate-string
-                                       {:gameid gameid
-                                        :diff spect-diff})])))
+    (ws/broadcast-to! [ws-id]
+                      :netrunner/diff (json/generate-string
+                      {:gameid gameid
+                        :diff (if (= side "Corp")
```
NoahTheDuke, line 25: indentation

src/clj/web/game.clj
```diff
@@ -20,41 +19,50 @@
...
+                                corp-diff
+                                runner-diff)})))
+  (ws/broadcast-to!
+                    (map #(:ws-id %) spectators)
```
NoahTheDuke, line 29: indentation

src/clj/web/game.clj
```diff
@@ -20,41 +19,50 @@
...
-  ([game states]
-   (send-state! :netrunner/state game states))
+  "Sends full states generated by public-states to either the client specified or all connected clients."
+  ([event
+    {:keys [gameid players spectators] :as game}
+    {:keys [type runner-state corp-state spect-state] :as states}
+    ws-id]
+   (let [player (some #(= (:ws-id %) ws-id) players)]
+     (ws/broadcast-to! [ws-id]
+                        event
```
NoahTheDuke, line 41: indentation

PR #5342, opened 2020-10-26
Reduce the number of regexs performed by the client on messages

dsimmer:
and speed up the interleaving of card results into messages

### DIFF
--- a/src/cljs/nr/utils.cljs
+++ b/src/cljs/nr/utils.cljs
@@ -167,15 +167,25 @@
 (defn card-patterns-impl []
   "A sequence of card pattern pairs consisting of a regex, used to match a card
   name in text, and the span fragment that should replace it"
-  (letfn [(span-of [title code] [:span {:class "fake-link" :id code} title])
-          (regex-of [card-title] (re-pattern (regex-escape card-title)))]
+  (letfn [(span-of [title code] [:span {:class "fake-link" :id code} title])]
     (->> @all-cards
          (filter #(not (:replaced_by %)))
          (map (juxt :title :code))
-         (map (fn [[k v]] [(regex-of k) (span-of k v)]))
+         (map (fn [[k v]] [k (span-of k v)]))
          (sort-by (comp count str first) >))))
 
 (def card-patterns (memoize card-patterns-impl))
+
+(defn contains-card-pattern-impl []
+  "A card pattern regex, used to match a card name in text to check if the rest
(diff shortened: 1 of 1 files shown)


src/cljs/nr/utils.cljs
```diff
@@ -167,16 +167,26 @@
...
          (filter #(not (:replaced_by %)))
          (map (juxt :title :code))
-         (map (fn [[k v]] [(regex-of k) (span-of k v)]))
+         (map (fn [[k v]] [k (span-of k v)]))
          (sort-by (comp count str first) >))))
 
 (def card-patterns (memoize card-patterns-impl))
 
+(defn contains-card-pattern-impl []
+  "A card pattern regex, used to match a card name in text to check if the rest
```
NoahTheDuke, line 180: Not gonna make you change this before merging, but function docstrings have to go before the argument list `[]`, otherwise it's just a discarded string literal within the function.

src/cljs/nr/utils.cljs
```diff
@@ -185,25 +195,16 @@
...
-  sequences of varying lengths by padding out the shorter of the two sequences,
-  in this case with empty strings."
+  interleaving `replacement`s into the context and dropping the last one as
+  interleave always weaves in excess"
   (if (string? element)
-    (let [context (split element regex)
-          match-count (count (re-seq regex element))
-          replacements (repeat match-count replacement)]
-      (->> (padded-interleave "" context replacements)
+    (let [context (.split element regex)
```
NoahTheDuke, line 204: any reason for `.split` instead of clojure's `split`?

dsimmer: Yes! When I was testing, it was quite a bit faster when splitting with other strings rather than a regex (although it works with both). I suspect this is because the browser is better placed to optimise this. This is why in card-patterns-impl I just returned the string of the card rather than the regex. Also, clojure's split has a weird edge case on a full match where is simply returns an empty array, which makes the logic slightly more complex.

src/cljs/nr/utils.cljs
dsimmer: Thanks, sloppy by me, I just copied card-patterns-impl which does the same thing

src/cljs/nr/utils.cljs
NoahTheDuke: Cool, that’s a great detail

PR #5358, opened 2020-11-09
Fix keep hand always mulliganing

dsimmer:
Caused by commit b5f84f4972dd0eab0e09dbd77f8401331ee235b7, missed a conversion of the prompt to use the full object instead


NoahTheDuke:
Ah shit, thank you.

PR #5359, opened 2020-11-09
Fix spectator chat

dsimmer:
In one of my other commits I mistakenly thought spectator? returned a boolean. Turns out, it is relied upon to return the actual user, so the spectator chat stopped working. I also cleaned up the functions to take a game object, since that is just as easy to pass as the game id and results in one less call in all those functions.

### DIFF
--- a/src/clj/web/lobby.clj
+++ b/src/clj/web/lobby.clj
@@ -109,22 +109,22 @@
   (defn reset-send-lobby
     []
     (let [[old] (reset-vals! send-ready true)]
-      (if-not old (send-lobby)))))
+      (when-not old (send-lobby)))))
 
 (defn player?
   "True if the given client-id is a player in the given game"
   [client-id game]
-    (some #(when (= client-id (:ws-id %)) %) (:players game)))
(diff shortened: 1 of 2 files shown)


src/clj/web/lobby.clj
```diff
@@ -112,29 +112,26 @@
...
+    (= client-id (-> game :players first :ws-id)))
 
 (defn spectator?
-  "True if the given client-id is a spectator in the given gameid"
-  [client-id gameid]
-  (when-let [game (game-for-id gameid)]
-    (not (some #(when (= client-id (:ws-id %)) %) (:players game))))) ; Faster to check players than to check spectators
+  "True if the given client-id is a spectator in the given game"
+  [client-id game]
+    (some #(when (= client-id (:ws-id %)) %) (:spectators game)))
```
NoahTheDuke, line 127: looks like some indentation stuff was left in here, along with the functions above

dsimmer: ughh, yeah thanks. Indentation for functions is normal for me so I didnt even notice.

PR #5370, opened 2020-11-10
Differ bug workaround to prevent junk spectator data

dsimmer:
Should I leave the comment in? I think I would be tempted to remove this code during a cleanup without the comment.


NoahTheDuke:
Mathew comment explaining the behavior is good, but the `prn` comment should probably go. That's purely for debugging and can be added later at a dev's need.

dsimmer:
Thanks, totally missed that!

Issue #5365, opened 2020-11-09
Opponents Leaving Lobby Remain in Lobby


analyzechris:
If someone leaves your game before it begins, the interface shows them as still being in your game lobby.

<img width="1253" alt="Screen Shot 2020-11-09 at 4 10 47 PM" src="https://user-images.githubusercontent.com/64614578/98596936-34c25a80-22a6-11eb-9888-04df07ceb4bc.png">
<img width="630" alt="Screen Shot 2020-11-09 at 4 06 11 PM" src="https://user-images.githubusercontent.com/64614578/98596938-355af100-22a6-11eb-82d8-113fab85725c.png">

NoahTheDuke:
Thanks for the report!

PR #5371, opened 2020-11-11
Consistent rejoin, join and watch states

dsimmer:
In my websocket cleanup commit, I removed some duplicate information being sent in the rejoin, turns out the client wasnt parsing the lobby/select state updates at all so rejoins are now broken. I have fixed the lobby/select message to correctly update the state and cleaned up some unnecessary params that were being sent.

When a user was removed from the game, the state was updated without notifying clients (in remove-users), preventing rejoins until the client was refreshed. This is now fixed to properly notify clients.

I also went through and made sure that the client full state update is sent before any of the diffs for watch, rejoin and join. This prevents the client receiving a game diff before they have received any state to diff against.

### DIFF
--- a/src/clj/web/game.clj
+++ b/src/clj/web/game.clj
@@ -35,30 +35,20 @@
     {:keys [type runner-state corp-state spect-state] :as states}
     ws-id]
    (let [player (some #(= (:ws-id %) ws-id) players)]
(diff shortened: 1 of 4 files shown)


dsimmer:
Fixed #5365 as well

PR #5379, opened 2020-11-12
Fix Jackson Howard black screen

### LINKED ISSUE #5374: Using Jackson Howard's re-shuffle ability breaks the game
Using Jackson Howard's second ability caused my browser window to permanently turn black, making it impossible to keep playing. Reloading the game, exiting and rejoining did not change anything. My whole game screen was completely black and there was nothing I could interact with.

jwarwick:
Arguments to shuffle function were in the wrong order. Also making sure not to pass nil to re-find.

Closes #5374

### DIFF
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -980,7 +980,7 @@
                {:label "Shuffle up to 3 cards from Archives into R&D"
                 :cost [:remove-from-game]
                 :async true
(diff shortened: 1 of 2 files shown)


Bovrill:
Seems to still be an issue. Perhaps this fix hasn't been made live yet?

jwarwick:
@Bovrill - it hasn't been merged and pushed to the live server yet.

PR #5380, opened 2020-11-12
Fix Tollbooth encounter message

### LINKED ISSUE #5377: Tollbooth log is not correctly showing encounter effect, although credits are subtracted
XXX encounters Tollbooth protecting Server 2 at position 0.
XXX pays 3  to install Black Orchestra.
tolaasin uses Tollbooth to force the Runner to .

jwarwick:
Original bug was that the entire message needed to be wrapped in a `str`, but the pay message is `pays 3 [Credits]`, which doesn't make sense for the log message.

So now we check if the price was paid and generate a hardcoded message.

Closes #5377

### DIFF
--- a/src/clj/game/cards/ice.clj
+++ b/src/clj/game/cards/ice.clj
@@ -3119,9 +3119,9 @@
   {:on-encounter {:async true
                   :effect (req (wait-for (pay state :runner card [:credit 3])
                                          (if (:cost-paid async-result)
-                                           (do (system-msg state :corp "uses Tollbooth to force the Runner to pay 3 [Credits]")
+                                           (do (system-msg state :runner (str (:msg async-result) " when encountering Tollbooth"))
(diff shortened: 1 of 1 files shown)


NoahTheDuke:
Good catch on the wording awkwardness. I think we should find a way to still log the `:msg` from payment because if the runner uses any credits from a card (Net Mercur, one of the companions, etc), we want to display that in the log. Maybe a second `system-msg` with `:runner` as the side? `(do (system-msg ...) (system-msg state :runner payment-str) (effect-completed ...)`

jwarwick:
I like the idea of generating the message from the runner. Maybe "Runner encounters Tollbooth and <payment string>" or "Runner <payment string> to encounter Tollbooth". Not sure if those have the same meaning rules-wise.

I can flip the unable to pay message to similar wording.

PR #5383, opened 2020-11-13
x-credits cost not being charged

### LINKED ISSUE #5376: Misdirection used for free
I used Misdirection twice in a game and we noticed that the credits weren't actually paid while using the card.

jwarwick:
Went with the two system message approach for Utae.

It appears we were updating the `credits` (plural) field instead of `credit` for a side. A number of cards implemented workarounds. I haven't addressed all of them.

Looking for feedback in case I'm missing something.

Closes #5376


src/clj/game/cards/programs.clj
```diff
@@ -2552,7 +2551,7 @@
                                    :effect (effect
                                              (continue-ability
                                                (when (pos? (cost-value eid :x-credits))
-                                                 (break-sub (cost-value eid :x-credits) (cost-value eid :x-credits) "Code Gate"))
+                                                 (break-sub nil (cost-value eid :x-credits) "Code Gate"))
```
NoahTheDuke, line 2554: this is going to mess up the "break sub" message. maybe we should have a `:msg` above? something like `:msg (msg "break " (quantify (cost-value eid :x-credits) "subroutine") " on " (card-str state current-ice))`? That way it'll print "Runner pays 4 credits to use Utae to break 4 subroutines on Hive protecting HQ at position 1. Runner uses Utae to break 4 subroutines (End the run, end the run, end the run, end the run) on Hive". It's kind of annoying to duplicate stuff, but we don't have a way to "override" the... hold on, just had a thought

NoahTheDuke: in fact, the cost exists on  the `eid` that's passed into `break-sub`. We could call `merge-costs-paid` in `break-subroutines`, which would handle printing "both" costs (one of which will be `nil` in this case). I bet this will break other stuff? lol but it's a way to handle it

PR #5384, opened 2020-11-13
Fix crash in Net Mercur when firing ability

jwarwick:
Was causing a stack overflow loop on `trigger-event-sync`

### DIFF
--- a/src/clj/game/cards/resources.clj
+++ b/src/clj/game/cards/resources.clj
@@ -1668,8 +1668,8 @@
                 :async true
                 :req (req (pos? (get-counters (get-card state card) :credit)))
                 :effect (req (add-counter state side card :credit -1)
-                             (wait-for (gain-credits state side eid 1)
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/resources.clj
```diff
@@ -1666,9 +1666,10 @@
 (defcard "Net Mercur"
   {:abilities [{:msg "gain 1 [Credits]"
                 :async true
+                :req (req (pos? (get-counters (get-card state card) :credit)))
                 :effect (req (add-counter state side card :credit -1)
                              (wait-for (gain-credits state side eid 1)
-                                       (trigger-event-sync state side eid :spent-credits-from-card card)))}]
+                                       (trigger-event-sync state side (make-eid state eid) :spent-credits-from-card card)))}]
```
NoahTheDuke, line 1672: ah fuck, I fucked this up. I'm sorry. It should be the other way around. Change `gain-credits` to call `make-eid`, and pass the `eid` directly into `trigger-event-sync`

PR #5385, opened 2020-11-14
Fix Jemison crash when forfeiting Gene Splicer

jwarwick:
Gene Splicer isn't an agenda (though it is worth agenda points). Removed the `agenda?` check when getting agenda value.

### DIFF
--- a/src/clj/game/core/agendas.clj
+++ b/src/clj/game/core/agendas.clj
@@ -39,9 +39,9 @@
 
 (defn get-agenda-points
   [card]
-  (when (agenda? card)
-    (or (:agendapoints card)
-        (:current-points card))))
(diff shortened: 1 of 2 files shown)


lostgeek:
Actually Gene Splicer _is_ an agenda, since it gets added to the score area "as an agenda worth 1 agenda point". I have no idea, how we'd actually implement this though

jwarwick:
@lostgeek - I don't actually *play* Netrunner 😿 
The `card-def` isn't an agenda, which was the root of the failure. The check for agenda points now doesn't have that as a requirement.
That might be causing an issue elsewhere in the code though. If we can think of any other interactions we could write test cases for them.

PR #5388, opened 2020-11-17
Allow non-ascii chars in deck names

jwarwick:
When computing the salt for a deck hash, non-ascii characters in the deck name will make `byte` throw a `java.lang.IllegalArgumentException`.

Using `slugify` to make the deckname all printable ascii.

No idea how to test this.

### DIFF
--- a/src/clj/web/decks.clj
+++ b/src/clj/web/decks.clj
@@ -34,6 +34,10 @@
              :status status
              :hash deck-hash)))
 
+(defn make-salt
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
Testing is hard, yeah. Could rebind the `encrypt` call to an fn that returns the resulting string, and then do a simple comparison. It's a pure function, with the salt being the deck name, so running it to get the expected result and then pasting that into the test is okay?

PR #5402, opened 2020-11-24
Fix deck status details

### LINKED ISSUE #5181: Format legality
Currently (as of 1/7/2020) regardless of what format you choose, decks will show as either standard legal or illegal. 

When playing a non-standard game, Jnet will only check if the deck is standard legal, ignoring the format's legality. 

When building a deck, Jnet will check if the deck is legal for the chosen format, but will display it as standard.

jwarwick:
Fixes #5181


NoahTheDuke:
I wish we had tests for the cljs stuff. I've looked into it and it's really annoying lol. This is great work, John. Thank you so much for handling it.

PR #5404, opened 2020-11-25
Fix Core Experience deck legality check

### LINKED ISSUE #5399: Core Experience deck checker does not check for card counts.
For example, I've been able to put 3 Hostile Takeovers into a deck, even though core experience should allow for only 2

jwarwick:
Fixes #5399


jwarwick:
Will require running the `update deck legality` task to correct existing decks.

PR #5409, opened 2020-11-26
Don't display canceled decks

jwarwick:
Creating a new deck, then canceling before saving, left the deck in your decklist display. Attempting to delete it would generate an exception on the server for an unknown deck id.

Also disabling the deck creation buttons until all decks are loaded; this path could also cause the above behavior.

### DIFF
--- a/src/cljs/nr/deckbuilder.cljs
+++ b/src/cljs/nr/deckbuilder.cljs
@@ -236,8 +236,12 @@
   (-> (:viewport @db-dom) js/$ (.removeClass "edit")))
 
 (defn cancel-edit [s]
-  (end-edit s)
-  (put! select-channel (:old-deck @s)))
+  (let [deck (:deck @s)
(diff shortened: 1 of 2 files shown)


src/cljs/nr/deckbuilder.cljs
```diff
@@ -796,15 +800,15 @@
      [:span.small "(Type or paste a decklist, it will be parsed)"]]]
    [edit-textbox s]])
 
-(defn collection-buttons [s user]
+(defn collection-buttons [s user decks-loaded]
   [:div.button-bar
-   [cond-button "New Corp deck" @user #(new-deck s "Corp")]
-   [cond-button "New Runner deck" @user #(new-deck s "Runner")]])
+   [cond-button  "New Corp deck" (and @user @decks-loaded) #(new-deck s "Corp")]
```
NoahTheDuke, line 805: extra space?

jwarwick: Thanks, updated.

PR #5421, opened 2020-11-29
Don't duplicate subs on ice when de-rezzing

jwarwick:


src/clj/game/core/initializing.clj
```diff
@@ -40,7 +40,7 @@
                   :current-strength :current-advancement-requirement :current-points
                   :runner-abilities :corp-abilities :rezzed :new
                   :added-virus-counter :subtype-target :server-target :extra-advance-counter)
-        c (assoc c :subroutines (subroutines-init c cdef) :abilities (ability-init cdef) :special nil)
+        c (assoc c :abilities (ability-init cdef) :special nil)
```
NoahTheDuke, line 43: I think the solution for this is to add a `:subroutines` to the `dissoc` call above, as the goal is to completely reset the subroutines vector on the card object. Printed subroutines exist on card objects, which we track.

jwarwick: I agree we probably need to `dissoc` `:subroutines`, but I don't think we want to add them back here. `make-card` calls `subroutines-init` which adds the subs from the `cdef`.

NoahTheDuke: Well, we don't call `make-card` when derezzing a card, which is when this particular function (`dissoc-card`) is called, and abilities that add subs (Sub Boost) don't remove their added subs on derez, so we have to reset the subroutines to only the printed subs at some point.

PR #5422, opened 2020-11-30
Httpkit and sente upgrade

dsimmer:


dsimmer:
If you are not logged in, the authentication popup now shows when you hit the play page and if you dismiss it you wont be able to see any games. You also wont get new chat messages in the general chats.

Logged in users should see no changes in behaviour.

src/clj/web/api.clj
```diff
@@ -81,11 +84,15 @@
...
   (wrap-routes user-routes auth/wrap-authentication-required)
   (wrap-routes tournament-routes auth/wrap-tournament-auth-required)
-  (wrap-routes admin-routes auth/wrap-authorization-required)
-  public-routes)
+  (wrap-routes admin-routes auth/wrap-authorization-required))
+
+(defroutes routes
+  (wrap-routes private-routes wrap-anti-forgery)
+  (wrap-routes public-CSRF-routes wrap-anti-forgery)
+  public-routes wrap-anti-forgery)
```
NoahTheDuke, line 95: Should you move the `wrap-anti-forgery` down to the `app` definition?

src/cljs/nr/ws.cljs
```diff
@@ -4,15 +4,19 @@
...
 
-(let [{:keys [chsk ch-recv send-fn state]}
-      (sente/make-channel-socket! "/ws" {:type :ws})]
-  (def chsk       chsk)
-  (def <ws-recv ch-recv) ; ChannelSocket's receive channel
-  (def ws-send! send-fn) ; ChannelSocket's send API fn
-  (def chsk-state state)   ; Watchable, read-only atom
-  )
+(if-not ?csrf-token
+  (println "CSRF token NOT detected in HTML, default Sente config will reject requests")
```
NoahTheDuke, line 11: Might want to leave this out as it will get printed _a lot_ lol

NoahTheDuke: nevermind, this is in the cljs

src/clj/web/api.clj
dsimmer: By default, the anti-forgery wrap will require the CSRF token on any endpoint that isnt a GET. If we apply it to all endpoints the login and register POST calls will require a CSRF... which you can only get by logging in.

What you are seeing here is a typo which I didnt pick up on in my testing because it is essentially ignored by the defroutes macro...

NoahTheDuke: cool, thanks

PR #5427, opened 2020-12-02
Add ranges on all chat commands

jwarwick:
Make it harder to generate an unsaveable state.

### DIFF
--- a/src/clj/game/core/commands.clj
+++ b/src/clj/game/core/commands.clj
@@ -269,7 +269,7 @@
         nil)
       (case command
         "/adv-counter" #(command-adv-counter %1 %2 value)
-        "/bp"         #(swap! %1 assoc-in [%2 :bad-publicity :base] (constrain-value value 0 1000))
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
Holy shit, this is so good.

src/clj/game/core/commands.clj
```diff
@@ -264,22 +269,22 @@
         nil)
       (case command
         "/adv-counter" #(command-adv-counter %1 %2 value)
-        "/bp"         #(swap! %1 assoc-in [%2 :bad-publicity :base] (max 0 value))
+        "/bp"         #(swap! %1 assoc-in [%2 :bad-publicity :base] (constrain-value value -1000 1000))
```
lostgeek, line 272: Why can bad pub be negative?

jwarwick: We aren't real consistent about what value gets modified when you use these commands. I figure if someone is stuck with a bp modifier, you can make the base value negative to correct for it. I think I do something similar in a few other cases as well.

PR #5428, opened 2020-12-03
Don't populate decks during load

### LINKED ISSUE #5425: Expand pronouns list
- Pronouns not set
- Any
- Prefer not to say
- [Blank]
- He/him
- She/her
- It
- They/them
- Ne/nem
- Ve/ver
- Ey/em
- Ze/hir
- Ze/zir
- Xe/xem

jwarwick:
Decks are only populated with real card data when selected for editing.

Also added more pronouns.
Fixes #5425

### DIFF
--- a/src/cljs/nr/deckbuilder.cljs
+++ b/src/cljs/nr/deckbuilder.cljs
@@ -4,7 +4,7 @@
             [clojure.string :refer [split split-lines join escape lower-case] :as s]
             [goog.string :as gstring]
             [goog.string.format]
-            [jinteki.cards :refer [all-cards] :as cards]
+            [jinteki.cards :refer [all-cards identity-cards] :as cards]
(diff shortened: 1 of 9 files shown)


src/cljs/nr/deckbuilder.cljs
```diff
@@ -63,48 +63,50 @@
...
 (defn- build-identity-name
   [title setname]
   (if setname
     (str title " (" setname ")")
     title))
 
+(defn lookup-identity
+  "Find an identity card from the card code or title"
+  [side card]
+  (if-let [ident (first (filter #(= (:code card) (:code %)) @identity-cards))]
```
NoahTheDuke, line 100: Why use `identity-cards` for this instead of just filtering on identities from `all-cards`?

PR #5435, opened 2020-12-07
Display all card images

jwarwick:
Display images for replaced cards.
Download all card images from NRDB, rather than using embedded url.

Reminder: You will need a newer version of `http-kit` in the `project.clj` file to get the `lein fetch` task to run. I've been using `2.4.0-alpha4`.


src/cljs/nr/cardbrowser.cljs
```diff
@@ -39,8 +41,37 @@
         (.setItem js/localStorage "cards" (.stringify js/JSON (clj->js {:cards cards :version server-version}))))
       (reset! all-cards cards)
       (swap! app-state assoc :cards-loaded true)
+      (swap! app-state assoc :previous-cards (generate-previous-cards cards))
```
NoahTheDuke, line 44: If you like this on a separate line for clarity, cool. If you like code golf like me, lol, `assoc` takes multiple key value pairs, so you can merge this with the line above: `(swap! app-state assoc :cards-loaded true :previous-cards ...)`

jwarwick: Would I program in clojure if I wasn't trying to optimize for loc?

src/cljs/nr/account.cljs
```diff
@@ -102,22 +82,14 @@
 
 (defn- remove-card-art
   [card s]
-  (swap! s update-in [:alt-arts] #(dissoc % (keyword (:code card))))
-  (when-let [replaces (:replaces card)]
-    (let [replaced-card (some #(when (= replaces (:code %)) %) (:cards @app-state))]
-      (when replaced-card
-        (remove-card-art replaced-card s)))))
+  (swap! s update-in [:alt-arts] #(dissoc % (keyword (:code card)))))
```
NoahTheDuke, line 85:
```suggestion
  (swap! s update :alt-arts dissoc (keyword (:code card))))
```

A long long time ago, Clojure didn't have `update`, only `update-in`. Now we do! `update` takes a keyword, a function, and any number of arguments. It then `apply`s the function to the argument list, passing in the current value as the first argument, making it p much: `(apply dissoc (:alt-arts @s) (keyword (:code card)))`. This is purely a "style" thing, lol, but one of those that I've been trying to clean up in the codebase. If you don't care to update this, no worries.

src/cljs/nr/cardbrowser.cljs
```diff
@@ -26,21 +28,58 @@
...
                           (filter #(= "standard" (:format %)))
                           (map (fn [e] (update e :date-start #(js/Date.parse %))))
                           (sort-by :date-start)
-                          last)]
-      (reset! cards/mwl latest_mwl)
+                          last)
+            alt-cards (->> cards
+                       (map #(select-keys % [:title :code :alt_art]))
+                       (filter :alt_art)
+                       (into {} (map (juxt :code identity))))
```
NoahTheDuke, line 39: I don't think this line does what you intend. The `->>` will transform it to be `(into {} (map (juxt :code identity)) (result-of-previous-lines))`. Meaning the `map` call will create a transducer and then be fed into the hashmap, along with the output of the previous lines. Maybe you meant: `(->> ... (filter :alt_art) (map (juxt :code identity)) (into {}))`?

src/cljs/nr/cardbrowser.cljs
```diff
@@ -84,12 +129,8 @@
...
       (->> alt-arts
            (concat [""])
-           (map (fn [art] (if art
-                            (assoc card :art art)
-                            card)))
-           (map (fn [c] (if (:art c)
-                          (assoc c :display-name (str (:code c) "[" (alt-art-name (:art c)) "]"))
-                          c)))
+           (map #(if % (assoc card :art %) card))
+           (map #(if (not= "" (:art %)) (dissoc % :previous-versions) %))
```
NoahTheDuke, line 133: `(not= "" (:art %))` will be `true` if there's no `:art` key. Is that okay?

src/cljs/nr/cardbrowser.cljs
```diff
@@ -221,15 +275,20 @@
...
+                 (contains? % :future-version)
+                 (contains? % :previous-versions))
+            cards)))
 
-(defn filter-alt-art-set [setname cards]
+(defn- filter-alt-art-set [setname cards]
   (when-let [alt-key (alt-version-from-string setname)]
+    (if (= alt-key :prev)
+      (filter #(or (contains? % :future-version) (contains? % :previous-versions)) cards)
     (let [sa (map first
```
NoahTheDuke, line 289: think spacing got messed up here, cuz the `let` block doesn't look like the else branch of the `if`

src/cljs/nr/cardbrowser.cljs
jwarwick: Yes, I think that's okay. We have a list of known arts and I'm appending  "" to the end. I'm open to a smarter way to remove the key from all copies of the card except the "" one.

NoahTheDuke: I haven't spent enough time with this to know what is supposed to happen, lol, so I trust your judgement.

PR #5455, opened 2020-12-25
Shame users that try to replace their identity with a card that doesn't exist

jwarwick:
Because, according to the server log, someone became `High-Risk Investment`...

### DIFF
--- a/src/clj/game/core/commands.clj
+++ b/src/clj/game/core/commands.clj
@@ -1,7 +1,7 @@
 (ns game.core.commands
   (:require
     [game.core.board :refer [all-installed get-zones server->zone]]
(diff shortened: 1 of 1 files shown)


NoahTheDuke:
There was a tournament a while ago where corps played as an agenda and runners played as a resource, which is why this happens lol.

jwarwick:
I'm not sure I even know what it means to use a Resource as an identity :sob:
I can take out the `identity?` check and only validate that the card name exists (which was the original error in the server log).

NoahTheDuke:
Here's the tournament link: https://forum.stimhack.com/t/okorina-a-tournament-on-sunday-may-31st/10706

PR #5460, opened 2020-12-29
Add deck list filters

jwarwick:
Added ability to filter decks in the `Deck Builder` page. Similar in behavior to the `Cards` page.

image


src/cljs/nr/cardbrowser.cljs
```diff
@@ -273,6 +273,7 @@
         corp-factions ["Jinteki" "Haas-Bioroid" "NBN" "Weyland Consortium" "Neutral"]]
     (case side
       "All" (concat runner-factions corp-factions)
+      "Any Side" (concat runner-factions corp-factions)
```
NoahTheDuke, line 276: Does this need "Neutral" as well?

jwarwick: No, Neutral is listed as one of the corp factions. We only need to add it if corp isn't present.

PR #5464, opened 2021-01-01
MU Rewrite, move installed runner cards to play area first, etc

### LINKED ISSUE #5458: Dadiana Chacon trashes card before it is installed
Runner with 3 cards in hand and credits equal to the install cost of one of those cards. Dadiana Chacon installed.
Attempting to install a card from hand (which brings you to 0 credits) causes Chacon to fire which trashes all 3 cards in hand; the card that was to be installed is trashed instead.

### LINKED ISSUE #5459: Dadiana Chacon doesn't fire if credits spent installed cards
Runner has Dadiana Chacon installed. If the runner spends credits (on an Event for example) but uses some number of credits from an installed card (or recurring credits) and the remainder from the credit pool, even if the credit pool goes to 0, Dadiana Chacon will not fire and cause damage.

NoahTheDuke:
Closes #5458
Closes #5459


src/clj/game/core/change_vals.clj
```diff
@@ -24,13 +24,13 @@
...
-  (system-msg state side
-              (str "sets unused MU to " (available-mu state)
-                   " (" (if (pos? delta) (str "+" delta) delta) ")")))
+; (defn- change-mu
+;   "Send a system message indicating how mu was changed"
+;   [state side delta]
+;   (free-mu state delta)
+;   (system-msg state side
+;               (str "sets unused MU to " (available-mu state)
+;                    " (" (if (pos? delta) (str "+" delta) delta) ")")))
```
jwarwick, line 33: Any reason to leave this in and commented out rather than just deleting?

src/clj/game/core/change_vals.clj
```diff
@@ -105,7 +105,7 @@
   "Increase/decrease a player's property (clicks, credits, MU, etc.) by delta."
   [state side {:keys [key delta]}]
   (case key
-    :memory (change-mu state side delta)
+    ; :memory (change-mu state side delta)
```
jwarwick, line 108: Delete this as well?

src/clj/game/core/change_vals.clj
NoahTheDuke: Whoops, I forgot to rewrite it. Lol thanks for catching this.

PR #5468, opened 2021-01-05
move flatlining by negative handsize before discard prompt 

osiccha:


src/clj/game/core/turns.clj
```diff
@@ -94,35 +94,36 @@
...
-                                         (string/join ", " (map :title targets))
-                                         (quantify (count targets) "card"))
-                                       " from " (if (= :runner side) "their Grip" "HQ")
-                                       " at end of turn"))
-                      (doseq [t targets]
-                        (move state side t :discard))
-                      (effect-completed state side eid))}
-        nil nil)
-      (effect-completed state side eid))))
+    (if (and (= side :runner)
```
NoahTheDuke, line 97: Thanks for this! Nested `if`s are best expressed with `cond`:
```clojure
(cond
  (and (= side :runner) (neg? (hand-size state side))
  (do ...)
  (> cur-hand-size max-hand-size)
  (continue-ability ...)
  :else
  (effect-completed ...))
```

PR #5472, opened 2021-01-06
Don't keep turn-state around recursively

jwarwick:


NoahTheDuke:
How does this interact with "/undo-turn"?

jwarwick:
Should work okay, since we can only back up one turn anyway.
And it passes all the tests...

NoahTheDuke:
Hell yeah, sounds good to me.

PR #5480, opened 2021-01-10
Only query game history on load

jwarwick:


jwarwick:
Pulled these changes into larger reagent cleanup PR

PR #5482, opened 2021-01-10
Add log message for Helheim Servers

### LINKED ISSUE #5479: Helheim Servers have no log entries
When you use the activated ability on a rezzed **Helheim Servers**, all of the ICE do correctly increase in strength.  However, there is no log entry; it confused my opponent who thought it was a bug that had carried the ICE strength over from last turn. I had to have him compare my current hand size with hand size that I had finished my turn out to convince him that there was no bug.

Could we please get a log message like "X uses Helheim Servers to discard a card and increase the strength of all ICE in server Y by +2" ?

jwarwick:
Fixes #5479

### DIFF
--- a/src/clj/game/cards/upgrades.clj
+++ b/src/clj/game/cards/upgrades.clj
@@ -569,6 +569,7 @@
 
 (defcard "Helheim Servers"
   {:abilities [{:label "All ice protecting this server has +2 strength until the end of the run"
+                :msg (msg "increase the strength of all ice protecting this server until the end of the run")
                 :req (req (and this-server
                                (pos? (count run-ices))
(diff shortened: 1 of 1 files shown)


src/clj/game/cards/upgrades.clj
```diff
@@ -569,6 +569,7 @@
 
 (defcard "Helheim Servers"
   {:abilities [{:label "All ice protecting this server has +2 strength until the end of the run"
+                :msg (msg "increase the strength of all ice protecting this server until the end of the run")
```
NoahTheDuke, line 572:
```suggestion
                :msg "increase the strength of all ice protecting this server until the end of the run"
```

PR #5484, opened 2021-01-10
Handle unknown Jinteki Biotech flip identity

jwarwick:
This shows up as an exception in the server logs. I suspect you get into this state by using `/replace-id`.

### DIFF
--- a/src/clj/game/cards/identities.clj
+++ b/src/clj/game/cards/identities.clj
@@ -745,6 +745,7 @@
                                (toast state :corp "Flip to: The Tank (Shuffle Archives into R&D)" "info")
                                "The Greenhouse"
                                (toast state :corp "Flip to: The Greenhouse (Place 4 advancement tokens on a card)" "info")
+                               ;; default case
                                (toast state :corp "No flip identity specified" "info")))}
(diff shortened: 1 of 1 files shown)


src/clj/game/cards/identities.clj
```diff
@@ -738,14 +738,14 @@
...
-                :req (req (:biotech-target card))
                 :effect (req (case (:biotech-target card)
                                "The Brewery"
                                (toast state :corp "Flip to: The Brewery (Do 2 net damage)" "info")
                                "The Tank"
                                (toast state :corp "Flip to: The Tank (Shuffle Archives into R&D)" "info")
                                "The Greenhouse"
-                               (toast state :corp "Flip to: The Greenhouse (Place 4 advancement tokens on a card)" "info")))}
+                               (toast state :corp "Flip to: The Greenhouse (Place 4 advancement tokens on a card)" "info")
+                               (toast state :corp "No flip identity specified" "info")))}
```
NoahTheDuke, line 748: Mind adding a `;; else` comment? Unlabeled default branch for `case` is one of my least favorite things in Clojure.

PR #5489, opened 2021-01-13
Implement Inazuma

### LINKED ISSUE #5434: Inazuma doesn't prevent from jacking out
According to comments I could find on the [netrunnerdb page](https://netrunnerdb.com/en/card/05016), Inazuma should prevent from jacking out when it is the innermost ice.

dkellner:
I'm still unfamiliar with the codebase and not sure if `make-subs-unbreakable-until-encounter-ends` is the best solution for the temporary effect. I've tried using `register-floating-effect`, but there seems to be no effect for making subs unbreakable yet. If this would have been the right way to go, let me know and I'll implement that.

Closes #5434 .

### DIFF
--- a/src/clj/game/cards/ice.clj
+++ b/src/clj/game/cards/ice.clj
@@ -1687,29 +1687,40 @@
    :runner-abilities [(bioroid-break 2 2)]})
 
 (defcard "Inazuma"
-  {:subroutines [{:msg "prevent the Runner from breaking subroutines on the next piece of ICE they encounter this run"
-                  :effect (effect
-                            (register-events
-                              card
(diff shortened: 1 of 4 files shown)


NoahTheDuke:
Hey hey! Thanks for the pr! Glad to see someone tackling this card. Gonna leave some comments inline.

dkellner:
Works like a charm, thanks for your guidance! Introducing that new floating effect was actually easier than I imagined and more elegant than my convoluted solution.

NoahTheDuke:
I'm so glad to hear that! It's been a real boon to us so far.

PR #5495, opened 2021-01-16
Replay system

### LINKED ISSUE #3438: Game Replay/Scrubbing
Record messages sent from client to server and back to perfectly replicate a game.

Game Replay interface which allows for scrubbing forwards and backwards.

Scrubber can be set to click- or turn-resolution.

Buttons to skip-to-next click or turn.

Replay speed - auto scrub through click or turns every X seconds, or in real time (provided we capture timestamps on actions).

Game Replays available through the History/Stats page (#3435).

Can be downloaded from a game's history plug.

Can be uploaded to play archived games of your own or others' shared games from within Play (Replay button next to New Game).

Recordings take into account the "Make players' hidden information visible to spectators" selection and will only show hidden information in the recordings as such. A Game Replay file could then potentially be different for each player. This would require the "hidden information" selection to be public in a game queue listing so that players can choose whether to join a game in which their hidden information will be available in the recording.

Users watching a replay can select between showing or hiding either player's hidden information.

Users can select which perspective the replay should be played at (runner/corp) i.e. which "side of the table" the replay is watched from.

Some way to share would be nice. A publish feature could give a link to be posted into an NRDB deck description. This could be used to showcase good/winning games with said deck. Can also be posted in a youtube stream/screenshare description so that viewers can replay the game and scrub through it themselves.

lostgeek:
Finally tackling this feature.

### ToDo

**Necessary**
- [x] Make scroll back on first node go to initial state
- [x] make forward go to next log entry
- [x] auto-scroll to current node
- [x] turn end of game into own node
- [x] Truncate to last 15 replays per player
- [x] Share function with timestamps
- [x] Replay not per player but per match
- [x] Side swapping
- [x] opt-in / out for replays
- [x] remove critical data
- [x] icon for undo-click / turn
- [x] non-emoji arrows

**Later on**
- [x] Display both prompts (won't do)
- [ ] Notetaking in the replay
- [x] display undo-click and undo-turn on timeline
- [ ] Remove chat logs from stats
- [ ] "Malkovich logs"

Closes #3438


msbeck:
Awesome to see this getting implemented! Kudos

PR #5502, opened 2021-01-20
Display translated card images and text

jwarwick:
Adds a `Language` selector in the user settings page. Pre-populated with languages supported in the latest NISEI packs (Simplified Chinese, Traditional Chinese, French, German, Italian, Japanese, and Korean).

Adds an `add-language` lein task to look for card images of the format `12345[lang-prefix].png` and associate the translated card image with that card in the database.

Updates the image display routines to use the language preference to display the appropriate card. Language preferences override alternate art settings.

Text translations live in `src/cljs/nr/translations.cljs`


src/cljs/nr/translations.cljs
```diff
@@ -694,7 +745,38 @@
...
    :close-shuffle "Osecla & Uffleshay"
    :heap "Eaphay"
-   }
+   :card-count (fn [[size]] (str size " ardcay" (when (not= 1 size) "s") "."))
+   :face-down-count (fn [[total face-up]] (str total " ardscay, " (- total face-up) " acefay-ownday."))
+   :up-down-count (fn [[total face-up]] (str face-up "↑ " (- total face-up) "↓"))
+   :initiation "Initiationyay"
+   :approach-ice "Approachyay iceyay"
+   :encouter-ice "Encounteryay iceyay"
+   :pass-ice "Asspay iceyay"
```
NoahTheDuke, line 754: Asspay

D58693:
Hi!

At first: Thanks alot for all your work! Jinteki.net is an insanely good project!

I found this article recently and saw, that there is already an language-selection-box in the profile-settings. If I understand correctly, it doesn't work yet, because the translations of the cards, etc. have not yet been included. Is that correct?

Is there a way to upload translated card images? Are there any plans to implement additional languages in the near future? Do you need help with the german translation?

Greetings
D58693

NoahTheDuke:
This set of changes only changed the interface. We don't have any language-specific card images, I don't think, and we don't have the translated card data either. We rely on netrunnerdb for both of those, and I don't believe they're providing them at this time.

D58693:
Thanks for your answer.

Noah ***@***.***> schrieb am Do., 15. Dez. 2022, 18:06:

D58693:
Hi NoahTheDuke,

you said, the "set of changes only changed the interface". What exactly is changing, when I switch the language in the profile-settings? I see no difference in the interface.

The translated card data of 57 data packs is already present. It is saved in "NetrunnerDB/netrunner-cards-json/translations/de". There are 7 other languages, too.

Is it possible to add the translated card data (without image) in the card preview zoom?

Greetings
D58693

PR #5507, opened 2021-01-21
Replace regex queries with case insensitive queries

dkellner:
Currently, we're using `{$regex "^...$" $options "i"}` queries to find users by
username and email. They "cannot use indexes effectively" [0] and in our
particular case fail to escape the user input, enabling some nasty edge cases.

Since version 3.4, MongoDB supports case insensitive indexes [1], which can
replace all of these queries for better performance.

This commit consists of these changes:
1. Upgrade Monger to support the "collation" option
2. Replace all `$regex` queries
3. Add task `create-indexes`

The new queries also work on a database without the indexes created by
`create-indexes`, but adding them improves performance.

[0] https://docs.mongodb.com/manual/reference/operator/query/regex/
[1] https://docs.mongodb.com/manual/core/index-case-insensitive/

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -10,7 +10,7 @@
                  [org.clojure/core.async "0.3.443"]
                  [cheshire "5.6.3"]
                  [stylefruits/gniazdo "1.1.4"]
-                 [danhut/monger "3.1.0"]
+                 [com.novemberain/monger "3.5.0"]
                  [differ "0.3.3"]
                  [com.taoensso/sente "1.16.0"]
                  [ring "1.7.1"]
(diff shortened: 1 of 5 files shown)


NoahTheDuke:
This is great! Thank you so much for researching this. Should we be worried about the failing match on special characters?

dkellner:
Right now, maybe :-). The failing matches basically mean that it is currently possible to register multiple users with the same username (e.g. `$foo`, or any other regex that never matches), because the check if the username already exists will always fail. But the actual login does not use a `$regex` query atm, meaning only the first such user can actually login.

With the new queries these usernames can still be used, but will be properly checked if they already exist.

PR #5511, opened 2021-01-25
Reload game history on game end

jwarwick:
Fix missing Neutral icon. 
Fix read-only reagent warning in replay link input.

### DIFF
--- a/src/cljs/nr/stats.cljs
+++ b/src/cljs/nr/stats.cljs
@@ -17,7 +17,7 @@
 
 (defn- fetch-game-history []
   (go (let [{:keys [status json]} (<! (GET "/profile/history"))
-            games (map #(assoc % :start-date (js/Date. (:start-date %))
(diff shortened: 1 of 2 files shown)


src/cljs/nr/stats.cljs
```diff
@@ -13,6 +13,15 @@
             [nr.ws :as ws]
             [reagent.core :as r]))
 
+(def state (r/atom {:games nil}))
+
+(defn- fetch-game-history []
+  (go (let [{:keys [status json]} (<! (GET "/profile/history"))
+            games (map #(assoc % :start-date (js/Date. (:start-date %))
```
NoahTheDuke, line 20:
```suggestion
            games (mapv #(assoc % :start-date (js/Date. (:start-date %))
```
`map` is lazy and we wanna make sure we "realize" the whole collection before inserting it into the state atom.

PR #5536, opened 2021-02-05
Add compound indexes on game-logs for usernames and start-date

dkellner:
Mongo cannot use the separate index on `start-date` for sorting after filtering
on the usernames, leading to slow execution times or memory exhaustion for
users with a lot of game-logs.


src/clj/tasks/db.clj
```diff
@@ -83,9 +83,8 @@
      ["cards" (array-map :type 1)]
      ["decks" (array-map :username 1)]
      ["game-logs" (array-map :gameid 1)]
```
NoahTheDuke, line 85: Should you include the `:start-date -1` here as well?

NoahTheDuke: Coolio, that's good to know.

PR #5537, opened 2021-02-05
Sample data and query profiling

dkellner:
This adds a new task `create-sample-data` and a section to the new `DEVELOPMENT.md` about how to use it to profile database queries locally.

Let me know if you find my writeup useful. I was not sure how much to "duplicate" from various sources and went for a basic "tutorial" on how to identify missing indexes and create them in our project.

We could also add the contents of `DEVELOPMENT.md` to the wiki, but I feel this kind of documentation should be as close to the code / developer as possible. Some already outdated information in the wiki supports this feeling ;-). And this way, if you grep for `create-sample-data` in the repo because you change its name, you will find the docs as well.


NoahTheDuke:
This is very cool! Thank you so much for writing it up. I think this is a great level of detail and will be helpful to everyone who interacts with database queries.

PR #5538, opened 2021-02-06
Add Chinese translation

bbbbbbbbba:
There are some things that couldn't be translated perfectly by modifying this strings; for example, there are quite a few colons and spaces added to those strings, and as you may know, Chinese usually uses full width colons ("：") and no spaces. But I think what I could do should be adequate.

I tried my best to check that every translation makes sense in context, but there are some strings I couldn't get the game to display at all (mostly error handlers like "Not allowed" or "No current run"), as well as some that I don't understand fully (e.g. what's the purpose for the admin-only "Feature" tab)?

Feedback, especially from Chinese speakers, is _very_ welcome. That includes minor nitpicks.


darmac (approved):

jwarwick:
That's amazing work. Will try to get this merged ASAP.
Some string will be tricky to generate, and there isn't any context (even if you look at the code).
"Features" is actually a blank page at the moment, I intend to have admin configurable values in there; things like number of replays to save, etc.

I didn't think about spaces/colons. I can probably pull more of the string construction into the translation file if that's useful.

PR #5547, opened 2021-02-10
Don't double count credits spent stats

jwarwick:
Previously counting both when picking counters from cards and for total spent value.


NoahTheDuke:
Nice tests!

PR #5556, opened 2021-02-12
WIP: Better start page for jnet

lostgeek:


NoahTheDuke:
I'm still thinking about this. Would love to see more done with it if you feel like spending the time on it.

PR #5585, opened 2021-02-18
Fixed cards disappearing from R&D in replays

### LINKED ISSUE #5553: Cards (particularly agendas) disappear during replays
Not sure why this is happening, but on Corp turn 9 in this replay  they should have drawn Cyberdex Sandbox, but instead they draw the Junebug below it, and the Sandbox disappears. By the end of the game another agenda has disappeared.
https://www.jinteki.net/replay/ff396031-1389-4e05-831d-d25c96cc9716

lostgeek:
The initial state was prepared incorrectly, which lead to the diffs doing weird things to R&D (drawing from the bottom, duplicating cards, etc.)

Unfortunately old replays are not fixed by this bugfix and will continue showing R&D incorrectly.

Closes #5553

### DIFF
--- a/src/clj/game/core/diffs.clj
+++ b/src/clj/game/core/diffs.clj
@@ -1,19 +1,28 @@
-(ns game.diffs
-  (:require [game.core :refer [card-is-public?]]
+(ns game.core.diffs
+  (:require [game.core.flags :refer [card-is-public?]]
             [game.core.card :refer [private-card]]
-            [game.core.set-up :refer [strip-for-replay]]
             [game.utils :refer [dissoc-in]]
             [differ.core :as differ]))
 
 (defn- strip [state]
   (-> state
-    (dissoc :eid :events :turn-events :per-turn :prevent :damage :effect-completed :click-state :turn-state :history)
-    (update-in [:corp :register] select-keys [:spent-click])
-    (update-in [:runner :register] select-keys [:spent-click])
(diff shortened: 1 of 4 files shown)


src/clj/game/core/diffs.clj
```diff
@@ -65,7 +74,7 @@
      (if (get-in @state [:options :spectatorhands])
        (assoc @state :corp corp-deck :runner runner-deck)
        (assoc @state :corp corp-private :runner runner-private))
-     (assoc @state :corp corp-deck :runner runner-deck)]))
+     @state]))
```
NoahTheDuke, line 77: What's this change do?

lostgeek: I first fixed the broken behavior with the change in `set_up.clj:130`. Now the initial state fits to the diffs that I save. But in this correct initial state the cards in the decks are not face-up, since that's also not the way it's done in game.

`corp-private` contains closed info for both R&D and HQ. `corp-deck` is the same, but with open info for HQ. R&D is only open info if `:view-deck` is set in `(:corp @state)`.

Therefore to get open info for R&D all the time, I need to grab the state before any info gets stripped, and that's simply just `@state`. That fourth entry of `private-states` is then available under `(:hist-state (public-states @state))` after getting stripped of other sensible information like the blocked-users list.

NoahTheDuke: excellent, thanks

PR #5586, opened 2021-02-18
Paperclip only shows auto-pump for Barriers

jwarwick:


NoahTheDuke:
jesus, these are getting complicated, lol

PR #5595, opened 2021-02-24
Convert @all-cards in client to use map instead of list

NoahTheDuke:


src/cljs/nr/gameboard.cljs
```diff
@@ -2234,7 +2232,7 @@
                                  #(card-highlight-mouse-over % value button-channel)
                                  :on-mouse-out
                                  #(card-highlight-mouse-out % value button-channel)
-                                 :id {:code value}}
+                                 :id (:title value)}
```
jwarwick, line 2235: Intentional change from a map to a function call?

NoahTheDuke: Yep. I honestly don't know how it was working before because it was setting the `id` string value in the HTML to `[object Object]` lol.

PR #5601, opened 2021-02-25
navigation in replays

osiccha:
ToDo

- use keyboard to navigate in a replay
- automatically display each change


src/cljs/nr/gameboard.cljs
```diff
@@ -1,6 +1,6 @@
 (ns nr.gameboard
-  (:require-macros [cljs.core.async.macros :refer [go]])
-  (:require [cljs.core.async :refer [chan put! <!] :as async]
+(:require-macros [cljs.core.async.macros :refer [go ]])
```
NoahTheDuke, line 2: Looks like the indentation was missed here, and there's some spacing left after `go`.

PR #5606, opened 2021-02-27
Add Artist Info

jwarwick:
Artist blurb box in the card browser. New section in About.
All data is populated from the promos.edn file.


NoahTheDuke:
This looks great!

PR #5610, opened 2021-03-04
High res card images

### LINKED ISSUE #5523: unable to see bottom of screen on tablet since update
Hi,

I just tried playing on a tablet for the first time since the jnet update, and i can no longer see or access the bottom of the screen at all. I can't scroll down to the bottom of the page, and if i zoom in and drag it there is just a black line there (see screenshot below). It's quite possibly a problem my end, but nothing has changed to my tablet as far as I know, so wondering if it might be something to do with the jnet update?
 
Edit: i have tried it on chrome and kiwi browsers.

Screenshot_20210129-122403_Chrome

jwarwick:
Restructure card image handling to support multiple languages, multiple image resolutions, and alternate art options.

Fixes #5523 (unrelated, but this PR messes with a bunch of css anyway)


src/clj/tasks/nrdb.clj
```diff
@@ -95,7 +92,7 @@
 
 (defn fetch-data
   [{:keys [card-images db local]}]
-  (let [edn (download-edn-data local)]
+  (let [edn (dissoc (download-edn-data local) :promos)]
```
NoahTheDuke, line 95: We can just remove the `:promos` data from netrunner-data.

jwarwick: Agree. Until it is removed from `netrunner-data` I need to filter it. After it's gone we should take this out.

src/clj/tasks/nrdb.clj
```diff
@@ -61,13 +61,17 @@
...
+(def ^:const cards-to-skip #{"08012" "09001" "26066" "26120"})
+
 (defn download-card-images
   "Download card images (if necessary) from NRDB"
   [cards]
   (let [img-dir (io/file "resources" "public" "img" "cards" "en" "default" "stock")]
     (io/make-parents img-dir)
     (let [previous-cards (generate-previous-card-stubs cards)
           total-cards (concat cards previous-cards)
+          total-cards (remove #(get cards-to-skip (:code %)) total-cards)
```
NoahTheDuke, line 74: This won't download those card images at all. Is this intended?

jwarwick: We don't want to download anything for the cards with multiple faces. Since we don't have a way to get all the faces. Currently those are in the repo. The way the image code works, it will prioritize a non-face image over a face image, so we don't want those in the path.

NoahTheDuke: Okay. You'll have to explain it in greater detail at some point.

Issue #4376, opened 2019-08-12
Inconsistency in the UI


lostgeek:
![Screenshot from 2019-08-12 09-18-25](https://user-images.githubusercontent.com/1409906/62850165-61be4100-bce2-11e9-8801-b789e5fc0b29.png)
- The textbox should be at the bottom of the whole server. Possibly happens because of the root of a central being handled differently.

![Screenshot from 2019-08-12 09-19-10](https://user-images.githubusercontent.com/1409906/62850296-bc579d00-bce2-11e9-9388-d69eb1121494.png)
![Screenshot from 2019-08-12 09-19-22](https://user-images.githubusercontent.com/1409906/62850300-bd88ca00-bce2-11e9-8cfd-1b4d675c3a84.png)
- Zones should have the same font size as other areas.

dkellner:
@NoahTheDuke I think this can be closed now.

PR #5633, opened 2021-03-12
Fix UI inconsistencies

dkellner:
Addresses #4376 and one other inconsistency I've found: previously Grip and HQ did not have a label showing the amount of cards. Especially for HQ I think this makes a lot of sense.

The three commits are independent of one another, so feel free to just merge some of them! I actually was just too lazy to create multiple sets of screenshots and kind of hoping you'll like all of the changes, anyway ;-).

jnet-ui-inconsistencies-pr
(I was using two different browsers, hence the slightly different fonts...)


NoahTheDuke:
Thanks for these!

PR #5634, opened 2021-03-13
Latest SG Spoilers

jwarwick:


src/clj/game/cards/events.clj
```diff
@@ -2036,6 +2036,15 @@
...
                        (unregister-events state side card {:events [{:event :runner-phase-12}]})))}))
 
+(defcard "Overclock"
+  {:async true
+   :makes-run true
+   :data {:counter {:credit 5}}
+   :interactions {:pay-credits {:type :credit}}
+   :prompt "Choose a server"
+   :choices (req runnable-servers)
+   :effect (effect (make-run eid target nil card))})
```
NoahTheDuke, line 2046:
```suggestion
   :effect (effect (make-run eid target card))})
```

src/clj/game/cards/identities.clj
```diff
@@ -1651,3 +1669,21 @@
...
                              (shuffle! :runner :deck)
                              (trigger-event :searched-stack nil))}]})
+
+(defcard "Zahya Sadeghi: Versatile Smuggler"
+  {:events [{:event :run-ends
+             :req (req (and (or (= :hq (target-server context))
+                                (= :rd (target-server context)))
+                            (pos? (total-cards-accessed context))))
+             :effect (effect (continue-ability
+                               (let [cards-accessed (total-cards-accessed context)]
```
NoahTheDuke, line 1679: This is exposed in the optional, because we pass in `targets` to both the prompt and the `yes-ability`. So the `continue-ability` nesting can be removed so it's purely `{:event :run-ends :optional {:req ...`

src/clj/game/cards/operations.clj
```diff
@@ -1544,6 +1549,21 @@
...
+   :msg (msg "force the Runner to " (decapitalize target))
+   :prompt "Pick one"
+   :choices (req ["Take 1 tag"
+                  (when (can-pay? state :runner (assoc eid :source card :source-type :ability) card (:title card) :credit 8)
+                    "Pay 8 [Credits]")])
+   :effect (req (if (= target "Pay 8 [Credits]")
+                  (wait-for (pay state :runner card :credit 8)
+                            (system-msg state :runner (:msg async-result))
+                            (effect-completed state side eid))
+                  (gain-tags state :corp eid 1 nil)))})
```
NoahTheDuke, line 1565:
```suggestion
                  (gain-tags state :corp eid 1)))})
```

src/clj/game/cards/upgrades.clj
```diff
@@ -23,6 +23,18 @@
...
+(defcard "Anoetic Void"
+  {:events [{:event :successful-run
+             :optional
+             {:prompt "Pay 2 [Credits] and trash 2 cards from HQ to end the run?"
+              :req (req (and (can-pay? state side eid card nil [:credit 2 :trash-from-hand 2])
+                             this-server))
+              :yes-ability
+              {:async true
+               :msg "pay 2 [Credits] and trash 2 cards from HQ to end the run"
+               :effect (req (wait-for (pay state :corp card [:credit 2 :trash-from-hand 2])
```
NoahTheDuke, line 35: This should be `:cost [:credit 2 :trash-from-hand 2]` and then the cost part can be removed from `:msg`, allowing for spending hosted credits to show up in the log.

PR #5636, opened 2021-03-16
Add Akira alt arts to promos

jwarwick:


NoahTheDuke:
Given that `:position` is updated every time something is added, do we even want to track it? Whatever we're using it for could be handled by either sorting or just using as presented, right?

Issue #5628, opened 2021-03-11
[Bug] In the Groove not giving resolving choice with Hayley Kaplan: Universal Scholar


Joseph7695:
No choice of effect to resolve, forced to resolve Hayley first

NoahTheDuke:
Thanks for the report!

francescopellegrini:
Fixed by #5640?

PR #5640, opened 2021-03-17
Update 'In the Groove' for Hayley Kaplan interaction

merrickbrown:
This is in response to issue #5628

### DIFF
--- a/src/clj/game/cards/events.clj
+++ b/src/clj/game/cards/events.clj
@@ -1386,7 +1386,12 @@
   {:events [{:event :runner-install
              :duration :end-of-turn
              :req (req (<= 1 (:cost target)))
-             :interactive (req (has-subtype? target "Cybernetic"))
+             :interactive (req (or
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
Thank you so much for tackling this issue! I don't want to get into the habit of making exceptions for single cards, so I think a better option here is to just remove the Hayley part and leave the `first-event?` check. That way, it will still pause for Hayley, and we can look at other interactions as they come in.

merrickbrown:
Yep - makes sense, I noticed there weren't any such exceptions (eg no tests for :title in conditions) so I'm not surprised this is a policy. I went ahead and updated the :interaction value as requested

merrickbrown:
One question, is it preferable to target `dev` or `master` for changes?

NoahTheDuke:
Master. A long long time ago, we tried to use dev as a staging/working branch, but it was too difficult to keep everything in sync so we abandoned it. I should probably remove that branch.

PR #5660, opened 2021-03-20
Made Precision Design non-mandatory + centralize corp recursion

### LINKED ISSUE #5651: HB:Precision Design effect display error
When the corp scores an agenda and triggers the ability of his ID(HB:Precision Design) it is never displayed to the runner which card was returned to hand. 

According to my understanding of the rules this information should only be hidden if the card returned from archives was facedown (in archives).

merrickbrown:
Also Fixes #5651

### DIFF
--- a/src/clj/game/core/def_helpers.clj
+++ b/src/clj/game/core/def_helpers.clj
@@ -1,17 +1,17 @@
 (ns game.core.def-helpers
   (:require
-    [game.core.card :refer [corp? get-card get-counters has-subtype?]]
+    [game.core.card :refer [corp? get-card get-counters has-subtype? in-discard? faceup?]]
     [game.core.damage :refer [damage]]
     [game.core.eid :refer [effect-completed]]
     [game.core.engine :refer [resolve-ability trigger-event-sync]]
     [game.core.gaining :refer [gain-credits]]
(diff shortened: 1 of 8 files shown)


src/clj/game/core/def_helpers.clj
```diff
@@ -179,6 +179,18 @@
...
+(defn corp-recur
+  ([] (corp-recur (constantly true)))
+  ([pred]
+   {:label "add card from Archives to HQ"
+    :prompt "Select a card to add to HQ"
+    :show-discard true
+    :choices {:card #(and (corp? %)
+                       (in-discard? %)
+                       (pred %))}
+    :msg (msg "add " (if (faceup? target) (:title target) "an unseen card") " to HQ")
```
NoahTheDuke, line 191: Wanna try something for me? Change the `(if (faceup? ...` part of the msg to instead use `card-str` and see how that reads. If it reads like shit, leave this as is.

NoahTheDuke:
This is a great PR! Thank you so much!

src/clj/game/core/def_helpers.clj
merrickbrown: Switched it over. Reads OK to me.

PR #5661, opened 2021-03-21
Move agenda points and hand size to more natural positions

dkellner:
This PR combines three small changes to the gameboard:

- Moving the "Agenda Points" line from the stats table to the scored area, directly under the scored agendas.
- Moving the "Max hand size" line from the stats table to the hand popup.
- Adding the current hand size to the hand panel's label (e.g. "HQ (4/5)").

See the screenshots below for details. I know we've discussed adding UI changes behind an option in settings, but I felt these changes here should be uncontroversal enough to just go through with them. For the other more radical changes (icons instead of text), I thought about introducing a "Compact UI" setting.

jnet-stats-ui
jnet-stats-ui-new-popup

### DIFF
--- a/src/cljs/nr/gameboard.cljs
+++ b/src/cljs/nr/gameboard.cljs
@@ -1423,6 +1423,14 @@
      (str (get-in opts [:opts :name])
           (when (not (get-in opts [:opts :hide-cursor])) (str " (" (fn cursor) ")")))]))
 
+(defn controls
+  "Create the control buttons for the side displays."
+  ([key] (controls key 1 -1))
+  ([key increment decrement]
+   [:div.controls
+    [:button.small {:on-click #(send-command "change" {:key key :delta decrement}) :type "button"} "-"]
+    [:button.small {:on-click #(send-command "change" {:key key :delta increment}) :type "button"} "+"]]))
+
 (defn- this-user?
   [user]
   (if (:replay @game-state)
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
This looks great! Where are the buttons to manually adjust hand size and agenda total?

dkellner:
Thanks! They are right next to the labels, but only shown when you are hovering over it. Just like it is in the stats panel.

NoahTheDuke:
That's great to hear.

NoahTheDuke (approved):

francescopellegrini:
If I may, I think that having both the Scored Area and Agenda Points totals show the same number is confusing: all other zones have their total correspond to the number of cards present.

I suggest to have the Scored Area total show the number of cards present while the Agenda Points total show the number of agenda points (duh).

dkellner:
You're right, good point! I will change that.

dkellner:
Done. It shows the number of cards again, e.g. "Scored Area (1)" for one agenda, no matter how many agenda points it has.

I've rebased on the current `master` as well.

src/cljs/nr/gameboard.cljs
```diff
@@ -1470,6 +1479,9 @@
             [:div
              [:a {:on-click #(close-popup % (:hand-popup @s) nil false false)} (tr [:game.close "Close"])]
              [:label (tr [:game.card-count] size)]
+             (let [{:keys [total]} @hand-size]
+               [:div.hand-size (str total " " (tr [:game.max-hand "Max hand size"]))
+                (controls :hand-size)])
```
NoahTheDuke, line 1484: Does this need a `(when me?` check?

dkellner: It's inside a `(when popup`, which is only `true` for your side as far as I can see. There's the special case with spectators, but they see (non-functional) controls now already.

NoahTheDuke: Cool, thanks!

PR #5679, opened 2021-03-24
Make project more docker friendly

jwarwick:


src/clj/tasks/db.clj
```diff
@@ -108,17 +108,18 @@
...
-  (try
-    (do
-      (doseq [index-args indexes]
-        (apply mc/create-index db index-args))
-      (println "Indexes successfully created."))
-    (catch Exception e (do
-                         (println "Create indexes failed" (.getMessage e))
-                         (.printStackTrace e)))
-    (finally (webdb/disconnect))))
+  ([] (create-indexes true))
```
NoahTheDuke, line 111: Did you mean to delete this function or maybe the whole file?

jwarwick: What the heck. Thanks for catching that.

PR #5682, opened 2021-03-24
Fix repeated text on Fully Operational's on-play message

locks:
Fixes mtgred#5681 by removing the repeated text.


NoahTheDuke:
Is this good to go?

locks:
@NoahTheDuke I wanted to add a test, but you can merge!

Issue #5685, opened 2021-03-25
Leech gives the runner the option to place a virus token, but the ability is mandatory.


wern212:
[Leech](https://netrunnerdb.com/en/card/30008) gives the runner a choice to place a virus counter (or not) when making a successful run. However, the ability is mandatory.

![image](https://user-images.githubusercontent.com/55791889/112521179-83c3e480-8d9c-11eb-89ee-2dc4f388f067.png)

![image](https://user-images.githubusercontent.com/55791889/112521192-88889880-8d9c-11eb-96a8-11cf0b5c690a.png)

NoahTheDuke:
Thanks for the report!

PR #5689, opened 2021-03-25
Make Leech event on successful run mandatory

locks:
Closes https://github.com/mtgred/netrunner/issues/5685

### DIFF
--- a/src/clj/game/cards/programs.clj
+++ b/src/clj/game/cards/programs.clj
@@ -1537,17 +1537,9 @@
 
 (defcard "Leech"
   {:events [{:event :successful-run
-             :interactive (get-autoresolve :autofire (complement never?))
-             :silent (get-autoresolve :autofire never?)
-             :optional
-             {:player :runner
-              :req (req (is-central? (target-server context)))
-              :waiting-prompt "Runner to decide if they will use Leech"
-              :autoresolve (get-autoresolve :auto-fire)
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/programs.clj
```diff
@@ -1537,17 +1537,10 @@
...
-             :optional
-             {:player :runner
-              :req (req (is-central? (target-server context)))
-              :waiting-prompt "Runner to decide if they will use Leech"
-              :autoresolve (get-autoresolve :auto-fire)
-              :prompt "Use Leech?"
-              :yes-ability {:msg "add 1 virus counter to Leech"
-                            :effect (req (add-counter state side card :virus 1))}
-              :no-ability {:effect (req (system-msg state side "does not add counter to Leech"))}}}]
+             :interactive (get-autoresolve :auto-fire (complement never?))
```
NoahTheDuke, line 1540:
```suggestion
             :interactive (req true)
```

NoahTheDuke, line 1540: Oops, I led you astray! You can remove this and the `:silent` entirely. They're only needed with `:optional` stuff.

locks: I removed both and added `:req` which I believe is necessary? Going to take a go at tests next!

locks:
Ready for review, I believe!

PR #5699, opened 2021-03-27
Convert `:add-counter` to use checkpoints

NoahTheDuke:


NoahTheDuke:
Closing this cuz I don't think it's workable in its current form. Unlike async functions like `runner-install`, `add-counters` isn't always awaited and thus we can't expect that a given ability will call `checkpoint` after invocation. This will probably have to be one of the last things I convert.

PR #5700, opened 2021-03-27
Some small styling and UI updates. Addressed some missing imports from #5701

### LINKED ISSUE #5609: Console scrolls up when leaving Play window
This looks like a regression bug.
When you leave Play window during play, for example checking Help window, when you come back, console has scrolled all the way to the top.
Also seeing, that if you resize window and go back to full screen, console will not have moved while being in windowed mode (not sure if those are the same issue)

### LINKED ISSUE #5703: Unable to scroll down completely in the cards tab
In the card tab, the last cards are cut and I can't scroll down further. Zooming in/out doesn't change anything.
I'm using Firefox 78.9.0esr 
Screenshot from 2021-03-28 10-19-38

merrickbrown:
- fixes #5609 - also resizes log when switching pages

- Changed the scrollbar styling to be a bit more in-line with the rest of the site (works on chrome, mac will allow it with system settings, but their scrollbars aren't so garish - firefox is a maybe I haven't checked it yet - I can't imagine this change will break it though)
settings-scroll
scroll-hover
^ scroll bar hovered/active

- 'Add to Deck' button is disabled in the deck builder when no card input (before it would error out or include a card with no title)
- Fixes #5703

### DIFF
--- a/src/cljs/nr/deckbuilder.cljs
+++ b/src/cljs/nr/deckbuilder.cljs
@@ -460,7 +460,7 @@
 (defn handle-add [s card-state event]
   (.preventDefault event)
   (let [qty (str->int (:quantity @card-state))
-        card (nth (:matches @card-state) (:selected @card-state))
+        card (nth (:matches @card-state) (:selected @card-state) nil)
         best-card (lookup (:side card) card)]
     (if (js/isNaN qty)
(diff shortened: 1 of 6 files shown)


merrickbrown:
Now fixes #5703

PR #5710, opened 2021-03-28
Added System Gateway and Startup formats

jwarwick:


NoahTheDuke (approved):
Looks great. We really gotta write some tests for this thing haha.

PR #5718, opened 2021-03-30
Enforce Actions

NoahTheDuke:


src/clj/game/core/actions.clj
```diff
@@ -32,11 +32,19 @@
 ;;; Neutral actions
 (defn- do-play-ability [state side card ability ability-idx targets]
   (let [cost (card-ability-cost state side ability card targets)]
-    (when (or (nil? cost)
-              (can-pay? state side (make-eid state {:source card :source-type :ability :source-info {:ability-idx ability-idx}}) card (:title card) cost))
+    (when (and (or (not (:action ability))
+                   (and (:action ability)
+                        (not (:resolving-action @state))))
```
merrickbrown, line 37: !P or (P and Q)  <=> (!P or P) and (!P or Q) <=> (!P or Q)

So I think this can be simplified to 
```clojure
(and 
  (or (not (:action ability)) (not (:resolving-action @state)))
  (or (nil? cost) (can-pay? ...)))

```

NoahTheDuke: Hell yeah, thanks.

Issue #5716, opened 2021-03-29
Harishchandra Ent. does not reveal the Runner's Grip


Waverhouse:
With the new update changing the Standard rotation it appears that the NBN Harishchandra ID is bugged. Was in a game where when the Runner became tagged through Ping or Funhouse (and then later, MCA Informant) the grip was not revealed. The Runner's cards in their grip were highlighted in blue, but did they not 'flip' over to reveal themselves, and hovering over them didd not reveal them either. It appears that this also caused the game to lose track of the Runner's grip - noticed later on that the runner's grip showed '(0/5)' even though the UI showed them with cards in their grip.

NoahTheDuke:
Thanks for the report!

PR #5729, opened 2021-03-31
Fix :openhands for runner

merrickbrown:
I don't think that #5716 was totally resolved by #5719 - the Harishchandra Ent. issue still occurred (I think) - this should fix that problem.

### DIFF
--- a/src/cljc/game/core/card.cljc
+++ b/src/cljc/game/core/card.cljc
@@ -325,7 +325,7 @@
   "Checks if the two cards are the same by :cid. Alternatively specify 1-function to use to check the card"
   ([card1 card2] (same-card? :cid card1 card2))
   ([func card1 card2]
-    (= (func card1) (func card2))))
+   (and (func card1) (func card2) (= (func card1) (func card2)))))
 
 (defn get-card-hosted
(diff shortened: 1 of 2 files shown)


src/cljc/game/core/card.cljc
```diff
@@ -325,7 +325,7 @@
   "Checks if the two cards are the same by :cid. Alternatively specify 1-function to use to check the card"
   ([card1 card2] (same-card? :cid card1 card2))
   ([func card1 card2]
-    (= (func card1) (func card2))))
+   (and (func card1) (func card2) (= (func card1) (func card2)))))
```
NoahTheDuke, line 328: If we're calling each function twice, let's use a `let` block. Something like:
```suggestion
   (let [r1 (func card1) r2 (func card2)] (and r1 r2 (= r1 r2))))))
```

jwarwick:
Ah, whoops. I blanked on the Harishchandra Ent. part of the issue. Thanks for catching that!

src/cljc/game/core/card.cljc
merrickbrown: nice catch, thanks!

PR #5733, opened 2021-03-31
Don't include card images in JAR files.

jwarwick:


project.clj
```diff
@@ -76,9 +76,12 @@
   :aot [#"game\.*"
         #"web\.*"
         #"tasks.fetch"
-        #"jinteki\.*"]
+        #"jinteki\.*"
+        #"web.core"]
```
NoahTheDuke, line 80: Isn't this handled by the `#"web\.*"` inclusion above?

jwarwick: Sadly, it still generates a warning message if the `main` function isn't called out explicitly.

jwarwick: But, yes, it was being included in the AOT from the regex.

NoahTheDuke: Oh hell yeah, that's annoyed me forever.

PR #5739, opened 2021-04-02
Added start-of-turn ability to "Smartware Distributor"

### LINKED ISSUE #5738: Runner Cannot Manually Drip Smartware Distributor
At the start of turn, the runner can click Daily Casts to order the drip to coincide with a timely Aesop's Pawnshop sale.

When you click Smartware Distributor during the start-of-turn window, it spends a click and places three credits on it.

merrickbrown:
Fixes #5738

### DIFF
--- a/src/clj/game/cards/resources.clj
+++ b/src/clj/game/cards/resources.clj
@@ -2345,16 +2345,21 @@
      :abilities [ability]}))
 
 (defcard "Smartware Distributor"
-  {:abilities [{:cost [:click 1]
-                :msg "place 3 [Credits]"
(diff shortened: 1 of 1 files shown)


merrickbrown:
I don't love this fix because this has the 'Take 1 [Credits]" ability on the card whenever it is clicked on - is there a way to only show that ability when we're in phase 1.2?

NoahTheDuke:
Sadly, no. Our Star of Turn stuff is a mess. The best thing to do is put "(start of turn)" in the label (which you've done), so folks know not to click on it unless it's the start of turn.

PR #5741, opened 2021-04-02
Added overflow to corp's board area. 

### LINKED ISSUE #5732: No horizontal scroll bar - can't use remote servers
jnet scroll problem 2

Hello. I was playing last night as the Corp and I had created 7 remote servers, but I couldn't scroll over to use servers 4 to 7. They were hidden behind the chat box, and server 7 was off screen completely. There was no way from me to scroll over to reach them. Would it be possible to get this fixed? Thank you!

merrickbrown:
This is a bit hacky due to the fact that you can't actually set overflow-x and overflow-y independently of each other (eg overflow-y: visible is not valid with overflow-x: auto) https://www.w3.org/TR/css-overflow-3/#overflow-properties 'computed values' 
This just gives enough padding to the corp-board area to have any content and subtracts a margin so that the content isn't too tall

I'm not 100% that this fix is worth it since you can just zoom out like a madman to see those servers. Fixes #5732

### DIFF
--- a/src/cljs/nr/gameboard/main.cljs
+++ b/src/cljs/nr/gameboard/main.cljs
@@ -1044,42 +1044,44 @@
 
 (defn board-view-corp [player-side identity deck deck-count hand hand-count discard servers run]
   (let [rs (:server @run)
-        server-type (first rs)]
-    [:div.corp-board {:class (if (= player-side :runner) "opponent" "me")}
-     (doall
-       (for [server (reverse (get-remotes @servers))
-             :let [num (remote->num (first server))
-                   similar-servers (filter #((compare-servers-for-stacking server) %) (get-remotes @servers))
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
I feel like @lostgeek ran into this a long time ago. Way back, you couldn't see servers behind the log pane, so he (or maybe it was someone else) changed it so they're visible underneath, but it looks like that breaks horizontal scrolling. I hope this fixes it!

merrickbrown:
FWIW,  after this change servers will not be visible under the log pane - there may be a way to achieve both but it's likely pretty hacky

merrickbrown:
Corp view (eg with scroll bar)
corp-with-scroll-view
Runner view
runner-with-scroll-view

PR #5742, opened 2021-04-02
Fix ping on-rez ability

### LINKED ISSUE #5730: Ping tags the runner if rezzed during a run on a different server (e.g. Send a Message)
If a Ping on a server other than the attacked server is rezzed during a run, (in this case a Ping on HQ rezzed because the runner stole a Send A Message from R&D) it will still tag the runner.

merrickbrown:
Ping's on-rez ability now checks that the run is against the server it is installed in. Fixes #5730

### DIFF
--- a/src/clj/game/cards/ice.clj
+++ b/src/clj/game/cards/ice.clj
@@ -2672,7 +2672,7 @@
    :strength-bonus (req (if (wonder-sub card 3) 5 0))})
 
 (defcard "Ping"
-  {:on-rez {:req (req (= (first (:server run)) (second (get-zone card))))
+  {:on-rez {:req (req (and run this-server))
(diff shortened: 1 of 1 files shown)


src/clj/game/cards/ice.clj
```diff
@@ -2672,7 +2672,7 @@
    :strength-bonus (req (if (wonder-sub card 3) 5 0))})
 
 (defcard "Ping"
-  {:on-rez {:req (req run)
+  {:on-rez {:req (req (= (first (:server run)) (second (get-zone card))))
```
NoahTheDuke, line 2675:
```suggestion
  {:on-rez {:req (req (and run this-server))
```

PR #5757, opened 2021-04-04
API Access to Game State

jwarwick:
Creation of an api-key to allow 3rd party extensions to access a user's game state.

### DIFF
--- a/src/clj/web/api.clj
+++ b/src/clj/web/api.clj
@@ -11,6 +11,7 @@
             [web.tournament :as tournament]
             [web.decks :as decks]
             [web.api-keys :as api-keys]
+            [web.game-api :as game-api]
             [compojure.route :as route]
             [ring.middleware.params :refer [wrap-params]]
             [ring.middleware.keyword-params :refer [wrap-keyword-params]]
@@ -55,7 +56,8 @@
            (POST "/register" [] auth/register-handler)
            (POST "/login" [] auth/login-handler)
            (POST "/forgot" [] auth/forgot-password-handler)
-           (POST "/reset/:token" [] auth/reset-password-handler))
+           (POST "/reset/:token" [] auth/reset-password-handler)
+           (GET "/game/deck" [] game-api/deck-handler))
 
 (defroutes admin-routes
            (POST "/admin/announce" [] admin/announce-create-handler)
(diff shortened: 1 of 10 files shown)


src/clj/web/lobby.clj
```diff
@@ -254,13 +254,14 @@
 (defn handle-lobby-create
   [{{{:keys [username] :as user} :user} :ring-req
     client-id                           :client-id
-    {:keys [title format allow-spectator save-replay spectatorhands password room side]} :?data :as event}]
+    {:keys [title format allow-spectator api-access save-replay spectatorhands password room side]} :?data :as event}]
```
merrickbrown, line 257: Maybe this happens elsewhere/as a matter of course, but should we check the that the user has-api-keys on the backend against the value passed from the frontend?

jwarwick: I'm really only using the `has-keys` to make the UI clearer (so someone doesn't check the box and not know how to make it work). Any real authorization will happen when a request comes in on the (yet-to-be-written) API.

jwarwick: Maybe another point that isn't clear - you need to have an API key and check the box in the Create Game screen to make the API work. So you can create games without allowing access with the API key.

src/clj/web/api.clj
```diff
@@ -106,8 +113,8 @@
 
 (defroutes routes
   (wrap-routes private-routes wrap-anti-forgery)
-  (wrap-routes public-CSRF-routes wrap-anti-forgery)
-  public-routes)
+  public-routes
```
NoahTheDuke, line 116: Any reason for this?

src/cljs/nr/gamelobby.cljs
```diff
@@ -528,7 +531,20 @@
...
+         (let [has-keys (:has-api-keys @user false)]
+           [:p
+            [:label
+             [:input {:disabled (not has-keys)
+                      :type "checkbox" :checked (:api-access @s)
+                      :on-change #(swap! s assoc :api-access (.. % -target -checked))}]
+             (tr [:lobby.api-access "Allow API access to game information"])
+             (when (not has-keys)
+               (str " " (tr [:lobby.api-requires-key "(Requires an API Key in Settings)"])))
+             ]])
```
NoahTheDuke, line 545: Hanging parens

NoahTheDuke (approved):
This is very interesting! Thanks for doing all this work! Can't wait to see bots on jnet

src/clj/web/api.clj
jwarwick: The `public-CSRF-routes` has this route handler `(GET "/*" [] pages/index-page)`, which blocks the `GET` in `public-routes`. You still want it with the CSRF token, so pulling it out doesn't make a ton of sense.

NoahTheDuke: That makes sense.

PR #5796, opened 2021-04-14
Use icons for player stats

dkellner:
jnet-player-stats-icons-comparison

## Summary

Add a new setting "Use icons for player stats" (default: `true`) and if it's `true`, use the icons embedded in the already existing font to reduce the amount of text in the stats panels.

The other changes are small refactorings that I deemed useful to keep the differences between the two versions as small as possible.

## Changes to the previous default layout
("Use icons for player stats" = `false`)

- Change stat color to blue on hover to indicate which stat will be changed by clicking the buttons.
- Slightly increased line height in the stats panel.

## Questions

- How do we handle translations? I've not touched `translations.cljs` in this PR at all. I could add the translations for english, but unfortunately not for the other languages.

### DIFF
--- a/src/cljs/nr/gameboard/player-stats.cljs
+++ b/src/cljs/nr/gameboard/player-stats.cljs
@@ -1,0 +1,88 @@
+(ns nr.gameboard.player-stats
+  (:require [clojure.string :as s :refer [capitalize join lower-case]]
+            [nr.avatar :refer [avatar]]
+            [nr.gameboard.actions :refer [send-command]]
+            [nr.gameboard.state :refer [game-state]]
+            [nr.translations :refer [tr tr-pronouns]]))
+
+(defn controls
+  "Create the control buttons for the side displays."
+  ([key] (controls key 1 -1))
+  ([key increment decrement]
+   [:div.controls
+    [:button.small {:on-click #(send-command "change" {:key key :delta decrement}) :type "button"} "-"]
+    [:button.small {:on-click #(send-command "change" {:key key :delta increment}) :type "button"} "+"]]))
+
+(defn- name-area
+  [user]
+  [:div.namearea [avatar user {:opts {:size 32}}]
+   [:div.namebox
+    [:div.username (:username user)]
+    (if-let [pronouns (get-in user [:options :pronouns])]
+      (let [pro-str (if (= "blank" pronouns) "" (tr-pronouns pronouns))]
+        [:div.pronouns (lower-case pro-str)]))]])
+
+(defmulti stats-view #(get-in @% [:identity :side]))
+
+(defn- display-memory
+  [memory]
+  (let [me? (= (:side @game-state) :runner)]
+    (fn [memory]
+      (let [{:keys [available used only-for]} memory
+            unused (- available used)]
+        [:div (tr [:game.mu-count] unused available)
+         (when (neg? unused) [:div.warning "!"])
+         (when me? (controls :memory))]))))
+
+(defn- display-special-memory
+  [memory]
+  (when-let [only-for (->> (:only-for memory)
+                           (filter #(pos? (:available (second %))))
+                           (into {})
+                           not-empty)]
+    [:div
+     (str "("
+          (join "), (" (for [[mu-type {:keys [available used]}] only-for
+                             :let [unused (max 0 (- available used))]]
+                         (str unused " of " available
+                              " " (capitalize (name mu-type))
+                              " MU unused")))
+          ")")]))
+
+(defmethod stats-view "Runner" [runner]
+  (let [me? (= (:side @game-state) :runner)]
+    (fn [runner]
+      (let [{:keys [user click credit run-credit memory link tag
+                    brain-damage active]} @runner]
+        [:div.panel.blue-shade.stats {:class (when active "active-player")}
+         (name-area user)
+         [:div (tr [:game.click-count] click)
+          (when me? (controls :click))]
+         [:div (tr [:game.credit-count] credit run-credit)
+          (when me? (controls :credit))]
+         [display-memory memory]
+         [display-special-memory memory]
+         [:div (str link " " (tr [:game.link-strength "Link Strength"]))
+          (when me? (controls :link))]
+         (let [{:keys [base total is-tagged]} tag
+               additional (- total base)
+               show-tagged (or is-tagged (pos? total))]
+           [:div (tr [:game.tag-count] base additional total)
+            (when show-tagged [:div.warning "!"])
+            (when me? (controls :tag))])
+         [:div (str brain-damage " " (tr [:game.brain-damage "Brain Damage"]))
+          (when me? (controls :brain-damage))]]))))
+
+(defmethod stats-view "Corp" [corp]
+  (let [me? (= (:side @game-state) :corp)]
+    (fn [corp]
+      (let [{:keys [user click credit bad-publicity active]} @corp]
+        [:div.panel.blue-shade.stats {:class (when active "active-player")}
+         (name-area user)
+         [:div (tr [:game.click-count] click)
+          (when me? (controls :click))]
+         [:div (tr [:game.credit-count] credit -1)
+          (when me? (controls :credit))]
+         (let [{:keys [base additional]} bad-publicity]
+           [:div (tr [:game.bad-pub-count] base additional)
+            (when me? (controls :bad-publicity))])]))))
--- a/src/cljs/nr/gameboard/player-stats.cljs
+++ b/src/cljs/nr/gameboard/player-stats.cljs
@@ -5,13 +5,21 @@
             [nr.gameboard.state :refer [game-state]]
             [nr.translations :refer [tr tr-pronouns]]))
 
-(defn controls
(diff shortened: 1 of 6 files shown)


src/cljs/nr/gameboard/player-stats.cljs
```diff
@@ -0,0 +1,127 @@
...
+  [:div.namearea [avatar user {:opts {:size 32}}]
+   [:div.namebox
+    [:div.username (:username user)]
+    (if-let [pronouns (get-in user [:options :pronouns])]
+      (let [pro-str (if (= "blank" pronouns) "" (tr-pronouns pronouns))]
+        [:div.pronouns (lower-case pro-str)]))]])
+
+(defmulti stats-area
+  (fn [player] {:side (get-in @player [:identity :side])
+                :icons (get-in @app-state [:options :player-stats-icons] true)}))
```
NoahTheDuke, line 27: This shouldn't be part of the multimethod. Inlining this check (or putting it as a cursor inside of a wrapping `let` block) is the better way, as it lets us re-use code instead of having mostly copied functions with only minor differences.

src/cljs/nr/gameboard/utils.cljs
```diff
@@ -0,0 +1,12 @@
+(ns nr.gameboard.utils
+  (:require [nr.gameboard.actions :refer [send-command]]))
+
+(defn inc-dec-overlay
```
NoahTheDuke, line 4: I'm not a fan of this function name. If `controls` is too general, this is too specific and obscure. Maybe a happy medium could be `stat-controls` or `stat-buttons` or something.

src/cljs/nr/gameboard/main.cljs
```diff
@@ -837,83 +830,8 @@
...
-        [:div.panel.blue-shade.stats {:class (when active "active-player")}
-         (name-area user)
-         [:div (tr [:game.click-count] click)
-          (when me? (controls :click))]
-         [:div (tr [:game.credit-count] credit -1)
-          (when me? (controls :credit))]
-         (let [{:keys [base additional]} bad-publicity]
-           [:div (tr [:game.bad-pub-count] base additional)
-            (when me? (controls :bad-publicity))])]))))
+     [:div.statsarea
```
NoahTheDuke, line 833:
```suggestion
     [:div.stats-area
```
Clojure uses kebab-case, so let's stick with that here as well.

src/cljs/nr/gameboard/player-stats.cljs
```diff
@@ -0,0 +1,127 @@
...
+                              " " (capitalize (name mu-type))
+                              " MU unused")))
+          ")")]))
+
+(defmethod stats-area {:side "Runner" :icons false} [runner]
+  (let [ctrl (inc-dec-overlay-for-side :runner)]
+    (fn [runner]
+      (let [{:keys [user click credit run-credit memory link tag
+                    brain-damage active]} @runner]
+        [:div.statsarea
```
NoahTheDuke, line 63:
```suggestion
        [:div.stats-area
```

src/cljs/nr/gameboard/player-stats.cljs
```diff
@@ -0,0 +1,127 @@
...
+          [:div (str brain-damage " " (tr [:game.brain-damage "Brain Damage"]))])]))))
+
+(defmethod stats-area {:side "Runner" :icons true} [runner]
+  (let [ctrl (inc-dec-overlay-for-side :runner)]
+    (fn [runner]
+      (let [{:keys [user click credit run-credit memory link tag
+                    brain-damage active]} @runner
+            base-credit (- credit run-credit)
+            plus-run-credit (when (pos? run-credit) (str "+" run-credit))]
+        [:div.statsarea
```
NoahTheDuke, line 85:
```suggestion
        [:div.stats-area
```

src/cljs/nr/gameboard/player-stats.cljs
```diff
@@ -0,0 +1,127 @@
...
+                       (when show-tagged [:div.warning "!"])]))
+         (ctrl
+          :brain-damage
+          [:div (str brain-damage " " (tr [:game.brain-damage "Brain Damage"]))])]))))
+
+(defmethod stats-area {:side "Corp" :icons false} [corp]
+  (let [ctrl (inc-dec-overlay-for-side :corp)]
+    (fn [corp]
+      (let [{:keys [user click credit bad-publicity active]} @corp]
+        [:div.statsarea
```
NoahTheDuke, line 105:
```suggestion
        [:div.stats-area
```

src/cljs/nr/gameboard/player-stats.cljs
```diff
@@ -0,0 +1,127 @@
...
+         (ctrl :click [:div (tr [:game.click-count] click)])
+         (ctrl :credit [:div (tr [:game.credit-count] credit -1)])
+         (let [{:keys [base additional]} bad-publicity]
+           (ctrl :bad-publicity [:div (tr [:game.bad-pub-count] base additional)]))]))))
+
+(defmethod stats-area {:side "Corp" :icons true} [corp]
+  (let [ctrl (inc-dec-overlay-for-side :corp)]
+    (fn [corp]
+      (let [{:keys [user click credit bad-publicity active]} @corp]
+        [:div.statsarea
```
NoahTheDuke, line 115:
```suggestion
        [:div.stats-area
```

src/css/gameboard.styl
```diff
@@ -287,7 +305,7 @@
     .scored
         height: 116px
 
-    .scored > .stats
+    .scored > .statsarea
```
NoahTheDuke, line 308:
```suggestion
    .scored > .stats-area
```

src/css/gameboard.styl
```diff
@@ -307,17 +325,20 @@
     .squeeze .card-wrapper
         position: absolute
 
-    .stats > div
-        height: 20px
+    .namearea
+        margin-bottom: 3px
+
+    .statsarea
```
NoahTheDuke, line 331:
```suggestion
    .stats-area
```

src/cljs/nr/gameboard/player-stats.cljs
```diff
@@ -0,0 +1,127 @@
...
+
+(defn- inc-dec-overlay-for-side
+  [side]
+  (if (= (:side @game-state) side)
+    inc-dec-overlay
+    (fn [key content] content)))
+
+(defn- name-area
+  [user]
+  [:div.namearea [avatar user {:opts {:size 32}}]
```
NoahTheDuke, line 18:
```suggestion
  [:div.name-area [avatar user {:opts {:size 32}}]
```
Looks like this is a holdover from the previous version of the code. Let's clean it up while we're here.

src/css/gameboard.styl
```diff
@@ -307,17 +325,20 @@
     .squeeze .card-wrapper
         position: absolute
 
-    .stats > div
-        height: 20px
+    .namearea
```
NoahTheDuke, line 328:
```suggestion
    .name-area
```

src/cljs/nr/gameboard/utils.cljs
NoahTheDuke: Additionally, any particular reason for moving this to a new file instead of keeping it with the other stats functions? Importing it from `nr.gameboard.stats` in `nr.gameboard.main` would work just as well. I'm generally not a fan of single-function namespaces.

NoahTheDuke:
Thanks so much! This looks exciting. I've left some comments.

dkellner:
Thanks for the thorough review! I will implement your feedback asap.

src/cljs/nr/gameboard/main.cljs
dkellner: I went with this for consistency with `namearea`, but as we will rename that one as well, kebab it is!

src/cljs/nr/gameboard/utils.cljs
dkellner: +1 for `stat-controls`.

I've moved it to `utils` because it's not just used in the `player-stats` module, and we could move other "util-like" functions there, too. But this probably only makes sense if we split up `main.cljs` further, and we can still move things around once we do that. I will put it in `player-stats` for now, keeping things simple.

merrickbrown:
This looks very nice - what do you think about bumping up the font-size on the click + credit values? Might make them stand out a bit more next to the icons

src/cljs/nr/gameboard/player-stats.cljs
dkellner: You're right. There were more differences initially and I failed to refactor it again after I finished. Looks cleaner now.

dkellner:
> This looks very nice - what do you think about bumping up the font-size on the click + credit values? Might make them stand out a bit more next to the icons

I've tried that, but it makes the whole panel look a bit messy IMO. Right now, "5 [credits]" in the stats panel will be displayed the same as in the chat log and the font sizes are the same as for other labels.

That said, while I think this PR is an improvement in clarity of the UI in general, there's still a lot of room for improvement.

dkellner:
@jwarwick I've rebased this on current `master` and tested it again.

PR #5798, opened 2021-04-14
Add state variables to trigger end of game screen

jwarwick:


NoahTheDuke:
Thanks for catching this.

PR #5822, opened 2021-04-18
Load generator

dsimmer:
Rewrote the load generator script to work with the new sample data and sente websocket CSRF requirements.

Removing the extra "game" argument in main.cljs seems to result in a dramatic speed up in chrome rendering a game with lots of spectators... for some reason...

Also, the duplicate defmethods of lobby/watch meant that the script could not join the game at the lobby stage, so I tweaked that to ensure that both methods do actually get called. Please comment on that - should I wrap them into one function?


src/clj/tasks/load_generator.clj
```diff
@@ -0,0 +1,141 @@
...
+      (ws/connect (str "ws://localhost:1042/ws?client-id=" runner-client-id "&csrf-token=" (URLEncoder/encode (runner-login "X-CSRF-Token")))
+                :client client
+                :on-error #(safe-println "runner error" %)
+                :on-connect (fn [n] (safe-println "Runner Connected"))
+                :on-close (fn [x y] (safe-println "Runner Disconnected"))
+                :headers runner-login)
+    (safe-println "Create lobby")
+
+    ;; Corp create lobby
+    (game-ws-handler/-msg-handler {:id :lobby/create :ring-req {:user {:username "sample2"}}
```
NoahTheDuke, line 60: spacing in these sections. maybe put the `:ring-req` on the next line for clarity.

src/clj/tasks/load_generator.clj
```diff
@@ -0,0 +1,141 @@
...
+                                    :?data {:gameid game-id
+                                            :password ""}})
+              socket))
+          (range maxUsers)))
+
+      (safe-println "Spectators connected")
+      (game-ws-handler/-msg-handler {:id :netrunner/start :ring-req {:user {:username "sample2"}}
+                          :client-id corp-client-id})
+      (safe-println "Started game"))
+      ))
```
NoahTheDuke, line 105: dangling parens

src/clj/tasks/load_generator.clj
```diff
@@ -0,0 +1,141 @@
...
+    (println msg))
+  (System/exit status))
+
+(defn command
+  "This is not really a load test per se, but a development tool to assist debugging performance issues.
+
+  This will start the server and create a single game on your environment between users sample2 (Corp) and sample3 (Runner).
+  One thousand spectators will be connected to the game by default, from users sample4-sample1003.
+
+  The load generator must be run AFTER creating sample users."
```
NoahTheDuke, line 133: Do the sample users have to be manually created?

NoahTheDuke: and if they do, should you include a method for automatically creating those users, maybe behind a flag in this task?

PR #5826, opened 2021-04-21
Loki fixes

nathanj:
When using Loki the other day, we noticed that the additional subs were not firing. They would get the green check as if fired, but there was no effect.

The problem was due to the `add-sub` function not working on subs that had already been added via `add-sub` during init. See the commit message for more details.

A few other Loki fixes:
 * Add a message when the card is chosen, because otherwise it is not obvious. 
 * Fix a FAQ entry regarding empty runner hand.

Finally, fix a Karuna sub label that I noticed was wrong when switching the test to use it.

### DIFF
--- a/test/clj/game/cards/ice_test.clj
+++ b/test/clj/game/cards/ice_test.clj
@@ -2695,24 +2695,33 @@
 (deftest loki
   (do-game
     (new-game {:corp {:deck [(qty "Hedge Fund" 5)]
-                      :hand ["Loki" "Archer"]
-                      :credits 100}})
-    (play-from-hand state :corp "Archer" "HQ")
+                      :hand ["Loki" "Karunā"]
+                      :credits 100}
+               :runner {:hand [(qty "Sure Gamble" 3) (qty "Easy Mark" 3)]}})
+    (play-from-hand state :corp "Karunā" "HQ")
     (rez state :corp (get-ice state :hq 0) {:ignore-cost :all-costs})
     (play-from-hand state :corp "Loki" "R&D")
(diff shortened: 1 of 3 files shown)


test/clj/game/cards/ice_test.clj
```diff
@@ -2693,28 +2693,71 @@
...
+                 :runner {:hand [(qty "Sure Gamble" 3) (qty "Easy Mark" 3)]}})
+      (play-from-hand state :corp "Karunā" "HQ")
+      (rez state :corp (get-ice state :hq 0) {:ignore-cost :all-costs})
+      (play-from-hand state :corp "Loki" "R&D")
+      (rez state :corp (get-ice state :rd 0) {:ignore-cost :all-costs})
+      (take-credits state :corp)
+      (run-on state "R&D")
+      (run-continue state)
+      (click-card state :corp "Karunā")
+      (is (last-log-contains? state "Corp chooses Karunā protecting HQ at position 0 for Loki's ability.")
```
NoahTheDuke, line 2744: I would prefer that this test be stripped down to the bare bones "does Loki's printed subroutine shuffle the grip into the stack?" So take out Karuna, take out all of the repeated assertions from the first test, have it be: "runner has cards in hand, make a run, encounter, fire subs, runner chooses to shuffle, grip is empty and stack has more cards"

NoahTheDuke:
Thanks for writing this up! I wonder how this happens.

test/clj/game/cards/ice_test.clj
nathanj: Good call, thanks. Fixed now. Now that the test is smaller I added testing for the zero cards in hands scenarios as well.

PR #5828, opened 2021-04-21
Add Currents back to state diff

### LINKED ISSUE #5823: "current" events not showing anymore
Since thursday (15.04.) played "Current" events are no longer displayed. Prior to that day they were being displayed in the playing area as long as they were active.

It is a bit cumbersome to search the chatlog for the last "current" card and find out if that card is still active or already trashed.

jwarwick:
Fixes #5823


NoahTheDuke:
Wow, I really hecked this up, didn't I?

PR #5839, opened 2021-04-24
Fix Neutralize All Threads / Trebuchet interaction

nathanj:
The prompt had no choices when these two cards were used together. NAT
sets must-trash? to true but Trebuchet sets can-trash to false. This was
leading to a scenario where no-action-str was empty leading to choices
being empty.


NoahTheDuke:
Damn good catch!

Issue #5642, opened 2021-03-17
Utae lacks "match strength" ability


francescopellegrini:
Hi,
it looks like Utae doesn't have the shortcut for matching the strength of the encountered Ice.
When I noticed it I had less than 3 `virtual` resources installed, FWIW.

francescopellegrini:
It looks like [this lack was intentional](https://github.com/mtgred/netrunner/blob/master/src/clj/game/cards/programs.clj#L2592)..

NoahTheDuke:
Yes it is. We can't currently do auto-pumping/breaking with X costs because of how auto-pumping/breaking is built. Maybe some day in the future we could do it, but it would require rewriting all of that stuff first.

Issue #5734, opened 2021-03-31
Unity does not seem to have "Fully break"


percomis:
I encountered an Enigma and Unity had no fully break option as usually icebreakers do.

NoahTheDuke:
Yes. The auto-icebreaker stuff require that the costs for a each ability are fixed, and because these programs don't have fixed costs, they can't use the auto-pump/break framework.

PR #5847, opened 2021-04-27
Special handling of dynamic abilities

lostgeek:
Fixes broken breakers


francescopellegrini:
Does this fix #5642 as well?

lostgeek:
I don't think so. This fix just changes the UI to send the correct command to the server. Utae with its different behavior on 3 virtual resources needs additional logic that isn't implemented yet.

francescopellegrini:
Ok, thank you for the clarification! And what about #5734 (Unity)?

PR #5862, opened 2021-05-01
Rework of the competitive lobby

lostgeek:
Things left to do:

- [x] cancel / claim system
  - [x] Trigger this on disconnect
- [x] stopping a run at 3 losses
- [x] automatically share all replays of Angel Arena games 
- [x] show latest runs
- publishing runs (will do later)
- [x] in `eligible-players` check for previous opponents

Ideas
- open decklists

### DIFF
--- a/src/clj/web/stats.clj
+++ b/src/clj/web/stats.clj
@@ -1,5 +1,6 @@
 (ns web.stats
-  (:require [web.mongodb :refer [object-id]]
+  (:require [clojure.string :refer [lower-case]]
+            [web.mongodb :refer [object-id]]
             [monger.collection :as mc]
             [monger.result :refer [acknowledged?]]
             [monger.operators :refer :all]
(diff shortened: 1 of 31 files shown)


NoahTheDuke:
This is looking good! Quick question: any particular reason to not hyphenate angel arena in the code?

lostgeek:
Mostly because I didn't like to hyphenate the file name. So I decided to drop it for all code-related names. Should I add it back for consistency and rename the folder `angel_arena`?

NoahTheDuke:
The style in the rest of the codebase is to hyphenate when joining two words. As file names are generally only referenced in require blocks, I don't mind if there is inconsistency between the variables/functions and the file name, but I think I'd prefer that we maintain consistency across the code itself.

src/clj/web/stats.clj
```diff
@@ -174,8 +175,26 @@
                 :runner.agenda-points (get-in @state [:runner :agenda-point])}
      :history (:history @state)}))
 
+(defn handle-angelarena-stats
+  [db {:keys [ending-players original-players state format] :as game}]
+  (try
+    (doall
+      (for [player original-players]
```
NoahTheDuke, line 182: If you're not using the return value of `for`, better to use `doseq` which has the same api but isn't lazy and returns `nil`.

src/clj/web/stats.clj
```diff
@@ -174,8 +175,26 @@
                 :runner.agenda-points (get-in @state [:runner :agenda-point])}
      :history (:history @state)}))
 
+(defn handle-angelarena-stats
+  [db {:keys [ending-players original-players state format] :as game}]
+  (try
```
NoahTheDuke, line 180: No catch block?

lostgeek:
This is ready to be pushed live now! :tada:

src/clj/web/utils.clj
```diff
@@ -30,3 +30,9 @@
        (.digest (MessageDigest/getInstance "MD5"))
        (BigInteger. 1)
        (format "%032x")))
+
+(defn average
+  [numbers]
+  (if (empty? numbers)
+    nil
```
NoahTheDuke, line 37: Not 0?

src/css/angel-arena.styl
```diff
@@ -0,0 +1,204 @@
...
+                    display-flex()
+                    width:100%!important
+                    height: 5rem
+                    margin: .25rem
+
+                    .info
+                        margin:auto
+                        flex-grow: 1
+
+
```
NoahTheDuke, line 185: Lots of open space in this file. Can we clean it up a little?

Issue #4050, opened 2019-02-26
Request: move or resize UI components


AndersRask:
It would be so cool to have some elements of customization on the UI. Moving layers or at least better resizing to fit larger screens so that when you zoom, you could still see your handsize etc!

NoahTheDuke:
This is a sweet idea, and something I am terrible at. I'd love for any help possible!

AndersRask:
A simple solution could be to implement so that the font scales with the zoom in css. I am by no means an expert (im a dev and architect) but I seem to remember if you use em instead of px you get better scaling?
I also found this https://github.com/yonran/detect-zoom

tsujigiri:
One quick-win I see is making the card in the upper-right corner more readable on high resolution screens by making it bigger and accept, that it will cover other things. It is not visible all the time after all. The chat could even go up to the top and the card be shown right next to it and bigger.

On my notebook (13", 1920×1080 px) I can zoom in one step, which just allows me to read the card. If I zoom in any further, things disappear from the screen and cover another.

armerpaul:
Hmmm so I think I have some ideas on this. I'll mess around with it this weekend and see what I can put together. 

(I have jnet running locally)

lallmon:
@AndersRask I would suggest using a REM (Root Element) scaling system, with the root being 16px, and using sizing steps of .125rem.

Reasons:
- 16px is the default of browsers (1rem = 16px)
- EM units are based on their parents sizing, so you'll get weird relative sizing bugs quickly compared to REM, which scale off that root.
- .125rem steps resolve into simple integers, which means lining elements up in layouts is easier (if you move element sizing to a REM system).
- That stops a lot of sizing bugs, and "magic numbers". (when something is say 237px for no other reason than "that's the closest")

I can possibly do a font attempt if @armerpaul isn't already doing this?

At this point in my life, the only pixel values I use if I can help it, are in border widths/radii. 😄

PR #5867, opened 2021-05-03
feature(REM): font sizing to rem units for UI scaling

lallmon:
Per conversation and then my suggestion in #4050 to make the fonts more scalable in the browser, using the REM (Root Element) units, based on 16px.

Reasons:
- 16px is the default of browsers (1rem = 16px), and typically the smallest font size you should use if you can otherwise help it.
- EM units are based on their parents sizing, so you'll get weird relative sizing bugs quickly compared to REM, which scale off that default.
- Using .125rem steps resolve into simple integers keeping in the powers of 2, which means lining elements up in layouts becomes easier and simpler, especially if you base all your measurements off a REM scale.
 - That stops a lot of sizing bugs, and "magic numbers". _(e.g. when something is say 237px for no other reason than, "that's the number that worked")_

This doesn't make the UI Movable or Resize-able as per the issue above, but helps with the scaling, **and the REM values still resolve to current values**, except in a couple cases, and those changes typically are 1 pixel in difference, stepped upward for readability.

### DIFF
--- a/src/css/base.styl
+++ b/src/css/base.styl
@@ -12,25 +12,25 @@ body
     color: text-color
     background-color: black
     font-family: "Titillium Web", "Helvetica Neue", sans-serif
-    font-size: 13px
-    line-height: 18px
+    font-size: 1rem
+    line-height: 1rem
 
 h1, h2, h3, h4
     line-height: normal
     margin: 0 0 8px
 
 h1
-    font-size: 36px
+    font-size: 2.25rem
 
 h2
-    font-size: 24px
+    font-size: 1.5rem
 
 h3
-    font-size: 20px
+    font-size: 1.25rem
     margin: 0 0 6px
 
 h4
-    font-size: 14px
+    font-size: .875rem
     line-height: 18px
     margin: 0 0 4px
 
@@ -79,7 +79,7 @@ section
 
 input, select, textarea
     font-family: "Titillium Web", "Helvetica Neue", sans-serif
-    font-size: 13px
+    font-size: .75rem
     border-radius: 4px
     padding: 0 5px
     margin: 0
@@ -106,7 +106,7 @@ label
 button, a.button
     display: inline-block
     border-radius: 4px
-    font-size: 12px
+    font-size: .75rem
     padding: 4px 12px
     margin: 0 8px 0 0
     background-color: transparent-orange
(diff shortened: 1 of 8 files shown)


src/css/base.styl
```diff
@@ -12,25 +12,25 @@ body
     color: text-color
     background-color: black
     font-family: "Titillium Web", "Helvetica Neue", sans-serif
-    font-size: 13px
-    line-height: 18px
+    font-size: 1rem
+    line-height: 1rem
```
NoahTheDuke, line 16: Does this change mess up when there are icons on each line?

src/css/chat.styl
```diff
@@ -49,8 +49,8 @@
         position: absolute
         z-index: 30
         width: 200px
-        font-size: 12px
-        line-height: 16px
+        font-size: 3rem
```
NoahTheDuke, line 52: 12px becomes 3rem?

src/css/gameboard.styl
```diff
@@ -101,15 +101,18 @@
...
-        height: 24px
-        line-height: 24px
+        font-size: 1rem
+        width: 2rem
+        height: 2rem
+        border-radius: 100%
+        line-height: 2rem
         text-align: center
-        margin: 2px
+        box-shadow:
```
NoahTheDuke, line 110: This looks like an error.

src/css/gameboard.styl
```diff
@@ -101,15 +101,18 @@
 
     .counter
         padding: 0
-        font-size: 12px
-        width: 24px
-        height: 24px
-        line-height: 24px
+        font-size: 1rem
+        width: 2rem
```
NoahTheDuke, line 105: 16 px is 1 rem. shouldn't 24px be 1.5rem or something similar?

src/css/base.styl
lallmon: Good catch on those.

src/css/gameboard.styl
```diff
@@ -101,15 +101,18 @@
...
+        height: 2rem
+        border-radius: 100%
```
NoahTheDuke, line 107: is this supposed to be included?

src/css/gameboard.styl
lallmon: This may have been from the rebase, I'll amend.

src/css/gameboard.styl
lallmon: Nope! That was a leak from an experiment I was doing with tokens. 😆

Issue #5843, opened 2021-04-26
Custom CSS colours for UI elements [nice to have]


vesperius:
At the moment of writing, the UI element colours are hard-coded throughout the site and seem to be set in CSS through:

> .blue-shade {
> 	background-color;
> 	box-shadow;
> }

Several approaches possible for this request:
1. The user could set their preferred RGB mix for those elements in "Settings" and have it saved like other preferences.
2. By default, playing on the Corp could mean using blue elements and Runner - red elements.
3. By default, playing with a faction could mean using faction-coloured elements.

lallmon:
Having theming of the factions, though there aren't terribly too many colored UI elements, which ones are you considering be changed?

Would it be the buttons that get the theme, or the UI background? I will say that the current Orange buttons would be a bit of a nightmare against some of the faction colors, NBN specifically. 😆

vesperius:
This is a great investigation you've linked to! Thank you for sharing.
I'll use the data you compiled there as inspiration to provide some sample screenshots in here a bit later.

For now, I am attaching my own default setup I have now for all games and a few links showing edits made for different events (they're not for factions, just for the whole event).

Disclaimer: I am not an artist/designer and my colour sensitivity might be very much skewed towards dark green being "the best colour ever" ;-)

- ![Screenshot_2021-05-10 Play Android Netrunner in your browser](https://user-images.githubusercontent.com/16701215/117683320-3ec10980-b1b4-11eb-91c6-9adf404cd42f.png)
- purple - https://youtu.be/COFc5n_kolk?list=PLe_k4DxJFd6_oF9S0GwXcCMp5o-wG6CzZ&t=8768
- blue - https://youtu.be/0zLa1qWGsvc?list=PLe_k4DxJFd6_oF9S0GwXcCMp5o-wG6CzZ&t=3106
- dark red - https://youtu.be/HESQ4kQVBwU?list=PLe_k4DxJFd6_oF9S0GwXcCMp5o-wG6CzZ&t=20058
- silver/gold - https://youtu.be/BARE2v1oc4Y?list=PLe_k4DxJFd6_oF9S0GwXcCMp5o-wG6CzZ&t=8819

PR #5872, opened 2021-05-07
chore (colors): consolidation of colors into variables

lallmon:
DRAFT DISCUSSION:

Like my last PR, this doesn't really change much, functionally. 😆 

I did a color audit and:
- matched hardcoded colors to existing variables.
- created new variables for hardcoded colors, that didn't exist.
- converted values to `rgb/rgba` because there's transparency (all good SciFi UI has transparency, right)

What this gives is a current big picture of the "bag 'o skittles" colors used in the application, and there's a lot of repetition, and some barely differentiated colors at all.
palette

A "good" UI is mostly based on consistency

This could possibly could move towards features like #5843, as well as developing less decision overhead when folks are developing (Color/Font/Spacing consistency).

Longer term proposals/ideas:
- For Faction colors, consider matching the Nisei color palettes, and using shades/tints of that
- Simplifying color naming into human readable.
- Using consistent transparency steps.
- adding drop shadow styles to the variables directly, in order to not rewriting them each time.

Would love some thoughts/feedback!


jwarwick:
I like it. Without having someone owning the UI/UX it gets a bit scattered. This should at least give us a place to start when adding new things.
+1 to stealing the NISEI colors.
I also like the idea of making things simpler for streamers, who are overriding the CSS today.

lallmon:
@jwarwick Yeah that happens a lot in projects without a CSS/Design lead. One huge benefit of using a consistent system, is less decision overhead for new features, which means iteration can be much quicker if you don't have to fiddle with magic numbers in CSS.

While I imagine some folks will/would be upset moving towards Nisei, it's probably better in regards to long term legal viability as well?

Can you possibly point me to (probably in Slack) to streamers that are overriding CSS? I'd like to see what they are doing.

I'll also keep this draft open, and work on a bit more consolidation into a palette/utility.

lallmon:
Ok @jwarwick and @NoahTheDuke 

Alright, that felt like 3000 lines of changes, not 300 haha.

Here's an overview:
- Unified the color pallette = I combed through all the styles and found all color definitions, and extracted them to the `palette.styl` (which was called variables, but there were only colors in there, so renamed)
- Converted colors to HSL, which is human understandable/writable compared to HEX.
- Named the colors a bit more directly, to know what they are while you are working in the css. 
- Reached out to Nisei and found the main colors used by the factions and mini factions, and somewhat based the naming around them. e.g. `orange-core` is the Anarch orange.
- Any colors that were within a step or so, I consolidated into less colors. There were a LOT of the Yellow/Orange/Gold spectrum, and a TON in the Blue spectrum.
- Added an `opacity.styl` which contains opacity variables `0.1` steps (there were lots of random values, and the difference between .71 and .76 is negligible. Any colors that use transparency run the stylus `alpha(color-variable, opacity-variable)` function.

*Issues I may have*
- BLUE! There were so many blue definitions, that I am sure I mixed dark and light. If you see one, just call out the closest blue it should be, and I will fix it.
- I had some trouble compiling Stylus, it needed `@require` in some of the files to access the variables. I was under the impression it compiled everything and gave it access to the variables and functions, but I'm not sure? There were a decent amount of default browser colors, so maybe those files never had access to the variables?



Current Palette File
jinteki-palette

src/css/base.styl
```diff
@@ -36,22 +38,22 @@ h4
...
 
 table
     border-collapse: collapse
 
 pre
     font-family: "Titillium Web", "Helvetica Neue", sans-serif
-    white-space: -moz-pre-wrap
-    white-space: -o-pre-wrap
-    white-space: pre-wrap
+    white-solid-space: -moz-pre-wrap
```
NoahTheDuke, line 54: Is this correct?

src/css/base.styl
```diff
@@ -313,7 +315,7 @@ span.small
     margin-left: 10px
 
 .ellipsis
-    white-space: nowrap
+    white-solid-space: nowrap
```
NoahTheDuke, line 318: Same here?

src/css/deckbuilder.styl
```diff
@@ -9,21 +10,21 @@
...
+            border-color: gold-base
 
         p
             margin-top: 0
 
         h4
             margin: 0
             text-overflow: ellipsis
-            white-space: nowrap
+            white-solid-space: nowrap
```
NoahTheDuke, line 27: Here

src/css/deckbuilder.styl
```diff
@@ -124,14 +125,14 @@
                 position: relative
                 overflow: visible
                 line-height: 20px
-                white-space:pre
+                white-solid-space:pre
```
NoahTheDuke, line 128: Here

src/css/deckbuilder.styl
```diff
@@ -124,14 +125,14 @@
...
 
                 button.small
                     margin: 3px 6px 3px 0
 
             .notes
                 margin-top: 20px
-                white-space: pre-wrap;
+                white-solid-space: pre-wrap;
```
NoahTheDuke, line 135: Here

src/css/gameboard.styl
```diff
@@ -186,13 +187,13 @@
         right: 3px
         padding-left: 3px
         padding-right: 3px
-        // white-space: nowrap
+        // white-solid-space: nowrap
```
NoahTheDuke, line 190: Here but also just delete this?

NoahTheDuke (commented):
Looks great! I think your find replace was over eager so look out for the white space being changed to white core space.

lallmon:
@NoahTheDuke Oh weird, I set it to search for just "white" but must have done it wrong, did a commit to put them back to correct.

src/css/gameboard.styl
```diff
@@ -384,15 +384,15 @@
             z-index:1
             font-family:Arial,sans-serif
             font-size:2.5rem
-            color:#c00
-            background: rgb(255, 255, 255, 0.5)
-            border:solid 4px #c00
+            color: red-core
+            background: alpha(black-solid, o-60)
```
NoahTheDuke, line 388: `o-60` vs .5? is that intentional?

PR #5883, opened 2021-05-13
Don't merge: Attempt at fixing scroll bars appearing mid run

lostgeek:
When the run arrow gets drawn typically at around 4-5 ice, the window gets a vertical scrollbar and the entire board becomes very jumpy. Unfortunately this approach of fixing it leads to the player not being able to click on certain cards on the opponent side...


lostgeek:
Added to #5885

PR #5897, opened 2021-05-19
standardize credits vs [Credits] usage in messages

francescopellegrini:
Fixes all places where "credits" and "[Credits]" were used wrong.


NoahTheDuke:
This is great, thank you so much!

PR #5899, opened 2021-05-20
Adding (start of turn) to additional cards

### LINKED ISSUE #5851: La Costa Grid isn't marked as "start of turn"
This makes the `:keep-menu-open` not recognize it as a start-of-turn fixing ability and the menu stays open after rezzing.

cardboardtech:
La Costa Grid's label is now tagged with (start of turn) so its menu will now close after rez. 

Related to this issue, I noticed other cards such as Cayambe Grid and Wall to Wall are also missing this tag. I was curious what the process is for fixing similar issues. Should I make individual issues for these cards or just fix them in one go as I see them?

### DIFF
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -297,7 +297,7 @@
 (defcard "C.I. Fund"
   {:derezzed-events [corp-rez-toast]
    :flags {:corp-phase-12 (req (pos? (:credit corp)))}
-   :abilities [{:label "Move up to 3 [Credit] from credit pool to C.I. Fund"
+   :abilities [{:label "Move up to 3 [Credit] from credit pool to C.I. Fund (start of turn)"
                 :prompt "Choose how many [Credit] to move"
                 :once :per-turn
                 :choices {:number (req (min (:credit corp) 3))}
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
Looks like this is _very new_ functionality done by @lostgeek, so the answer is probably "go for it!" We love getting help with stuff like this.

cardboardtech:
I think I've covered all cards with start of turn abilities. Notably, C.I. Fund and Executive Boot Camp have a mix of start of turn and active abilities, so the menu will stay open on rez, but the start of turn ability will at least be labeled to indicate what it is for. Clyde Van Rite and Daily Quest did not have a ":label" for their ability so I added one. 

I also noticed that we have several cards with start of turn effects, but no ability associated with it (i.e. Public Support, Personalized Portal). All of the automation is done in the event itself. I wonder if it would be a good idea to create abilities to represent these effects so people can manually trigger the effect if needed, though there might be a reason they are automated this way. That being said, I think I'd want to handle that outside this pull request either way.

NoahTheDuke:
You have discovered an area where our automation isn't great and the solution (use "normal" abilities) is also not great, haha. It's a mish-mash of abilities and events, lots of little hacks to make everything work right. I think what you've got here is good, but if you want to add manual abilities for Public Support and Personalized Portal, you're welcome to. Probably better to have and not need, at this point.

cardboardtech:
Haha, I see. If that's the case then I think I'm going to call it good and move on to some other issues.

cardboardtech:
I realized I jacked up my branches (forgot to branch off my forked main initially for this). Going to close this pull request, get everything sorted out, and redo this.

cardboardtech:
Resubmitted in pull request #5904

Issue #5892, opened 2021-05-17
Keiko triggers twice if a companion is installed and used in the same turn


AceEmpress:
Specifically, it triggered when I installed a paladin poemu and again when I ran and used trickster taka credits later in the turn - it should only trigger the first time either condition is met, not the first time a companion is installed and the first time credits from a companion are used.
![image](https://user-images.githubusercontent.com/81694351/118546445-e0fe6580-b74f-11eb-9ed9-5d2664f77c1d.png)

PR #5900, opened 2021-05-21
Fix Keiko trigger condition

h4pp7:
Fix for #5892 

Keiko gave credits a second time in a turn, when credits where used from
a companion after a companion was installed that turn.

`event-count` with `:runner-install` now gets a different filter
predicate than the call with `:spent-credits-from-card`.

### DIFF
--- a/src/clj/game/cards/hardware.clj
+++ b/src/clj/game/cards/hardware.clj
@@ -942,10 +942,12 @@
    :events [{:event :spent-credits-from-card
              :req (req (and (not (facedown? target))
                             (has-subtype? target "Companion")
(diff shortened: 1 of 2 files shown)


h4pp7:
There is probably potential to make this less boilerplat, though.

NoahTheDuke:
Good catch! Yeah, the engine is kind of "in between" right now where most of the conditional abilities use an old system and a select group of them use the newer system. Leads to dumb bugs like this. I wouldn't worry too much about boilerplate just yet, still some things to figure out first.

h4pp7:
Made a new pull request (#5918 ) with the same changes from a different branch.

Issue #2964, opened 2017-12-19
Create game as "Any Side"


Dalmadog:
Hi,

Is there a possibility to create "any side" games, instead of having to pick runner or corp ? Sometimes I feel like just wanting to play whatever side.

The guy who join an existing lobby would choose which side he wants. Giving him the choice would just contribute to more agreement before starting the game (game-design concerns).

francescopellegrini:
Closed by #5905

PR #5905, opened 2021-05-23
Create game as Any Side

cardboardtech:
I saw issue #2964 and thought I'd give it a shot especially since this sort of functionality is one that I've wished existed. I decided to keep things simple for now and just have sides be picked at random.

- Added a new "Any Sides" radio button when a player is creating a game. 
- If "Any Sides" is selected, the "Swap Sides" button is disabled
- The "Select Deck" link only appears if the player has a side selected
- When a second player joins an "Any Side" game, the sides are picked at random.

Potential Improvements:
- Let the 2nd player pick a side if they want (maybe through a dialogue or perhaps just additional Join buttons on the game list)
- Reset to "Any Sides" if the 2nd player leaves (could be difficult?).
- Keep "Swap Sides" button active and have it rotate between the three options. This could also be a solution for resetting to "Any Sides" when a player leaves.

NOTE: My experience with Clojure and functional programming languages as a whole are practically zero (come from a .NET background), so I've been learning on the fly. This feature is a big step up from adding some text to labels, so I'd appreciate feedback on whether I'm doing things correctly or if there are better ways to pull off what I'm going for.

Demonstration:
CreateGameRandomSides

### DIFF
--- a/src/clj/web/game.clj
+++ b/src/clj/web/game.clj
@@ -144,7 +144,7 @@
         (ws/broadcast-to! [client-id] :lobby/select {:gameid gameid
                                                      :started started
                                                      :state (json/generate-string (side (public-states (:state game))))}) ;side works here because user cannot rejoin as a spectator
-        (lobby/join-game user client-id gameid)
+        (lobby/join-game user client-id gameid "Any Side")
         (main/handle-rejoin state user)
         (swap-and-send-diffs! (lobby/game-for-id gameid))))))
 
--- a/src/clj/web/lobby.clj
+++ b/src/clj/web/lobby.clj
@@ -334,7 +334,7 @@
   (let [game (game-for-id gameid)
         first-player (first (:players game))]
     (when  (= (:ws-id first-player) client-id)
-      (if (> (count (:players game)) 1)
+      (if (< 1 (count (:players game)))
         (refresh-lobby-update-in gameid [:players] (partial mapv swap-side))
         (let [updated-player (change-side first-player side)]
           (refresh-lobby-assoc-in gameid [:players] [updated-player]))))))
--- a/src/clj/web/lobby.clj
+++ b/src/clj/web/lobby.clj
@@ -253,6 +253,13 @@
                        "Corp"))
       (dissoc :deck)))
 
+(defn change-side
+  "Returns a new player map with the player's :side set to a new side"
+  [player side]
+  (-> player
+      (assoc :side side)
+      (dissoc :deck)))
+
 (defn blocked-users
   [{:keys [players]}]
   (mapcat #(get-in % [:user :options :blocked-users]) players))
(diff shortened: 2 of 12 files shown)


cardboardtech:
Added Split Dropdown Button to replace the Join button when the game is for Any Side. There are three options. 

1. Clicking the Join button will randomly pick sides
2. Clicking "As Corp" in the dropdown will set the joining player as the Corp
3. Clicking "As Runner" in the dropdown will set the joining player as the Runner

Here is an image of how I decided to style the button and drop down. I am absolutely open to feedback.
image

Here are some previous iterations I implemented but decided not to go with.
Button based:
Join-Button

Block-link based (styled after existing drop down menu)
Join-BlockLink

Block-link based with border
image

cardboardtech:
After getting some feedback, I decided to update the dropdown menu's style for the split button to be more inline with our existing dropdown menu. I did add a more visible border to all of our dropdown menus to make them more readable. 

image
image

src/clj/web/game.clj
```diff
@@ -144,7 +144,7 @@
         (ws/broadcast-to! [client-id] :lobby/select {:gameid gameid
                                                      :started started
                                                      :state (json/generate-string (side (public-states (:state game))))}) ;side works here because user cannot rejoin as a spectator
-        (lobby/join-game user client-id gameid)
+        (lobby/join-game user client-id gameid "Any Side")
```
NoahTheDuke, line 147: No need for this.

NoahTheDuke: jk misread

src/css/palette.styl
```diff
@@ -21,10 +21,11 @@ orange-core = hsl(19, 75%, 55%) // Nisei Anarch
 
 // Golds
 //Mainly Jinteki.net button styles
-gold-dark = hsl(19, 56%, 40%)
```
NoahTheDuke, line 24: Is this something you need to update in the rest of the code?

NoahTheDuke:
This looks sick as hell, thank you so much.

src/css/palette.styl
cardboardtech: nowhere else uses gold-dark (that I could find) but I found the current value of gold-dark was more of a weird orange I decided to change it. That being said, I didn't even end up using this new value in my code so it might not be worth changing.

cardboardtech:
Thanks! I've got one more commit that is almost ready to go around changing the Swap Sides button to let you pick Any Side, Corp, or Runner when there is only one player. Would be helpful because currently if you make a game as Any Side, someone joins and sides are assigned, and that person leaves the lobby there is no way for you to go back to Any Side without remaking the lobby.

src/clj/web/lobby.clj
```diff
@@ -312,11 +330,14 @@
...
   [{client-id :client-id
-    gameid :?data}]
+    {:keys [gameid side]} :?data}]
   (let [game (game-for-id gameid)
-        fplayer (first (:players game))]
-    (when (= (:ws-id fplayer) client-id)
-      (refresh-lobby-update-in gameid [:players] (partial mapv swap-side)))))
+        first-player (first (:players game))]
+    (when  (= (:ws-id first-player) client-id)
+      (if (> (count (:players game)) 1)
```
NoahTheDuke, line 337:
```suggestion
      (if (< 1 (count (:players game)))
```
this is dumb but i prefer to keep all inequalities smaller on the left, helps keep things consistent (and i find it to be more natural).

PR #5907, opened 2021-05-24
Fix fawkes double charging

jove-m:
Fawkes double charges you when you try to pump it. This is due to its x-credit cost ability firing another ability that's given that same cost.


NoahTheDuke:
Should this be rolled into #5908?

NoahTheDuke:
Handled in #5908.

PR #5909, opened 2021-05-25
Runner can only spend virus counters on their cards

cardboardtech:
Now that we have a Corp card that has virus counters on it (Sandstone), we needed to filter out corp cards when selecting cards to take virus counters from via The Nihilist or Freedom. 

Another issue I discovered while looking into this is that several cards including The Nihilist and Freedom use the function "number-of-virus-counters" in game.core.virus to determine if they should be able to activate their abilities. That function includes viruses on Corp cards as well which means The Nihilist will trigger if Sandstone has 2 virus counters. We'll want to make a new function or pass in an additional parameter so we only check for runner viruses where appropriate.

### DIFF
--- a/src/clj/game/core/pick_counters.clj
+++ b/src/clj/game/core/pick_counters.clj
@@ -1,6 +1,6 @@
 (ns game.core.pick-counters
   (:require
-    [game.core.card :refer [get-card get-counters installed?]]
+    [game.core.card :refer [get-card get-counters installed? runner?]]
     [game.core.card-defs :refer [card-def]]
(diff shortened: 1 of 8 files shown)


src/clj/game/core/pick_counters.clj
```diff
@@ -40,6 +40,7 @@
                                 (= "Hivemind" (:title %)))
                             true)
                           (installed? %)
+                          (= "Runner" (:side %))
```
NoahTheDuke, line 43:
```suggestion
                          (runner? %)
```

NoahTheDuke: Will have to import as well.

PR #5915, opened 2021-05-27
Fix Gatekeeper message. It now includes origin of chosen cards (issue 5753)

### LINKED ISSUE #5753: Gatekeeper should show origin of cards in log message
Gatekeeper doesn't show whether cards came from HQ or Archives.

h4pp7:
Closes #5753.

I changed the message of Gatekeeper's first sub to include the zone (HQ or
Archives) of the chosen Agendas. It now works basically the same way
as Attitude Adjustment.

Which made me wonder if this should be a seperate function
`reveal-and-shuffle-into-rd-effect`, or even if `game.core.shuffling/shuffle-into-rd-effect` could
be enhanced. But AFIK only Attitude Adjustment and Drudge Work reveal and
shuffle from Archives and/or HQ. And these also have an additional credit gain
effect.

Maybe the `enemurate-str` helper function could be useful somewhere else. But I 
can also just do that in the card definition.

### DIFF
--- a/test/clj/game/cards/ice_test.clj
+++ b/test/clj/game/cards/ice_test.clj
@@ -1685,37 +1685,109 @@
 
 (deftest gatekeeper
   ;; Gatekeeper
-  (do-game
-    (new-game {:corp {:deck [(qty "Ice Wall" 10)]
-                      :hand ["Gatekeeper" "Posted Bounty" "Hostile Takeover"]
-                      :discard [(qty "Ice Wall" 2) "Hostile Takeover"]}})
-    ;; Actual test
-    (play-from-hand state :corp "Gatekeeper" "New remote")
-    (take-credits state :corp)
-    (let [gate (get-ice state :remote1 0)
-          hand (-> (get-corp) :hand count)
-          deck (-> (get-corp) :deck count)
-          num-shuffles (count (core/turn-events state :corp :corp-shuffle-deck))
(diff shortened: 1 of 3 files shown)


test/clj/game/cards/ice_test.clj
```diff
@@ -1685,37 +1685,116 @@
...
+                                   "NEXT Wave 2"
+                                   "Accelerated Beta Test"]
+                            :discard ["Efficiency Committee"
+                                      "Hyperloop Extension"]}})
+          (play-from-hand state :corp "Gatekeeper" "HQ")
+          (take-credits state :corp)
+          (let [hand    (:hand (get-corp))
+                discard (:discard (get-corp))
+                gate    (get-ice state :hq 0)
+                log-str (str "Corp uses Gatekeeper to shuffle"
```
NoahTheDuke, line 1782: I understand the desire for clarity, but this could easily be one or two lines, not 7 lol.

NoahTheDuke:
I think it's fine how you have it right now. If we come across another card that needs this functionality, we can export to a function later. And `enumerate-str` has use and could be used in other places, tho I'm forgetting exactly where at the moment.

h4pp7:
I reduced the number of lines a bit now 😅
I'll look for other places where we could use enumerate-string when I find the time after this is merge.

Issue #5902, opened 2021-05-22
Trick of Light current implementation and SU21 wording


francescopellegrini:
Trick of Light, up to [System Core 2019](https://netrunnerdb.com/en/card/25102), was referring to 2 cards, the former being the source of advancement tokens and the latter being their destination.

The [SU21 version](https://netrunnerdb.com/en/card/31058) inverted this pattern, thus making the current implementation of the card invalid or confusing at the very least.

PR #5916, opened 2021-05-27
Fix reversed order of card choice for Trick of Light

h4pp7:
See issue #5902

The player now chooses the card, to which to move advancement tokens
onto, and then the source of the advancement tokens. System Update 2021
reversed the wording compared to before.


src/clj/game/cards/operations.clj
```diff
@@ -2494,25 +2494,26 @@
...
-                                                        (can-be-advanced? %))}
-                                  :msg (msg "move " c " advancement tokens from "
-                                            (card-str state fr) " to " (card-str state target))
-                                  :effect (effect (add-prop :corp target :advance-counter c {:placed true})
-                                                  (add-prop :corp fr :advance-counter (- c) {:placed true}))})
+                               (let [card-to-take target]
+                                 {:async true
+                                  :prompt "Move how many advancement tokens?"
+                                  :choices (take (inc (get-counters card-to-take :advancement)) ["0" "1" "2"])
+                                  :msg (msg "move " (str->int target) " advancement tokens from "
```
cardboardtech, line 2513: No need to convert to int for the message.

```suggestion
                                  :msg (msg "move " target " advancement tokens from "
```

cardboardtech (commented):
Haha, looks like we were working on the same thing. #5917 
Our solutions are pretty much the same, though I think I like your variable names more.

h4pp7:
Ah, I was afraid that could happen 😅  Is there a way I we can let people know, that we're working on an issue?

I see that you already incorporated one of the variable names and I like your other one more. Together with your improved requirement checking I prefer your pull request over mine!

PR #5918, opened 2021-05-29
Fix Keiko trigger condition (supersedes #5900 )

### LINKED ISSUE #5892: Keiko triggers twice if a companion is installed and used in the same turn

h4pp7:
I realized that thehead branch for my pull request #5900 was my master branch, which was a bit annoying. I made a new branch and this new pull request. Sorry for that!

Original message:
Fixes #5892 

Keiko gave credits a second time in a turn, when credits where used from
a companion after a companion was installed that turn.

`event-count` with `:runner-install` now gets a different filter
predicate than the call with `:spent-credits-from-card`.

### DIFF
--- a/src/clj/game/cards/hardware.clj
+++ b/src/clj/game/cards/hardware.clj
@@ -942,10 +942,12 @@
    :events [{:event :spent-credits-from-card
              :req (req (and (not (facedown? target))
                             (has-subtype? target "Companion")
-                            (let [f #(and (not (facedown? (first %)))
-                                          (has-subtype? (first %) "Companion"))]
-                              (= 1 (+ (event-count state :runner :spent-credits-from-card f)
-                                      (event-count state :runner :runner-install f))))))
+                            (= 1 (+ (event-count state :runner :spent-credits-from-card
+                                                 #(and (not (facedown? (first %)))
+                                                       (has-subtype? (first %) "Companion")))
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/hardware.clj
```diff
@@ -942,10 +942,12 @@
...
                             (has-subtype? target "Companion")
-                            (let [f #(and (not (facedown? (first %)))
-                                          (has-subtype? (first %) "Companion"))]
-                              (= 1 (+ (event-count state :runner :spent-credits-from-card f)
-                                      (event-count state :runner :runner-install f))))))
+                            (= 1 (+ (event-count state :runner :spent-credits-from-card
+                                                 #(and (not (facedown? (first %)))
+                                                       (has-subtype? (first %) "Companion")))
+                                    (event-count state :runner :runner-install
+                                                 #(and (not (:facedown (first %)))
```
NoahTheDuke, line 949:
```suggestion
                                                 #(and (not (facedown? (first %)))
```

NoahTheDuke: jk not this one

PR #5919, opened 2021-05-29
add MU icon where it was missing

francescopellegrini:
Haven't tested the rendered output, hope this doesn't break anything..


NoahTheDuke:
We try to stick to how the text is written on cards, so we use MU instead of the icon for all non-gain text.

PR #5920, opened 2021-05-30
changed calls to is' with potential to print the whole state

h4pp7:
When `game.utils-test/is'` is called with a function in the first
position of the assertion (like in `(is' (prompt-is-type? state side
:select))`) and that test fails, the sub-forms of the assertion finally
get evaluated in `game.utils-test/error-wrapper` when given to
`do-report`. If state is one of those sub-forms, the whole state appears
in the report.

If tests get run in the repl or with lein, the whole state gets printed
into the console or log. `lein test` even crashes with a Java
OutOfMemory error after a while.

To prevent this, for all of the occurences of that pattern, we test the
assertion indirectly.

In addidtion, the test if a prompt is shown in `play-run-event-impl` did
not return an exception when no prompt is shown (because `(get-in @state
[:runner :prompt])` returns an empty list, when no prompt is shown).

### DIFF
--- a/test/clj/game/core_test.clj
+++ b/test/clj/game/core_test.clj
@@ -439,7 +439,7 @@
     (when (and (:run @state)
                (= [server] (get-in @state [:run :server]))
                (run-continue state))
-      (is' (some? (not-empty (get-in @state [:runner :prompt]))) "A prompt is shown")
+      (is' (seq (get-in @state [:runner :prompt])) "A prompt is shown")
       (is' (true? (get-in @state [:run :successful])) "Run is marked successful"))))
 
 (defmacro play-run-event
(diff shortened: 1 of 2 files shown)


test/clj/game/core_test.clj
```diff
@@ -439,8 +439,8 @@
     (when (and (:run @state)
                (= [server] (get-in @state [:run :server]))
                (run-continue state))
-      (is' (get-in @state [:runner :prompt]) "A prompt is shown")
-      (is' (get-in @state [:run :successful]) "Run is marked successful"))))
+      (is' (some? (not-empty (get-in @state [:runner :prompt]))) "A prompt is shown")
```
NoahTheDuke, line 442: Better to use seq, which returns nil if empty.

NoahTheDuke (approved):

PR #5922, opened 2021-05-31
Fixes atmans counters being payable with the wrong type of recurring credits

### LINKED ISSUE #5772: Atman does not work with recurring credits
When you select how many power counters to place on Atman, it does not take into account recurring credits from e.g. Mantle or Multithreader, and does not prompt to use recurring credits from those sources when resolving the ability.

So if you have two Multithreaders installed and 1 c in your credit pool, you should be allowed to go up to 5 power counters, but the prompt will only give you the option for 1, and when you resolve the ability it automatically takes your 1c from your credit pool.

jove-m:
Closes #5772 Atman was using install credits (.e.g sahasrara) for its ability. Using continue-ability with the :source type set properly is the same way engolo's on encounter ability handles this.

### DIFF
--- a/src/clj/game/cards/programs.clj
+++ b/src/clj/game/cards/programs.clj
@@ -407,12 +407,11 @@
                                               (continue :runner nil))}]}))
 
 (defcard "Atman"
-  {:on-install {:prompt "How many power counters?"
-                :choices {:number (req (total-available-credits state :runner eid card))}
-                :msg (msg "add " target " power counters")
-                :async true
-                :effect (effect (add-counter card :power target)
-                                (lose-credits eid target))}
+  {:on-install {:effect (effect
+                          (continue-ability {:eid (assoc eid :source-type :ability :source card)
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/programs.clj
```diff
@@ -407,12 +407,11 @@
...
-                :choices {:number (req (total-available-credits state :runner eid card))}
-                :msg (msg "add " target " power counters")
-                :async true
-                :effect (effect (add-counter card :power target)
-                                (lose-credits eid target))}
+  {:on-install {:effect (effect
+                          (continue-ability {:eid (assoc eid :source-type :ability :source card)
+                                             :cost [:x-credits]
+                                             :msg (msg "add " (cost-value eid :x-credits) " power counters")
+                                             :effect (effect (add-counter card :power (cost-value eid :x-credits)))} card targets))}
```
NoahTheDuke, line 414: let's put the `card targets` on it's own line

PR #5924, opened 2021-05-31
move clearing of :this-turn to start-turn

### LINKED ISSUE #5888: political dealings + seamless launch
If the corp installs an agenda after the runners last click using political dealings (e.g. by rezzing a spin doctor during the last PA window on runner's turn) the game will not allow seamless launch to be used on that agenda.

h4pp7:
fixes #5888 

Cards that get installed after the "End Turn" button is clicked count as
installed this turn until the end of the next turn. Agendas that get
installed between end of turn and start of turn cannot be targeted by
Seamless Launch that turn because of that. Vulnerability Audit is
affected by the same problem (both check for `(not :this-turn)`.

To fix this, the `:this-turn` flag now gets cleared at start of turn
(like the fix for Hayley with `:turn-events`).

But this might create similar problems now for other szenarios,
where you need to install cards in your turn, but before the draw step.
They will now loose the :this-turn flag.

So I don't know if this is a good idea...

TODO:
- [ ] tests for these kind of scenarios

### DIFF
--- a/src/clj/game/core/turns.clj
+++ b/src/clj/game/core/turns.clj
@@ -54,7 +54,20 @@
   ; Don't clear :turn-events until the player clicks "Start Turn"
   ; Fix for Hayley triggers
   (swap! state assoc :turn-events nil)
-
+  (doseq [card (all-active-installed state :runner)]
+    ;; Clear :installed :this-turn as turn has ended
+    (when (= :this-turn (installed? card))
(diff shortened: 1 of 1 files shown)


NoahTheDuke:
Hmmmmm I can't think of a scenario off-hand that breaks because of this, but I don't think it's the correct way to handle this issue.

Issue #5764, opened 2021-04-06
Sneakdoor is still declaring a successful HQ run when HQ has crissum on it.


BlueHairedHackerCatgirl:
This shouldn't be the case, what should happen is you get a not-successful run on HQ and enter the access phase.

NoahTheDuke:
Thanks for the report!

NBKelly:
@NoahTheDuke this one has a test asserting the correct behavior and can be closed (see #6238)

NBKelly:
This was fixed, and we've got unit tests for it now via #6238

PR #5927, opened 2021-06-01
Rejigs the successful run logic to better account for sneakdoor/crisium

### LINKED ISSUE #5713: Omar, DJ Fenris, Steve, Crisium Grid
I know the above combination creates a whole set of issues, but: I made an Omar run with DJ Fenris installed set to Steve on Archives, which I diverted to HQ, which had a rezzed Crisium Grid. Jnet prompted me to choose cards for Steve, which it shouldn't do.

jove-m:
Closes #5713 #5764. This changes the flow of declaring a run successful. 

Previously it would check the server didn't have crisium, then mark the run as successful and register it. Then it would run sneakdoor/omar, which would remove the registration, change the server, then add a new registration, then the run code would trigger any run successful events. I haven't looked too closely into this as I've removed the code here anyway but I believe the "remove the registration" step was bugged in that it removed the first registration in the list rather than the last. 

The new flow starts with the same check for crisium, if we don't find it we go straight to running sneakdoor. Sneakdoor just changes the server. Then we check again for crisium, if we don't find it we do everything like marking the run successful, registering it and running any events.

### DIFF
--- a/src/clj/game/cards/identities.clj
+++ b/src/clj/game/cards/identities.clj
@@ -1251,10 +1251,8 @@
              :prompt "Treat as a successful run on which server?"
              :choices ["HQ" "R&D"]
              :effect (req (let [target-server (if (= target "HQ") :hq :rd)]
-                            (swap! state update-in [:runner :register :successful-run] next)
                             (swap! state assoc-in [:run :server] [target-server])
                             (trigger-event state :corp :no-action)
-                            (swap! state update-in [:runner :register :successful-run] conj target-server)
                             (system-msg state side (str "uses Omar Keung: Conspiracy Theorist to make a successful run on " target))))}
             {:event :run-ends
              :effect (effect (update! (dissoc-in card [:special :omar-run])))}]})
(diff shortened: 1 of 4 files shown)


NoahTheDuke:
Good catch! You're dropping the `eid` in `register-successful-run`, so it'll need to be:

```clojure
(defn- register-successful-run
  [state side eid server]
  ;; TODO: :pre-successful-run exists merely for Omar Keung and Sneakdoor Beta
  ;; Needs prevention system to remove
  (queue-event state :pre-successful-run (select-keys (:run @state) [:server :run-id]))
  (wait-for (checkpoint state nil (make-eid state eid))
            (if (any-effects state side :block-successful-run)
              (do (swap! state update-in [:runner :register :successful-run] conj (-> @state :run :server first))
                  (swap! state assoc-in [:run :successful] true)
                  (queue-event state :successful-run (select-keys (:run @state) [:server :run-id]))
                  (checkpoint state nil eid)))))
```

PR #5933, opened 2021-06-02
standardize waiting prompt messages

francescopellegrini:
Rationales:
- use `use` if the subject is who originally played the card
- use `choose an option for` if there is a finite number of choices
- use `decide on` if either there is a non-finite number of choices or there is only 1 (e.g. do X to prevent Y)
- always mention the name of the card in full, even for IDs.

Suggestions are welcomed on the wording for "ambush" ICEs (e.g. Archangel, Herald, Chrysalis, Sapper).

### DIFF
--- a/src/clj/game/cards/events.clj
+++ b/src/clj/game/cards/events.clj
@@ -384,7 +384,7 @@
                  :ability
                  {:msg "force the Corp to add all cards in HQ to the top of R&D"
                   :player :corp
-                  :waiting-prompt "Corp to decide on CBI Raid"
+                  :waiting-prompt "Corp to make a decision"
                   :async true
                   :effect (effect
                             (continue-ability
@@ -525,7 +525,7 @@
                               (continue-ability
                                 state :corp
                                 {:optional
-                                 {:waiting-prompt "Corp to decide on Credit Crash"
+                                 {:waiting-prompt "Corp to choose an option"
                                   :prompt (msg "Spend " cost " [Credits] to prevent the trash of " title "?")
                                   :player :corp
                                   :yes-ability {:async true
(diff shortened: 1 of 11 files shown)


NoahTheDuke:
Thank you so much for this PR! Love to see this kind of consistency. We're really lax about this stuff, so it's great to see care and consideration.

Because we always display the title of the card at the top of the prompt window, there's no need to include the name of the card in the waiting prompt message. (Not writing the full name was the primary reason to add that functionality, lol).

francescopellegrini:
Oh my, this is embarrassing.. I totally missed that detail, I'm sorry!

What could be some alternative wordings?

- `Corp/Runner to choose an option` if there is a finite number of choices
- `Corp/Runner to decide` if either there is a non-finite number of choices or there is only 1 (e.g. do X to prevent Y)

NoahTheDuke:
No worries! I think those alternatives are great too. I don't have strong feelings one way or the other about the wording, so whichever you feel is best is good for me.

If you're not on Stimslack, https://stimslackinvite.herokuapp.com/ and then join #jinteki and holler at me, and I'll add you to our jnet dev slack server so we don't have to chat purely through PRs

src/clj/game/cards/events.clj
```diff
@@ -2273,8 +2273,8 @@
 (defcard "Push Your Luck"
   (letfn [(corp-choice [spent]
             {:player :corp
-             :waiting-prompt "Corp to guess even or odd"
-             :prompt "How many credits were spent?"
+             :waiting-prompt "Corp to choose an option"
+             :prompt "Even or odd?"
```
francescopellegrini, line 2277: Changed the text of this prompt because the answer is binary and not related to the amount of credits spent.

PR #5934, opened 2021-06-02
Remove alpha of typeahead in the Deckbuilder

cardboardtech:
Since this is a white dropdown with black text that appears over a white textbox with black text, I decided to remove the alpha entirely from the typeahead. Most of our other controls have at least an alpha of o-90. I did test the typeahead with an alpha of o-90 and I think it looks fine and could work if we wanted to go that way and I don't remember what it looked like before so if we want the alpha its an easy add.

Without alpha (committed)
image

Alpha of o-90
image

### DIFF
--- a/src/css/deckbuilder.styl
+++ b/src/css/deckbuilder.styl
@@ -161,7 +161,7 @@
                     padding: 2px
                     margin: 2px 0 0 0
                     width: 308px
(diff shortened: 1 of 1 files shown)


NoahTheDuke:
This is a good change, absolutely no reason to have text fields transparent.

PR #5952, opened 2021-06-10
Refactors check-damage-prevention to use trigger-event-simult

jove-m:
Hopefully the first of many.

### DIFF
--- a/src/clj/game/core/damage.clj
+++ b/src/clj/game/core/damage.clj
@@ -139,13 +139,15 @@
              state player nil (str "Prevent any of the " (- n already-prevented) " " (name type) " damage?") ["Done"]
              (fn [_] (let [prevent (get-in @state [:damage :damage-prevent type])
                            damage-prevented (if prevent (- prevent already-prevented) false)]
-                       (if damage-prevented (trigger-event state side :prevented-damage type prevent) nil)
                        (system-msg state player
                                    (if damage-prevented (str "prevents "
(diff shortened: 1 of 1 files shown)


src/clj/game/core/damage.clj
```diff
@@ -139,13 +139,15 @@
...
                            damage-prevented (if prevent (- prevent already-prevented) false)]
-                       (if damage-prevented (trigger-event state side :prevented-damage type prevent) nil)
                        (system-msg state player
                                    (if damage-prevented (str "prevents "
                                                              (if (>= damage-prevented Integer/MAX_VALUE) "all" damage-prevented)
                                                              " " (name type) " damage") "will not prevent damage"))
                        (clear-wait-prompt state other-player)
-                       (effect-completed state side eid)))))
+                       (if damage-prevented
+                           (wait-for (trigger-event-simult state side (make-eid state eid) :prevented-damage type prevent)
```
NoahTheDuke, line 148: If the only thing inside of the wait-for is `effect-completed`, just use the awaited function directly: `(trigger-event-simult state side eid :prevented-damage type prevent)`. It will call `effect-completed` for you.

NoahTheDuke: also, `trigger-event-simult` requires that the arg after the event keyword is a map (or `nil`) that carries various options. so at the very least, `(trigger-event-simult state side eid :prevented-damage nil type prevent)`

PR #5956, opened 2021-06-11
Refactor/event cleanup

jove-m:
Whole bunch of stuff. Purging, shuffling, start of game effects, ice and breakers, damage and accessing have all been updated to better async. It all basically works and passes the test suite though it needs a lot of cleanup. 
Changes to the engine so that cancel-effect and leaves-play can take an ability map and so be async and the addition of the effect-seq macro and the effect-sequence function that it uses.
Also random bugfixes that I should really pull out to their own branches. Power to the people now uses the access trigger, I figure it using steal cost was an old workaround that's no longer needed and Wyvern now doesn't count as searching the stack.


NoahTheDuke:
I won't be on my computer much this weekend but I will read through this and leave comments on Monday.

src/clj/game/core/access.clj
```diff
@@ -203,69 +203,69 @@
...
-                            ability (access-ab ability-card)]
-                        (when (and (:run @state)
-                                   (:trash? ability true))
-                          (swap! state assoc-in [:run :did-trash] true))
-                        (wait-for (resolve-ability state side (make-eid state ability-eid) ability ability-card [card])
-                                  (let [card (first async-result)]
-                                    (trigger-event state side :no-steal card)
-                                    (access-end state side eid card {:stolen (in-scored? card)}))))))}
-      card nil)))
+  (wait-for (trigger-event-simult state side (make-eid state eid) :pre-steal-cost nil card)
```
NoahTheDuke, line 206: Should stay `trigger-event`.

src/clj/game/core/commands.clj
```diff
@@ -40,7 +40,7 @@
   (set-prop state side target :advance-counter value)
   (system-msg state side (str "sets advancement counters to " value " on "
                               (card-str state target)))
-  (trigger-event state side :advancement-placed target))
+  (wait-for (trigger-event-simult state side (make-eid state) :advancement-placed nil target)))
```
NoahTheDuke, line 43: Should be left as `trigger-event`. If `set-adv-counter` isn't async, this will make things worse.

src/clj/game/core/damage.clj
```diff
@@ -139,13 +139,14 @@
...
                            damage-prevented (if prevent (- prevent already-prevented) false)]
-                       (if damage-prevented (trigger-event state side :prevented-damage type prevent) nil)
                        (system-msg state player
                                    (if damage-prevented (str "prevents "
                                                              (if (>= damage-prevented Integer/MAX_VALUE) "all" damage-prevented)
                                                              " " (name type) " damage") "will not prevent damage"))
                        (clear-wait-prompt state other-player)
-                       (effect-completed state side eid)))))
+                       (if damage-prevented
+                           (trigger-event-simult state side eid :prevented-damage nil type prevent)
```
NoahTheDuke, line 148: `:prevented-damage` isn't used in any cards so it should just be removed.

src/clj/game/core/damage.clj
```diff
@@ -156,10 +157,11 @@
...
    (swap! state update-in [:damage :damage-prevent] dissoc type)
    ;; alert listeners that damage is about to be calculated.
-   (trigger-event state side :pre-damage type card n)
    (let [active-player (get-in @state [:active-player])]
-     (if unpreventable
-       (resolve-damage state side eid type (damage-count state side type n args) args)
-       (wait-for (check-damage-prevention state side type n args active-player)
-                 (wait-for (check-damage-prevention state side type n args (if (= active-player :corp) :runner :corp))
-                           (resolve-damage state side eid type (damage-count state side type n args) args)))))))
+     (wait-for (trigger-event-simult state side (make-eid state eid) :pre-damage nil type card n)
```
NoahTheDuke, line 161: Should stay `trigger-event`.

src/clj/game/core/eid.clj
```diff
@@ -38,6 +39,14 @@
       (swap! state update :effect-completed dissoc (:eid eid))
       results)))
 
+
```
NoahTheDuke, line 42: Extra space

src/clj/game/core/eid.clj
```diff
@@ -38,6 +39,14 @@
...
+(defn effect-sequence
+  [state side eid args func]
```
NoahTheDuke, line 44: Spacing is a little messed up in this function. I also think the `func` should be first, and the `args` should be last.

src/clj/game/core/eid.clj
```diff
@@ -38,6 +39,14 @@
...
+    (if (empty? args)
+        (effect-completed state side eid)
+        (wait-for (func state side (make-eid state eid) (first args))
+                  (effect-sequence state side eid (rest args) func))))
```
NoahTheDuke, line 48: `rest` should be `next` to avoid laziness.

src/clj/game/core/engine.clj
```diff
@@ -439,9 +439,12 @@
   ([state side card message choices ability args]
    (let [f #(resolve-ability state side ability card [%])]
      (show-prompt state side (:eid ability) card message choices f
-                  (if-let [f (:cancel-effect args)]
-                    (assoc args :cancel-effect #(f state side (:eid ability) card [%]))
-                    args)))))
+                  (let [c (:cancel-effect args)]
+                    (assoc args :cancel-effect 
+                             (if c (if (fn? c)
+                                       #(do (c state side (:eid ability) card [%]) (effect-completed state side (:eid ability)))
```
NoahTheDuke, line 445: Don't pass in the `eid` if the function won't complete it. Use `(make-eid state (:eid ability))` instead.

src/clj/game/core/gaining.clj
```diff
@@ -45,47 +46,49 @@
...
-  (doseq [[cost-type amount] (partition 2 args)]
-    (cond
-      ;; amount is a map, merge-update map
-      (map? amount)
-      (doseq [[subtype amount] amount]
-        (swap! state update-in [side cost-type subtype] (safe-inc-n amount))
-        (swap! state update-in [:stats side :gain cost-type subtype] (fnil + 0) amount))
+(defn gain [state side eid & args]
+  (effect-seq state side eid [[cost-type amount] (partition 2 args)]
+    (let [continuation #(trigger-event-simult state side eid (if (= side :corp) :corp-gain :runner-gain) nil [cost-type amount])]
```
NoahTheDuke, line 51: I'm not a fan of this kind of code reuse. Makes it harder to see what's happening in each sub-form.

src/clj/game/core/gaining.clj
```diff
@@ -45,47 +46,49 @@
...
-(defn gain [state side & args]
-  (doseq [[cost-type amount] (partition 2 args)]
-    (cond
-      ;; amount is a map, merge-update map
-      (map? amount)
-      (doseq [[subtype amount] amount]
-        (swap! state update-in [side cost-type subtype] (safe-inc-n amount))
-        (swap! state update-in [:stats side :gain cost-type subtype] (fnil + 0) amount))
+(defn gain [state side eid & args]
+  (effect-seq state side eid [[cost-type amount] (partition 2 args)]
```
NoahTheDuke, line 50: If the goal of `effect-seq` is to act like `doseq`, we gotta change how it's written cuz this isn't correct for the implementation.

src/clj/game/core/ice.clj
```diff
@@ -68,11 +68,11 @@
...
 (defn add-sub!
-  ([state side ice sub] (add-sub! state side ice sub (:cid ice) nil))
-  ([state side ice sub cid] (add-sub! state side ice sub cid nil))
-  ([state side ice sub cid args]
+  ([state side eid ice sub] (add-sub! state side eid ice sub (:cid ice) nil))
+  ([state side eid ice sub cid] (add-sub! state side eid ice sub cid nil))
+  ([state side eid ice sub cid args]
    (update! state :corp (add-sub ice sub cid args))
-   (trigger-event state side :subroutines-changed (get-card state ice))))
+   (trigger-event-simult state side eid :subroutines-changed nil (get-card state ice))))
```
NoahTheDuke, line 75: Should stay `trigger-event`.

src/clj/game/core/ice.clj
```diff
@@ -375,14 +376,15 @@
...
+  ([state side breaker] (update-breaker-strength state side (make-eid state) breaker))
+  ([state side eid breaker]
   (let [breaker (get-card state breaker)
         old-strength (get-strength breaker)
         new-strength (breaker-strength state side breaker)
         changed? (not= old-strength new-strength)]
     (update! state side (assoc (get-card state breaker) :current-strength new-strength))
-    (trigger-event state side :breaker-strength-changed (get-card state breaker) old-strength)
-    changed?))
+    (trigger-event-simult state side eid :breaker-strength-changed nil (get-card state breaker) old-strength)
```
NoahTheDuke, line 386: Should stay `:trigger-event`

NoahTheDuke (changes requested):
I sent a message on slack, but I think this should be closed and split into individual PRs for each events (or maybe two events at a time if they're small).

src/clj/game/core/eid.clj
NoahTheDuke: jk, having seen how it's being used (as a drop-in replacement for `doseq`, this is fine. Function body needs work tho.

PR #5962, opened 2021-06-19
Adds gain/lose-clicks

jove-m:
Main change is adding gain/lose-clicks and making cards use them. They trigger the new events :corp/runner-click-gain/lose.

Changes to cards include adjusted chronotype using the new event, changes to mca austerity policy to be more in line with the card text and so still trigger chronotype. Also an incidental bugfix in that riot suppression will no longer incorrectly trigger chronotype and I've added a test for this. I also added a test for stacking mca. It's something I wanted to check with my changes and it's nice to have. 

I looked at making chronotype trigger from bioroids as well which isn't a problem, but apparently there's no ruling for if it should yet.

### DIFF
--- a/src/clj/game/cards/resources.clj
+++ b/src/clj/game/cards/resources.clj
@@ -108,14 +108,13 @@
                             (effect-completed state :runner eid)))}]})
 
 (defcard "Adjusted Chronotype"
-  {:events [{:event :runner-lose
-             :req (req (and (some #{:click} target)
-                            (let [click-losses (count (filter #(= :click %) (mapcat first (turn-events state side :runner-lose))))]
-                              (or (= 1 click-losses)
-                                  (and (= 2 click-losses)
-                                       (has-flag? state side :persistent :genetics-trigger-twice))))))
+  {:events [{:event :runner-click-loss
+             :req (req (let [click-losses (count (filter #(= :click %) (mapcat first (turn-events state side :runner-lose))))]
+                            (or (= 1 click-losses)
+                                (and (= 2 click-losses)
+                                     (has-flag? state side :persistent :genetics-trigger-twice)))))
              :msg "gain [Click]"
-             :effect (effect (gain :runner :click 1))}]})
+             :effect (effect (gain-clicks :runner 1))}]})
 
 (defcard "Aeneas Informant"
   {:events [{:event :no-trash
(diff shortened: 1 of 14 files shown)


src/clj/game/cards/resources.clj
```diff
@@ -108,14 +108,13 @@
...
 
 (defcard "Adjusted Chronotype"
-  {:events [{:event :runner-lose
-             :req (req (and (some #{:click} target)
-                            (let [click-losses (count (filter #(= :click %) (mapcat first (turn-events state side :runner-lose))))]
-                              (or (= 1 click-losses)
-                                  (and (= 2 click-losses)
-                                       (has-flag? state side :persistent :genetics-trigger-twice))))))
+  {:events [{:event :runner-click-lose
+             :req (req (let [click-losses (count (filter #(= :click %) (mapcat first (turn-events state side :runner-lose))))]
```
NoahTheDuke, line 112: Should this be checking `:runner-click-lose`?

src/clj/game/core/gaining.clj
```diff
@@ -102,6 +102,23 @@
...
+          (trigger-event state side (if (= :corp side) :corp-click-gain :runner-click-gain) amount args)))))
+
+(defn lose-clicks
+  ([state side amount] (lose-clicks state side amount nil))
+  ([state side amount args]
+    (when (and amount
+               (or (= :all amount)
+                   (pos? amount)))
+      (do (lose state side :click amount)
+          (trigger-event state side (if (= :corp side) :corp-click-lose :runner-click-lose) amount args)))))
```
NoahTheDuke, line 120: Can this be changed to `:X-click-loss` to match losing credits above?

src/clj/game/cards/resources.clj
jove-m: Yeah that's adjusted chronotype. Leaving it on runner-loss is what causes the bug where it triggers from riot suppression.

At the moment it's moot and I'd prefer to wait on the bioroid ruling, but at some point I'd like to cleanup the way clicks are spent. At the moment, most engine click spending uses (deduct that fires no events at all, presumably to avoid triggering chronotype. Obviously that's not needed any more and you can just use (lose-clicks to trigger abilities and (lose to avoid it, the same as credits or tags.

PR #5966, opened 2021-06-24
Mark abilities as Playable

### LINKED ISSUE #5858: Sync ID when flipped doesn't enable the trash resource menu when you have 0 or 1 credits
Many people don't realise that Sync has a flip ability. It turns out that the Sync ID can be flipped, and when it's flipped, you can trash resources for 2 credits fewer. Amazing!
Unfortunately, it looks like jnet tests for the corp having at least 2 credits to trash things, even when the Sync ID is flipped.

cardboardtech:
Like how we mark cards as playable, mark abilities as playable. This lets us disable abilities in the UI easily. For now, I only apply this to basic abilities as hard coding the costs on the client-side did not let us check for additional costs or cost reduction via cards like SYNC. I did not apply this to initiating a run since that seemed to be handled differently. Might look into that more in the future. This fix should allow us to disable any abilities. I chose not to implement that in this pull request as I was worried about side effects and want to do more extensive testing.

Fixes #5858

### DIFF
--- a/src/clj/game/core/diffs.clj
+++ b/src/clj/game/core/diffs.clj
@@ -5,6 +5,7 @@
     [game.core.board :refer [installable-servers]]
     [game.core.card :refer :all]
     [game.core.cost-fns :refer [card-ability-cost]]
+    [game.core.engine :refer [can-trigger?]]
     [game.core.installing :refer [corp-can-pay-and-install? runner-can-pay-and-install?]]
     [game.core.payment :refer [can-pay?]]
     [game.core.play-instants :refer [can-play-instant?]]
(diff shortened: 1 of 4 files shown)


src/clj/game/core/diffs.clj
```diff
@@ -40,6 +42,25 @@
...
 
+(defn ability-playable? [state side card ability]
+  (let [cost (card-ability-cost state side ability card)]
+    (if (can-pay? state side nil cost)
+      (assoc ability :playable true)
+      ability)))
+
+(defn abilities-playable? [state side card ability-kw]
+  (into [] (for [ability (get card ability-kw)]
+             (ability-playable? state side card ability))))
```
NoahTheDuke, line 53:
```suggestion
  (mapv #(ability-playable? state side card %) (get card ability-kw))
```

NoahTheDuke: nvm

PR #5979, opened 2021-07-04
Implements gamenet. 

jove-m:
Changes to the engine are rather than merging up agenda steal costs as they come in, we store them as is alongside an optional source. They're then merged as usual when they're used but the stored list is also given to any events triggered.
Events now pass the source information from their eid through to their :req functions. To do this gather-events now takes an optional eid argument that it passes through to can-trigger? rather than just using (make-eid state). Neither is actually async of course, they just want the source info.
When a triggered event is resolved, it will now have its source info replaced with the correct info for the card the ability came from. 
There's also a bugfix where ice subroutines would have :source (:title card) instead of just :source card. 
Most of the changes to cards are just passing source info through to payment functions or agenda steal costs and removing now unneeded workarounds. Note that a lot of :req functions still need to use the workaround, mostly when calling can-pay?, because :req still has the info of the triggering event.

### DIFF
--- a/src/clj/game/cards/identities.clj
+++ b/src/clj/game/cards/identities.clj
@@ -49,7 +49,7 @@
                    :no-ability {:effect (req (clear-wait-prompt state :corp))}
                    :yes-ability
                    {:async true
-                    :effect (req (if (not (can-pay? state :corp (assoc eid :source card :source-type :ability) card nil :credit 1))
+                    :effect (req (if (not (can-pay? state :corp eid card nil :credit 1))
                                    (do
                                      (toast state :corp "Cannot afford to pay 1 [Credit] to block card exposure" "info")
                                      (expose state :runner eid (:card context)))
@@ -66,7 +66,7 @@
                                        {:async true
                                         :effect
                                         (req (wait-for
-                                               (pay state :corp card [:credit 1])
+                                               (pay state :corp (make-eid state eid) card [:credit 1])
                                                (system-msg state :corp
                                                            (str (:msg async-result)
                                                                 " to prevent "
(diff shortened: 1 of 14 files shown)


src/clj/game/cards/identities.clj
```diff
@@ -505,10 +505,18 @@
...
 (defcard "GameNET: Where Dreams are Real"
-  {:implementation "Credit gain not implemented. You can use shortcut ability."
-   :abilities [{:msg "gain 1 [Credits] (shortcut)"
-                :async true
-                :effect (effect (gain-credits :corp eid 1))}]})
+  (let [gamenet-ability {:req (req (and (:run @state)
+                                        (= (:side (:source eid)) "Corp")
+                                        (or (and (not= (:source-type eid) :runner-trash-corp-cards)
+                                                 (not= (:source-type eid) :runner-steal))
+                                            (some (fn [[cost source]] (and (some #(or (= (first %) :credit) (= (first %) :x-credits)) (merge-costs cost))
```
NoahTheDuke, line 512:
```suggestion
                                            (some (fn [[cost source]] (and (some #(or (= (cost-name %) :credit) (= (cost-name %) :x-credits)) (merge-costs cost))
```

Better to use the multi-methods for accessing cost related stuff, just in case we ever decide to change how they're handled.

src/clj/game/core/access.clj
```diff
@@ -205,9 +204,10 @@
   [state side eid card]
   (trigger-event state side :pre-steal-cost card)
   (swap! state update-in [:stats :runner :access :cards] (fnil inc 0))
-  (let [cost (steal-cost state side card)
+  (let [additional-costs (steal-cost state side card)
+        cost (merge-costs (mapv first additional-costs))
```
NoahTheDuke, line 208:
```suggestion
        cost (merge-costs (mapv value additional-costs))
```

NoahTheDuke: Same reasoning as above.

src/clj/game/core/engine.clj
```diff
@@ -618,15 +618,16 @@
   This is essentially Phase 9.3 and 9.6.7a of CR 1.1:
   http://nisei.net/files/Comprehensive_Rules_1.1.pdf"
   ([state side event targets] (gather-events state side event targets nil))
-  ([state side event targets card-abilities]
+  ([state side event targets card-abilities] (gather-events state side event targets nil (make-eid state)))
+  ([state side event targets card-abilities eid]
```
NoahTheDuke, line 622: Even if they're not being completed, `eid`s should be in the third position for consistency sake.

src/clj/game/core/ice.clj
```diff
@@ -242,7 +242,7 @@
 
 (defn resolve-subroutine!
   ([state side ice sub]
-   (let [eid (make-eid state {:source (:title ice)
+   (let [eid (make-eid state {:source ice
```
NoahTheDuke, line 245: lol thanks for this, it was a dumb decision back in [2018](https://github.com/mtgred/netrunner/commit/34be048e63b0f1975256bd168cd672991d2ab602) and I'm surprised no one has fixed it yet.

test/clj/game/cards/identities_test.clj
```diff
@@ -1363,6 +1363,106 @@
     (click-prompt state :runner "No action") ; Dismiss trash prompt
     (is (last-log-contains? state "Caprice") "Accessed card name was logged")))
 
+(deftest gamenet-where-dreams-are-real
+  (testing "Gain credits from gold farmer"
+    (do-game
+      (new-game {:corp {:id "GameNET: Where Dreams are Real" :hand ["Gold Farmer"]} :runner {:hand ["Corroder"] :credits 10}})
```
NoahTheDuke, line 1369: Let's keep each pair on a separate line to make set-up easier to read.

src/clj/game/core/access.clj
jove-m: This one is selecting the whole cost from the costs and sources list rather than pulling the costs themselves apart like the one above.

NoahTheDuke: Oh good call.

PR #5980, opened 2021-07-05
Prevent using illegal decks outside of casual

jove-m:
What it says on the tin. Just checks that the decks are legal for the format (or that you're playing casual) when listing them for selecting, accepting them as a selection server side or disabling the create game button in the deck builder.

### DIFF
--- a/src/cljc/jinteki/validator.cljc
+++ b/src/cljc/jinteki/validator.cljc
@@ -280,3 +280,13 @@
     (if status
       status
       (calculate-deck-status deck)))
+
+(defn legal-deck?
+ ([deck] (legal-deck? deck (:format deck)))
+ ([deck fmt]
+   (get-in deck
+     [:status (keyword fmt) :legal]
+     (get-in (trusted-deck-status (assoc deck :format fmt))
+         [(keyword fmt) :legal]
+         false))))
+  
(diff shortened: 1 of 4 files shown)


src/cljc/jinteki/validator.cljc
```diff
@@ -280,3 +280,13 @@
     (if status
       status
       (calculate-deck-status deck)))
+
+(defn legal-deck?
+ ([deck] (legal-deck? deck (:format deck)))
+ ([deck fmt]
+   (get-in deck
```
NoahTheDuke, line 287: Instead of nesting `get-in`s, let's break each of these out into a `let` block for readability.

src/cljc/jinteki/validator.cljc
```diff
@@ -280,3 +280,13 @@
...
+     [:status (keyword fmt) :legal]
+     (get-in (trusted-deck-status (assoc deck :format fmt))
+         [(keyword fmt) :legal]
```
NoahTheDuke, line 290: Spacing.

src/cljc/jinteki/validator.cljc
```diff
@@ -280,3 +280,13 @@
...
+         false))))
+
```
NoahTheDuke, line 292: Empty line

src/cljc/jinteki/validator.cljc
NoahTheDuke: Looks like you took this out of the gamelobby file, which is great, so thanks for that. Still good to reformat when moving code like this. Better to get it clean while we're here!

PR #5885, opened 2021-05-14
Splitting of the right-pane into multiple parts

lostgeek:
Goal is to move the visual settings into a tab on the log for easier access and immediate visibility of the result.

https://user-images.githubusercontent.com/1409906/118323993-b4eda500-b501-11eb-8af9-99988dad26c5.mp4

PR #5982, opened 2021-07-07
Fix Command Pop-up

### LINKED ISSUE #5945: Command pop-up no longer shows up
Pop-up with command auto-completion no longer shows up when typing in the chat :(

cardboardtech:
The command pop-up no longer displayed after PR #5885. The overflow settings that were used to prevent clipping prevented the command pop-up from displaying. I moved the overflow settings down a level to the first div of the content panel. This displays all of the inner content correctly while allowing the command pop-up to appear. From my research, this was an issue because the position property and overflow properties were on the same level ([an example I found](https://css-tricks.com/popping-hidden-overflow/)) 

fixes #5945

### DIFF
--- a/src/css/gameboard.styl
+++ b/src/css/gameboard.styl
@@ -587,7 +587,7 @@
 
                                 >div:first-child
                                     display:inline
-                                    width: 168px!important
+                                    width: 168px !important
                                     height: 234.833px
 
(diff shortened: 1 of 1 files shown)


francescopellegrini:
I'd love to see this merged @NoahTheDuke

PR #5984, opened 2021-07-09
Various stability improvements

jove-m:
There are a bunch of different things here, all with the intent of improving stability.

The simplest is the fix to the zombie games problem. The response to a game being created getting dropped will put the client in a bad state, which allows you to spam the new game button actually causing the problem. Now the server ignores new game requests if you already have a game. 

We still want something in the client to help with the bad state though so the full game list now includes a :selected property, this indicates that this is the game the client should be in. 

Then there's the error handling and state rollback. If there's an error in an action handler the client will be informed and the state will be rolled back to before the action started. When the client gets the error it displays a toast and releases the interface lock. There's no actual error detail being sent to the client, just that an error happened. This does mean all exceptions will now use the error reporting stuff that was previously only used by the one custom exception. There's no stack trace there but I'm not sure how much use a stack trace in the logs has been. It can always be added again if wanted. I also worry a little about the performance impact of snapshotting the state before every action for rollback. I'm not sure how clojure handles that internally. Whatever the case it shouldn't be too bad, no more than the snapshots taken for /undo-click etc. 

Upon disconnect the client will lock itself. When reconnecting it will request the full game list and state of any game it's in. When it gets the new state it will release the lock. The game state is sent via a new netrunner/resync request followed by a previously deprecated netrunner/state response.

### DIFF
--- a/src/cljs/nr/gameboard/actions.cljs
+++ b/src/cljs/nr/gameboard/actions.cljs
@@ -7,6 +7,11 @@
             [nr.utils :refer [toastr-options]]
             [nr.ws :as ws]))
 
+(defn reset-game [state]
+  (reset! game-state (assoc state :side (get-side state)))
+  (reset! last-state @game-state)
+  (reset! lock false))
+
 (defn init-game [state]
   (let [side (get-side state)]
     (.setItem js/localStorage "gameid" (:gameid @app-state))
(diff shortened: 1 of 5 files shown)


src/cljs/nr/gameboard/actions.cljs
```diff
@@ -33,10 +39,15 @@
   (when (= gameid (:gameid @game-state))
     (toast (tr [:game.inactivity "Game closed due to inactivity"]) "error" {:time-out 0 :close-button true})))
 
-(defmethod ws/-msg-handler :netrunner/state [{data :?data}] (init-game (parse-state data)))
+(defn handle-error []
+  (toast (tr [:game.error "Internal Server Error. Please type /bug in the chat and follow the instructions."]) "error" {:time-out 0 :close-button true})
+  (reset! lock false))
+
+(defmethod ws/-msg-handler :netrunner/state [{data :?data}] (reset-game (parse-state data)))
```
NoahTheDuke, line 46: When is init-game being called now?

jove-m: From launch-game, so as a result of :netrunner/start.

PR #5987, opened 2021-07-14
Prevents an infinite loop in toasts

jove-m:
Currently it's possible for diff generation to cause a toast. playable? checks if you can run to play a run event. If you can't it will toast you telling you why. This leads to out of place toasts when you can't run with run events in hand. If you generate some other toast while in this state it will lead to an infinite loop. The client will acknowledge the toast, this will cause a diff and a new toast to be generated and it loops forever. 

Easiest way to reproduce is to play as liza and run a central through excalibur with a run event in hand. When you get the toast for taking the liza tag the loop will begin. 

This patch adds a silent argument to the can-run? function that suppresses the toast and passes it through when generating diffs.

### DIFF
--- a/src/clj/game/core/diffs.clj
+++ b/src/clj/game/core/diffs.clj
@@ -35,7 +35,7 @@
                   (operation? card))
               (can-play-instant?
                 state side {:source :action :source-type :play}
-                card {:base-cost [:click 1]})])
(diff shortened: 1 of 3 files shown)


jove-m:
Fixes #5989

PR #5996, opened 2021-07-24
Auto pump and break variable cost and strength

### LINKED ISSUE #5734: Unity does not seem to have "Fully break"

### LINKED ISSUE #5736: Marjanah not getting discount
The cost to use Marjanah is not changing after the first run of a turn.

### LINKED ISSUE #5994: Marjanah no offering reduction
Run completion replacement from Security Testing leads to Marjanah not providing break reduction on later runs in the turn.

cardboardtech:
Expanded out the usage of `:break-cost-bonus` so cost changes will display properly in the abilities window and be taken into account when calculating auto pump and break abilities.
image image

Following a similar pattern, I expanded this out to the pump strength with `:pump-bonus` so breakers with non-static pump strength such as Unity can use auto pump and break abilities.
image image

fixes #5736 
fixes #5994 
resolves #5734

### DIFF
--- a/src/clj/game/cards/programs.clj
+++ b/src/clj/game/cards/programs.clj
@@ -2554,16 +2554,12 @@
                               :effect (effect (gain-credits :corp eid 2))}]}))
 
 (defcard "Unity"
-  {:abilities [(break-sub 1 1 "Code Gate")
-               {:label "1 [Credits]: Add 1 strength for each installed icebreaker"
-                :async true
-                :effect (effect (continue-ability
-                                  (strength-pump
-                                    1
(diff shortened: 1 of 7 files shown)


NoahTheDuke:
This looks great. Can you add a test or two for unity or others to show it works correctly?

francescopellegrini:
Does this fix #5642 as well?

cardboardtech:
I think some additional work would need to be done to fix Utae, though I think getting the "Match Strength" ability active even if you don't have an active break ability wouldn't be too difficult. I'll poke around at it.

PR #6007, opened 2021-07-29
Don't close discard/deck popup when clicking in the popup

### LINKED ISSUE #6001: Labor Rights prompt minimises on select and makes it impossible to select the correct cards
When you play labor rights and select the first card it minimises the archives window, if you click archives to open it back up it selects the top card from archives as one of your targets, you then can't de-select that because doing so closes the prompt again

### LINKED ISSUE #6003: Selection of cards for Steve Cambridge trigger is hampered by autoclose of popup window
First card is selected, window closes, second card seems to be autoselected and corp gets to choose from wrong cards.

### LINKED ISSUE #6009: Spin Doctor Bug
Once you select the first cards to shuffle back in the archives window closes. If you click on archives to open it back up again it shuffles the top card of archives. If you click specifically on the bottom half of archives it opens back up archives without selecting the top card to be returned with spin.

### LINKED ISSUE #6011: Preemptive Action Issues
Hi, When using preemptive action, you are not able to openly select three cards.

When playing, the archives pop up appears as normal, but one the first card is selected, the pop up vanishes. Clicking archives reopens the box, but also selects the card 'on top' of archives. If you try to deselect this card, it closes archives and then you are trapped in a cycle, where opening the archives pop up selects that card. As such, you can only elect two of your choice plus the top card of archives and the pop up box needs to open and close multiple times.

I do not know if this same issue occurs with other multi-recursion effects, such as Genotyping.

cardboardtech:
Created a container around the discard and deck zones so we can have the toggle click event for the popups only occur when clicking on the main section. Clicking with in the popup will not close it unless you click the "Close" link. 

Also disabled the ability to "select" the top card of the discard during selections. You have to select from the popup. I think the way I did this was kinda awkward so if you think of a better way of handling this then please let me know. Another solution would be to just disable pointer events for the card through CSS. This is a cleaner solution as it still allows you to drop cards into the discard from that zone, but you can't drag the top card out of the discard

Another minor fix is that I adjusted the margins of the cards that appear in the popup so the backing text isn't poking out ever so slightly from the top left corner. 

fixes #6001
fixes #6003 
fixes #6009
fixes #6011

### DIFF
--- a/src/clj/game/core.clj
+++ b/src/clj/game/core.clj
@@ -272,7 +272,8 @@
 
   [game.core.diffs
    public-states
-   public-diffs]
+   public-diffs
+   finalize-state]
 
   [game.core.drawing
    draw
--- a/src/clj/game/core.clj
+++ b/src/clj/game/core.clj
@@ -272,8 +272,7 @@
 
   [game.core.diffs
    public-states
-   public-diffs
-   finalize-state]
+   public-diffs]
 
   [game.core.drawing
    draw
--- a/src/clj/game/core/diffs.clj
+++ b/src/clj/game/core/diffs.clj
@@ -90,6 +90,7 @@
    :choices
    :card
    :prompt-type
+   :show-discard
    ;; traces
    :player
    :base
@@ -131,7 +132,6 @@
    :quote
    :register
    :prompt-state
-   :show-discard
    :agenda-point
    :agenda-point-req])
 
@@ -146,7 +146,6 @@
       (update :scored card-summary-vec state side)
       (update :register select-keys [:spent-click])
       (prompt-summary same-side?)
-      (update :show-discard #(when same-side? %))
       (prune-null-fields)))
 
 (defn corp-keys []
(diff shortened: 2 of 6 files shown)


cardboardtech:
With this last commit, I think I have everything working. The discard view opens when an ability with `:show-discard: true` is active and closes when that ability completes. I did this by adding `:show-discard` key to a player. This value is updated right before we send the state to the clients. The value is set to `:show` if the active prompt has the `:show-discard` tag. If it does not or there is no prompt, we update the value in stages (:show -> :close -> nil). 

I tried to do this elsewhere such as when we updated the current prompt in `game.core.prompt-state` or by just trying to determine the correct value when gathering the diffs to send to the client in `game.core.diffs` but ran into problems with the `:close` getting wiped out before actually being sent to the client or not persisting in the state long enough to determine if we need to close the discard view or not. 

I'm happy for feedback on if there is a better way to handle this. I'm also not sure if the `finalize-state` function that handles this should live in `web.game` where it currently is or perhaps a different namespace like `game.core.state` or `game.core.diffs`

cardboardtech:
After some more poking about, I managed to find a purely client-side fix to closing the discard view. Most of my previous comment can be ignored haha. So now things are back to having `:show-discard` only exist on prompts and not on players. Instead of updating the state and decrementing the `:show-discard` key in stages, the client simply keeps track if a card ability opened the discard view. Once it gets a new prompt that doesn't have `:show-discard = true` then it closes the discard view.

PR #6014, opened 2021-08-05
Implement point system for Eternal

cardboardtech:
This implements the point system outlined [here](https://nisei.net/blog/a-new-beginning-for-eternal/) for the Eternal format. This fix assumes the changes made in [this PR](https://github.com/NoahTheDuke/netrunner-data/pull/20) are implemented in the netrunner-data repository since it assumes that format legality is represented in a map instead of just a value.

For example, Jinteki: Potential Unleashed's legality looks like this:
```clojure
:format{:standard {:banned true}, 
        :system-gateway {:rotated true}, 
        :startup {:rotated true}, 
        :eternal {:legal true, :points 3}, 
        :snapshot {:legal true, :restricted true}, 
        :snapshot-plus {:legal true, :restricted true}, 
        :classic {:banned true}}
```
This let's us store a point value for a format as well as mark cards that are playable with restrictions as legal cards. This let's us not need to check several different keywords when the card is essentially legal. 
 
I'm representing deck points in the UI as a green "❖" (U+2756) symbol. Absolutely open to suggestions on this one. 

Though cards can have different point values I didn't want to tack multiple on like influence dots as it could get very cluttered. Instead the point value will display if you hover over the symbol for a second or two. In addition, I list the current deck points and limit in the deck header. If you mouse over the line a tool tip will display giving you a breakdown.
image

If your deck is invalid due to deck points, the "Why?" span will tell you on mouse over.
image

Legality of cards viewed in the card browser will mark the card with the deck points symbol and will display the value on mouse over.
image

As I was updating the code to use the new legality structure, I ended up fixing some miscellaneous style issues. This includes changing the color of the ID text if the ID is banned or rotated from the format in the same way we color those cards in the main deck building section, keeping the pointer look when mousing over card text that isn't the standard link color, and sorting the format status legality for a deck in the order we typically display formats around jnet.
(basic deck building was in the middle of this list before) 
image

### DIFF
--- a/src/cljc/jinteki/cards.cljc
+++ b/src/cljc/jinteki/cards.cljc
@@ -6,7 +6,8 @@
 (defonce all-cards #?(:clj (atom {})
                       :cljs (r/atom {})))
 
-(defonce mwl (atom []))
+(defonce mwl #?(:clj (atom {})
+                :cljs (r/atom {})))
 
 (defonce sets (atom []))
 
--- a/src/cljs/nr/translations.cljs
+++ b/src/cljs/nr/translations.cljs
@@ -970,6 +970,7 @@
      :max "aximummay"
      :influence "Influenceyay"
      :agenda-points "Agendayay ointspay"
+     :deck-points "Eckday ointspay"
      :hash "Ournamenttay ashhay"
      :why "Whyay?"
      :legal "egallay"
--- a/src/cljs/nr/translations.cljs
+++ b/src/cljs/nr/translations.cljs
@@ -160,6 +160,7 @@
      :max "maximum"
      :influence "Influence"
      :agenda-points "Agenda points"
+     :deck-points "Deck points"
      :hash "Tournament hash"
      :why "Why?"
      :legal "legal"
--- a/src/cljs/nr/utils.cljs
+++ b/src/cljs/nr/utils.cljs
@@ -13,6 +13,7 @@
 (def restricted-dot (str "🦄" zws)) ; on the restricted list
 (def alliance-dot (str "○" zws))    ; alliance free-inf dot
 (def rotated-dot (str "↻" zws))     ; on the rotation list
+(def deck-points-dot (str "❖" zws)) ; costs deck points
 
 (def banned-span
   [:span.invalid {:title "Removed"} " " banned-dot])
@@ -22,6 +23,11 @@
 
 (def rotated-span
   [:span.casual {:title "Rotated"} " " rotated-dot])
+
+(defn deck-points-card-span [points]
+  [:span.legal {:title (when points
+                         (str "Deck points: " points))}
+   " " deck-points-dot])
 
 (defn- make-dots
   "Returns string of specified dots and number. Uses number for n > 20"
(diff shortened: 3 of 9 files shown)


src/cljs/nr/translations.cljs
```diff
@@ -160,6 +160,7 @@
      :max "maximum"
      :influence "Influence"
      :agenda-points "Agenda points"
+     :deck-points "Deck points"
```
NoahTheDuke, line 163: Make sure to add an entry to the pig latin translation file so you can verify it works.

src/cljs/nr/utils.cljs
```diff
@@ -23,6 +24,11 @@
 (def rotated-span
   [:span.casual {:title "Rotated"} " " rotated-dot])
 
+(defn deck-points-card-span [points]
+  [:span.legal {:title (when points
```
NoahTheDuke, line 28: Should you make the span when there are no points to display?

src/css/base.styl
```diff
@@ -297,6 +289,14 @@ button.off
 button .fake-link
     color: white-solid
 
+.casual
```
NoahTheDuke, line 292: Any reason to move these away from the other legality-related colors?

src/cljc/jinteki/cards.cljc
```diff
@@ -6,7 +6,8 @@
 (defonce all-cards #?(:clj (atom {})
                       :cljs (r/atom {})))
 
-(defonce mwl (atom []))
+(defonce mwl #?(:clj (atom {})
```
NoahTheDuke, line 9: This is a good change.

src/cljs/nr/utils.cljs
cardboardtech: I think it is useful since it allows you to display the symbol without a value when you hover over it. I use this for example in the header `Deck points: 4/7 ❖`. Since that symbol is just visual on that line I don't give it a value so nothing displays when you hover over it. Could be worth just displaying the point limit even though it is listed next to it. I feel like there was one other place I do this, but I can't remember right now.

src/css/base.styl
cardboardtech: Yeah, its due to the way CSS chooses what property to use when classes are equal. For background, for all of our card lines in the deck builder we use the `.fake-link` tag to color them orange and make the mouse cursor turn into a pointer when you hover over it. When a card in the deck builder is rotated or banned we change the span's class to `.inactive` or `.casual` to change the color, but in the process would lose the pointer on hover. I wanted to keep the pointer aspect of `.fake-link` while still changing the color, but due to the CSS priority of the two classes being the same it would prioritize the color of `.fake-link` because it was defined later in the CSS file. That was a long way of saying that I moved those below `.fake-link` so I can use the colors from it.

This was not the cleanest change. I should revisit this to at least organize it better. Maybe move these to their own file or move all the legality colors to the same spot. Maybe even better, just remove the color aspect from `.fake-link` and default the color to the orange elsewhere.

NoahTheDuke: Hmm yeah, that's definitely awkward. css sucks, lol. I'd say move all of the legality colors in that case, better to keep everything grouped.

PR #6037, opened 2021-09-06
subliminal recur should reveal

### LINKED ISSUE #5744: Subliminal Messaging should trigger Hyoubu
[Subliminal](https://netrunnerdb.com/en/card/31082):
> When your turn begins, if this card is in Archives and the Runner did not initiate any runs during their last turn, you may reveal this card and add it to HQ.

[Hyoubu](https://netrunnerdb.com/en/card/26039):
> The first time each turn you reveal a card, gain 1 credit.

kjchiu:
fixes #5744

### DIFF
--- a/src/clj/game/cards/operations.clj
+++ b/src/clj/game/cards/operations.clj
@@ -2306,7 +2306,9 @@
               :prompt "Add Subliminal Messaging to HQ?"
               :yes-ability
               {:msg "add Subliminal Messaging to HQ"
-               :effect (effect (move card :hand))}}}]})
+               :effect (effect (reveal eid card)
+                               (move card :hand))}}}]})
+
 
 (defcard "Success"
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/operations.clj
```diff
@@ -2306,7 +2306,9 @@
               :prompt "Add Subliminal Messaging to HQ?"
               :yes-ability
               {:msg "add Subliminal Messaging to HQ"
-               :effect (effect (move card :hand))}}}]})
+               :effect (effect (reveal eid card)
```
NoahTheDuke, line 2309: as this is an `async` function, you gotta either call it last or write something like `:effect (req (wait-for (reveal state side (make-eid state eid) card)) (move state side card :hand) (effect-completed state side eid)))`, and of course you gotta add `:async true` to the ability map.

PR #6038, opened 2021-09-07
CR1.5 Run Timing

### LINKED ISSUE #5555: options for change to UI on passing ice
at the moment, if the first ice is not rezzed, then the first prompt for a jackout window is after you pass that ice
this creates an ambiguous situation for the corp, since they might assume you wish to continue and rez the second ice, when actually you want to jack out
eg you are farming counters
or you want to provoke a rez on the first ice but then not contiune
so as runner to avoid this, you have to 'continue' then hit 'jack-out' really fast before corp reacts to the pass ice
could we have a 'continue then jack-out' option on the first ice if there's a no-rez
there is obviously a 'farm wheel counters' option but that's universally relevant. and the runner exposing the inner ice because you reacted too slow is a big issue.

### LINKED ISSUE #5911: Omar ability prevents Turning Wheel additional accesses
Link to bug replay: https://jinteki.net/bug-report/f9f5fa81-ebef-497d-8361-eae1ea6bfa44?b=0

Description:

When the runner uses Omar's ID ability to run archives and switch to HQ or R&D, spending tokens on Turning Wheel does not cause additional cards to be accessed.

cardboardtech:
This updates our run structure to be in-line with the current CR1.5 rules for runs and breaching servers.

The big thing on the front end is that there is an extra stop in between ice for the Movement phase. Options to jack out were removed from the Approach Ice and Encounter Ice phase since the Runner can only choose to jack out during the Movement phase. Due to this several abilities that explicitly let the runner jack out needed to be changed to give a jack out prompt. Auto-pass for the Corp will continue all phases except when approaching an unrezzed ice (as before) and on the last Movement phase before breaching a server.

Accessing cards has been reworked. The structure is mostly the same, but instead of determining the total number of cards to access and then decrementing that value after each access, we instead check for any valid candidates and continue accessing until there are no longer any candidates. This allows for infinite looping Ganked! and Drafter/Ansel. 

Terminology in the code has been updated for breaching a server.
`:pre-access` has been renamed to `:breach-server`.

Certain cards also needed to be updated to better fit the terminology. For example, cards like R&D Interface and HQ Interface no longer provide a global access bonus and instead trigger on `:breach-server` like Docklands Pass. This prevents them from providing bonus accesses outside of a breach (Counter Surveillance). 

I also did some more significant rewrites for:
- Counter Surveillance
- Rip Deal
- Dedicated Neural Net (corp chooses accesses from HQ)

Issues:
resolves #5555
resolves #5911


src/clj/game/core/access.clj
```diff
@@ -1066,84 +1135,56 @@
...
 
-(defn access-bonus-count
-  [state side s]
-  (sum-effects state side nil :access-bonus [s]))
-
 (defn access-count
   [state side kw]
-  (let [run (:run @state)
-        s (case kw
+  (let [s (case kw
```
NoahTheDuke, line 1200: I think this is unnecessary because R&D Interface et al now say `:rd` instead of `:rd-access`.

src/clj/game/core/access.clj
```diff
@@ -1160,20 +1201,31 @@
...
                  (swap! state assoc-in [:run :did-access] true))
-               (when-not (zero? (:total access-amount))
-                 (swap! state assoc-in [:runner :register :accessed-cards] true))
                (wait-for (resolve-ability state side (choose-access access-amount server (assoc args :server server)) nil nil)
-                         (wait-for (trigger-event-sync state side :end-access-phase {:from-server (first server)})
+                         (wait-for (trigger-event-sync state side :end-breach-server {:from-server (first server)})
                                    (unregister-floating-effects state side :end-of-access)
                                    (unregister-floating-events state side :end-of-access)
-                                   (effect-completed state side eid)))))))
+                                   (effect-completed state side eid)))))))
```
NoahTheDuke, line 1291: i'd prefer to not change the end of file characters, it leads to noisy diffs with no functional changes

test/clj/game/cards/agendas_test.clj
```diff
@@ -745,8 +745,7 @@
   ;; Dedicated Neural Net
   (testing "Corp chooses card to access. Issue #4874"
     (do-game
-      (new-game {:corp {:deck [(qty "Hedge Fund" 5)]
```
NoahTheDuke, line 748: One of the reasons I include these is to keep the game from being considered "won" during the course of normal games. I know we `clear-win` during set up, but I generally think it's better to not have to worry about that.

test/clj/game/core/access_test.clj
```diff
@@ -8,114 +8,133 @@
...
     (do-game
-      (new-game {:corp {:deck ["Hedge Fund"]
-                        :hand [(qty "Hedge Fund" 4)]}})
-      (click-draw state :corp)
-      (take-credits state :corp)
-      (run-empty-server state "R&D")
-      (is (nil? (get-run)))
-      (is (empty? (:prompt (get-runner))) "Runner has no access prompt")
-      (is (empty? (:prompt (get-corp))))))
+     (new-game {:corp {:deck ["Hedge Fund"]
```
NoahTheDuke, line 11: looks like indentation got messed up

PR #6044, opened 2021-09-14
fix special MU display

### LINKED ISSUE #5964: Special MU doesn't show MU icon
This bug was caused by #5919, sorry.. :(

Moreover, special MU should abide to the settings introduced by #5796.

image

francescopellegrini:
Fixes #5964

### DIFF
--- a/src/cljs/nr/gameboard/player-stats.cljs
+++ b/src/cljs/nr/gameboard/player-stats.cljs
@@ -45,20 +45,18 @@
 (defn- display-special-memory
   ([memory] (display-special-memory memory false))
   ([memory icon?]
-  (when-let [only-for (->> (:only-for memory)
-                           (filter #(pos? (:available (second %))))
-                           (into {})
-                           not-empty)]
-    [:div
-     [:<> "("
-          (join "), (" (for [[mu-type {:keys [available used]}] only-for
-                             :let [
-                               unused (max 0 (- available used))
-                               mu-type-name (capitalize (name mu-type))
-                              ]]
-                         (if icon? [:<> unused "/" available " " mu-type-name " " [:span.anr-icon.mu]]
-                             (tr [:game.special-mu-count] unused available mu-type-name))))
-          ")"]])))
+   (when-let [only-for (->> (:only-for memory)
+                            (filter #(pos? (:available (second %))))
+                            (into {})
+                            not-empty)]
+     [:<>
+      (doall
+       (for [[mu-type {:keys [available used]}] only-for
+             :let [unused (max 0 (- available used))
+                   mu-type-name (capitalize (name mu-type))]]
+         ^{:key mu-type-name}
+         [:div (if icon? [:<> unused "/" available " " mu-type-name " " [:span.anr-icon.mu]]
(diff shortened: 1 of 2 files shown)


francescopellegrini:
Note that I've **not** tested this visually...

src/cljs/nr/gameboard/player-stats.cljs
```diff
@@ -43,18 +43,21 @@
...
-                         (str unused " of " available
-                              " " (capitalize (name mu-type))
-                              " [mu] unused")))
+                             :let [
+                               unused (max 0 (- available used))
+                               mu-type-name (capitalize (name mu-type))
+                              ]]
+                         (if icon? [:<> unused "/" available " " mu-type-name " " [:span.anr-icon.mu]]
+                             (tr [:game.special-mu-count] unused available mu-type-name))))
           ")")]))
```
cardboardtech, line 61: You're missing a parenthesis at the end of `display-special-memory` since you added the second signature. Won't build without it.

This needs some more work since it doesn't display properly after it builds. I think the issue is that this is wrapped entirely in a `(str)` function.   
image

NoahTheDuke:
> Note that I've **not** tested this visually... 

Haha I would suggest doing that.

francescopellegrini:
> > Note that I've **not** tested this visually...
> 
> Haha I would suggest doing that.

I'm having a hard time creating a containerized environment with all the prerequisites.. :/

src/cljs/nr/gameboard/player-stats.cljs
```diff
@@ -43,19 +43,22 @@
...
-                              " " (capitalize (name mu-type))
-                              " [mu] unused")))
-          ")")]))
+                             :let [
+                               unused (max 0 (- available used))
+                               mu-type-name (capitalize (name mu-type))
+                              ]]
+                         (if icon? [:<> unused "/" available " " mu-type-name " " [:span.anr-icon.mu]]
+                             (tr [:game.special-mu-count] unused available mu-type-name))))
+          ")"]])))
```
cardboardtech, line 61: Looks like since we are needing to use span elements to display the icon, we can't use any string functions like join like we used to. Instead we'll have to loop through the mu-types and generate an element for each one. I messed around with it for a bit and managed to get it to display using the code below.
image

```suggestion
(defn- display-special-memory
  ([memory] (display-special-memory memory false))
  ([memory icon?]
   (when-let [only-for (->> (:only-for memory)
                            (filter #(pos? (:available (second %))))
                            (into {})
                            not-empty)]
     [:<>
      (doall
       (for [[mu-type {:keys [available used]}] only-for
             :let [unused (max 0 (- available used))
                   mu-type-name (capitalize (name mu-type))]]
         ^{:key mu-type-name}
         [:div (if icon? [:<> unused "/" available " " mu-type-name " " [:span.anr-icon.mu]]
                   (tr [:game.special-mu-count] unused available mu-type-name))]))])))
```

francescopellegrini: Thank you so much @cardboardtech !

cardboardtech: happy to help

PR #6050, opened 2021-09-17
Make run abilities persist after trash

### LINKED ISSUE #5754: Stargate does not trigger when trashed during a run
When Stargate is used to make a run and is then trashed during that run, it fails to trigger when the run is successful (instead I get an ordinary R&D access). I believe this behaviour is incorrect.

### LINKED ISSUE #6049: Stargate trashed by Balista mid-run
Link to bug replay: https://jinteki.net/bug-report/6a2d86ea-c963-4df0-b3be-1bba4192b765?b=0

Description:

I used Stargate to run R&D, which was protected by Ballista. Corp trashed Stargate then I breached, but I accessed cards like normal instead of draw 3/trash 1. Somebody in Discord said this wasn't working correctly.
This was a demo game where I was controlling both sides.

cardboardtech:
Updated Stargate, Keyhole, Analog Dreamers, and Expert Schedule Analyzer to have their effects persist even if they are trashed during the course of the run. Modeled the fix off of how we implemented Counter Surveillance.

fix #5754
fix #6049

### DIFF
--- a/src/clj/game/cards/programs.clj
+++ b/src/clj/game/cards/programs.clj
@@ -371,22 +371,22 @@
                               :effect (effect (lose-credits :corp eid 1))}]}))
 
 (defcard "Analog Dreamers"
-  {:abilities [{:cost [:click 1]
-                :msg "make a run on R&D"
-                :makes-run true
-                :async true
-                :effect (effect (make-run eid :rd card))}]
-   :events [(successful-run-replace-breach
-              {:target-server :rd
-               :this-card-run true
-               :ability
-               {:prompt "Choose a card to shuffle into R&D"
-                :choices {:card #(and (not (ice? %))
-                                      (not (rezzed? %))
-                                      (zero? (get-counters % :advancement)))}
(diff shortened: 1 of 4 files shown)


src/clj/game/cards/programs.clj
```diff
@@ -1139,19 +1139,19 @@
...
+                  :ability
+                  {:msg (msg "reveal all of the cards cards in HQ: "
+                             (string/join ", " (map :title (:hand corp))))
+                   :async true
+                   :effect (effect (reveal eid (:hand corp)))}})]
+    {:abilities [{:cost [:click 1]
+                  :msg "make a run on HQ"
+                  :makes-run true
+                  :async true
+                  :effect (effect (register-events card [(assoc ability :duration :end-of-run)])
```
NoahTheDuke, line 1153: You can just put the duration directly on the ability. (Same with the other card abilities in this PR.) the ability maps aren't used elsewhere so you don't have to worry about the duration leaking into another usage.

NoahTheDuke (approved):

PR #6055, opened 2021-09-21
Checkpoint draws

### LINKED ISSUE #5974: Glitch with Political Dealings + Daily Business Show + Near-Earth Hub
Bug reproduced multiple times today, it seems solid.

Steps:

(1) while being NEH, install Political Dealings and one or more Daily Business Show.
(2) Draw an agenda at the start of your turn.
(3) Choose to install the agenda in a new remote server for the PD trigger. This draws you a card with NEH.
(4) You now get the Daily Business Show trigger for the draw at step 2. BUT, you are only allowed to discard the card you drew at step 3. If you had multiple DBSs, you are unable to select enough cards to satisfy the trigger.

Note that the game doesn't allow you to order PD and DBS triggers the way you want (which would be the correct behavior, imo): it always puts PD first.

My hunch is that DBS asks you to select one of the cards that you have 'just drawn', which doesn't work well if you have drawn other cards by other means before it resolves.

I would post a replay, but I don't think you can see the problem (which is mostly a UI problem) from the replay anyway.

thx :heart:

NoahTheDuke:
Change `draw` to use checkpoints instead of trigger-events. Updated relevant cards. Not a lot changed cuz we don't have a good way of doing `:set-aside` yet and adding a whole section will be annoying.

Closes #5974


src/clj/game/core/drawing.clj
```diff
@@ -27,45 +28,56 @@
...
-                     (effect-completed state side eid))
-                   (when (safe-zero? (remaining-draws state side))
-                     (prevent-draw state side))))
-               (when (< draws-after-prevent draws-wanted)
-                 (let [prevented (- draws-wanted draws-after-prevent)]
-                   (system-msg state (other-side side) (str "prevents "
-                                                            (quantify prevented "card")
-                                                            " from being drawn"))))))))
-
+   (swap! state update-in [side :register] dissoc :currently-drawing)
```
cardboardtech, line 45: Since we are clearing `:currently-drawing` at the start here, I believe issue https://github.com/mtgred/netrunner/issues/5974 is still going to be an issue. I'm wondering if the test you wrote for this scenario is giving a false positive since the only cards being drawn are Hedge Funds.

I'm thinking we should only clear `:currently-drawing` at the end of this function and only if the call to the function is the initial draw. If an additional draw occurs while a draw is resolving those additional cards should be added to the existing `:currently-drawn` set. That's at least how I interpret 8.4.3 

> Abilities that resolve while drawn cards are set aside can move those cards or add new cards to that set.

src/clj/game/core/moving.clj
```diff
@@ -448,6 +451,10 @@
                  (or (ice? a)
                      (ice? b)))
         (set-current-ice state))
+      (when (in-hand? a) (remove-from-currently-drawing state a-side a))
+      (when (in-hand? b) (remove-from-currently-drawing state b-side b))
+      (when (in-hand? moved-a) (add-to-currently-drawing state a-side moved-a))
+      (when (in-hand? moved-b) (add-to-currently-drawing state b-side moved-b))
```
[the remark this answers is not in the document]
NoahTheDuke: good point

src/clj/game/core/drawing.clj
NoahTheDuke: this is a mistake, i meant to leave it at the end.

test/clj/game/cards/assets_test.clj
```diff
@@ -1108,7 +1108,42 @@
...
+      (is (empty? (:prompt (get-corp))) "DBS doesn't trigger on mandatory draw")))
+  (testing "Interaction with NEH and Political Dealings and nested draws. Issue #5974"
+    (testing "DBS first"
+      (do-game
+        (new-game {:corp {:identity "Near-Earth Hub: Broadcast Center"
+                          :deck [(qty "Hedge Fund" 10)]
+                          :hand ["Daily Business Show" "Political Dealings" "Merger"]
+                          :credits 20}})
+        (core/move state :corp (find-card "Merger" (:hand (get-corp))) :deck {:front true})
+        (play-from-hand state :corp "Daily Business Show" "New remote")
```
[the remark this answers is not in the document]
NoahTheDuke: hmm maybe. i'll move the `move` call down to the runner's turn.

src/clj/game/core/drawing.clj
```diff
@@ -27,45 +28,56 @@
...
+       (let [to-draw (take draws-after-prevent (get-in @state [side :deck]))
+             drawn (mapv #(move state side % :hand) to-draw)
+             drawn-count (count drawn)]
+         (swap! state update-in [side :register :drawn-this-turn] (fnil #(+ % drawn-count) 0))
+         (swap! state update-in [:stats side :gain :card] (fnil + 0) n)
+         (swap! state update :bonus dissoc :draw)
+         (if suppress-event
+           (effect-completed state side eid)
+           (let [draw-event (if (= side :corp) :corp-draw :runner-draw)]
+             (swap! state assoc-in [side :register :currently-drawing] drawn)
```
cardboardtech, line 73: We should `conj` `drawn` to `:currently-drawing` here so we don't overwrite any pending draws.

src/clj/game/core/drawing.clj
```diff
@@ -27,45 +28,56 @@
...
+             (queue-event state draw-event {:cards drawn
+                                            :count drawn-count})
+             (wait-for
+               (checkpoint state nil (make-eid state eid) nil)
+               (wait-for (trigger-event-sync state side (make-eid state eid) (if (= side :corp) :post-corp-draw :post-runner-draw) drawn-count)
+                         (let [eid (make-result eid (-> @state side :register :currently-drawing))]
+                           (swap! state update-in [side :register] dissoc :currently-drawing)
+                           (effect-completed state side eid))))))
```
cardboardtech, line 81: I'm trying to think if there are any negative side-effects of potentially triggering `:post-side-draw` events twice. 

In the scenario where an NEH draw is triggered during DBS we'd fire post draw from the NEH draw. We would then clear `:currently-drawing` before returning to the initial call to `draw` where we'd fire post draw events again but this time with 0 cards based to the event. 

Currently we only have 2 cards that care about post draw events. Jinja City Grid clears its fake "once per turn" so that isn't a big deal to be called twice. There is also Find the Truth which reveals the cards drawn which could potentially trigger on reveal events twice; however, I'm not sure the Runner can event create a scenario like the NEH/DBS/PD scenario. At the very least a `req` could be added to Find the Truth to not fire if there aren't any cards in `:currently-drawn`.

src/clj/game/core/drawing.clj
NoahTheDuke: I don't think its's possible to draw during the resolution of a draw.

src/clj/game/core/drawing.clj
```diff
@@ -70,14 +70,14 @@
          (if suppress-event
            (effect-completed state side eid)
            (let [draw-event (if (= side :corp) :corp-draw :runner-draw)]
-             (swap! state assoc-in [side :register :currently-drawing] drawn)
+             (swap! state update-in [side :register :currently-drawing] conj drawn)
```
cardboardtech, line 73: I don't think this is going to give us the list of lists that we want. Calling `conj` on two list will merge the lists together into one. When we first populate `:currently-drawing` we need to wrap the list in another list. I think that's why the test is failing in the latest build.

I guess it needs to look something like this
```
(conj [[1 2] [3 4]] [5 6])       
;;=> [[1 2] [3 4] [5 6]]
```

NoahTheDuke: If `:currently-drawing` is `nil`, `conj` will wrap the given object in a list. If it's a list, `conj` will append the object.

cardboardtech (approved):

PR #6071, opened 2021-09-29
Playable Abilities Fixes

cardboardtech:
Identities with active abilities (Earth Station, Sync, etc.) need to check that their abilities can be used.

### DIFF
--- a/src/clj/game/core/diffs.clj
+++ b/src/clj/game/core/diffs.clj
@@ -49,7 +49,6 @@
              :source-type :ability
              :source-info {:ability-idx ability-idx}}]
     (if (and (or (active? card)
-                 (condition-counter? card)
                  (:autoresolve ability))
              (can-pay? state side eid card nil cost)
              (can-trigger? state side eid ability card nil))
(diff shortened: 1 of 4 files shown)


cardboardtech:
Found a fix for hosted cards, condition counters with active abilities, and corp abilities on runner cards (PAD Tap).

Hopefully this is the rest of the issues. The only other thing I saw was that break sub abilities on ice (Bioroid, F2P) are always showing as active even if they should be disabled. That's a better problem than always being disabled.

src/clj/game/core/diffs.clj
```diff
@@ -49,6 +49,7 @@
              :source-type :ability
              :source-info {:ability-idx ability-idx}}]
     (if (and (or (active? card)
+                 (condition-counter? card)
```
NoahTheDuke, line 52: should this check be in `active?`? Seems like it. Might not be worth fixing in this PR, tho.

src/clj/game/core/diffs.clj
```diff
@@ -199,11 +203,11 @@
    :brain-damage])
 
 (defn rig-summary
-  [state]
+  [state side]
```
NoahTheDuke, line 206: Good catch!

src/clj/game/core/diffs.clj
```diff
@@ -71,7 +72,10 @@
     [(not (is-public? card side))
      (prune-null-fields (private-card card))]
     [(:hosted card)
-     (update card :hosted (partial mapv #(card-summary % state side)))]
+     (-> card
```
NoahTheDuke, line 75: Instead of always updating `:hosted`, maybe something like:
```clojure
    (-> (if (:hosted card) (update card :hosted (partial mapv #(card-summary % state side))) card)
        (playable? state side)
        (card-abilities-playable? state side)
        (prune-null-fields))))
```

Issue #6075, opened 2021-09-30
v96 Earth Station ID does not work


pwoloszun:
Link to bug replay: https://jinteki.net/bug-report/784a090e-c1f3-414e-a2ae-8ea1a1c422b1?b=0

Description:

Cant use Earth Station ID ability

NoahTheDuke:
Thanks for the report!

Issue #6074, opened 2021-09-30
v96 bug: Regolith Mining License does not work


pwoloszun:
Regolith Mining License (https://netrunnerdb.com/en/card/30071) does not work properly.

When I try to use it to get LAST 3c from it nothing happens.

NoahTheDuke:
Thanks for the report!

Issue #6073, opened 2021-09-30
v96 bug: Paule's Cafe does not work


pwoloszun:
Paule's Cafe (https://netrunnerdb.com/en/card/26092) does not work.

After I host 1st card (either program or hardware) on it, its submenu is disabled all the time.
From then I can't neither install hosted card, nor host another card on it.

Issue #6068, opened 2021-09-29
Trashing nico campaign results in internal server error.


rubenpieters:
Link to bug replay: https://jinteki.net/bug-report/bebd0ab5-2ad9-46c8-876e-6174054b4e6d?b=0

Description:

Trashing nico campaign results in internal server error.

PR #6077, opened 2021-09-30
Rewrite diff generation

NoahTheDuke:
there's a bug where as runner you see the "corp" side, so idk what i'm doing wrong


cardboardtech:
These are all issues that came up from the last release that we have fixed, but since you are reworking diffs I figure we shouldn't close them until you finish this in case they crop back up.

https://github.com/mtgred/netrunner/issues/6068
https://github.com/mtgred/netrunner/issues/6073
https://github.com/mtgred/netrunner/issues/6074
https://github.com/mtgred/netrunner/issues/6075

Issue #6124, opened 2021-11-03
Hidden information is available to spectators through browser inspection


NBKelly:
Spectators may view hidden information on any face-down installed cards through the inspector on their browser - the card-name field is filled.

I have checked that a player is not able to to view this information, and as far as I can tell:
* A player can't see this information while in game
* Leaving and reconnecting does not allow the player to see this information
* Leaving a joining as a spectator does allow the player to see this information
* Any spectator is capable of seeing this information

Though I don't imagine it's likely, this does provide an avenue for for players to cheat through the use of a third party or sock-puppet account.

Steps to reproduce:
Join any game as a spectator, open up your inspector, and select a face-down card.

![inspect_01](https://user-images.githubusercontent.com/9095245/140008979-26a8f314-dba1-4688-aa16-44ca1754b229.jpg)

![inspect_02](https://user-images.githubusercontent.com/9095245/140008989-90a6b959-a081-4ed1-98b2-27f3908eabad.jpg)

NoahTheDuke:
Excellent work, detective! Thanks so much, we'll get this cleaned up shortly.

Issue #6120, opened 2021-11-02
On the lam is installed face down


tolaasin:

NoahTheDuke:
Thanks for the report!

Issue #6117, opened 2021-11-01
Fake points appear face down in score areas


kalotay:
Fake points (e.g. Mad Dash, Gene Splicer) appear facedown in score areas.
<img width="135" alt="Screenshot 2021-11-01 at 18 01 47" src="https://user-images.githubusercontent.com/661871/139715373-ac843fa3-ff78-48ce-89f3-e5fa004bf988.png">
.

mikeprosser:
Aah whoops, I raised a duplicate of this here https://github.com/mtgred/netrunner/issues/6119 which I've now closed. As I commented in there, Eavesdrop also seems to be installing face down on unrezzed ice - although that might be a different issue as the fake points still "work", Eavesdrop encounter doesn't trigger

NoahTheDuke:
Thanks for the report!

PR #6125, opened 2021-11-03
Diff Fixes

cardboardtech:
Working on fixing some diff issues. I'm going to keep adding to this PR, but feel free to merge things in whenever. 

fixes https://github.com/mtgred/netrunner/issues/6124 - Fix spectator diffs so they don't see hidden information (this also has the fixes for PR https://github.com/mtgred/netrunner/pull/6123, just without the message changes on the ID itself)
fixes https://github.com/mtgred/netrunner/issues/6120
fixes https://github.com/mtgred/netrunner/issues/6117

### DIFF
--- a/src/clj/game/core/installing.clj
+++ b/src/clj/game/core/installing.clj
@@ -12,7 +12,7 @@
     [game.core.flags :refer [turn-flag?]]
     [game.core.hosting :refer [host]]
     [game.core.ice :refer [update-breaker-strength]]
-    [game.core.initializing :refer [ability-init card-init]]
+    [game.core.initializing :refer [ability-init card-init corp-ability-init runner-ability-init]]
     [game.core.moving :refer [move trash]]
     [game.core.payment :refer [build-spend-msg can-pay? merge-costs]]
     [game.core.rezzing :refer [rez]]
@@ -435,13 +435,17 @@
   (assert (or (event? card) (operation? card)) "condition counter must be event or operation")
   (let [cdef (card-def card)
         abilities (ability-init cdef)
+        corp-abilities (corp-ability-init cdef)
+        runner-abilities (runner-ability-init cdef)
         card (convert-to-condition-counter card)
         events (filter #(= :hosted (:condition %)) (:events cdef))]
     (if (corp? card)
       (wait-for (corp-install state side (make-eid state eid)
                               card target {:host-card target
                                            :ignore-all-cost true})
-                (let [card (update! state side (assoc async-result :abilities abilities))]
+                (let [card (update! state side (assoc async-result
+                                                      :abilities abilities
+                                                      :runner-abilities runner-abilities))]
                   (unregister-events state side card)
                   (unregister-constant-effects state side card)
                   (register-events state side card events)
@@ -450,7 +454,9 @@
       (wait-for (runner-install state side (make-eid state eid)
                                 card {:host-card target
                                       :ignore-all-cost true})
-                (let [card (update! state side (assoc async-result :abilities abilities))]
+                (let [card (update! state side (assoc async-result
+                                                      :abilities abilities
+                                                      :corp-abilities corp-abilities))]
                   (unregister-events state side card)
                   (unregister-constant-effects state side card)
                   (register-events state side card events)
--- a/src/clj/game/core/moving.clj
+++ b/src/clj/game/core/moving.clj
@@ -3,7 +3,7 @@
     [clojure.string :as string]
     [game.core.agendas :refer [update-all-agenda-points]]
     [game.core.board :refer [all-active-installed]]
-    [game.core.card :refer [active? card-index convert-to-agenda corp? facedown? fake-identity? get-card get-title get-zone has-subtype? ice? in-hand? in-play-area? installed? resource? rezzed? runner?]]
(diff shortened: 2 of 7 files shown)


src/clj/game/core/moving.clj
```diff
@@ -76,9 +76,10 @@
         ;; Set :seen correctly
         c (if (= :corp side)
             (cond
-              ;; Moving rezzed card to discard, explicitly mark as seen
+              ;; Moving rezzed card or condition counter to discard, explicitly mark as seen
               (and (= :discard (first dest))
-                   (rezzed? card))
+                   (or (rezzed? card)
+                       (is-type? card "Counter")))
```
NoahTheDuke, line 82: We should have a `condition-counter?` in `game.core.card`.

src/cljc/game/core/card.cljc
```diff
@@ -437,10 +437,11 @@
 ;;    forfeit, the card is immediately trashed. See rule 8.2.5.
 
 (defn convert-to-agenda
-  [{:keys [cid host hosted side title zone]} n]
+  [{:keys [cid code host hosted side title zone]} n]
   (map->Card
     {:agendapoints n
      :cid cid
+     :code code
```
NoahTheDuke, line 444: What's this needed for?

src/cljc/game/core/card.cljc
```diff
@@ -456,11 +457,11 @@
 ;;    characteristics, and is trashed.
 
 (defn convert-to-condition-counter
-  [{:keys [cid side title zone]}]
+  [{:keys [cid code side title zone]}]
   (map->Card
     {:cid cid
+     :code code
      :printed-title title
-     :seen true
```
NoahTheDuke, line 463: This isn't needed? We have a check for displaying all card counters in the gameboard?

src/cljs/nr/gameboard/board.cljs
```diff
@@ -658,107 +658,109 @@
   "Returns true if the installed card should be drawn face down."
   [{:keys [side type facedown rezzed host] :as card}]
   (if (= side "Corp")
-    (and (not= type "Operation")
+    (and (and (not= type "Operation")
```
NoahTheDuke, line 661: nested `and` can be flattened to a single `and`.

src/clj/game/core/installing.clj
```diff
@@ -435,13 +435,17 @@
   (assert (or (event? card) (operation? card)) "condition counter must be event or operation")
   (let [cdef (card-def card)
         abilities (ability-init cdef)
+        corp-abilities (corp-ability-init cdef)
```
NoahTheDuke, line 438: this reminds me, `corp-abilities` and `runner-abilities` are dumb, we should just have a flag on abilities about which player can use them lmao

src/cljc/game/core/card.cljc
cardboardtech: This is needed to retrieve card images.

src/cljc/game/core/card.cljc
cardboardtech: The `:seen` property is removed whenever a corp card is installed. I had the option to either add a check during install to not remove the `:seen` or on moving to the discard. Since we already have logic to add `:seen` to rezzed cards being moved into the discard I chose to just expand that existing check to include condition counters.

src/cljs/nr/gameboard/board.cljs
cardboardtech: whoops, I had originally put this as an `or` before realizing that was wrong and just flipped it to an `and`.

src/clj/game/core/installing.clj
cardboardtech: Agreed, these `corp-abilities` and `runner-abilities` being separate has bit us quite a bit with some of the work we've done lately.

NoahTheDuke: I'll tackle it at some point, no need for it to be done here, just thinking out loud lol.

src/cljs/nr/gameboard/board.cljs
```diff
@@ -3,7 +3,7 @@
   (:require [cljs.core.async :refer [chan put! <! timeout] :as async]
             [clojure.string :as s :refer [capitalize includes? join lower-case split blank? starts-with? ends-with?]]
             [differ.core :as differ]
-            [game.core.card :refer [has-subtype? asset? rezzed? ice? corp?
+            [game.core.card :refer [condition-counter? get-title has-subtype? asset? operation? rezzed? ice? corp?
```
NoahTheDuke, line 6: Double check that this actually works lol. The card namespace can be accessed in the clojurescript but because we convert the game state to json, certain properties change from keywords to strings and get hecked up.

cardboardtech: This works fine for everything I've tested. Types are stored as strings and not keywords so I believe it doesn't run into the json conversion issue.

cardboardtech:
I ran into an issue where if the Runner clicked on Mad Dash while it was in the score area it would transfer to the Corp. This was because the UI would give Mad Dash the "Score" action since it me the conditions for it (type = Agenda, advancement requirement of zero met). Since Score was the only Action/Ability on Mad Dash it would send the Score action to the server which didn't check if the Agenda was already scored/stolen thus giving it to the Corp.

NoahTheDuke:
Good catch!

Issue #6133, opened 2021-11-07
"If you made a Successful Run" effects not checking until after a run is complete.


Tamijo-Netrunner:
![Screenshot_20211107-094211_Discord](https://user-images.githubusercontent.com/39098450/140640074-a99277e2-cb69-4d22-977c-a5850f390fd5.jpg)

As per this screenshot; the runner made a successful run on HQ, then during the access phase installed a Penumbral Toolkit. This card cost them 2c instead of the 0c it should've cost.

cardboardtech:
I was unable to recreate this issue at least in the basic case of stealing an agenda from HQ and using Pantograph to install Penumbral Toolkit. Maybe there was more to the scenario? 

I did write a test for this in PR #6135 so we at least catch if this becomes an issue in the future.

PR #6135, opened 2021-11-09
Rework Pantograph

cardboardtech:
I was looking into Issue #6133 and was unable to reproduce the issue. The cost reduction applied when installing Penumbral Toolkit through a Pantograph trigger. I wrote a test to make sure this continues to work in the future. As I was looking at this, I realized that Pantograph's ability could be streamlined to reduce the number of inputs needed per trigger, so I made those changes.

Pantograph changes:
1. Removed optional prompt for gaining a credit since gaining a credit is not optional 
2. Removed optional wrapper from the install ability as the select prompt itself is optional (clicking "Done")
3. Added message in the cancel function to indicate that the Runner declined to install a card

This reduced the number of inputs needed for Pantograph to just 1 per trigger.

resolves https://github.com/mtgred/netrunner/issues/6133

### DIFF
--- a/src/clj/game/cards/hardware.clj
+++ b/src/clj/game/cards/hardware.clj
@@ -1283,37 +1283,26 @@
 
 (defcard "Pantograph"
   (let [install-ability
-        {:optional
-         {:waiting-prompt "Runner to make a decision"
-          :prompt "Install card with Pantograph ability?"
-          :player :runner
-          :yes-ability
-          {:async true
-           :prompt "Choose a card to install"
-           :choices
-           {:req (req (and (runner? target)
-                           (in-hand? target)
-                           (not (event? target))
-                           (can-pay? state side (assoc eid :source card :source-type :runner-install)
-                                     target nil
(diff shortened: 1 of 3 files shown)


src/clj/game/cards/hardware.clj
```diff
@@ -1283,37 +1283,26 @@
...
-                                  state side
-                                  {:optional
-                                   {:prompt "Gain 1 [Credits] with Pantograph ability?"
-                                    :yes-ability
-                                    {:async true
-                                     :msg "gain 1 [Credits]"
-                                     :effect (req (gain-credits state :runner eid 1))}}}
-                                  card nil)
+         :msg "gain 1 [Credits]"
+         :effect (req (wait-for (gain-credits state :runner 1)
```
NoahTheDuke, line 1305: What's the reason for removing the optional here? The ability isn't mandatory.

NoahTheDuke: jk, it is, lmao. ignore me!

NoahTheDuke (approved):
i like this

Issue #6143, opened 2021-11-12
Runner not moving forward after Trypano trashes current ICE


Joseph7695:
1. Trypano trashes inner ICE.
2. BUG: Encountering the outer ICE

Expected: 
2. Breaching server.

jtfq99999 spends  to make a run on Server 2.
jtfq99999 approaches Anansi protecting Server 2 at position 1.
jtfq99999 encounters Anansi protecting Server 2 at position 1.
jtfq99999 pays 3  to use Aumakua to break all 3 subroutines on Anansi.
jtfq99999 has no further action.
jtfq99999 passes Anansi protecting Server 2 at position 1.
jtfq99999 trashes Simulchip to use Simulchip to install a program.
jtfq99999 pays 0  to install Trypano on Kakugo protecting Server 2 at position 0.
jtfq99999 uses Cookbook to place 1 virus counter on Trypano hosted on Kakugo protecting Server 2 at position 0.
jtfq99999 uses Trypano to trash Kakugo protecting Server 2 at position 0.
jtfq99999 will continue the run.
jtfq99999 approaches Anansi protecting Server 2 at position 0.
jtfq99999 encounters Anansi protecting Server 2 at position 0.
jtfq99999 has no further action.
Chillsection uses Anansi to do 3 net damage.
jtfq99999 trashes due to net damage.
jtfq99999 is flatlined.

Chillsection:
*bump

confirming this just happened in our game

PR #6149, opened 2021-11-16
Handle Updating Run Position When Ice Is Added/Removed

cardboardtech:
Added some code in `moving.clj` to handle changing the position of a run when a piece of ice is added/removed from an inward position. Before any card ability that could add/remove pieces of ice would have to manually update position and an ice being trashed/removed would be handled by the Approach/Encounter phase going into the Movement phase. The issue was if a piece of ice was trashed/added during the Movement phase it would cause problems where the runner would re-approach passed ice. Now the only abilities that need to handle moving the Runner's position are cards like Mti or Formicary which explicitly make the Runner move back to encounter newly added ice on Approach when the Runner's position normally cannot be moved.

fixes https://github.com/mtgred/netrunner/issues/6143

To Do:
Update some UI logic as this change can cause some run controls to appear incorrectly.


src/clj/game/core/diffs.clj
```diff
@@ -352,7 +353,9 @@
 (defn run-summary
   [state]
   (when-let [run (:run @state)]
-    (select-non-nil-keys run run-keys)))
+    (-> run
+        (update :current-ice #(some? (get-card state %)))
```
NoahTheDuke, line 357: I think we should either have `:current-ice` actually hold the current ice, or we should name it something else. Better to keep consistency between front and back end. It's extra work to remember that `:current-ice` is a card object in the engine but a boolean in the cljs.

src/clj/game/cards/agendas.clj
```diff
@@ -1722,12 +1722,6 @@
                                                              (wait-for (corp-install
                                                                          state side chosen-ice chosen-server
                                                                          {:ignore-all-cost true :index target})
-                                                                       (when (and run
```
NoahTheDuke, line 1725: Any time you remove lines like this and you're left with `(wait-for (X) (effect-completed))`, you can condense it to a single `(X state side eid ...)` call.

src/clj/game/core/diffs.clj
cardboardtech: Yeah, I wasn't sure what the best way to handle it is. I really just need to know if the approached ice is still there. I think the most intuitive thing is to just put `:current-ice` through `card-summary` but it's more than I really need and we have to make sure hidden information is stripped so the Runner can't see what an unrezzed approached ice is. 

I think our options are:
1. Rename this to like `:current-ice-exists?` (or some better name I cannot think of)
2. Move the `run-summary` out of `strip-state` and into `state-summary` where we can pass it the side and properly strip the card

I'm leaning towards 2, but I feel like moving `run-summary` to `state-summary` goes against some of the work you did in refactoring this file.

NoahTheDuke: I think I'd prefer 1 because there's no reason to include the `:current-ice` when it can be/is derived by the client. I think I added `:current-ice` in the engine to handle when an ice is trashed but we still need to refer to it, but it's an ugly hack. Better to not send anything at all and rely on existing stripping to know if a given card is visible or not.

PR #6158, opened 2021-11-24
Lobby rewrite

NoahTheDuke:
Might be the dumbest rewrite I've embarked on, but the sente handlers and front-end code are all a goddamn mess and I cannot handle working with it anymore, lmao.


src/clj/web/admin.clj
```diff
@@ -5,7 +5,7 @@
    [monger.operators :refer :all]
    [monger.result :refer [acknowledged? updated-existing?]]
    [web.config :refer [frontend-version]]
-   [web.lobby :refer [all-games]]
+   ; [web.lobby :refer [all-games]]
```
cardboardtech, line 8: any reason to keep this commented line?

awildturtok (commented):
Sorry for the ping, but i'm looking at this code as there's suspicion this PR might be responsible for spikes as of late.

awildturtok:
My lisp is not strong enough to warrant any merit, so I've mostly looked at where you are allocating long lived objects. Your handling of `app-state users` was the only thing I noticed, though I've no clue how big that object will eventually get.

I've commented here: https://github.com/mtgred/netrunner/pull/6158#discussion_r1258358850

I dunno how sensible it is, but if you deregister all user that haven't pinged in the last 2 ping epochs that might clear up some space. (I'm still not sure if `app-state users` is all users, active user ids, or data of active users) 

---

If it's just used uuids then it's just ~[720Kbyte](https://stackoverflow.com/questions/29518372/how-much-memory-java-hashsetlong-should-take) assuming 10k different people log in over a period.

PR #6184, opened 2021-12-24
add log containing chosen card type to Biased Reporting

### LINKED ISSUE #6180: Biased Reporting doesn't log which card type was chosen
I think having an info log would be nice.

Cards already logging the chosen card type: Engram Flush, Peeping Tom, Saisentan, Focus Group, Standard Procedure.

francescopellegrini:
Fixes #6180

### DIFF
--- a/src/clj/game/cards/operations.clj
+++ b/src/clj/game/cards/operations.clj
@@ -233,11 +233,9 @@
       :prompt "Choose a card type"
       :choices ["Hardware" "Program" "Resource"]
       :async true
+      :msg (msg "choose " target)
       :effect (req (let [t target
                          n (num-installed state t)]
-                     (system-msg
-                       state :corp
-                       (str "uses Biased Reporting to choose " target))
                      (wait-for
                        (resolve-ability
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/operations.clj
```diff
@@ -235,6 +235,9 @@
       :async true
       :effect (req (let [t target
                          n (num-installed state t)]
+                     (system-msg
+                       state :corp
+                       (str "uses Biased Reporting to choose " target))
```
NoahTheDuke, line 240: This can go into a :msg entry.

francescopellegrini: I thought so as well, but then I noticed `system-msg` has been used twice in the same card implementation, so I assumed there had to be a reason for that.

NoahTheDuke: I haven't looked at the rest of the implementation, but any time you have a "uses Card Name to X" phrase in a system message, you can move the X part to a :msg, which will create the rest of the message for you.

PR #6195, opened 2022-01-01
fix Compile and Mayfly effects fixed resolution order

### LINKED ISSUE #5781: Mayfly & Compile Need Window to Choose Ordering
If runner uses compile to install Mayfly, then uses it to break a subroutine during that run, when the run ends Mayfly and Compile will meet their conditions simultaneously, and mayfly should go to heap or bottom of the stack depending on how the runner chooses to order their triggers (it should have the box that lets you pick which trigger goes first).  Currently, no box appears and mayfly always goes to the bottom of the stack.

francescopellegrini:
Fixes #5781


test/clj/game/cards/events_test.clj
```diff
@@ -1207,6 +1207,57 @@
       (is (= 1 (-> (prompt-map :runner) :choices count)) "Runner has 1 choice")
       (is (= ["Stack"] (prompt-buttons :runner)) "Runner's only choice is Stack")))
 
+(deftest compile-effects-ordering-1
+  ;; Compile & Mayfly - should get order of choice
+  (do-game
+      (new-game {:corp {:deck ["Masvingo"]
```
NoahTheDuke, line 1213: Check spacing and hanging parentheses.

NoahTheDuke (approved):

PR #6196, opened 2022-01-01
fix Armed Intimidation and Malapert Data Vault fixed resolution order

### LINKED ISSUE #5702: Malapert Data Vault + Armed Intimidation - "when you score" forced resolution order
When scoring an Armed Intimidation from a server with _Malapert Data Vault_, the Corp player is forced to resolve _Malapert Data Vault_ first, before the Runner is given the choice between taking 2 tags or taking 5 meat damage.

**Expected Behaviour:** when scoring an agenda with a "when you score" ability (such as _Armed Intimidation_) from a server with _Malapert Data Vault_ installed and rezzed, the Corp player should be able to choose the order of ability resolution.

**Actual Behaviour:** scoring some agendas works as expected, scoring _Armed Intimidation_ forces the order of resolution

Standoff - OK
Hostile Takeover - OK
Armed Initimidation - NOT OK

francescopellegrini:
Fixes #5702

### DIFF
--- a/src/clj/game/cards/agendas.clj
+++ b/src/clj/game/cards/agendas.clj
@@ -161,9 +161,11 @@
                       card nil))}})
 
 (defcard "Armed Intimidation"
-  (letfn [(ab []
+  {:on-score
    {:player :runner
+    :interactive (req true)
     :async true
+    :waiting-prompt "Runner to choose an option"
     :prompt "Choose Armed Intimidation score effect"
     :choices ["Suffer 5 meat damage" "Take 2 tags"]
     :effect (req (case target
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/agendas.clj
```diff
@@ -175,7 +174,12 @@
                    (do (system-msg state :runner "chooses to take 2 tags from Armed Intimidation")
                        (gain-tags state :runner eid 2 {:card card}))
                    ; else
-                   (effect-completed state side eid)))}})
+                   (effect-completed state side eid)))})]
+    {:on-score
+     {:interactive (req true)
```
NoahTheDuke, line 179: I think the only change that's necessary is putting `:interactive true` in the `:on-score` ability. Moving the main ability to a function doesn't actually change anything given that there's no recursion.

NoahTheDuke (approved):

Issue #5075, opened 2020-05-03
Maya trigger occasionally fires when it shouldn't


mgsholte:
The text for Maya says the decision to move a card to the bottom of R&D must be made immediately after accessing that card from R&D. If the runner access a card this way and declines to activate Maya and then accesses it again later, then they are re-prompted to activate the Maya trigger.

Example: 
- run R&D
- access a card from R&D
- trash that card
- decline to use the Maya trigger
- run archives
- the game prompts the runner to trigger Maya again for the card that was just trashed

A more contrived scenario:
- ensure the corp has no cards in hand
- run R&D
- access a card from R&D
- decline to trash that card
- decline to use the Maya trigger
- install Equivocation
- run R&D again
- use Equivocation to force the corp to draw the card that was previously accessed
- run HQ
- when the runner accesses the card in HQ, the game will prompt to use Maya

I think this only happens if the runner re-accesses a card on the same turn that they declined to use Maya on it, but I can't confirm that.

NoahTheDuke:
Excellent, thank you!

Diogened:
2020-08-26
I recently noticed that Maya give a prompt after a successful run on HQ. It allowed to use that card ability on a card located in HQ.

NBKelly:
Closed as completed via 71b5c9b and #6198

PR #6198, opened 2022-01-02
add test for Maya triggering on R&D accesses only

francescopellegrini:
I tried to reproduce the issue reported in #5075 but failed doing it.

Maybe it had been already fixed by https://github.com/mtgred/netrunner/commit/71b5c9b7606d19c2fc5178d237a9b0ae193df739.


NoahTheDuke (approved):

NBKelly:
Yeah, it was fixed by that one - it now picks up that the card accessed started in R&D. A previous failing case would be accessing any card in archives that had been trashed from R&D (only agendas and traps had access events wired up at this point) - since they had `:rd` as their previous zone.

PR #6201, opened 2022-01-02
fix Aeneas Informant not giving credits when cards are moved to RFG

### LINKED ISSUE #6161: Aeneas Informant trigger does not fire when a card is removed from game using Salsette Slums
There's a ruling that if a card is removed from game with Salsette Slums that should trigger the condition of Aeneas informant that the runner may gain 1 credit for accessing a card and not trashing it. 

Currently this doesn't fire on jnet though. Other triggers like Aumakua and Maw, and use similar wording do fire, so there is likely something specific in the implementation of Aeneas Informant that doesn't handle this case fully.  

image

image

francescopellegrini:
Fixes #6161.

### DIFF
--- a/src/clj/game/cards/resources.clj
+++ b/src/clj/game/cards/resources.clj
@@ -120,10 +120,10 @@
              :effect (effect (gain-clicks :runner 1))}]})
 
 (defcard "Aeneas Informant"
-  {:events [{:event :no-trash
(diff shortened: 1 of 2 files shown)


NoahTheDuke (approved):
This reminds me that I gotta try to make the access events use the queue event/checkpoint system.

Issue #6190, opened 2021-12-29
Bailiff doesn't trigger with (Rebirth-ed) Queztal's new wording


tvaduva:
Link to bug replay: https://jinteki.net/bug-report/bfb55719-ada4-4890-a5f0-ac05be12a6aa?b=0

Bailiff doesn't trigger with (Rebirth-ed) Queztal's new wording. See linked replay. I used Rebirth to be Queztal and when I broke Bailiff the Corp didn't gain a credit. Not sure if it's new wording of Queztal has updated to work with Bailiff or if it's because I Rebirth-ed into the ID.

francescopellegrini:
The gain-credit trigger on Bailiff [is not yet implemented](https://github.com/mtgred/netrunner/blob/master/src/clj/game/cards/ice.clj#L596).

NoahTheDuke:
I don't know why it isn't, we have the technology for tracking breaking subs. Hopefully someone can get around to it at some point.

tvaduva:
Ah, that makes sense! I didn't even think to check if it was implemented in the game. Just expecting that it was, as Noah noted, since so many of the breaking subs logic works so well. No big deal; I thought I caught an edge case that was missed. Thanks for all your effort to get so many of these cards working correctly.

PR #6207, opened 2022-01-09
Implementation and tests for Deep Dive, Bailiff

NBKelly:
Deep Dive will skip the menu/asking for a second access when there is exactly one card, which seems correct.

There are exactly two flaws with this implementation that I know about:
* You shouldn't be able to play deep dive when there are 0 cards in R&D (it can't change the game state)
* Bacterial Programming is incorrect when used during resolution of Deep Dive

The first issue is easily fixable, I just don't know exactly how (I assume it's simple oneliner in the req function). The second one would require a 'set-aside' implementation, which is currently a bit out of my league.

I believe I have got all the edge cases in testing, specifically:
* Trashing and stealing work
* The happy path (accessing 8 cards)
* It works with additional costs (strongbox, ikawah, etc)

Though it is possible I have still overlooked something.

### DIFF
--- a/src/clj/game/cards/events.clj
+++ b/src/clj/game/cards/events.clj
@@ -648,24 +648,29 @@
                                                          {:duration :end-of-run})]))}]})
 
 (defcard "Deep Dive"
- (letfn [(deep-dive-access-singleton [cards]
- 	  ;; access exactly one card without showing a menu
+  (letfn [(deep-dive-access [cards]
+ 	   ;; Accesses a card from a set of given cards.
+	   ;; Returns the set of unchosen cards as async-result
 	   {:async true
-	    :msg (msg "access " (:title (first cards)))
-	    :effect (req (wait-for
-	     	     	    (access-card state side (first cards))			    
-			    (effect-completed state side eid)))})
-	 (deep-dive-select [cards]	 
-	   ;; access a card, then return the list of cards without the accessed card
-	   {:prompt "Select a card to access"
-	    :waiting-prompt "Runner to access a card"
-	    :not-distinct true
-	    :choices cards
-	    :async true
-	    :effect (req (wait-for
-	    	    	   (access-card state side target)
-			   (let [new-cards (remove #(same-card? % target) cards)]
-			     (effect-completed state side (make-result eid new-cards)))))})]
+	    :effect (req (wait-for (resolve-ability state side
+	       (if (= 1 (count cards))
+	             ;; Only show a menu if there's more than one card
+	     	     {:async true
+		      :msg (msg "access " (:title (first cards)))
+	              :effect (req (wait-for
+	     	                     (access-card state side (first cards))
+		                     (effect-completed state side eid)))}
+		     {:prompt "Select a card to access"
+	              :waiting-prompt "Runner to access a card"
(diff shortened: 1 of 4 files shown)


NBKelly:
Whoops, was unaware github would automatically add new commits to the PR - I'll actually use branches in the future.

In any case, there's an automated implementation and tests for bailiff, which should close #6190.

src/clj/game/cards/ice.clj
```diff
@@ -593,9 +593,17 @@
                  end-the-run]})
 
 (defcard "Bailiff"
+  (letfn [(bailiff-gain-credits [state side eid n]
+  	    (if (pos? n)
+	      (wait-for (gain-credits state :corp (make-eid state eid) 1)
+	      		(bailiff-gain-credits state side eid (dec n)))
+	      (effect-completed state side eid)))]
   {:implementation "Gain credit is manual"
```
francescopellegrini, line 601: This can be dropped, I assume. ;)

src/clj/game/cards/events.clj
```diff
@@ -647,6 +647,102 @@
...
+	   :async true
+	   :effect (req (wait-for (access-card state side target)
+			(system-msg state side (str " shuffles R&D!"))
+			(shuffle! state :corp :deck)
+			(effect-completed state side eid)
+	   	   ))})
+
+	 ;;access exactly one card from a set of cards
+	 ;;then, potentially access a different card from that set
+	 ;;there is probably a better way to do this
```
lostgeek, line 675: I think you can break down Deep Dive like this:
- Choose and access a card
- Ask for a second pick
- Choose and access again
- Shuffle.

The `wait-for` mechanism has a place for keeping track of information called `async-result`. The awaited function can pass on a value `X` by calling `(effect-completed state side (make-result eid X))`. The next ability then has a binding `async-result` available to access it.

You can use this to create a `choose-and-access` ability that passes on the remaining set-aside cards as an `async-result`. Then your main effect looks something like this:
```
(wait-for (resolve-ability state side choose-and-access card set-aside)
          (resolve-ability state side {:optional
                                       {:prompt "Access another card?"
                                        :yes-ability choose-and-access}}
                           card async-result)
          (shuffle! state side :deck)
          (system-msg state :corp "shuffles R&D"))
```
Here I also use the fact that the 5-fn used all over the engine have "targets" as their fifth parameter, which I pass the `set-aside` cards and the `async-result` returned by the first ability through.

src/clj/game/cards/events.clj
```diff
@@ -647,6 +647,102 @@
                                                          (max 0 (min 4 (available-mu state)))
                                                          {:duration :end-of-run})]))}]})
 
+(defcard "Deep Dive"
+ (letfn [(deep-dive-single [cards state side eid shuffle]
+ 	  ;;access the top card from cards - to skip menu when there is only one valid choice
```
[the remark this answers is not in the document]
NoahTheDuke: I'll note that this one is pure preference, I use `;;` all the time, cuz I find it easier to notice. Either one you choose, just add a space between the final `;` and the start of your comment: `;; access the top`. this gives the comment a bit of breathing room and makes reading it easier.

PR #6220, opened 2022-01-20
add missing gain-tags and make-eid import

francescopellegrini:
Fixes the test CI broken since #6194 was merged, my bad :(


NoahTheDuke:
It's all good, we've all broken master at some point.

PR #6222, opened 2022-01-21
fix Rebranding Team being active in the Runner's score area

### LINKED ISSUE #6219: Rebranding Team active in runner score area
Link to bug replay: https://jinteki.net/bug-report/cebfd068-c284-4966-bef9-7daa9da1bac4?b=0

Description: Rebranding Team shouldn't be active in the runner score area, but it applied in this game.

francescopellegrini:
Fixes #6219.

Also fixes Remote Data Farm and Self-Destruct Chips not logging when swapped or added to the Corp's score area.

### DIFF
--- a/src/clj/game/cards/agendas.clj
+++ b/src/clj/game/cards/agendas.clj
@@ -1447,8 +1447,9 @@
                 :effect (effect (add-prop target :advance-counter 1 {:placed true}))}]})
 
 (defcard "Remote Data Farm"
-  {:on-score {:silent (req true)
-              :msg "increase their maximum hand size by 2"}
+  {:move-zone (req (when (and (in-scored? card)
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
This doesn't work if the card is added to the corp's score area and not scored (Lady Liberty, for example). I think all that's needed is removing the flag?

francescopellegrini:
> This doesn't work if the card is added to the corp's score area and not scored (Lady Liberty, for example). I think all that's needed is removing the flag?

That's a good point. Although there are a few other agendas which might suffer from this bug as well: Remote Data Farm, Self-Destruct Chips, Superconducting Hub (constant effect), Water Monopoly.

NoahTheDuke (approved):
Looks great, thank you!

PR #6224, opened 2022-01-22
Small updates to messages

francescopellegrini:
- replace 'add counter to' with 'place counter on'
- replace 'power token' with 'power counter'
- replace card self-reference with 'itself'
- replace 'choose not' with 'decline'
- replace 'put on' with 'add to'
- replace 'due to' with 'for'


src/clj/game/cards/agendas.clj
```diff
@@ -876,9 +876,8 @@
                 :effect (effect (damage eid :net 1 {:card card}))}]})
 
 (defcard "Hyperloop Extension"
-  (let [he {:async true
-            :effect (req (system-msg state side (str "uses Hyperloop Extension to gain 3 [Credits]"))
-                         (gain-credits state :corp eid 3))}]
+  (let [he {:msg "gain 3 [Credits]"
```
NoahTheDuke, line 879: Lost `:async true` here

src/clj/game/cards/assets.clj
```diff
@@ -188,8 +188,8 @@
 (defcard "Anson Rose"
   (let [ability {:label "Place 1 advancement token on Anson Rose (start of turn)"
                  :once :per-turn
-                 :effect (effect (system-msg (str "places 1 advancement counter on Anson Rose"))
-                                 (add-prop card :advance-counter 1 {:placed true}))}]
+                 :msg "place 1 advancement counter on Anson Rose"
```
NoahTheDuke, line 191: Itself?

src/clj/game/cards/assets.clj
```diff
@@ -1777,13 +1778,13 @@
                           (continue-ability
                             (let [move-to target]
                               {:async true
-                               :prompt "Move how many tokens?"
+                               :prompt "Move how many advancement tokens?"
                                :choices {:number (req (get-counters card :advancement))
                                          :default (req (get-counters card :advancement))}
                                :cost [:trash]
                                :effect (effect (add-counter move-to :advancement target {:placed true})
                                                (system-msg (str "trashes Reconstruction Contract to move "
```
NoahTheDuke, line 1785: Itself?

src/clj/game/cards/operations.clj
```diff
@@ -1152,13 +1152,13 @@
                      :prompt "Choose an agenda, asset or upgrade to install from Archives and place 2 advancement tokens on"
                      :show-discard true
                      :not-distinct true
+                     :msg "place 2 advancements counters on the installed card"
```
NoahTheDuke, line 1155: I think this should stay after the install. Otherwise it'll be out of order.

src/clj/game/cards/resources.clj
```diff
@@ -859,10 +859,10 @@
                   :choices (req (eligible-cards runner))
                   :effect (effect (move target :hand))
                   :msg (msg "add " (:title target) " to grip")}
-                 {:label "Add a power counter manually"
+                 {:label "Place a power counter"
                   :once :per-turn
                   :effect (effect (add-counter card :power 1))
-                  :msg "manually add a power counter"}]
+                  :msg "manually place a power counter on District 99"}]
```
NoahTheDuke, line 865: Itself

src/clj/game/cards/upgrades.clj
```diff
@@ -123,7 +123,7 @@
                                             :req (req (get-card state ice))
                                             :effect (effect (trash eid (get-card state ice) nil))}])
                                         (force-ice-encounter state side eid ice))))}
-              :no-ability {:effect (effect (system-msg :corp (str "declines to rez a hosted piece of ice")))}}}]})
+              :no-ability {:effect (effect (system-msg :corp (str "declines to rez a hosted piece of ice with Awakening Center")))}}}]})
```
NoahTheDuke, line 126: Declines to use?

NoahTheDuke (approved):
Some ideas, none are required

src/clj/game/cards/assets.clj
francescopellegrini: We're still using `system-msg` here, so the log doesn't have the "Runner uses XXX to" prefix. Using "itself" would result in "The runner trashes *itself*", which would be awkward 😆

francescopellegrini:
@NoahTheDuke on second thought, which of the following wording is more correct? (I'm not a native speaker...)

- X uses Y to put a credit on itself
- X uses Y to put a credit on it

I found some card using the latter: Mammon, Nfr, Sūnya, Chatterjee University, Temple of the Liberated Mind.

francescopellegrini:
> @NoahTheDuke on second thought, which of the following wording is more correct? (I'm not a native speaker...)
> 
> * X uses Y to put a credit on itself
> * X uses Y to put a credit on it
> 
> I found some card using the latter: Mammon, Nfr, Sūnya, Chatterjee University, Temple of the Liberated Mind.

I'm going with the latter because the Runner/Corp is the subject is all of our messages.

NoahTheDuke:
You've already done all the work so i I don't wanna demand you do more work, but I would say that "X uses Y to put a credit on itself" is the better sounding phrase because "itself" applies when a thing performs an action that applies to itself, even if it's something else causing that's action. Which is to say, even tho a player is the subject of the sentence, the card is performing an action on itself.

francescopellegrini:
No worries, it's just a matter of reverting the commit ;)

francescopellegrini:
I will split some changes in separate PRs as well.

NoahTheDuke:
Only if you want to. I don't mind reviewing PRs this big and have already done the work

PR #6230, opened 2022-01-23
Fix Saraswati Mnemonics and Mushin No Shin allowing rez out of turn

### LINKED ISSUE #6208: BUG: Saraswati Mnemonics (Identity: Jinteki) + NGO Front (Neutral Asset)
Saraswati Mnemonics - Jinteki Corp - click+1c: Install a card from HQ in a remote server, then place 1 advancement token on it. You cannot score or rez that card until your next turn begins.

NGO Front - Neutral Asset - rez cost 0 - NGO Front can be advanced. Trash+1adv token: Gain 5c. Trash+2adv tokens: Gain 8c

If you install NGO Front with Saraswati Mnemonics you are still able to rez and use NGO Front before your next turn begins.

francescopellegrini:
Fixes #6208

### DIFF
--- a/src/clj/game/cards/operations.clj
+++ b/src/clj/game/cards/operations.clj
@@ -1347,7 +1347,7 @@
     :effect (req (wait-for (corp-install state side target "New remote" nil)
                            (let [installed-card async-result]
                              (add-prop state side installed-card :advance-counter 3 {:placed true})
-                             (register-turn-flag!
+                             (register-persistent-flag!
                                state side
                                card :can-rez
                                (fn [state _ card]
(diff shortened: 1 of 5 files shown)


src/clj/game/cards/operations.clj
```diff
@@ -1360,7 +1360,11 @@
                                  (if (same-card? card installed-card)
                                    ((constantly false) (toast state :corp "Cannot score due to Mushin No Shin." "warning"))
                                    true)))
-                             (effect-completed state side eid))))}})
+                             (effect-completed state side eid))))}
+   :events [{:event :corp-turn-begins
+             :duration :until-corp-turn-begins
```
NoahTheDuke, line 1365: This will persist (lol) past when the operation is trashed at the end of the run?

francescopellegrini: I'm not sure I understood your question: operations are trashed immediately after resolving..

NoahTheDuke: `:events` entries are registered when an operation is played and before it resolves, and then after it's resolved they are removed. I think the `:duration` part means it won't be unregistered after resolution but I don't remember.

francescopellegrini: All I can say is the updated test is now green :)

PR #6231, opened 2022-01-25
Rewrite server code again

NoahTheDuke:
Got fed up with how I'd written the "new" server code, cuz it's messy and awkward. So I've rewritten it all again, this time using re-frame as a base (but making all dispatch synchronous). On the one hand, it can look really weird and feel a little magical, but on the other, it makes separating the "pure" from the "impure" logic _really nice_.

I feel like I want a middle-ground, but I haven't spent the time to build a "clean room" version of re-frame that does only what we need, so for now we'll use `executor`, my adaption of re-frame.


NoahTheDuke:
I wanted this to work so bad but it wasn't meant to be. 😭

PR #6232, opened 2022-01-26
Overclock

### LINKED ISSUE #6216: Overclock should not be able to be used as an additional cost for running Earth Station flipped remote
Runner has too few credits to run ES remote that is flipped (cost 6)

He is allowed to use Overclock and use those credits even though cost comes before card being played

NBKelly:
fixes overclock and cold read credits being able to be spent after playing the event but before being in a run (ie to pay for earth station).

Closes #6216 , but also deals with:

* Cold read vs. earth station
* This same issue was happening with reduced service and cold site server too, and should be resolved

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -75,7 +75,7 @@
                  [metosin/malli "0.7.0"]]
 
   :plugins [[lein-eftest "0.5.9"]
-            [cider/cider-nrepl "0.27.3"]]
+            [cider/cider-nrepl "0.26.0"]]
 
   :profiles {:dev {:dependencies [[binaryage/devtools "1.0.4"]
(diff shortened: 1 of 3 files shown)


project.clj
```diff
@@ -75,7 +75,7 @@
                  [metosin/malli "0.7.0"]]
 
   :plugins [[lein-eftest "0.5.9"]
-            [cider/cider-nrepl "0.27.3"]]
+            [cider/cider-nrepl "0.26.0"]]
```
NoahTheDuke, line 78: Let's not downgrade libraries

NoahTheDuke (approved):

NoahTheDuke (approved):

PR #6234, opened 2022-01-27
Adding Polish to the translations file

vesperius:
1st version of Polish strings, as per currently listed in the :en array


NoahTheDuke (approved):
Hell yeah, thank you so much for taking the time to write all of this out.

vesperius:
My pleasure :) I just hope it displays nicely with some phrases being a bit longer (and will fix it if it won't).
Thanks for keeping the servers ready to be ran ;-)

Issue #5282, opened 2020-09-11
Reset username feature


aseixas:
After a long absence, I was trying to get into the game again, but couldn't sign up on jinteki.net as my email had been used. So I reset the password, but now I still can't log in as the emails didn't mention what username I just reset the password to.

Dirjel:
Could we make the "reset password" email include the username?  This is a common problem.

PR #6254, opened 2022-02-15
Add username in password reset email and sanitize usernames

lf94:
Should resolve https://github.com/mtgred/netrunner/issues/5282 .

Untested.


src/clj/web/auth.clj
```diff
@@ -198,19 +198,28 @@
...
+                      {:email email})
+  )
+)
+
 (defn forgot-password-handler
   [{db :system/db
     email-settings :system/email
     {:keys [email]} :params
     headers         :headers}]
   (if (find-non-banned-user db {:email email})
```
jwarwick, line 214: Instead of querying the db again for the username, I'd make this an `if-let` to bind the `user`.

src/clj/web/auth.clj
```diff
@@ -198,19 +198,28 @@
...
     (let [code (set-password-reset-code! db email)
+          user (get-user-of-email db email)
           msg (mail/send-message
                 email-settings
                 {:from    "support@jinteki.net"
                  :to      email
                  :subject "Jinteki Password Reset"
-                 :body    (str "You are receiving this because you (or someone else) have requested the reset of the password for your account.\n\n"
+                 :body    (str "You are receiving this because you (or someone else) have requested the reset of the password for your account " (user :username) ".\n\n"
```
jwarwick, line 222: We probably need to sanitize this somehow. Otherwise we would send user generated text to the email address. Not sure what rules we would want to sanitize against.

NoahTheDuke: Do we do any sanitization for usernames already? If so, should be good to go. If not, then hmmm

lf94: It appears there is no sanitation for usernames. See: [netrunner/src/clj/web/auth.clj:66`](https://github.com/mtgred/netrunner/pull/6254/files#diff-9457c792e3eb8cde9946d207495013396702673433db6ff9d8676e49a7834f4fL66)

But also, I'm pretty sure the email is sent as `plain/text` and is not an HTML email. In order to have an HTML email a different `:content_type` must be specified via `postal` it appears.

src/clj/web/auth.clj
lf94: Fixed - hopefully it's correct. I've literally never used Clojure before today.

lf94:
@jwarwick @NoahTheDuke I've added the username sanitation, which uses patterns suggested by Noah from the Discourse project. Please test as I have nothing setup to test!

src/clj/web/user.clj
```diff
@@ -9,6 +9,28 @@
...
+                        invalid-leading-char-pattern,
+                        invalid-trailing-char-pattern,
+                        repeated-special-char-pattern
+                        ])
+
+(defn valid-username?
+  "Validate a username"
+  [username]
+  (if (<= (count username) max-username-chars)
+    (every? (fn [re] (= (re-matches re username) false)) patterns-username)
```
NoahTheDuke, line 30: (= X false) can be (not X). But also, (every? (not X) Y) should be (not (any? X Y)). So:

```suggestion
    (not (any? #(re-matches % username) patterns-username))
```
I'm on my phone so test this first but it should be equivalent while also short-circuiting on the first match.

src/clj/web/user.clj
```diff
@@ -9,6 +9,28 @@
...
+                        unicode-invalid-char-pattern,
+                        invalid-leading-char-pattern,
+                        invalid-trailing-char-pattern,
+                        repeated-special-char-pattern
+                        ])
+
+(defn valid-username?
+  "Validate a username"
+  [username]
+  (if (<= (count username) max-username-chars)
```
NoahTheDuke, line 29: Instead of (if X Y false), you can do (and X Y).

src/clj/web/user.clj
```diff
@@ -9,6 +9,28 @@
    :isadmin :ismoderator :tournament-organizer
    :special :options :stats :has-api-keys :banned])
 
+(def max-username-chars 20)
+(def ascii-invalid-char-pattern #"[^\w.-]")
+(def unicode-invalid-char-pattern #"[^\p{Alnum}\p{M}._-]")
+(def invalid-leading-char-pattern #"^[^\p{Alnum}\p{M}_]+")
+(def invalid-trailing-char-pattern #"[^\p{Alnum}\p{M}]+$")
+(def repeated-special-char-pattern #"[-_.]{2,}")
+(def patterns-username [
```
NoahTheDuke, line 18: I'd prefer this to be on the next line and the white space trimmed so it's:
```clojure
(def patterns
  [pattern-1
   ...
   pattern-X])
```

NoahTheDuke (commented):
Looking good! I'd love to see some tests in the appropriately matched test namespace that show failure on invalid usernames.

src/clj/web/user.clj
lf94: Heh, I was going to use function literal shorthand here but was worried it was unconventional.

I see you prefer to compose functions vs using something like every - sounds good to me :slightly_smiling_face: 

I'll look around the codebase to see where to add tests...

src/clj/web/user.clj
lf94: Yeah, I just did that because my editor auto-did it, and figured it knows better than me for Clojure code :laughing:

src/clj/web/user.clj
lf94: Roger o7

lf94:
@NoahTheDuke ready for further comments. :slightly_smiling_face:

test/clj/web/user/test.clj
```diff
@@ -0,0 +1,16 @@
...
+(deftest username
+  (testing "valid-username?"
+    (is (valid-username? "test"))
+    (is (not (valid-username? "abcdefghijklmnopqrstu")) "Username is invalid when more than 20 characters")
+    (is (valid-username? "abcdefghijklmnopqrst") "Username is valid when less than or equal to 20 characters")
+    (is (valid-username? "🐸") "Emoji are valid in usernames")
+    (is (valid-username? "🐸🐸🐸🐸🐸🐸🐸🐸🐸🐸🐸🐸🐸🐸🐸🐸🐸🐸🐸🐸") "Make sure 20 glyphs is still valid")
+    (is (valid-username? "<script src=\"t.js\"") "HTML-like text is valid")
+    (is (valid-username? "<h1>hello</h1>") "HTML is not valid")
+    (is (valid-username? "http://example.org") "URLs are not valid")))
```
NoahTheDuke, line 15: This should fall so use not

lf94: Blah silly me. Fixed it and found out I needed `re-find` instead of `re-matches` - all passing now :slightly_smiling_face:

test/clj/web/user/test.clj
```diff
@@ -0,0 +1,16 @@
+(ns web.user.test
+  (:require 
+   [web.user :refer :all]
+   [clojure.test :refer :all]))
```
NoahTheDuke, line 4:
```suggestion
   [clojure.test :refer [deftest is]]))
```

src/clj/web/user.clj
```diff
@@ -9,6 +9,22 @@
    :isadmin :ismoderator :tournament-organizer
    :special :options :stats :has-api-keys :banned])
 
+(def character-length #(.codePointCount % 0 (count %)))
```
NoahTheDuke, line 12: Doesn't `count` call String's `length`, which returns the number of code points in the Java String?

test/clj/web/user/test.clj
```diff
@@ -0,0 +1,16 @@
...
+
+(deftest username
+  (testing "valid-username?"
```
NoahTheDuke, line 7: Remove this testing block cuz it doesn't add anything relevant to the failure message.

NoahTheDuke (commented):
Looking good! Sorry for all the back and forth.

src/clj/web/user.clj
lf94: No, I originally used `count` and it was doing a byte count, not glyph count.

NoahTheDuke: hmmmmmmmm. This will probably also throw a reflection warning, so maybe
```suggestion
(defn character-length [^String username]
  (.length username))
```

LukaszKundys:
Hello, what's the status of this feature?

lf94:
It's been done for awhile now. There are arguments on what it should accept. I no longer have time to touch this though.

NoahTheDuke:
For some reason, it got missed in the most recent server update. Thankfully, the code hasn't gone away so one of us will hopefully take a look at some point and update/merge it and then it'll go live!

PR #6275, opened 2022-03-13
Midnight sun updates

NBKelly:
Here's all the changes I made for midnight sun.

Features include:

* prevent-ability renamed to prevent-paid-ability
* additional-cost function added for agendas
* `trash` to `trash-can` for self-trash costs, which disambiguates it from the other `trash` key
* `trash-installed` and `trash-other-installed` costs for cards
* Implementations for anemone, revolver, light the fire, hakarl, azef protocol

Note that currently, azef protocol has one failing test, for the case where you attempt to trash itself as part of scoring it. The actual scenario works, but I need to fix the test - I'll have a go at tackling that a bit later.


src/clj/game/cards/agendas.clj
```diff
@@ -1828,3 +1834,5 @@
                                       (not (has-subtype? target "Virtual"))
                                       (not (:facedown (second targets)))))
                        :value 1}]})
+
```
NoahTheDuke, line 1837: Mind removing these additional spaces at the end?

src/clj/game/core/actions.clj
```diff
@@ -546,26 +546,48 @@
...
-       (when-let [on-score (:on-score (card-def c))]
-         (register-pending-event state :agenda-scored c on-score))
-       (queue-event state :agenda-scored {:card c
-                                          :points points})
-       (checkpoint state nil eid {:duration :agenda-scored})))))
+     (let [additional-costs (score-additional-cost-bonus state side card)
+           cost (merge-costs (mapv first additional-costs))
+           cost-strs (build-cost-string cost)
+           can-pay (can-pay? state side (make-eid state (assoc eid :additional-costs additional-costs)) card (:title card) cost)]
+       (if (string/blank? cost-strs)
```
NoahTheDuke, line 580: Nested ifs can be a `cond`.

PR #6288, opened 2022-03-16
Timestamp cleanup

jwarwick:
Now sending all timestamps in UTC string format and displaying times in localtime.


NoahTheDuke:
Damn dude, thanks for this. What a mess I left.

PR #6291, opened 2022-03-16
Fix time for decks

jwarwick:


NoahTheDuke:
Should we be coercing everything in the database to handle this?

jwarwick:
@NoahTheDuke - maybe? 🤷 There's a lot of cruft in the db and I figured it was safer to catch it here, though dates for other items could have this same issue.

PR #6300, opened 2022-03-18
More updates

### LINKED ISSUE #6116: Password-protected lobbies should *not* use a "password" field
I don't care if someone reads my j-net lobby password over my shoulder when I create a private lobby, and making it a cleartext field would:
a) let me make 100% sure I have the right password
b) make chrome stop trying to update my j-net password every time I make a private lobby

### LINKED ISSUE #6298: User management tab is blank
The new features for user management are blank post-March 17th upgrade.

jwarwick:
Fixes #6116 
Fixes #6298

### DIFF
--- a/src/clj/web/admin.clj
+++ b/src/clj/web/admin.clj
@@ -76,7 +76,10 @@
     {:keys [action user-type username] :as data} :?data
     uid :uid}]
   (if (and (active-user? user)
-           (:isadmin user)
+           (or (:isadmin user)
+               (and (:ismoderator user)
+                    (or (= user-type :specials)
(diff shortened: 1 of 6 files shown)


src/clj/web/admin.clj
```diff
@@ -76,7 +76,10 @@
     {:keys [action user-type username] :as data} :?data
     uid :uid}]
   (if (and (active-user? user)
-           (:isadmin user)
+           (or (:isadmin user)
```
NoahTheDuke, line 79: Don't we have a helper function for this stuff?

PR #6305, opened 2022-03-20
fix Customized Secretary prompt not closing for the Corp

### LINKED ISSUE #6303: Customized secretary, when installed, tells the corp that the runner is taking an action when the runner has no prompt
On install of customized secretary, the corp gets a "runner is taking an action" prompt that persists even when the runner ends their turn, preventing the corp from beginning their next turn. This is solved by the corp manually using the /close-prompt command. The runner has no additional prompts on their screen.

francescopellegrini:
Fixes #6303

### DIFF
--- a/src/clj/game/cards/programs.clj
+++ b/src/clj/game/cards/programs.clj
@@ -848,8 +848,7 @@
 (defcard "Customized Secretary"
   (letfn [(custsec-host [cards]
             (if (empty? (filter program? cards))
-              {:async true
-               :msg "shuffle the stack"
+              {:msg "shuffle the stack"
                :effect (effect (shuffle! :deck))}
               {:prompt "Choose a program to host"
                :choices (concat (filterv program? cards) ["Done"])
--- a/test/clj/game/cards/programs_test.clj
+++ b/test/clj/game/cards/programs_test.clj
@@ -1817,6 +1817,32 @@
               (= (:title (nth (-> @state :runner :deck) 2)) "Deuces Wild")
               (= (:title (nth (-> @state :runner :deck) 3)) "Encore"))))))
 
+(deftest customized-secretary-shuffles-stack-when-no-program-is-hosted
+  ;; Customized Secretary - shuffles the stack when no program is hosted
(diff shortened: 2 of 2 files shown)


src/clj/game/cards/programs.clj
```diff
@@ -868,7 +867,8 @@
                   :waiting-prompt "Runner to make a decision"
                   :effect (req (let [from (take 5 (:deck runner))]
                                  (wait-for (reveal state side from)
-                                           (continue-ability state side (custsec-host from) card nil))))}
+                                           (continue-ability state side (custsec-host from) card nil)
+                                           (effect-completed state side eid))))}
```
NoahTheDuke, line 871: This is incorrect. Effect-completed should never be called after continue-ability; the ability being confused should call effect-completed.

src/clj/game/cards/programs.clj
```diff
@@ -856,8 +856,7 @@
                :async true
                :effect (req (if (= target "Done")
                               (do (shuffle! state side :deck)
-                                  (system-msg state side "shuffles the stack")
-                                  (effect-completed state side eid))
```
NoahTheDuke, line 860: This line needs to stay here. All branches of an async ability need to call effect-completed.

NoahTheDuke (changes requested):
The ability on line 851 when there are no programs is tagged as async but doesn't called effect-completed. Remove async and it'll work.

src/clj/game/cards/programs.clj
```diff
@@ -850,7 +850,8 @@
             (if (empty? (filter program? cards))
               {:async true
                :msg "shuffle the stack"
-               :effect (effect (shuffle! :deck))}
+               :effect (req (shuffle! state side :deck)
+                            (effect-completed state side eid))}
```
NoahTheDuke, line 854: Because the ability only called system-msg, it doesn't need to be async, so remove this line and the :async true a couple lines above.

NoahTheDuke (changes requested):

PR #6313, opened 2022-03-23
Get replay working

### LINKED ISSUE #5957: Feature Request: /save-replay command
It would be nice to have an in-game command to allow players to save the replay if they did not check it in the lobby.

Potentially this could require both players to consent by typing it similar to /undo-turn

jwarwick:
Fixes #5957

### DIFF
--- a/src/cljs/nr/lobby.cljs
+++ b/src/cljs/nr/lobby.cljs
@@ -24,10 +24,10 @@
   (swap! app-state assoc :games data))
 
 (defmethod ws/event-msg-handler :lobby/state [{data :?data}]
-  ;; XXX - the empty state sent with the lobby list is blowing up the replay's started flag
-  (swap! app-state assoc :current-game data)
-  (when (:started data)
-    (ws/ws-send! [:game/resync {:gameid (:gameid data)}])))
+  (when-not (= "local-replay" (:gameid @app-state))
+    (swap! app-state assoc :current-game data)
+    (when (:started data)
+      (ws/ws-send! [:game/resync {:gameid (:gameid data)}]))))
 
 (defmethod ws/event-msg-handler :lobby/notification [{data :?data}]
   (play-sound data))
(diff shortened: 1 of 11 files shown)


src/cljs/nr/lobby.cljs
```diff
@@ -84,10 +85,14 @@
                                "error" {:time-out 0 :close-button true}))))))))
 
 (defn leave-game []
-  (ws/ws-send! [:game/leave {:gameid (current-gameid app-state)}]
-               8000
-               #(when (sente/cb-success? %)
-                  (leave-game!))))
+  (if (= "local-replay" (:gameid @app-state))
+    (do
+      (swap! app-state assoc :gameid nil)
```
NoahTheDuke, line 90: We shouldn't be tracking :gameid on app-state anymore, which is why there's the current-gameid function. This line might not be "bad" but it's not necessary.

src/cljs/nr/routes.cljs
```diff
@@ -40,6 +40,8 @@
                       :view deck-builder}]
      ["/play" {:name :nav/lobby
                :view lobby-or-game}]
+     ["/replay/:rid" {:name :nav/replay-lobby
```
NoahTheDuke, line 43: Is this being used somewhere?

src/cljs/nr/lobby.cljs
jwarwick: It seems like we still use `:gameid` at the top level of the state in a few places. I agree that we should pick one and stick with it. Looking at all the various flows I can imagine how we aren't setting things correctly in all cases and sending people to fake games and lobbies.

src/cljs/nr/routes.cljs
jwarwick: We redirect to a `/replay/...` path when launching a replay and without this it was confusing the router. I don't use `rid` since we actually take it from an element in the page, but wanted to keep it for path matching purposes.

src/cljs/nr/lobby.cljs
NoahTheDuke: Ah geez, really? Damn, sorry about that. One of these days I'll find time and look at that mess. I thought caught everything.

PR #6319, opened 2022-03-26
WIP Set aside

lostgeek:
PR opened for nbkelly

### DIFF
--- a/src/clj/game/cards/hardware.clj
+++ b/src/clj/game/cards/hardware.clj
@@ -763,16 +763,16 @@
                        (doseq [c shuffle-back]
                          (move state side c :deck))
                        (shuffle! state side :deck))})
-          (shuffle-next [set-aside target to-shuffle]
-            (let [set-aside (remove-once #(= % target) set-aside)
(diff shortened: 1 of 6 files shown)


NBKelly:
This closes #6324

NBKelly:
A summary for these changes:
* There's now a 'set-aside' zone
* Some cards (deep dive, gachapon) interact with the set aside zone
* This prevents messy interactions with Bacterial Programming, paige piper, that we had before by pretending to have a set-aside zone
* Cards that are in set-aside still belong to the set of 'all-cards'
* the zone pops up in the corner (one for each side) when in use, like the play area and rfg
* Gachapon and Deep Dive have been updated to use the Set Aside zone

Issues with this currently:
* Some tests for gachapon fail (specifically, the corporation does not see a waiting prompt)
* The corporation never sees the set-aside zone. As far as I can tell, only the runner sees it. I'm not sure why this happens.
* We should be able to easily specify if the cards in the zone are public (face-up), or private (face-down to opponent). This doesn't matter yet (as far as I know), but may later with future cards.

NBKelly:
See #6362

PR #6326, opened 2022-04-07
Stop go loops in deckbuilder when leaving page

jwarwick:
Fixes problem requiring multiple clicks to select a deck after leaving the page and no longer displaying card zoom images


NoahTheDuke:
Oh nicely caught!

PR #6337, opened 2022-04-14
fixes kyuban/inversificator interaction

### LINKED ISSUE #6336: Inversificator + Kyuban: Kyuban Does Not Trigger When Inversificator is Used to Swap the Passed Host Ice
Kyuban does not trigger when Inversificator is used to swap the passed host ice. 

Kyuban text: "Install Kyuban only on a piece of ice. **Gain 2[c] whenever you pass host ice.**"

Inversificator text: "**The first time each turn you pass a piece of ice** after an encounter during which this program fully broke that ice, **you may swap it** with another installed piece of ice."

When the host ice is passed, I believe Kyuban should trigger, giving the runner 2 credits, and then the runner should be prompted if they would like to use Inversificator to swap the ice. 

I believe part of why Kyuban doesn't trigger is the logic used to fix Inversificator's interaction with Kakugo (https://github.com/mtgred/netrunner/issues/5143). I first noticed this issue in 2020 after the Kakugo interaction was corrected - apologies for not reporting it then!

**Reproduction**
- Have the Corp install two ice protecting a server. 
- Have the Runner, Rielle "Kit" Peddler: Transhuman, install Inversificator and then install Kyuban on the outermost ice (henceforth, the "host ice").
- Have the Runner complete a successful run on the iced server. During this run, the Corp should rez the host ice so that the Runner can use Inversificator to break all its subroutines and swap it with the innermost ice. In completing this run, the Runner will pass the host ice twice, so Kyuban should fire twice, but Kyuban will only fire the second time host ice is passed; Kyuban will fail to fire the first time host ice is passed, when Inversificator's swap ability is used.

NBKelly:
Fixes #6336 

Just had to mark them as interactive.

### DIFF
--- a/src/clj/game/cards/programs.clj
+++ b/src/clj/game/cards/programs.clj
@@ -1462,6 +1462,7 @@
   (auto-icebreaker {:abilities [(break-sub 1 1 "Code Gate")
                                 (strength-pump 1 1)]
                     :events [{:event :pass-ice
+                              :interactive (req true)
                               :req (req (and (all-subs-broken-by-card? (:ice context) card)
                                              (first-event? state side :end-of-encounter
                                                            (fn [targets]
@@ -1547,6 +1548,7 @@
   {:hosting {:card #(and (ice? %)
                          (can-host? %))}
    :events [{:event :pass-ice
+             :interactive (req true)
              :req (req (same-card? (:ice context) (:host card)))
              :msg "gain 2 [Credits]"
              :async true


kevnburg:
Tested in game and ran into the issue again:
Kyuban Inversificator Bug

NoahTheDuke:
Sorry, we've not updated in a while. Hopefully we'll get it live soon.

kevnburg:
Ah okay - v106.1 was just a hat art update. No worries!

PR #6338, opened 2022-04-18
Prepaid/mystic QOL

### LINKED ISSUE #5778: Networking buyback can be paid with Prepaid VoicePad
I dont think this is actually possible by the rules, and a check in GLC said as much.

### LINKED ISSUE #6335: Public Trail can be paid using Prepaid VoicePAD
image
Tested and confirmed, Runner with [Prepaid VoicePAD](https://netrunnerdb.com/en/card/31038) can used hosted credits to pay the ability of [Public Trail](https://netrunnerdb.com/en/card/30057).

Since Public Trail is an Operation and not an Event, this shouldn't be allowed (and even if it was, I believe Prepaid VoicePAD is only supposed to pay for the play cost and not the cost of any abilities, but that's irrelevant).

Reproduction: 
- Have the runner install one or more Prepaid VoicePADs. 
- Make sure they are hosting a recurring credit, have the runner make a run. Runner should end their turn with at least 8 credits.
- Then have the corp play Public Trail. Runner choose to pay to avoid the tag, then gets the option to spend credits from Prepaid VoicePAD.

NBKelly:
* Prepaid and mystic maemi can be used for X-cost events (bribery, brute-force-hack).
* Prompts pass an EID into maybe-pay (which could be relevant later).
* Networking can no longer be paid with prepaid/mystic (closes #5778).
* Public Trail can no longer be paid with prepaid/mystic (closes #6335).


src/clj/game/cards/events.clj
```diff
@@ -220,7 +220,8 @@
     :msg (msg "increase the rez cost of the first unrezzed piece of ice approached by " target " [Credits]")
     :effect (effect
               (continue-ability
-                (let [bribery-x target]
+               (let [bribery-x target
+                     eid (assoc eid :x-cost true)]
```
NoahTheDuke, line 224: I don't know how we want to do it but this should be programmatic instead of individually placed in each necessary card.

src/clj/game/core/actions.clj
```diff
@@ -143,14 +143,16 @@
...
         (do (remove-from-prompt-queue state side prompt)
-            (wait-for (maybe-pay state side card choices choice)
-                      (when (:counter choices)
-                        ;; :Counter prompts deduct counters from the card
-                        (add-counter state side card (:counter choices) (- choice)))
-                      ;; trigger the prompt's effect function
-                      (when effect
-                        (effect (or choice card)))
-                      (finish-prompt state side prompt card)))
+            (system-msg state side (str (:eid prompt)))
```
NoahTheDuke, line 146: Looks like a debug here

src/clj/game/cards/events.clj
NBKelly: ^^ I agree. There isn't currently any information on the card being x-cost as far as I can tell (it's listed as 0 - I think it's stripped out in netrunner-data). What do you suggest?

NoahTheDuke: Damn, is it really? That's a shame cuz it makes this harder. Maybe we use this fix for now and spend some hammock time thinking about a good abstraction for this.

PR #6342, opened 2022-04-27
fixed ltf/reduced service interaction, added test

### LINKED ISSUE #6340: Reduced Service should still charge Light the Fire
Reduced Service:

> As an additional cost to run this server, the Runner must pay 2credit for each hosted power counter.

Light the Fire:

> Run a remote server. During that run, cards in the root of the attacked server lose all abilities.

As costs to run come before the run starts, Reduced Service should still require the runner to pay the cost even when using Light the Fire. However, this doesn't appear to be happening. In the screenshot below I had a remote with an RS and a single counter on it, the runner ran it once normally and was correctly charged 2c, and the second time they used LtF and weren't charged.

Screenshot 2022-04-26 at 16 49 56

NBKelly:
Changed order of events in light the fire (make run, then disable cards). All tests seem to still pass.
Also added a test to see that the runner does pay before running.

Fixes #6340

### DIFF
--- a/src/clj/game/cards/resources.clj
+++ b/src/clj/game/cards/resources.clj
@@ -1614,7 +1614,7 @@
                                        (when (and (some #{:content} (:zone second-card))
                                                   (some #{server} (:zone second-card)))
                                          (disable-card state :corp second-card))
-                                       ;; disable cards that have left the server
+                                       ;; enable cards that have left the server
                                        (when (and (some #{:content} (:zone first-card))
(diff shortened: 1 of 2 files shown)


NBKelly:
Note that although this enforces the runner needing to pay the cost, we don't (seem) to have any blanket 'can afford to run' functions (can-run? only looks at the flags), so players may end up taking brain damage for no reason trying to run a server that they shouldn't be able to run (reduced service, earth station).

Issue #6347, opened 2022-04-29
Class act leaves a card in the stack when drawing your entire deck (unless it's one card)


NBKelly:
I only noticed because I've been playing a lot of Obelus.

If you draw cards equal to (or greater than) your deck size, you have to bottom one, then it stays. I think we can probably skip the class act trigger altogether if `cards-drawn >= cards-in-deck` (right now we skip it if there's exactly one card in deck).

francescopellegrini:
I tried to fix the behaviour of Class Act in #6265 but I probably did it wrong..
Feel free to fix it :)

NBKelly:
hmm - so it seems like the eid and context on `draw-cards` has no notion of how many cards it intended to draw (only how many it was able to draw). I wonder if that's the angle of attack we need to find a solution.

PR #6349, opened 2022-04-30
[WIP] Class act lets you draw your whole deck

NBKelly:
See #6347, #6265

Looking at the netrunnerdb page on the class act, it appears that the wording has been changed significantly from what's actually printed on the card.

Here's the original (printed) text:
* Immediately before you draw for the first time each turn, look at the top X cards of your stack. Add 1 of these to the bottom of the stack. X is equal to the number of cards you will draw plus 1.

Here's the updated text:
* The first time each turn you would draw any number of cards, increase the number of cards you will draw by 1. When you draw those cards, if you drew at least 2, add 1 of them to the bottom of your stack.

It turns out our implementation is actually correct with the updated text, but I am going to bring this up with rules to figure out of this is intended, or if they're going to rewrite it again.

------------

Previously, class act would force you to bottom a card if you were drawing your whole deck (ie Diesel on 3/2 cards gave 2/1 in hand).
Short of rewriting the card to act exactly as written (which I think we already moved away from), the only solution I could think of was making the `draw-cards` function actually pass through the number of cards it intended to draw.
I've made the following assumptions:
* If the number of cards you intend to draw is equal to the number of cards you actually draw, then you should bottom a card 

**OR**
* If the number of cards in the deck is greater than 0 after drawing, then you should bottom a card

The first one ensures most of the regular cases, and the second one ensures that interactions with `:prevent-draw` effects now work exactly the same as they did before. I'm not sure if they're actually correct, but all tests pass, so I believe they have the same behavior as they did before this change. 

The only time this will trigger is if:
* You planned to draw more cards than you did (ie target = 9, but you actually drew 7)
 
 **AND**
* This leaves 0 cards in the stack (which means that bottoming a card would decrease your draws by another 1).

### DIFF
--- a/src/clj/game/cards/resources.clj
+++ b/src/clj/game/cards/resources.clj
@@ -2783,7 +2783,8 @@
               (first-time-draw-bonus :runner 1)
               {:event :runner-draw
                :req (req (and (first-event? state :runner :runner-draw)
-                              (< 1 (count runner-currently-drawing))))
+                              (or (= (count runner-currently-drawing) (:target-draws context))
+                                  (< 0 (count (:deck runner))))))
                :once :per-turn
                :once-key :the-class-act-put-bottom
                :async true
--- a/src/clj/game/core/drawing.clj
+++ b/src/clj/game/core/drawing.clj
@@ -76,6 +76,7 @@
                (let [draw-event (if (= side :corp) :corp-draw :runner-draw)]
                  (swap! state update-in [side :register :currently-drawing] conj drawn)
                  (queue-event state draw-event {:cards drawn
+                                                :target-draws draws-wanted
                                                 :count drawn-count})
                  (wait-for
                   (checkpoint state nil (make-eid state eid) nil)


NBKelly:
I've asked about this on discord, if rules tells me they don't plan to update the class act again then I guess I'll close this issue.

NBKelly:
Started an email chain with rules to figure this out (and a few other rules questions too), I'll update this one again a while later.

NoahTheDuke:
lmao How about now?

NBKelly:
> lmao How about now?

I had forgotten about this, but the most recent CR actually reverted the behavior - class act now has you look at the top x+1 again, bottom 1, then do the draw step. I'll make a PR for that a little bit later :)

francescopellegrini:
>  I'll make a PR for that a little bit later

@NBKelly how's that PR going? I tried to implement the latest CR behavior but with no success.. :/

NBKelly:
Hi, I totally blanked on this (tagme doesn't feel good in the current meta).

I think we can pick up the old class act implementation from the commit history.

Issue #5959, opened 2021-06-16
Show implementation notes in game log


lostgeek:
When a card with an implementation note is rezzed / installed face-up / scored, print the log to the game log as an information for players.

francescopellegrini:
Closed by #6359

PR #6359, opened 2022-05-09
Show implementation messages in game log

### LINKED ISSUE #5210: A UI element for errata.
Currently, there is no way to tell at all from the Jinteki.net interface if there is errata for the card. I personally think it would be ideal if it showed up along with the card art when moused over, but any sort of notification would be nice

NBKelly:
This allows for implementation messages to be shown for relevant cards when they're rezzed, installed face-up (runner, or corp agendas), scored (by the corp), or played from hand (operation/events).

There is a single `implementation-msg` command in `say` which takes `state` and `card`, and if the card has implementation information, prints it.

impl_04

impl_03

impl_02

impl_01

I'm not 100% sold on the format I'm using, but it seems alright for now. I can't use a capital I on implementation, because that matches Imp.

I think there should be implementation notes on Identities, but I have no clue how (where) to sanely implement that.

Closes #5210, #5959

### DIFF
--- a/src/clj/game/core/play_instants.clj
+++ b/src/clj/game/core/play_instants.clj
@@ -36,6 +36,7 @@
                    "play "
                    (build-spend-msg payment-str "play"))]
     (system-msg state side (str play-msg title (when ignore-cost " at no cost")))
+    (implementation-msg state card)
     (play-sfx state side "play-instant")
     ;; Select the "on the table" version of the card
     (let [card (current-handler state side card)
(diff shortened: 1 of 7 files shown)


src/clj/game/core/play_instants.clj
```diff
@@ -48,6 +48,8 @@
                             card)
                           {:resolve-effect true :init-data true});;:resolve-effect is true as a temporary solution to allow Direct Access to blank IDs
           play-event (if (= side :corp) :play-operation :play-event)]
+      ;; print card errata/implementation if it has any
+      (implementation-msg state card)
```
NoahTheDuke, line 52: Move to directly below the `system-msg`.

NoahTheDuke:
I think this is great. I left comments saying to move these messages directly below the existing message that is printed so 1) the message "position" is consistent and 2) the implementation details are the latest/most prominent thing. If you think it should be the other way around, let's do that then, but in either case we should be printing them consistently either before or after.

NBKelly:
Alright, I think I got those suggestions. 

Is there a reasonable place to put implementation messages for ids?

NBKelly:
The only reasonable place I could find for id messages was `set-up`.  It's called directly after the id loads when the game starts.

impl_05

PR #6360, opened 2022-05-10
New commands

### LINKED ISSUE #6256: Improve Derez/Rerez Handling
Turning this conversation on Discord into an issue:

image

The rerez issue is well-explained in the image.

For further context on the Derez suggestion:
Currently Derez is available as a menu item when the corp clicks on a rezzed card. Since derezzing is almost entirely a "manually fix a broken game state" command, it makes sense to only make the derez command available as a slash-command (where the other "fix the game state" commands live). This would reduce accidental derezzes (partially mitigating the rerez issue) and help new players who sometimes are confused about whether the corp can voluntarily derez cards.

### LINKED ISSUE #6266: Command to reveal hand
Would be great to have a command like `reveal-hq` and `reveal-grip` so that we can see each other hand, similar to how Engram's ability is done. Great for teaching, or just out of curiosity post-game.

### LINKED ISSUE #6358: Refresh-ID command for dev/testing
I'm not sure if something like this exists already, but I've noticed that troubleshooting issues with identities seems a lot more cumbersome than other cards. 

I would like to see something like 'refresh-id' as a command, which just reloads the identity card.

NBKelly:
I implemented a few commands mentioned in issues. Specifically:
* `rez-free`: rezzes a card in a disabled state, then enables it. This lets you fix gamestate with cards that have on-rez effects, like spin doctor
* `reveal-hand`: reveals your hand in the game log. Does not trigger reveal effects in game.

reveal-hand-01_000

* `reload-id`: reloads the currently active identity. This is primarily for dev/testing, but sometimes it can fix mirrormorph.
* `derez`: I'm surprised this didn't exist as a command to start with.

Worth discussion: should `rez-free` just replace the `rez` command?

Closes #6358, closes #6266.

Maybe closes #6256 - do we want to keep the derez button on cards?

### DIFF
--- a/src/clj/game/core/commands.clj
+++ b/src/clj/game/core/commands.clj
@@ -20,7 +20,7 @@
    [game.core.prompts :refer [show-prompt]]
    [game.core.props :refer [set-prop]]
    [game.core.psi :refer [psi-game]]
-   [game.core.rezzing :refer [rez]]
+   [game.core.rezzing :refer [rez derez]]
    [game.core.runs :refer [end-run get-current-encounter jack-out]]
    [game.core.say :refer [system-msg system-say unsafe-say]]
    [game.core.set-up :refer [build-card]]
@@ -292,6 +292,16 @@
                    nil nil))}
       nil nil)))
 
+(defn command-derez
+  [state side]
+  (when (= :corp side)
+    (resolve-ability
+     state side
+     {:prompt "Choose a card to derez"
+      :choices {:card #(rezzed? %)}
+      :effect (effect (derez target))}
+     nil nil)))
+
 (defn command-trash
   [state side]
   (let [f (if (= :corp side) corp? runner?)]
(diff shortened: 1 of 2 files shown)


NBKelly:
Fair enough - I also like the convenience of the derez buttons. I do wonder if it's worth putting a confirmation menu on them, like
* Click derez on Spin Doctor
* Prompt: Derez Spin Doctor? [yes/yes, also refund me/no]

I think that would tackle the original usability case while still having the discoverability/convenience.

For now, I've changed `reveal-hand` to `show-hand`, and I noted the `/host` command had no documentation so I added it. (the `/error` command also has no documentation, but I can't see any possible use for it, so I didn't add documentation for it.)

Dan-Garwood:
My argument for removing Derez from the menu is that it's one of the only menu commands that is always illegal (except to fix gamestate), and no other menu commands are there to fix gamestate issues, AFAIK. Thus, the discoverability should be the same as the other slash-commands that are used to fix gamestate, unless there's a very good reason for derezzing to be more easily discovered than any of the other ways of fixing the gamestate. The difference in discoverability creates natural categorization: things on buttons are normal game actions; things in slash-commands are ways to fix a broken game. In my experience, /undo-click, /undo-turn, /counters n, etc. all get used far more than the derez button.

Bad-Epsilon:
My argument for removing Derez from the menu is that it's one of the only menu commands that is always illegal (except to fix gamestate), and no other menu commands are there to fix gamestate issues, AFAIK. Thus, the discoverability should be the same as the other slash-commands that are used to fix gamestate, unless there's a very good reason for derezzing to be more easily discovered than any of the other ways of fixing the gamestate. The difference in discoverability creates natural categorization: things on buttons are normal game actions; things in slash-commands are ways to fix a broken game. In my experience, /undo-click, /undo-turn, /counters n, etc. all get used far more than the derez button.

Issue #6361, opened 2022-05-10
Deepdive into Bacterial programming freezes up game


Zerothmaxima:
After the corp programs their bacterials, the runner gets an option button to resumer run and a greyed out button. Neither do anything and /close-prompt does not help. 

I assume there is a conflict with bacterial ordering R&D and DD shuffling?

NBKelly:
Duplicate of #6324

Issue #6324, opened 2022-04-04
Deep Dive + Bacterial Programming bug


groenkaaf:
as runner, i played deep dive vs saraswati. accessed bact prog, corp resolved, then i spent a click to access a snare, which i trashed. we both saw a button that said "Continue to No current run." couldn't use /close-prompt. game pretty much locked up. not sure what would have happened if i had undone the click as my opponent conceded.

francescopellegrini:
Fixed by #6362?

PR #6362, opened 2022-05-11
Set aside implementation

NBKelly:
I'm making another pull request for this because github appears to be doing something funny with the other one.

I thought the set-aside box wasn't working for one side, but it turned out to actually be one of my browsers caching content so it didn't appear. 

Here's the state of set-aside right now:
* Everything is face-up (the current rules say everything in set-aside is face-up unless otherwise specified - we don't need to worry about it yet)
* It works, and things are removed/returned to the deck (it fixes the bacterial interaction)

Gachapon no longer displays a 'waiting' prompt for the corp. I'm not sure why.

In the future, I think it would be worth implementing tags for visibility, so we can set-aside cards and specify who is allowed to see them face-up.


NBKelly:
Closes #6324, Closes #6361

NBKelly:
I managed to figure out a solution/technique for setting visibility on set-aside cards.

My solution is:
* When cards are set-aside, we set `corp-can-see` and `runner-can-see` keys in a special `set-aside-visibility` map for the card.
* In `card.cljc`, a card is `set-aside-visible` if it is set-aside, and can be seen by the input side
* By default, cards will be visible to both sides (this is the behavior defined in the rules doc)

As a proof of concept, here's deep dive with the four visibility options:

set-aside-vis-03
set-aside-vis-02
set-aside-vis-01
set-aside-double-vis

The exact implementation could probably be tidier, but I think the concept itself seems reasonable.

src/cljc/game/core/card.cljc
```diff
@@ -144,6 +144,20 @@
...
+  "Checks if the specific card is in a set-aside area."
+  [card]
+  (= (get-zone card) [:set-aside]))
+
+(defn set-aside-visible?
+  "Checks if the specific card is in set aside and visible to this side"
+  [card side]
+  (and (in-set-aside? card)
+       (if (= :corp side)
+         (true? (:corp-can-see (:set-aside-visibility card)))
```
NoahTheDuke, line 157: Don't need true?

test/clj/game/cards/hardware_test.clj
```diff
@@ -1845,7 +1846,9 @@
       (is (= 3 (count (:deck (get-runner)))) "3 cards remain in deck")
       (is (= 2 (count (:rfg (get-runner)))) "Removed 2 cards from game")
       (is (no-prompt? state :runner) "No more prompts")
-      (is (no-prompt? state :corp) "Waiting prompt cleared")))
+      (is (no-prompt? state :corp) "Waiting prompt cleared")
+      (is false "boop")
```
NoahTheDuke, line 1850: boop

NBKelly:
I eventually figured out why the gachapon test was failing - the set-aside function was calling effect-completed on the eid given. I believe it shouldn't, but if the correct approach is that we should be making a new eid each time we call set-aside, I can adapt it back to that too.

NBKelly:
I think that covers it for now - I added a couple of extra functions

* `set-aside-for-me` - puts the cards in set aside visible only to their owners (this will save time later)
* `add-to-set-aside` - adds a card into a group of cards already set-aside (given you know the eid)
* `swap-set-aside-cards` - handles swaps where one of the cards (or both) were in set-aside

Additionally, each card in set-aside keeps track of the eid that put it there - this is so you can refer to a card to figure out which group of set-aside cards it belongs to. This is enough so that (later) we could work on changing draw behaviors to use set-aside as in the CR.

PR #6363, opened 2022-05-12
Bulk bugfixes 2

### LINKED ISSUE #3714: Mwanza and Shiro (different problem)
Issue 1: Upon firing of second Shiro sub mandating access, all access effects are fired at once and the chat log shows both players what cards were accessed regardless of access effect or not.
Issue 2: Credits are not granted to Corp following the end of this access.
Issue 3: Upon successful run and access of R&D following a Shiro access, the runner accesses an additional 3 cards on top of the  extra 3 mandated by Mwanza.

### LINKED ISSUE #4455: Aumakua nolonger works with Gang Sign
If no cards are trashed turtle does not get a counter from Gang Sign triggers. Is this a rule change?

### LINKED ISSUE #5616: Shiro + Mwanzo gives way too many accesses and credits
If the runner is forced to access R&D multiple times with Mwanza Grid installed and Shiro's second sub, then jnet gives way too many accesses on the 2nd or later access. More specifically, it gives 1+3n accesses on the nth time that they access (4, 7, 10, 13...). Similarly, the payout in credits is way too much, yet different. It instead gives the sum of (2+6n) from 1 to n (8, 22, 42, 68...). To reproduce, simply rez a Mwanza on R&D and fire the Shiro second sub as however many times during a single run. It becomes relevant in normal play when using cards like Ganked and Marcus Batty to fire the subroutine multiple times.

### LINKED ISSUE #6253: Using Salsette Slums to rfg Mwanza City Grid results on a non-trigger of the 'gain credits' component
Runner has Salsette Slums installed. Corp has Mwanza CIty Grid installed on HQ.
Runner runs HQ, accesses Mwanza City Grid first, and removes it from the game with Salsette Slums.
Additional accesses on HQ are triggered successfully.
The 'gain 2 credits for every card accessed' clause does not trigger.
According to the rules team, both effects should trigger.

### LINKED ISSUE #6355: Divide and Conquer not triggering Aumakua.
When you play divide and conquer, you should get one counter on Aumakua for each server you breach that you don't steal or trash from but atm if you steal or trash something from any server you get no counters.

NBKelly:
Going through a few more old issues.
* Lewi/lovegood was fixed at some point, so I added a test.
* I added an implementation for Zed 1.0/2.0 that crawls through the run log, and finds out if any break abilities that cost :lose-click have been used. The implementation looks ugly, I would appreciate advice on how to make it nicer.
* I wanted to make aumakua better, so I tried using `:end-breach-server`, only to find that it tracked literally nothing except the server that was breached.  So I added some state tracking for each breach, and that information is returned with `:end-breach-server`. Right now, aumakua should work with divide and conquer, gang sign, kitsune, shiro, etc, properly now. I'm going to add some tests for this soon. Once that's done, this closes #6355 and closes #4455
* I expanded breach tracking enough that we can fix mwanza city grid (closes #5616, closes #6253, closes #3714)

### DIFF
--- a/src/clj/game/cards/ice.clj
+++ b/src/clj/game/cards/ice.clj
@@ -308,7 +308,8 @@
 (defn spent-click-to-break-sub
   "Checks if the runner has spent(lost) a click to break a subroutine this run"
   [state run]
-  (let [events (:events run)
+  (let [all-cards (get-all-cards state)
+        events (:events run)
         breaks (filter #(= :subroutines-broken (first %)) events)
         cards (vec (map #(first (second %)) breaks))
         subs (map #(:subroutines %) cards)
@@ -321,9 +322,9 @@
         ;; If Adjusted Matrix ever gets correctly implemented, there will be
         ;; a minor edge case here if the runner uses a non-click break ab on a
         ;; card hosting adjusted matrix -nbkelly, 2022
-        actual-breakers (map #(find-cid % (get-all-cards state)) breakers)
-        abs (map #(if (= (:side %) "Runner") (:abilities %) (:runner-abilities %)) actual-breakers)
-        costs (map :break-cost (apply concat abs))
+        actual-breakers (map #(find-cid % all-cards) breakers)
+        abs (mapcat #(if (= (:side %) "Runner") (:abilities %) (:runner-abilities %)) actual-breakers)
+        costs (map :break-cost abs)
         clicks (filter #(some #{:lose-click} %) costs)]
     (not-empty clicks)))
 
@@ -3585,20 +3586,26 @@
 
 (defcard "Zed 1.0"
   {:subroutines [{:label "Do 1 brain damage"
+                  :async true
                   :effect (req (if (spent-click-to-break-sub state run)
-                                 (resolve-ability state side (do-brain-damage 1) card nil)
-                                 (system-msg state side "does not do brain damage with Zed 1.0")))}
+                                 (continue-ability state side (do-brain-damage 1) card nil)
+                                 (do (system-msg state side "does not do brain damage with Zed 1.0")
+                                     (effect-completed state side eid))))}
                  {:label "Do 1 brain damage"
+                  :async true
                   :effect (req (if (spent-click-to-break-sub state run)
-                                 (resolve-ability state side (do-brain-damage 1) card nil)
-                                 (system-msg state side "does not do brain damage with Zed 1.0")))}]
+                                 (continue-ability state side (do-brain-damage 1) card nil)
+                                 (do (system-msg state side "does not do brain damage with Zed 1.0")
+                                     (effect-completed state side eid))))}]
    :runner-abilities [(bioroid-break 1 1)]})
 
 (defcard "Zed 2.0"
   {:subroutines [trash-hardware-sub
                  trash-hardware-sub
-                 {:label "Do 2 brain damage"
+                 {:label "Do 1 brain damage"
+                  :async true
                   :effect (req (if (spent-click-to-break-sub state run)
-                                 (resolve-ability state side (do-brain-damage 2) card nil)
-                                 (system-msg state side "does not do brain damage with Zed 2.0")))}]
+                                 (continue-ability state side (do-brain-damage 2) card nil)
+                                 (do (system-msg state side "does not do brain damage with Zed 2.0")
+                                     (effect-completed state side eid))))}]
    :runner-abilities [(bioroid-break 2 2)]})
--- a/src/clj/game/cards/ice.clj
+++ b/src/clj/game/cards/ice.clj
@@ -313,7 +313,7 @@
         cards (vec (map #(first (second %)) breaks))
         subs (map #(:subroutines %) cards)
         broken-subs (map #(filter :broken %) subs)
-        breakers (flatten (map #(map :breaker %) broken-subs))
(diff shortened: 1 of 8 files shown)


src/clj/game/cards/programs.clj
```diff
@@ -424,10 +424,9 @@
                                  :msg "manually place a virus counter on itself"
                                  :effect (effect (add-counter card :virus 1))}]
                     :strength-bonus (req (get-virus-counters state card))
-                    :events [{:event :run-ends
-                              :req (req (and (not (or (:did-trash target)
-                                                      (:did-steal target)))
-                                             (:did-access target)))
+                    :events [{:event :end-breach-server
+                              :req (req (and (not (or (:did-steal target)
```
NoahTheDuke, line 428: Unnecessary `and` here.

src/clj/game/cards/ice.clj
```diff
@@ -304,6 +304,29 @@
...
+;;; For Zed 1.0 and 2.0
+(defn spent-click-to-break-sub
+  "Checks if the runner has spent(lost) a click to break a subroutine this run"
+  [state run]
+  (let [events (:events run)
+        breaks (filter #(= :subroutines-broken (first %)) events)
+        cards (vec (map #(first (second %)) breaks))
+        subs (map #(:subroutines %) cards)
+        broken-subs (map #(filter :broken %) subs)
+        breakers (flatten (map #(map :breaker %) broken-subs))
```
NoahTheDuke, line 316:
```suggestion
        breakers (mapcat #(map :breaker %) broken-subs)
```

src/clj/game/cards/ice.clj
```diff
@@ -304,6 +304,29 @@
...
+        ;; this is the list of every breaker the runner used to
+        ;; break a subroutine this run. If we check that it has a
+        ;; 'lose-click' break ability, we should be safe most of the
+        ;; time. This can only be: adjusted matrix, ABR, and corp cards
+        ;; If Adjusted Matrix ever gets correctly implemented, there will be
+        ;; a minor edge case here if the runner uses a non-click break ab on a
+        ;; card hosting adjusted matrix -nbkelly, 2022
+        actual-breakers (map #(find-cid % (get-all-cards state)) breakers)
```
NoahTheDuke, line 324: maybe something like `(let [all-cards (get-all-cards state)] (map #(find-cid % all-cards) breakers))` so we're only calling `get-all-cards` once instead of once per breaker.

src/clj/game/cards/ice.clj
```diff
@@ -304,6 +304,29 @@
...
+        abs (map #(if (= (:side %) "Runner") (:abilities %) (:runner-abilities %)) actual-breakers)
+        costs (map :break-cost (apply concat abs))
```
NoahTheDuke, line 326:
```suggestion
        abs (mapcat #(if (= (:side %) "Runner") (:abilities %) (:runner-abilities %)) actual-breakers)
        costs (map :break-cost abs)
```

src/clj/game/cards/ice.clj
```diff
@@ -3561,14 +3584,21 @@
                  (do-net-damage 1)]})
 
 (defcard "Zed 1.0"
-  {:implementation "Restriction on having spent [click] is not implemented"
-   :subroutines [(do-brain-damage 1)
-                 (do-brain-damage 1)]
+  {:subroutines [{:label "Do 1 brain damage"
+                  :effect (req (if (spent-click-to-break-sub state run)
+                                 (resolve-ability state side (do-brain-damage 1) card nil)
+                                 (system-msg state side "does not do brain damage with Zed 1.0")))}
```
NoahTheDuke, line 3590:
```suggestion
                  :async true
                  :effect (req (if (spent-click-to-break-sub state run)
                                 (continue-ability state side (do-brain-damage 1) card nil)
                                 (do (system-msg state side "does not do brain damage with Zed 1.0")
                                     (effect-completed state side eid))}
```
Something like this, you'll probably have to clean it up. `do-brain-damage` is an `:async` ability, so it has to be resolved with `continue-ability` and the calling ability has to be `:async` too.

NoahTheDuke: Oh yeah, and this applies to the other Zed 1 and 2 subroutines.

NoahTheDuke:
Lookin good, nbkelly! thanks for doing all this work.

NBKelly:
I think I tackled all of the suggestions. I also added `from-server` back into `end-breach` (I accidentally took it out before).

NBKelly:
The one above should fix all the instances where a single mwanza forces +3/+6/+9... accesses from triggering multiple times. 
I think for the economic ability, we can probably tackle all #5616, #6253 and #3714 by having the 'gain-credits' ability be registered when breaching, instead of just being there forever.
This would also require each breach tracking the cards accessed (which is next on my list of things to do).

NBKelly:
The errata text for turtle now states "whenever a breach ends and...", so it's actually incorrect to add counters for accesses which do not breach a server (Such as otoroshi, or Quest Complete).

I think turtle is 100% correctly implemented now, and any issues leftover would happen to be cards which shouldn't breach a server but do, or cards that do but shouldn't.

I've added tests for turtle vs. gang sign, divide & conquer. I'll be done with this PR once I've added tests for mwanza.

NBKelly:
I think that caps out all the changes for this one :)

Issue #6339, opened 2022-04-24
Pelangi lasts until end of run, not end of encounter


nishadham:
Pelangi's card text says it should last until end of encounter, but J.net makes it last until end of run. Issues with multiple encounters per run e.g. Ganked!, Letheia Nisei... 
<img width="431" alt="Screenshot 2022-04-24 at 14 55 19" src="https://user-images.githubusercontent.com/58484015/164979938-448af9e0-e0a0-4cca-9355-d063b40c27fa.png">

francescopellegrini:
Fixed by #6364

PR #6364, opened 2022-05-14
Bulk bugfixes 3

NBKelly:
Another batch of fixes.
* Flame out only creates on trigger now/doesn't quash other triggers that are applied to flame out
* Pelangi last for the encounter instead of the whole run
* Bioroid break, f2p and negotiator now require the runner is encountering them for their abilities to be used (so you can't use any random bioroid ability to break whatever ice you're encountering)
* The second subroutine on shiro now has a label


NBKelly:
Closes #6339 (I got the wrong one in the commit message, but it also gets closed)

Issue #6237, opened 2022-01-29
Baba Yaga is copying restrictions on cards (e.g. Passport)


lostgeek:
Situation: Baba Yaga with a hosted Passport.
Expected Behaviour: The copied ability from Passport on Baba Yaga can be used to break ice on remotes.
Actual Behaviour: The ability only works on centrals.

Since we copy over the entire ability, we also copy the `:req` condition that is used to implement the central-only restriction on Passport. A potential solution might look like this (coming from discussions with Noah):

- Add a constant-ability function in the card `:cannot-use-ability`
- Add a constant ability check in `play-ability` like `(not (any-constant-ability? ... :cannot-use-ability))`
- Remove `:req` from any card using it for this reason and convert it into a constant ability.
- Check in the diff generation for how "usable" cards were communicated and adjust it as necessary

NBKelly:
Note to myself for later: Since there's only a tiny number of relevant cards, it might be reasonable to have the `req` fns for these cards assert their card name for the restrictions.

PR #6366, opened 2022-05-20
WIP/proof of concept - Cannot use ability

NBKelly:
See #6237 - A flat 'cannot-use-ability' applying to the whole card didn't seem reasonable, because you should be able to use abilities on the cards that aren't interfacing (ie you can boost strength while running remotes if you needed to tank your credits, or spend credits for net mercur). We can't put the `cannot-use-ability` fn in the ability itself either, because then it gets copied by baba yaga (or maybe yaga should strip that?).

I found a sort of compromise was placing the :cannot-use-ability key in the card, and referring to it in the :req function of the ability, like:
`
:req (req (not (:cannot-use-ability ...)))
`
This means that cards which copy the abilities will end up getting nil when trying to call that, which should always be true. 

Maybe a better option would be actually putting a specific key in with the card, like:

`
:cannot-use-ability {:central-breaker (req ...)}
`

### DIFF
--- a/src/clj/game/cards/programs.clj
+++ b/src/clj/game/cards/programs.clj
@@ -232,11 +232,29 @@
   (Central suite: Alias, Breach, Passport)"
   [break pump]
   (auto-icebreaker
-    {:abilities [(break-sub (:break-cost break) (:break break) (:breaks break)
-                            {:req (req (and (#{:hq :rd :archives} (target-server run))
-                                            (<= (get-strength current-ice) (get-strength card))
-                                            (has-subtype? current-ice (:breaks break))))})
-                 pump]}))
+   {:cannot-use-ability (req
+                         ;(system-msg state side (str "cannot-use check"))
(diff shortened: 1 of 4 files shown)


NoahTheDuke:
Sorry I've not reviewed this yet. I think this can be accomplished with a `:constant-effects`. I have a branch that changes what we pass to `get-effects` and the like to make this easier. The goal is that you could write the effect like `:req (req (and (same-card? card (:icebreaker context)) (protecting-central? (:ice context))))`, and then the interface abilities can be freely copied.

NoahTheDuke:
Any updates on this?

NoahTheDuke:
Looks like I mentioned a branch, I don't think I ever pushed that branch or fully developed it, I'll try to get to that at some point.

NoahTheDuke:
This is a good idea but it should be possible to track through the existing `:static-abilities`. Closing for inactivity and a potentially better solution.

PR #6395, opened 2022-06-17
Draws are set aside

### LINKED ISSUE #6369: Stricter rules enforcement for drawing cards
According to [CR 8.4](https://nisei.net/wp-content/uploads/2021/10/NISEI-Comprehensive-Rules-v1.5-clean.pdf#h.egppjt2pqyi), cards should be in the set-aside zone (facedown) during the act of drawing - this is when abilities which act on the drawn cards can fire - and then they should be moved to the hand.

Once set-aside is merged, we should be able to work on this. In most cases, it should be unnoticeable - but it should make cards like Daily Business Show and The Class Act more intuitive/straightforward to use, from a usability standpoint.

This has implications for a few cards - any ability which forces the runner to play with a revealed grip (ie harischandra) should not be able to see cards in the set-aside zone (and as a consequence, may not make inferences about what the runner chose not to draw). I believe the card moved to the bottom should also not be visible with find the truth (unless the runner chooses to order it the other way, which may actually have strategic benefits).

NBKelly:
I implemented the draws-pass-through-set-aside rule in CR 8.4, using the open set-aside PR as a base.

The only cards that needed explicit changes were:
* Mr. Li (he also had an existing bug where his effect would linger if the corp prevented draws, which is now fixed)
* Raman Rai (To swap the card into/out of the set-aside zone)
* Bug (to search for CIDs in the set-aside cards, rather than the hand)

Then about 80 or so tests needed to be changed to either
1) count cards in the set-aside zone, or
2) click on cards in the set-aside zone, rather than the hand

Closes #6369


NBKelly (commented):
Ignore the goofy commit message

PR #6398, opened 2022-06-21
Behaviour for refusing to access a card

### LINKED ISSUE #6392: Aeneas Informant gives a credit in Gagarin even when runner doesn't pay to access
Set Aeneas Informant to autoresolve
Run Gagarin remote
Don't pay to access the remote (it was Overseer Matrix and rezzed before access)
Aeneas gives a credit

NBKelly:
Previously, refusing to access a card (ie with gag) would still call "access-end" with that card as an argument.

This enabled maw and aeneas informant to work on gag cards that were not accessed (and also incremented cards accessed, enabled 'accessed but not stolen/trashed' etc conditions, and so on).

The simple fix is to just not to trigger these events when we don't access the card.

Closes #6392

### DIFF
--- a/src/clj/game/core/access.clj
+++ b/src/clj/game/core/access.clj
@@ -358,6 +358,12 @@
   [state side]
   (merge-costs (get-in @state [:bonus :access-cost])))
 
+(defn- refused-access-cost
+  "The runner refused to pay (or could not pay) to access"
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
Looks great, thanks for solving this.

PR #6400, opened 2022-06-22
Fix wrong username validation check

PietroCarrara:
It was mistyped as a username having to be, at the same time, greater than AND smaller than 20 characters.

Since `valid-username?` checks for both size and invalid characters, I've changed the message to a more generic one.


PietroCarrara:
(Still trying to figure out some errors related to testing inside Docker, so this is still a WIP)

PR #6402, opened 2022-06-24
Surround calls to "import-vars" with an anonymous function

### LINKED ISSUE #6397: Build Fails (IndexOutOfBoundsException: Method code too large)
Steps to reproduce:
- Make a fresh clone of the project
- `lein fetch`

Expected behavior: Code is compiled and fetch is run

Observed behavior: Compilation fails
```
Compiling web.core
Syntax error macroexpanding at (agendas.clj:1:1).
Execution error (IndexOutOfBoundsException) at clojure.asm.MethodWriter/computeMethodInfoSize (MethodWriter.java:2061).
Method code too large!

Full report at:
/tmp/clojure-7174747273417157914.edn
Compilation failed: Subprocess failed (exit code: 1)
```
Running `Arch Linux 5.18.3-arch1-1` from looking at circle-ci, this seems to be an issue since 639ff9265d27263f3b034ac5a6ab8bb317dae6c0.

Since I have 0 clojure experience, I have no idea what this error means or a useful way to fix it, but for now disabling AOT compilation seems to fix it:
[project.clj](../blob/master/project.clj)
```clojure
(defproject netrunner "1.0"
  :description "Browser implementation of Android: Netrunner card game."
  :url "https://github.com/mtgred/netrunner"
  :license {:name "The MIT License (MIT)"
            :url "https://opensource.org/licenses/MIT"}
  :min-lein-version "2.7.1"

  :source-paths ["src/clj" "src/cljs/nr" "src/cljc"]
  :test-paths ["test/clj"]

  ;;    :aot [web.core]
  :jar-name "netrunner.jar"
  :jar-exclusions [#"public/img/cards/*"]
  :uberjar-name "netrunner-standalone.jar"
  :uberjar-exclusions [#"public/img/cards/*"]
  :omit-source true
  :main web.core

  ...
```

PietroCarrara:
This causes the compiler to put each call in a separate method. If it tried to put them all in one single method, it would grow beyond the JVM's maximum of 64kb. Fixes #6397. This should also get circle-ci working again.

This is my first time coding clojure, so if you don't feel this is a good solution or spot anything weird, please do point it out!

Steps to reproduce:

- Make a fresh clone of the repo
- Switch to this branch (`fix-build`)
- `$ lein repl`

Old behaviour: The compilation would fail (as seen in the issue)
New Behaviour: The compilation succeeds (and so do the tests that are supposed to succeed with `$ lein test`)

### DIFF
--- a/src/clj/game/core.clj
+++ b/src/clj/game/core.clj
@@ -63,7 +63,12 @@
     [game.macros]
     [potemkin :refer [import-vars]]))
 
-(import-vars
+(defmacro expose-vars
+  "Surrounds a import-vars call with an anonymous function to get arround the 64kb limit on the method size in the JVM" 
+  [symbols]
+  `((fn ~[] (import-vars ~symbols))))
+
+(expose-vars
   [game.core.access
    access-bonus
    access-bonus-count
@@ -98,7 +103,7 @@
    steal-cost-bonus
    turn-archives-faceup])
 
-(import-vars
+(expose-vars
   [game.core.actions
    advance
    click-advance
@@ -128,19 +133,19 @@
    trash-resource
    view-deck])
 
-(import-vars
+(expose-vars
   [game.core.agendas
    update-advancement-requirement
    update-all-advancement-requirements
    update-all-agenda-points])
 
-(import-vars
+(expose-vars
   [game.core.bad-publicity
    bad-publicity-prevent
    gain-bad-publicity
    lose-bad-publicity])
 
-(import-vars
+(expose-vars
   [game.core.board
    all-active
    all-active-installed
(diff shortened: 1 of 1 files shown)


src/clj/game/core.clj
```diff
@@ -63,7 +63,12 @@
     [game.macros]
     [potemkin :refer [import-vars]]))
 
-(import-vars
+(defmacro expose-vars
+  "Surrounds a import-vars call with an anonymous function to get arround the 64kb limit on the method size in the JVM" 
+  [symbols]
+  `((fn ~[] (import-vars ~symbols))))
```
NoahTheDuke, line 69: I don't think you need the unquote in front of the arg vector, but maybe you do.

PR #6408, opened 2022-07-03
chiyashi automatically trashes cards

### LINKED ISSUE #6404: Eater did not trigger Chiyashi's trash top 2 cards of Runner's stack
When I broke Chiyashi's subs while Eater was installed, it did not trigger the "trash 2 cards from top of stack."

Here's the [log history url](https://jinteki.net/profile/history/075e5fed-0cdf-4e26-848e-15412e36ba45).

NBKelly:
Chiyashi automatically mills cards if there is an installed AI. 

The mills happen 2 at a time (so the interaction with buffer drive will always be correct), even when 2+ subroutines are broken at once.

This seems to work as expected when hippo is used (mills twice, then ice is trashed before the third mill) - not sure if there are any other cases I need to catch.

Closes #6404

### DIFF
--- a/src/clj/game/cards/ice.clj
+++ b/src/clj/game/cards/ice.clj
@@ -873,15 +873,20 @@
    :subroutines [end-the-run]})
 
 (defcard "Chiyashi"
-  {:implementation "Trash effect when using an AI to break is activated manually"
-   :abilities [{:async true
-                :label "Trash the top 2 cards of the Runner's Stack"
-                :req (req (some #(has-subtype? % "AI") (all-active-installed state :runner)))
-                :msg (msg (str "trash " (string/join ", " (map :title (take 2 (:deck runner)))) " from the Runner's Stack"))
-                :effect (effect (mill :corp eid :runner 2))}]
-   :subroutines [(do-net-damage 2)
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/ice.clj
```diff
@@ -873,15 +873,20 @@
...
-                :msg (msg (str "trash " (string/join ", " (map :title (take 2 (:deck runner)))) " from the Runner's Stack"))
-                :effect (effect (mill :corp eid :runner 2))}]
-   :subroutines [(do-net-damage 2)
-                 (do-net-damage 2)
-                 end-the-run]})
+  (letfn [(chiyashi-auto-trash [state side eid n]
+            (if (pos? n)
+              (wait-for (mill state :corp :runner 2)
+                        (system-msg state side "uses Chiyashi to trash the top 2 cards of the Stack")
+                        (chiyashi-auto-trash state side eid (- n 1)))
```
NoahTheDuke, line 880:
```suggestion
                        (chiyashi-auto-trash state side eid (dec n)))
```

NoahTheDuke (approved):

PR #6409, opened 2022-07-03
add label to Mirrormorph ability to fix tests

francescopellegrini:
cc @NBKelly


NoahTheDuke:
Could also remove the `msg` call in `:msg`.

PR #6413, opened 2022-07-06
Permit users to set custom gameboard backgrounds

presheaf:
(Disclaimer - I'm not a frontend person, so if something looks weird, there's probably no good reason for it. :) Please feel free to just tell me if something needs to be changed.)

Custom backgrounds are implemented by letting the user set an custom image URL in a new field under the settings tab, and having the gameboard look up the custom URL if the background is also set to custom. It might be cleaner to have the user upload an image into local storage and use that instead, but as the image fetching happens entirely in the frontend, I couldn't really see a security issue with it, so I decided it was fine as is.

Unfortunately, I haven't actually been able to test this in the master branch, due to some difficulty building a dev environment - seems like the docker images are out of date, and I couldn't fix them following the instructions on the wiki. It ran fine on an older fork, so I'm hopeful it'll work there too, but until I get my dev environment setup I'd appreciate it if someone with a dev environment could try to run it and see if it works.


NoahTheDuke:
Hey @presheaf! I merged the docker update PR so if you want, give this another shot? I know a lot of people would be happy to see this completed.

presheaf:
Sorry about the delay here, but thanks to both for fixing the Docker build - I'll get on this now. :)

presheaf:
I fixed the issues mentioned here and on Slack, making custom backgrounds scale to fill the window I set them to cover, meaning that parts of the image might stick out if the aspect ratio of the gameboard and the image differ. I think it'll look nicer than stretching, which is the other option. Incorporating some suggestions from Noah, I also cleaned up the setting page a little, to only show the input box if needed. It now oks like this:

tmp

PR #6421, opened 2022-07-16
Light the fire triggers on successful run (and lets you breach the empty server)

### LINKED ISSUE #6419: Light the Fire! interaction with Reduced Service
**Reduced Service**
When you rez this upgrade, you may pay up to 4credit to place that many power counters on it.
As an additional cost to run this server, the Runner must pay 2credit for each hosted power counter.
Whenever the Runner makes a successful run on a central server, remove 1 hosted power counter.

**Light the Fire!**
click, trash, suffer 1 core damage: Run a remote server. During that run, cards in the root of the attacked server lose all abilities. When that run is successful, trash all cards in the root of the attacked server.

On Jinteki.net, when running a server protected by Reduced Service with Light the Fire! the runner is not forced to pay the additional tax from Reduced Service, but they should, because it is a cost to run and Light the Fire!'s ability does not come into play until the run begins.

NBKelly:
Closes #6419 - but also importantly, allows you to breach the attacked server (so it combos with Bank Job, SecTesting, Turtle).


src/clj/game/cards/resources.clj
```diff
@@ -1584,12 +1584,13 @@
...
               (enable-card state :corp c)))]
-    (let [breach-ability (successful-run-replace-breach
-                          {:mandatory true
-                           :duration :end-of-run
-                           :ability {:async true
-                                     :msg "trash all cards in the server at no cost"
-                                     :effect (effect (trash-cards eid (:content run-server)))}})
+    (let [successful-run-trigger {:event :successful-run
+                                  :duration :end-of-run
+                                  :mandatory true
```
NoahTheDuke, line 1589: No longer needed.

src/clj/game/cards/resources.clj
```diff
@@ -1603,12 +1603,13 @@
...
-    (let [breach-ability (successful-run-replace-breach
-                          {:mandatory true
-                           :duration :end-of-run
-                           :ability {:async true
-                                     :msg "trash all cards in the server at no cost"
-                                     :effect (effect (trash-cards eid (:content run-server) {:cause-card card}))}})
+    (let [successful-run-trigger {:event :successful-run
+                                  :duration :end-of-run
+                                  :mandatory true
+                                  :async true
```
NoahTheDuke, line 1609:
```suggestion
                                  :async true
```

Issue #5965, opened 2021-06-24
Some player controls displayed to spectators


the-Bruce:
**Overview:**
Certain UI elements that are specific to active players are presented to spectators.
Spectators are presented with +/- buttons when hovering over Corp Agenda Points, both players max hand size (in the extended Grip/HQ panel). Shuffle and Show are displayed when clicking R&D.
These buttons don't appear to do anything when clicked (although due to the risk of disrupting games this has not been extensively tested).
![image](https://user-images.githubusercontent.com/44057980/123301839-fd8e7a00-d513-11eb-937f-6b13da4da330.png)
![image](https://user-images.githubusercontent.com/44057980/123301959-1ac34880-d514-11eb-971c-9c1dc17aa506.png)
![image](https://user-images.githubusercontent.com/44057980/123302103-41817f00-d514-11eb-9ff6-d37933e00cef.png)

**Steps to reproduce:**
1. Spectate a game.
2. Hover over (or click) one of the identified areas

_Expected behaviour:_
No buttons displayed to spectator
_Observed behaviour:_
Buttons Displayed

**Tested on:**
Firefox 78.11.0esr, Chrome 91.0.4472.114 on Linux Debian
Safari on iPadOS 14.6 (iPad Air 3rd gen)

PR #6433, opened 2022-07-21
#5965 Hide all stat-controls from spectator + deck on-click also not …

queueseven:
…allowed for spectators


NoahTheDuke:
Thank you! Make sure you say "Closes X" in either the commit message or the PR message so the related issue will be closed when this gets merged.

Issue #6431, opened 2022-07-20
Flag the runners Mark in the UI (not just chat)


tolaasin:
Provide a visual indication of the current Mark target in the UI, as is done with Azmari.

francescopellegrini:
I implemented it like Azmari or Sec Testing, I hope it works and feels good!

tolaasin:
wonderful! Thank you so much!

PR #6434, opened 2022-07-22
fix mark triggering multiple times

### LINKED ISSUE #6425: Virtuoso and Sable triggers the mark selection twice.
ChrisFerg started their turn 12 with 12  and 5 cards in their Grip.
ChrisFerg identifies their mark to be R&D.
ChrisFerg identifies their mark to be R&D.

francescopellegrini:
Fixes #6425 and should address #6431 🤞

### DIFF
--- a/src/clj/game/core/mark.clj
+++ b/src/clj/game/core/mark.clj
@@ -2,6 +2,7 @@
   (:require
     [game.core.say :refer [system-msg]]
     [game.core.servers :refer [central->name]]
+    [game.core.update :refer [update!]]
     [game.macros :refer [req]]))
 
 (defn set-mark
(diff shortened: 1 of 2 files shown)


src/clj/game/core/mark.clj
```diff
@@ -19,5 +20,5 @@
     (system-msg state :runner (str "identifies their mark to be " (central->name new-mark)))))
 
 (def identify-mark-ability
-  {:req (req (nil? (:mark @state)))
```
NoahTheDuke, line 22: I'm only okay with this if the CR says that "Identify Your Mark" always happens but game state changes only if there's no existing mark.

NBKelly: I think that's what the text says for mark.
> When your turn begins, identify your mark. (If you don’t have a mark, a random central server becomes your mark for this turn.)

So identify should always happen, but the gamestate only changes the first time each turn.

NoahTheDuke: Ah yeah, god point. I forgot that's how they worded the reminder text. Cool!

Issue #6436, opened 2022-07-22
Implement commands for keywords introduced in Midnight Sun


locks:
There have been 3 new keywords introduced in the most recent NISEI expansion: [sabotage, mark and charge](https://nisei.net/blog/midnight-sun-mechanics/).

## Tasks
- Implement command for `sabotage`
- Implement command for `mark`
- Implement command for `charge`

francescopellegrini:
Closed by #6439, #6442, and #6445 :)

locks:
oh wow, that was fast 😱

PR #6439, opened 2022-07-23
add charge command

francescopellegrini:
Partial work for #6436

### DIFF
--- a/src/clj/game/core/commands.clj
+++ b/src/clj/game/core/commands.clj
@@ -6,6 +6,7 @@
                            has-subtype? ice? in-hand? installed? map->Card rezzed?
                            runner?]]
    [game.core.change-vals :refer [change]]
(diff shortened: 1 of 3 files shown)


src/clj/game/core/commands.clj
```diff
@@ -348,6 +349,11 @@
                                                                           ": " (get-card state target))))
                                           :choices {:card (fn [t] (same-side? (:side t) %2))}}
                                         (map->Card {:title "/card-info command"}) nil)
+        "/charge"     #(resolve-ability %1 %2
+                                        {:prompt "Choose an installed card to charge"
+                                         :effect (req (charge-card %1 %2 (make-eid state) target))
```
NoahTheDuke, line 354: Should be `:async true` and pass in the `eid` here.

NoahTheDuke (changes requested):

PR #6442, opened 2022-07-23
implement mark command

francescopellegrini:
Additional work for #6436


francescopellegrini:
See also #6439

NoahTheDuke (approved):
Add a test for this and the other command PR?

francescopellegrini:
> Add a test for this and the other command PR?

Do commands have tests? 😅 

I tested them running the Docker compose and creating a new game. :)

NoahTheDuke:
Yeah, there's a commands_test.clj file, should be pretty easy to just verify with a single call that it works.

PR #6445, opened 2022-07-23
implement sabotage command

francescopellegrini:
Final work for #6436

See also #6439 and #6442

### DIFF
--- a/src/clj/game/core/commands.clj
+++ b/src/clj/game/core/commands.clj
@@ -22,6 +22,7 @@
    [game.core.psi :refer [psi-game]]
    [game.core.rezzing :refer [rez derez]]
    [game.core.runs :refer [end-run get-current-encounter jack-out]]
+   [game.core.sabotage :refer [sabotage-ability]]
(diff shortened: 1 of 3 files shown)


src/clj/game/core/commands.clj
```diff
@@ -428,6 +429,12 @@
                                          :choices {:card (fn [t] (same-side? (:side t) %2))}}
                                         (map->Card {:title "/rfg command"}) nil)
         "/roll"       #(command-roll %1 %2 value)
+        "/sabotage"   #(when (= %2 :runner)
+                          (resolve-ability
+                            %1 %2
+                            {:async true
+                             :effect (effect (continue-ability (sabotage-ability (constrain-value value 0 1000)) nil nil))}
```
NoahTheDuke, line 436: This whole ability map could be `(sabotage-ability (constrain-value value 0 1000))`. No need to wrap in the ability map and continue-ability.

NoahTheDuke (changes requested):

PR #6446, opened 2022-07-23
fix Élivágar Bifurcation score effect

### LINKED ISSUE #6443: Élivágar Bifurcation Prompt doesnt close
The prompt for scoring Bifurcation doesnt close when clicking done. The derez ability is a may so I expected this to be the way to "refuse" the trigger. Clicking "done" multiple times didnt remove the window for me either. It has happened every time I score the agenda. This issue has occured to me while playing sportsmetal which also triggers on score so the issue might be related to multiple triggers on score happening.

There is a workaround by just using the /close-prompt command. Hasnt caused any further issues for me after using it.

To replicate the issue you start the game with sportsmetal and just score the agenda normally (install adv adv).

francescopellegrini:
Fixes #6443

### DIFF
--- a/src/clj/game/cards/agendas.clj
+++ b/src/clj/game/cards/agendas.clj
@@ -670,7 +670,6 @@
 (defcard "Élivágar Bifurcation"
   {:on-score
    {:interactive (req true)
-    :async true
     :waiting-prompt "Corp to make a decision"
     :prompt "Choose a card to derez"
     :choices {:card #(rezzed? %)}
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/agendas.clj
```diff
@@ -670,10 +670,12 @@
 (defcard "Élivágar Bifurcation"
   {:on-score
    {:interactive (req true)
+    :async true
```
NoahTheDuke, line 673: Gotta leave this out.

NoahTheDuke (changes requested):

PR #6448, opened 2022-07-24
fix Big Deal not RFG after resolution

francescopellegrini:
Apparently this tiny bit of code was left out :)


NoahTheDuke:
Any reason to not put this directly on the ability map? It should only happen when the corp doesn't cancel the prompt.

francescopellegrini:
> Any reason to not put this directly on the ability map? It should only happen when the corp doesn't cancel the prompt.

Nope, you're totally right, thanks!

PR #6450, opened 2022-07-24
make Extract additional trash cost optional

francescopellegrini:
Reworked to resemble other _may_ abilities.

### DIFF
--- a/src/clj/game/cards/operations.clj
+++ b/src/clj/game/cards/operations.clj
@@ -8,7 +8,7 @@
    [game.core.board :refer [all-active-installed all-installed
                             get-all-installed get-remote-names get-remotes
                             installable-servers server->zone]]
-   [game.core.card :refer [agenda? asset? can-be-advanced? card-index corp?
+   [game.core.card :refer [agenda? asset? can-be-advanced? card-index corp? corp-installable-type?
                            event? facedown? faceup? get-advancement-requirement
                            get-card get-counters get-zone hardware? has-subtype? ice? identity? in-discard?
                            in-hand? installed? is-type? operation? program? resource? rezzed? runner?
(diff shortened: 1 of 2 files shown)


NoahTheDuke (changes requested):
If we're moving these implicit optional to explicit, they should have reqs in the :optional map to make sure they can be used.

francescopellegrini:
> If we're moving these implicit optional to explicit, they should have reqs in the :optional map to make sure they can be used.

In the end I refactored it to resemble Lateral Growth. I think it's way more readable now

src/clj/game/cards/operations.clj
```diff
@@ -1322,13 +1319,15 @@
     :effect (req (wait-for (gain-credits state side 4)
                            (continue-ability
                              state side
-                             {:player :corp
-                              :prompt "Choose a card to install"
+                             {:prompt "Choose a card to install"
+                              :req (req (not-empty (filter #(corp-installable-type? %) (:hand corp))))
```
NoahTheDuke, line 1323:
```suggestion
                              :req (req (not-empty (filter corp-installable-type? (:hand corp))))
```

src/clj/game/cards/operations.clj
```diff
@@ -756,24 +756,21 @@
...
-                                              (gain-credits state side eid 3)))}
-                      card nil))
-                   ;; no cards to trash -> skip prompt, no info is given away
-                   (do (system-msg state side "uses extract to gain 6 [credit]")
-                       (gain-credits state side eid 6))))}})
+    :msg "gain 6 [Credit]"
+    :effect (req (wait-for (gain-credits state side 6)
+                           (continue-ability
+                             state side
+                             {:prompt "Choose an installed card to trash"
```
NoahTheDuke, line 763: Maybe add a `:waiting-prompt` too?

NoahTheDuke (changes requested):

PR #6452, opened 2022-07-24
Small fixes to messages

francescopellegrini:
Add "decline" messages to improve clarity, fix Regenesis typo, remove leftover `println` from test


NoahTheDuke:
I gotta fix cancel-effect lol. It's awkward to use. Thanks for cleaning these up.

PR #6456, opened 2022-07-24
fix Azef Protocol pending test

francescopellegrini:
Figured out why it was failing :)

### DIFF
--- a/src/clj/game/core/costs.clj
+++ b/src/clj/game/core/costs.clj
@@ -366,7 +366,7 @@
      :choices {:all true
                :max (value cost)
                :card #(and (installed? %)
-                           (not (same-card? % card))
+                           (not (partial same-card? % card))
                            (if (= side :runner)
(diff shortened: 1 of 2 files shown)


src/clj/game/core/costs.clj
```diff
@@ -366,7 +366,7 @@
      :choices {:all true
                :max (value cost)
                :card #(and (installed? %)
-                           (not (same-card? % card))
+                           (not (partial same-card? % card))
```
NoahTheDuke, line 369: I don't know what you think this does, but it's not correct lol. `partial` returns a function, so this returns `false` always. What is the goal?

NoahTheDuke (changes requested):

src/clj/game/core/costs.clj
francescopellegrini: 🤦

PR #6457, opened 2022-07-24
Use :not-self key

francescopellegrini:
Also let Chop Bot 3000 trigger like Aesop's Pawnshop at the beginning of the Runner's turn.


NoahTheDuke (approved):
Love love love this change.

PR #6458, opened 2022-07-24
fix Wave not revealing the fetched card

### LINKED ISSUE #6455: Wave (Midnight Sun #37) on-rez effect does not trigger Hyoubu Institute's (Downfall #39) ability.
image

When [Wave](https://netrunnerdb.com/en/card/33037) is rezzed during a run, it searches R&D for a piece of ICE and reveals it, then adds it to HQ. This all fuctions correctly. 

However, if the Corp ID is [Hyoubu Institute](https://netrunnerdb.com/en/card/26039), this reveal should trigger the ID ability. It does not.

francescopellegrini:
Fixes #6455

### DIFF
--- a/src/clj/game/cards/ice.clj
+++ b/src/clj/game/cards/ice.clj
@@ -3697,6 +3697,7 @@
                              :cancel-effect (effect (system-msg "uses Wave to shuffle R&D")
                                                     (shuffle! :deck))
                              :effect (effect (shuffle! :deck)
+                                             (reveal eid target)
                                              (move target :hand))}
                :no-ability {:msg "decline to search for a piece of ice"}}}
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/ice.clj
```diff
@@ -3697,6 +3697,7 @@
                              :cancel-effect (effect (system-msg "uses Wave to shuffle R&D")
                                                     (shuffle! :deck))
                              :effect (effect (shuffle! :deck)
+                                             (reveal eid target)
```
NoahTheDuke, line 3700: Gotta be a `wait-for`, add `:async true` to the ability map, and then add `effect-completed` after `move`

NoahTheDuke (changes requested):

PR #6465, opened 2022-07-25
Fix patchwork ability counting for "spending credits from installed cards" effects

### LINKED ISSUE #6462: Newly installed The Twinning incorrectly does not trigger if Patchwork is used earlier in the turn
Reproduction steps:

1. Have Patchwork and Mystic Maemi in play, with a credit on Maemi.
2. Install The Twinning from the grip, using Patchwork's credit reduction ability.
3. Play an event using Maemi, as your first spend of a credit from an installed card - The Twinning does not get a counter.

Twinning also gets a counter if the Patchwork ability is used while it is in play, which is not correct.

image

It seems very possible that other sources of cost reduction that use the "hosted credit" interface are broken as well if this is a bug.

artemmikhalitsin:
Fixes #6462 

Since patchwork is the only card that works this way, I felt that this solution is a little easier than implementing a whole :reduce-cost interaction. Please let me know if I need to make any changes.

### DIFF
--- a/src/clj/game/core/pick_counters.clj
+++ b/src/clj/game/core/pick_counters.clj
@@ -81,6 +81,11 @@
   (req (update! state side (assoc-in card [:counter counter-type] (dec (get-counters card counter-type))))
        (complete-with-result state side eid 1)))
 
+(defn- not-cost-reduction-payment-cards
+  "Returns cards which do not have the :cost-reduction tag"
+  [cards]
+  (filter #(not (-> (card-def %) :interactions :pay-credits :cost-reduction)) cards))
+
 (defn pick-credit-providing-cards
   "Similar to pick-virus-counters-to-spend. Works on :recurring and normal credits."
   ([provider-func outereid] (pick-credit-providing-cards provider-func outereid nil 0 (hash-map)))
@@ -111,7 +116,7 @@
                                            (when (and card-strs remainder-str)
                                              " from their credit pool"))]
                           (lose state side :credit remainder)
-                          (let [cards (map :card (vals selected-cards))]
+                          (let [cards (not-cost-reduction-payment-cards (map :card (vals selected-cards)))]
                             (wait-for (trigger-spend-credits-from-cards state side cards)
                                       ; Now we trigger all of the :counter-added events we'd neglected previously
                                       (pick-counter-triggers state side eid selected-cards selected-cards target-count message))))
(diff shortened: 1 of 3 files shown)


test/clj/game/cards/hardware_test.clj
```diff
@@ -3375,13 +3375,27 @@
       (new-game {:runner {:deck ["Patchwork" "Sure Gamble" "Film Critic"]}})
       (take-credits state :corp)
       (play-from-hand state :runner "Patchwork")
-      (is (= 1 (:credit (get-runner))) "Runner has 1 credit")
```
artemmikhalitsin, line 3378: Oops this was an accident will fix in the morning

NoahTheDuke:
Hmm this is an interesting fix. I don't like flags as a general rule, prefer to write the better abstraction. But I also think that to write the correct version would take an undue amount of code.

This works as a stopgap for now, but we should definitely spend some hammock time thinking about the best way to structure the code.

src/clj/game/core/pick_counters.clj
```diff
@@ -111,7 +116,7 @@
                                            (when (and card-strs remainder-str)
                                              " from their credit pool"))]
                           (lose state side :credit remainder)
-                          (let [cards (map :card (vals selected-cards))]
+                          (let [cards (not-cost-reduction-payment-cards (map :card (vals selected-cards)))]
```
NoahTheDuke, line 119: I think I would prefer to inline this function call: `(->> (vals selected-cards) (map :card) (remove #(-> (card-def %) :interactions :pay-credits :cost-reductions)))`. It's not being used anywhere else and the name is almost the same length as the implementation.

artemmikhalitsin: Thanks. I scoped it into a function because it seemed hard to read but `->>` definitely makes this much easier to understand

artemmikhalitsin:
I suppose the proper way of doing this is to rework patchwork as a "cost-reducing" effect with a prompt when the runner is about to play a card, but prior to selecting any  hosted credits. This would also mean also including any cost-reducing effects when calculating whether the card is playable. Since patchwork is so far the only card with this sort of "active cost-reducing" ability, I felt that it was probably not worth the time.

NoahTheDuke:
Exactly my thought too. All other cost reducing abilities are static and there's nothing to be gained from changing them either.

PR #6474, opened 2022-07-26
Regenesis fix: Don't prevent Regenesis effect when runner card is trashed

### LINKED ISSUE #6466: Regenisis Don't Trigger if Runner has Taken Damage That Turn
Regenisis triggers so long as no cards have been added to archives on the turn when its scored, but right now if the runner takes damage (for example, bladderwort at start of turn), it does not.  Not sure if this extends to any runner discard/trashes.  To duplicate:

Corp has Agenda in Archives
Corp has Bladderwort on the board and 2 credits.
Corp has Regenisis on the board.
Corp starts turn, does 1 net damage, and gains a credit.  
Corp Advances Regeneisis 3 times.  This should allow you to add the face down archives agenda to score area, but it does not.

artemmikhalitsin:
Fixes #6466 

When checking for presence of :card-moved to discard events for regenesis, only consider corp cards

### DIFF
--- a/src/clj/game/cards/agendas.clj
+++ b/src/clj/game/cards/agendas.clj
@@ -1529,7 +1529,7 @@
 (defcard "Regenesis"
   {:on-score {:req (req (and (some #(not (faceup? %)) (:discard corp))
                              (no-event? state side :card-moved #(and (= [:discard] (:zone (second %)))
-                                                                      (= "Corp" (:side (second %)))))))
+                                                                     (corp? (second %))))))
               ;; we want a prompt even if there are no valid targets,
               ;; to make sure we don't give away hidden info
               :prompt "Select a face-down agenda in Archives?"
--- a/src/clj/game/cards/agendas.clj
+++ b/src/clj/game/cards/agendas.clj
@@ -1528,7 +1528,7 @@
 
 (defcard "Regenesis"
   {:on-score {:req (req (and (some #(not (faceup? %)) (:discard corp))
-                             (no-event? state side :card-moved #(and (= [:discard] (:zone (second %)))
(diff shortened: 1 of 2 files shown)


artemmikhalitsin:
on a side note, I noticed that `turn-events` ignores `side` when looking up events... should it filter based on side as well? should this maybe done somewhere else?

src/clj/game/cards/agendas.clj
```diff
@@ -1528,7 +1528,8 @@
 
 (defcard "Regenesis"
   {:on-score {:req (req (and (some #(not (faceup? %)) (:discard corp))
-                             (no-event? state side :card-moved #(= [:discard] (:zone (second %))))))
+                             (no-event? state side :card-moved #(and (= [:discard] (:zone (second %)))
```
NoahTheDuke, line 1531: Mind doing some clean-up and changing this to use `in-discard?`?

NoahTheDuke:
Abilities need to be able to look at all events. If a given event needs to communicate who caused/initiated the event, then it's generally passed in as a "target". (We're slowly transitioning from a list of "targets" to a context map that has named values, but it's a slow process lol, so my apologies for how messy this all is.) Is there something specific here you're thinking of?

NoahTheDuke:
Also, if you wanna join our dev slack, email me at noah.bogart@hey.com (or post your email in here but some people don't like that lol)

artemmikhalitsin:
Hi Noah, I emailed you earlier this week from artemmikhalitsin@gmail.com, not sure if you've seen it

NoahTheDuke:
Oops, that's my bad lol.

PR #6478, opened 2022-07-26
Fixes #6159 inside job's effect should only be active for its own run

### LINKED ISSUE #6159: Inside Job gives double bypass with Doppelgänger
Link to bug replay: https://jinteki.net/bug-report/73576b07-7b8e-4ae3-8047-21aadcd46758?b=0

Description:

Inside Job gives double bypass  with Doppelgänger

queueseven:


queueseven:
I guess more cards are affected and needs fixing. I guess every event card that `:makes-run true` and does an `:event` without `:duration :end-of-run`, like *Leave No Trace*

NoahTheDuke:
I think this is a fine change for now.

PR #6482, opened 2022-07-27
:watch-path so shadow-cljs picks up and injects changed css correctly

queueseven:
This should fix css hot reloading.

### DIFF
--- a/shadow-cljs.edn
+++ b/shadow-cljs.edn
@@ -3,6 +3,7 @@
  {:app {:target :browser
         :output-dir "resources/public/js"
         :devtools {:watch-dir "resources/public/css"
+                   :watch-path "/css"
                    :repl-init-ns nr.dev}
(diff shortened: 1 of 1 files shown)


queueseven:
I don't want to push changes myself, even if they are so small. I think it would be better if at least one other person confirms that this works now, even if this is such a minor thing.... 🤔

NoahTheDuke (commented):
I thought these overlapped. You need both?

queueseven:
@NoahTheDuke Well, we could remove `:watch-dir`, then shadow-cljs will watch for changes in the entire root folder (as far as I understand it), but since we know that our css is in the `resources/public/css`, it doesn't hurt to keep it. The important part is `:watch-path "/css"`

NoahTheDuke:
And it knows to search subdirectories for that path?

PR #6486, opened 2022-07-28
Fix: Pay moon pool RFG cost even if no cards are discarded

### LINKED ISSUE #6483: Moon Pool Doesn't RFG if No cards are trashed from HQ
As in title.  If corp chooses to trash zero cards from HQ with Moon Pool (so they can shuffle cards already in archives), it stays on the board instead of rfging as it should.

artemmikhalitsin:
Fixes #6483 

now the child ability has the :cancel-effect, so cost can still be paid

### DIFF
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -1390,6 +1390,7 @@
              :choices {:card #(and (corp? %)
                                    (in-hand? %))
                        :max 2}
+             :async true
              :msg (msg "trash " (count targets) " cards from HQ ")
              :effect (req (wait-for (trash-cards state :corp targets {:cause-card card})
                                     (continue-ability
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -1371,7 +1371,7 @@
                                  (in-discard? %)
                                  (not (faceup? %)))
                      :max 2}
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/assets.clj
```diff
@@ -1385,21 +1385,27 @@
...
-                              :max 2}
-                    :msg (msg "trash " (count targets) " cards from HQ ")
-                    :effect (req (wait-for (trash-cards state :corp targets {:cause-card card})
-                                           (continue-ability
-                                             state side
-                                             moon-pool-reveal-ability
-                                             card nil)))
-                    :cancel-effect (effect (system-msg "declines to use Moon Pool to trash any cards from HQ")
-                                           (continue-ability moon-pool-reveal-ability card nil))}]})))
+      (let [moon-pool-discard-ability
```
NoahTheDuke, line 1388: There's no need for a second let block, move this into the enclosing let block.

NoahTheDuke (changes requested):

src/clj/game/cards/assets.clj
artemmikhalitsin: Thanks, I agree. It was a remnant of when I was messing with letfn

NoahTheDuke (approved):

PR #6489, opened 2022-07-29
Don't trigger Ob Superheavy when trashed card has no cost

### LINKED ISSUE #6488: Cannot trash Public Agendas with Ob Superheavy ability available
Attempting to trash Oaktown Renovation causes an Internal Error. Fails both through card abilities (ie. Stavka, Extract) and Drag+Drop or /trash. Have done a bit of testing and it seems to happen with all Public Agendas. 

I'm guessing it hangs because it doesn't know what to do with a non-rezzed+non-derezzed card, or the agenda is seen as rezzed but doesn't have a rez cost.

There is no issue if the Ob trigger has already been fired that turn though.

artemmikhalitsin:
Fixes #6488 

Maybe should also not fire when cost of trashed card is 0, since it will always have no legal targets?

### DIFF
--- a/src/clj/game/cards/identities.clj
+++ b/src/clj/game/cards/identities.clj
@@ -1483,11 +1483,13 @@
                                   :effect (resolve-install)}
                                  card nil))}}})]
     {:events [{:event :corp-trash
-               :req (req (and
-                           (installed? (:card context))
-                           (rezzed? (:card context))
-                           (some? (trash-cause eid target))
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
This is good but it's not quite right. I didn't see you had started on this or I wouldn't have written my own fix in #6490, should have given you advice on how to do the same.

artemmikhalitsin:
Totally agree. Just looked up and learned that face-up agendas are neither rezzed nor unrezzed.

PR #6490, opened 2022-07-29
Bug fixes 22-07-28

### LINKED ISSUE #6488: Cannot trash Public Agendas with Ob Superheavy ability available

NoahTheDuke:
Closes #6488

### DIFF
--- a/src/clj/game/cards/operations.clj
+++ b/src/clj/game/cards/operations.clj
@@ -549,7 +549,8 @@
   {:on-play
    {:prompt "Choose a faceup card"
     :choices {:card #(or (and (corp? %)
-                              (faceup? %))
+                               (installed? %)
+                               (faceup? %))
                          (and (runner? %)
                               (or (installed? %)
                                   (:host %))
--- a/src/clj/game/cards/operations.clj
+++ b/src/clj/game/cards/operations.clj
@@ -549,7 +549,7 @@
   {:on-play
    {:prompt "Choose a faceup card"
     :choices {:card #(or (and (corp? %)
-                              (rezzed? %))
+                              (faceup? %))
                          (and (runner? %)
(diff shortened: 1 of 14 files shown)


src/clj/game/cards/operations.clj
```diff
@@ -549,7 +549,7 @@
   {:on-play
    {:prompt "Choose a faceup card"
     :choices {:card #(or (and (corp? %)
-                              (rezzed? %))
+                              (faceup? %))
```
artemmikhalitsin, line 552: This lets you dedi cards in archives

artemmikhalitsin (changes requested):

src/clj/game/cards/operations.clj
artemmikhalitsin, line 552:
```suggestion
                               (installed? %)
                               (faceup? %))
```
Otherwise you can dedi cards in archives lol

artemmikhalitsin:
Just tried it out in a game and oaktown installs face-down

src/clj/game/cards/operations.clj
artemmikhalitsin: Looking at the wording on a few cards (Dedi, Mumbad Construction Co.), it seems "faceup" implies "installed", unless otherwise specified.

PR #6503, opened 2022-08-01
Fix "Swap sides" chat messages in pending lobby

locks:
This took me longer than I'm comfortable admitting.

### DIFF
--- a/src/clj/web/lobby.clj
+++ b/src/clj/web/lobby.clj
@@ -523,21 +523,17 @@
         (->> (assoc lobbies gameid)))
     lobbies))
 
-(defn- player-swap-text [players]
-  (apply
-   str
-   (map #(str (-> % :user :username) " is now " (-> % :side) ". ")
-        players)))
+(defn swap-text [players current-side]
+  (let [[player1 player2] (mapv swap-side players)
+        player1-username (-> player1 :user :username)
+        player2-username (-> player2 :user :username)]
+    (str player1-username " has swapped sides. "
+         (if (= current-side "Any Side")
+           "Waiting for opponent."
+           (str player1-username " is now " (:side player1) ". "))
+         (when player2
+           (str player2-username " is now " (:side player2) ".")))))
 
-(defn swap-text [players side]
-  (let [[{player-1-side :side
-          {player-1-username :username} :user}] players]
-    (str
-     player-1-username " has swapped sides. "
-     (cond
-       (= side "Any Side") "Waiting for oppenent."
-       (= side player-1-side) (player-swap-text players)
-       :else (player-swap-text (mapv swap-side players))))))
 
 (defmethod ws/-msg-handler :lobby/swap
   [{{user :user} :ring-req
--- a/src/clj/web/lobby.clj
+++ b/src/clj/web/lobby.clj
@@ -523,13 +523,21 @@
         (->> (assoc lobbies gameid)))
     lobbies))
 
-(defn swap-text [players]
-  (let [[player1 player2] (mapv swap-side players)
-        player1-username (-> player1 :user :username)
-        player2-username (-> player2 :user :username)]
-    (str player1-username " has swapped sides. "
-         player1-username " is now " (:side player1) ". "
-         player2-username " is now " (:side player2) ".")))
+(defn- player-swap-text [players]
+  (apply
+   str
+   (map #(str (-> % :user :username) " is now " (-> % :side) ". ")
+        players)))
+
+(defn swap-text [players side]
+  (let [[{player-1-side :side
+          {player-1-username :username} :user}] players]
+    (str
+     player-1-username " has swapped sides. "
+     (cond
+       (= side "Any Side") "Waiting for oppenent."
+       (= side player-1-side) (player-swap-text players)
+       :else (player-swap-text (mapv swap-side players))))))
 
 (defmethod ws/-msg-handler :lobby/swap
   [{{user :user} :ring-req
@@ -538,7 +546,7 @@
   (let [lobby (app-state/get-lobby gameid)]
     (when (and lobby (first-player? uid lobby))
       (let [swap-message (core/make-message {:user user
-                                             :text (swap-text (:players lobby))})
+                                             :text (swap-text (:players lobby) side)})
             new-app-state (swap! app-state/app-state
                                  update :lobbies
                                  #(-> %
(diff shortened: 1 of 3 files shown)


NoahTheDuke:
Thanks for tackling this!

I prefer that indentation for entire files isn't changed. We don't keep strict indentation/styling rules, but we generally hew to the Clojure Style Guide except that we do 2-space indentation for functions. I personally think it's a dumb rule to have differing indentation for function calls and macros with bodies or rest arguments and the rest of the codebase is (generally) aligned on this as well.

src/clj/web/lobby.clj
```diff
@@ -523,13 +523,21 @@
...
+
+(defn swap-text [players side]
+  (let [[{player-1-side :side
+          {player-1-username :username} :user}] players]
+    (str
+     player-1-username " has swapped sides. "
+     (cond
+       (= side "Any Side") "Waiting for oppenent."
+       (= side player-1-side) (player-swap-text players)
+       :else (player-swap-text (mapv swap-side players))))))
```
NoahTheDuke, line 540: I think this is overly complicated for what it's trying to do. I'm not entirely sure why it's written this way, so my apologies for not understanding. When would the code reach the `:else` block? No wait, the odd one is when the first player has chosen to swap to their own/same side? I would expect that we don't print anything from such a move.

Outside of that, I think that I would prefer this to go back to a variation on how it was originally written. Maybe something like this? If you want the destructuring, you can use it. But I don't think we need another layer of function calls.

```clojure
(defn swap-text [players current-side]
  (let [[player1 player2] (mapv swap-side players)
        player1-username (-> player1 :user :username)
        player2-username (-> player2 :user :username)]
    (str player1-username " has swapped sides. "
         (if (= side "Any Side")
           "Waiting for opponent."
           (str player1-username " is now " (:side player1) ". "))
         (when player2
           (str player2-username " is now " (:side player2) ".")))))
```

NoahTheDuke: The reason is that this is a very procedural function: take the two players, print their usernames with opposite side strings. The indirection of the `cond` and `player-swap-text` makes it much harder to see that simple string generation.

NoahTheDuke (changes requested):

src/clj/web/lobby.clj
```diff
@@ -1,20 +1,20 @@
...
-   [game.utils :refer [server-card]]
-   [jinteki.utils :refer [select-non-nil-keys side-from-str superuser?]]
-   [jinteki.validator :as validator]
-   [medley.core :refer [find-first random-uuid]]
-   [monger.collection :as mc]
-   [time-literals.read-write :as read-write]
-   [web.app-state :as app-state]
-   [web.mongodb :as mongodb]
-   [web.stats :as stats]
-   [web.ws :as ws]))
```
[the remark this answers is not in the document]
locks: Sorry, VSCode just reformatted it on save.

src/clj/web/lobby.clj
locks: So, this is a bit niche but the problem is mostly when there's only one person in the lobby. When there's two people in the lobby there isn't a "Any Side", since we just keep swapping back and forth.
However, if there's only one person, here's the current behaviour:

### Swap sides -> Any Side

- `locks has swapped sides. locks is now Corp. is now .` 

### Swap sides -> Corp

- `locks has swapped sides. locks is now Corp. is now .`
- Swap sides -> Corp
- `locks has swapped sides. locks is now Runner. is now .
- Swap sides -> Any Side
- `locks has swapped sides. locks is now Runner. is now .

### Swap sides -> Runner

- `locks has swapped sides. locks is now Runner. is now .`
- Swap sides -> Runner
- `locks has swapped sides. locks is now Corp. is now .`
- Swap sides -> Any Side
- `locks has swapped sides. locks is now Corp. is now .`

NoahTheDuke: Right, so my suggestion is to rely on `when` to return `nil` (which `str` converts to an empty string) if player 2 doesn't exist, and to either print "Waiting on opponent" if player 1 is "Any Side" or player 1's new side (which is set by calling `(mapv swap-side players)` already). There shouldn't be any other logic here, right?

locks: So if you pick the side that was already picked, we return nothing?

NoahTheDuke: Should we even show the side you currently are? Or just show the ones you're not? In what circumstances is it helpful to be "Corp" and then click "Corp" and have it say "You've swapped sides and are now Corp."?

locks: I definitely don't think it's useful! :P What do you think of disabling the button, so that the listing is always consistent?

NoahTheDuke: Yeah, if that's possible, let's do that. Then it simplifies our logic so we only have to deal with actually moving from Corp->Runner or Runner->Corp.

src/cljs/nr/pending_game.cljs
```diff
@@ -85,12 +88,16 @@
...
        (into
-         [:ul.dropdown-menu.blue-shade]
-         (for [side ["Any Side" "Corp" "Runner"]]
-           [:a.block-link
-            {:on-click #(ws/ws-send! [:lobby/swap {:gameid @gameid
-                                                   :side side}])}
-            [:li (tr-side side)]]))])))
+        [:ul.dropdown-menu.blue-shade]
+        (for [side ["Any Side" "Corp" "Runner"]]
+          (if (current-side? @players side)
```
locks, line 93: @NoahTheDuke if you know of a more succint way to do this let me know, it's my first time touching Reagent!

NoahTheDuke: Two thoughts: 1) Lift the `(-> players first :side)` into a let block so all you're checking is `(if (= player-side side) ,,,)`, and 2) move the `if` to only wrap the properties map: `[:a.block-link (if (= player-side side) {:style ...} {:on-click ...}) [:li (tr-side side)]]`

Then we don't even need the `current-side?` function.

NoahTheDuke:
Looks good but check your indentation changes. Seems your autoformatter needs to be updated to use 2 spaces for function calls.

locks:
@NoahTheDuke I have "format on save" disabled (reasons), I hope it's ok now.

Issue #6493, opened 2022-07-29
Move from Circle CI to GitHub Actions


locks:
I need to figure out why `lein kaocha` (I don't usually use clj) is not a task and it should be ready to go.

PR #6504, opened 2022-08-01
Implement CI as a GitHub Action

locks:
I haven't removed the Circle CI configuration yet because I wanted to confirm with maintainers that this is the way to go.
Closes https://github.com/mtgred/netrunner/issues/6493.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,31 @@
...
+    branches: ["master"]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install dependencies
+        run: lein deps
```
NoahTheDuke, line 16: Unnecessary as all calls to `lein` will resolve deps first.

locks: Interesting. It was like that in the default Clojure workflow and I think all the ones I saw during my research. Removing this step!

locks: @NoahTheDuke people might do this so as not to get the dependency spam when running other steps:
image

I'd suggest keeping it separate even if not strictly necessary to make it easier to troubleshoot.

NoahTheDuke: It looks like the cache is only being set when the tests don't fail, so maybe include the fix in this PR to see what happens.

NoahTheDuke:
Looks like this works, so go ahead and remove the CircleCI folder. rip in pepperonis circle!

PR #6523, opened 2022-08-08
Closes #6522: fix z-index

### LINKED ISSUE #6522: Trying to install a card: prompt hidden
grafik

Workaround:

.gameboard .leftpane .inner-leftpane .right-inner-leftpane .button-pane .panel {
    z-index: 0 !important;
}

queueseven:
The *-container are position: relative, so they create a new stacking context. Setting .servers-menu's z-index to 30 therefore does not render it above .panel's z-index of 20; the positioned parent of .server-menu needs the z-index.

### DIFF
--- a/src/css/gameboard.styl
+++ b/src/css/gameboard.styl
@@ -337,6 +337,7 @@
 
     .hand-container, .deck-container, .discard-container
         position: relative
+        z-index: 30
 
     .hand-controls
         display: inline-block


NoahTheDuke:
This isn't the best fix. Instead of juggle z-orders, the server-menu should not render _at all_ if there's something in prompt-state. There is no moment when it should be available at the same time a prompt is available, as a prompt means that we are in the midst of resolving an ability or effect and don't have click priority.

queueseven:
Yes, but when you want to install a card and the server menu opens, there's actually nothing in the prompt state, so I'm not sure we're talking about the same thing here 🙂. 

There's somehow also a way to break the 'close-server-menu-on-click-anywhere' functionality... 🤔

What's more confusing is that I was not able to reproduce the bug today; but it certainly happened to other peope, too. Somebody talked about it on jnet's chat, and the screenshot in the issue is from discord....

not very satisfactory....

NoahTheDuke:
> Yes, but when you want to install a card and the server menu opens, there's actually nothing in the prompt state, so I'm not sure we're talking about the same thing here 🙂.

lol we're not, my apologies. I misunderstood the issue. This is probably fine, tho I'm always worried about z orders.

PR #6526, opened 2022-08-09
Enforce MU limits when installing

NoahTheDuke:
see #6469

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -85,10 +85,10 @@
                    :resource-paths ["target"]
                    :clean-targets ^{:protect false} ["target"]
                    :jvm-opts ["-Djdk.attach.allowAttachSelf"
-                              "-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5010"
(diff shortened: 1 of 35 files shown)


queueseven:
I took the liberty to play aorund with these changes. There's an edge case that does not work correctly:

I have all my memory used up, have 2 or more link and want to install [ZU.13 Key Master](https://netrunnerdb.com/en/card/02007), a cloud breaker.

I can't install it as the 'Insufficient MU'-prompt will always be opened, but Key Master should have a memory cost of 0.

NoahTheDuke:
Thanks so much for looking at this. I'll check that out!

queueseven:
I'm also currently looking, but I don't really know how the :constant-effects stuff works.... 😂🤔

queueseven:
I'm looking forward to seeing what your solution looks like, lol

NBKelly:
One edge case that might be worth testing - if I'm replacing my console with another console (swapping out masterwork for another masterwork to cycle a card, for example) I shouldn't be forced to trash any programs under the current CR.

PR #6527, opened 2022-08-09
Add :implementation to condition-counters

queueseven:
This fixes at least two things:

a) When installing such a card (like 'On the Lam'), since :implementation is not :full and :title is empty, the message 'implementation: .' will be logged.

b) When hovering over an installed card, it shows the warning 'not implemented' (even if the card is implemented)

### DIFF
--- a/src/cljc/game/core/card.cljc
+++ b/src/cljc/game/core/card.cljc
@@ -495,10 +495,11 @@
 ;;    characteristics, and is trashed.
 
 (defn convert-to-condition-counter
-  [{:keys [cid code side title zone]}]
+  [{:keys [cid code side title zone implementation]}]
   (map->Card
     {:cid cid
      :code code
+     :implementation implementation
      :printed-title title
      :side side
      :type "Counter"


NoahTheDuke:
Good catch! Should we maybe fix this in our printing of `:implementation` too? I suspect this is not the only place we create an adhoc card that will have the same issue.

queueseven:
That was my first idea as well, but then I noticed that the card preview also shows the 'not implemented ' warning wrongly. So it's better to fix it where the "fake card" is created. I could take a look tomorrow if there's another place where we create ad-hoc cards ... 🤔

NoahTheDuke:
Almost every single slash command creates an adhoc card, I think `as-agenda` does too.

queueseven:
Ah, ok, I'll check if this is an issue, tooVon: Noah ***@***.***>Gesendet: Dienstag, 9. August 2022, 19:02An: mtgred/netrunner ***@***.***>Cc: queueseven ***@***.***>; Author ***@***.***>Betreff: Re: [mtgred/netrunner] Add :implementation to condition-counters (PR #6527)
Almost every single slash command creates an adhoc card, I think as-agenda does too.

queueseven:
I checked, and indeed, as-agenda needs :implementation, too (for cards like Fan Site etc.).

Issue #6521, opened 2022-08-08
Emotes


distributive:
Hearthstone has emotes (instead of in-game chat) which could be a useful feature in jnet for speeding up some common interactions:
![image](https://user-images.githubusercontent.com/26557961/183405006-bb36df66-7513-4bfd-b36c-95b81a236044.png)

Suggestions:
- gl;hf
- Yes/no
- Action?/Response?
- Can I undo this click?
- Can we undo turn?
- Nice!
- Ping (for testing connection)
- gg! (maybe lock this during games to prevent bm)

ggDropbear:
- I'm thinking...
- Uh oh./I'm in trouble...
- [That's bait.](https://thumbs.gfycat.com/FilthyWelcomeCrossbill-max-1mb.gif)
- Have to faceplant./Faceplanting./Faceplant.
- I swear I have money in this deck.

PR #6532, opened 2022-08-10
#6521 add emotes to user profile

queueseven:
First step: allow adding emotes


src/cljs/nr/account.cljs
```diff
@@ -49,6 +49,7 @@
   (swap! app-state assoc-in [:options :log-width] (:log-width @s))
   (swap! app-state assoc-in [:options :log-top] (:log-top @s))
   (swap! app-state assoc-in [:options :blocked-users] (:blocked-users @s))
+  (swap! app-state assoc-in [:options :emotes] (:emotes @s))
```
NoahTheDuke, line 52: I _hate_ this whole system, have I said this before? Lmao I don't want you to change it here, just voicing my displeasure instead of bottling it up so we can start brainstorming a better way.

src/cljs/nr/account.cljs
```diff
@@ -478,27 +539,8 @@
...
-                                    :name "block-user-button"
-                                    :on-click #(add-user-to-block-list user s)}
-            (tr [:settings.block "Block user"])]]
-          (doall (for [bu (:blocked-users @s)]
-                   [:div.line {:key bu}
-                    [:button.small.unblock-user {:type "button"
-                                                 :on-click #(remove-user-from-block-list % s)} "X" ]
-                    [:span.blocked-user-name (str "  " bu)]]))]
-
+         (blocked-users user s)
```
NoahTheDuke, line 542: Use brackets so they're handled by reagent's reactive rendering logic.

src/cljs/nr/account.cljs
```diff
@@ -236,6 +254,49 @@
...
+                     :block-user-input :blocked-users
+                     add-user-to-block-list
+                     remove-user-from-block-list)
+   ])
+
+(defn- emotes [user s]
+  [:section
+   [:h3 (tr [:settings.emotes "Emotes"])]
+   (account-list-box user s
+                     (tr [:settings.:emote "Emote"])
```
NoahTheDuke, line 293: Remove colon from inside keyword.

NoahTheDuke:
So the idea is that emotes are user defined? I thought Hearthstone had predefined emotes, so people couldn't write "go fuck yourself" as an emote lmao

src/cljs/nr/account.cljs
queueseven: TBH I don't really know much about reagent lol

queueseven:
> So the idea is that emotes are user defined? I thought Hearthstone had predefined emotes, so people couldn't write "go fuck yourself" as an emote lmao

It has predefined emotes because it has no chat. Offering a chat function opens a huge can of worms of problems, at least if you have a huge (toxic and competitive) user base.

Since we don't stop users from writing insults (or worse) in chat, there's no reason to not offer user defined emotes. 

At least that's my opinion.

I could imagine adding a rate limiting, but... You could also just copy a text spam paste/enter in chat... But I don't think the chat was/is a problem since our users are usually nice? 😂

src/cljs/nr/account.cljs
NoahTheDuke: You should at least read this page to get a sense of what's possible: http://reagent-project.github.io/

NoahTheDuke:
Any updates on this? I'm good with adding it, just gotta get it cleaned up a bit I think.

NoahTheDuke:
Gonna close this for inactivity. Holler if you want to get it working!

PR #6548, opened 2022-08-17
Fix API access

queueseven:


NoahTheDuke:
This is a pretty dumb move on my part, lol. Thanks for catching it. Hopefully the rest of the code isn't littered with such changes. :grimacing:

Issue #6518, opened 2022-08-08
Pravdivost Consulting: Political Solutions doesn't close the prompt....


queueseven:
...when the corp declines to use the ability

NoahTheDuke:
Thanks for the report!

Issue #6530, opened 2022-08-10
Softlock when using Prav's ability


locks:
I'm using [Pravdivost Consulting](https://netrunnerdb.com/en/card/33048).
The runner makes a successful run, and I'm shown the yes/no prompt about using Prav's ability.
I choose yes, but then I don't have any card that can be advanced so I click done. The game softlocks at this point.

queueseven:
related: #6518

francescopellegrini:
#6559 should've fixed this.

PR #6559, opened 2022-08-24
remove additional confirmation prompt from Pravdivost Consulting's ability

francescopellegrini:
After playing with it for a while, I can see why Aesop's Pawnshop didn't ask you for confirmation every time.. 🙄


francescopellegrini:
Could fix #6518 and #6530

PR #6565, opened 2022-08-27
remove all uses of prompt! since it should not be called directly

francescopellegrini:
I found [this note in the code](https://github.com/mtgred/netrunner/blob/master/src/clj/game/core/engine.clj#L436) so I though a small refactor could be appreciated.


NoahTheDuke:
At this point, try removing it from engine.clj and see what breaks.

francescopellegrini:
> At this point, try removing it from engine.clj and see what breaks.

I removed some more references and turned the function private, but I didn't remove it completely. Is it ok for you?=

NoahTheDuke:
Yeah, making it a helper in `game.core.engine` is great.

PR #6574, opened 2022-08-29
Makes Endurance non-silent on runs with Into the Depths.

### LINKED ISSUE #6556: Into the Depth + Endurance
When you use Into the Depth with an empty Endurance and you passed a piece of ice, you should be able to place 1 power counter on Endurance before you trigger Into the Depth (they both have the same trigger timing). However, on jnet, it only lets you resolve Into the Depth first, so it's impossible to place power counter on the ship before you can charge it.

Saintis:
Fixes #6556 by making Endurance non-silent when ItD is the current run event.

Have tested manually with DL and ItD.

### DIFF
--- a/src/clj/game/cards/hardware.clj
+++ b/src/clj/game/cards/hardware.clj
@@ -674,7 +674,8 @@
    :constant-effects [(mu+ 2)]
    :events [{:event :successful-run
              :req (req (first-event? state :runner :successful-run))
-             :silent (req true)
+             ;; Specifically non-silent on ItD runs because then order of Endurance trigger and ItD charge matters
+             :silent (req (not= "Into the Depths" (-> @state :run :source-event :title)))
              :msg "place 1 power counter on itself"
              :async true
              :effect (effect (add-counter eid card :power 1 nil))}]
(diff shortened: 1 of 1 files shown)


francescopellegrini:
> Fixes #6556 by making Endurance non-silent when ItD is the current run event.
> 
> Have tested manually with DL and ItD.

Have you tried making it `interactive` instead?

NoahTheDuke:
Yeah, we've done targeted fixes like this in the past (Bank Job has one or two), but I don't like it and don't want to introduce more. They're brittle and don't fix the fundamental problem. If the issue is "Endurance needs to know if an active card has a 'charge' effect" then maybe we can think about labeling those abilities with `:charge true` or something.

In the meantime, I'd probably say that `:interactive (req true)` is best here.

Saintis:
`:interactive (req true)` is not needed and Endurance doesn’t need to be interactive. ItD has an interactive flag so the simplest fix would be to just remove the `:silent` flag on endurance. A more robust fix would be to add `:charge true` or similar to the ItD ability that Endurance can look for. The best fix would be to rewrite the `:interactive` system to be more designed for trigger interplays. 

I’ll change the PR tonight to just remove the silent flag. Then endurance will still be interactive with ItD but not with cards like dirty laundry.

NoahTheDuke:
> The best fix would be to rewrite the :interactive system to be more designed for trigger interplays.

Agreed, the limiter as always is the time and effort to change everything over. If you wanna do it, I'd love to review and merge!

Saintis:
`:silent` has been removed, this is ready to be merged now :)

PR #6583, opened 2022-09-01
Street peddler set-aside

NBKelly:
Changes street peddler to set-aside the cards. This also means we can use tech trader properly (since peddler is trashed before the install/prompt). Also added the sanity check to usage that you can legally install at least one of the cards (tech trader credits can't be used for the sanity check, as per the rules).

For some reason I can't figure out, trash-cards is giving me a stack overflow. I'll figure that out sooner or later - then I need to delete the old street peddler code.


NoahTheDuke:
You're getting a stack overflow because you're re-using `eid`s. The `eid` that is passed into `req` needs to be passed to `effect-completed` on it, or an async function that at some point will call `effect-completed` on it, and _nowhere else_. If you want to re-use the data attached to an eid to make a new eid, write `(make-eid state eid)` (and then `(assoc new-eid :blah blah)` as you wish).

It looks like `set-aside` isn't an async function (we gotta find a way to mark whether a function merely takes an eid or is async :sob:), but down in `(runner-install state side (assoc eid ...))`, that's gotta be `(runner-install state side (make-eid state (assoc eid ...)))`.

Also looks like there's no need for you to create `set-aside-eid` because you're immediately using it in `set-aside-cards`. Just say `(let [set-aside-cards (get-set-aside state side eid)] ...)`.

Additionally, the nested ability needs to be async. Idk why you commented out `:async true`, but you should add it back in.

Hope that helps solve the issue.

NBKelly:
Yeah, that was it. I was saving that eid earlier because I was calling `get-set-aside ...` again, but I switched to just filtering out the chosen card instead.

Right now this should be fully functional, and I just need to clean up a couple of tests (notably, having this work properly means the cancel button is gone - but you also can't use the ability unless there's at least one valid target now).

NBKelly:
Alright, this is now fully functional. I noticed the "street peddler trashes" pending test was for trying to dismiss the menu by manually trashing peddler during resolution for some reason.

That pending test is still there, but peddler is already trashed at that point now, so I think it's kind of an irrelevant test now?

Peddler also properly enforces rules for being able to use it now, which is nice, even if it is a few metas too late :joy: .

PR #6593, opened 2022-09-03
ixodidae, lamprey, tapwrm, etc are all trashed by the runner now

### LINKED ISSUE #6582: Purgin Tapwrm should cause the runner to trash
When the corp purges virus counters to trash Tapwrm, it should be the runner who is trashing it, triggering cards like Wastelands and Reaver. As is, the card gets trashed and those cards weren't getting triggered

NBKelly:
This means that "trash cards" triggers like wasteland will trigger like they're meant to (like they do with clot).

I think that side was injecting the corporation before, since they were the one triggering the event.

Closes #6582

### DIFF
--- a/src/clj/game/cards/events.clj
+++ b/src/clj/game/cards/events.clj
@@ -2476,8 +2476,9 @@
    :events [{:event :purge
              :condition :hosted
              :async true
-             :effect (req (wait-for (trash state side card {:cause :purge
-                                                            :cause-card card})
+             :msg (msg "trash itself")
+             :effect (req (wait-for (trash state :runner card {:cause :purge
+                                                               :cause-card card})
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/events.clj
```diff
@@ -2476,8 +2476,9 @@
    :events [{:event :purge
              :condition :hosted
              :async true
-             :effect (req (wait-for (trash state side card {:cause :purge
-                                                            :cause-card card})
+             :msg (msg "trash itself")
```
NoahTheDuke, line 2479:
```suggestion
             :msg "trash itself"
```

NoahTheDuke: And same with the rest of them.

PR #6594, opened 2022-09-03
compile completes its eid

### LINKED ISSUE #6393: Issues with Compile triggers preventing turns proceeding - possibly interaction with Apoc
I  just had a game where I used Compile to run HQ on an Apoc turn, installing an engolo, and from then on it would not let my opponent start their turn unless they /close-prompt, even though I had no triggers. Then, on a later turn after I installed paperclip  from the bin, every time i finished a run it would give me an option for a paperclip and a compile trigger but they didn't seem to do anything. Unfortunately I don't have a replay to share.

NBKelly:
Closes #6393


NoahTheDuke:
I think the easier solution is to remove `:async true`.

PR #6596, opened 2022-09-04
ob allows you to shuffle when trashing a 0 cost card (and a test)

### LINKED ISSUE #6595: Ob should let you search (and fail) when trashing a 0 cost card
I think it's basically a 1-liner to fix (I had an "if cost is not 0" there), and I just never considered that use case :joy:

NBKelly:
You're allowed to search for a card with a -1 printed cost (even though none have been printed).

To avoid wasting time, searching with a cost of -1 or less will just instantly shuffle the deck. If they make a card with a printed cost of less than 0, they've earned the five minutes it takes to fix it :joy:.

Also I noticed the ordering when using rashida (the test case) will say what it does when it trashes rashida, rather than when it actually draws 3/gains 3. I've moved the rashida log message a tiny bit (this also means it orders nicely with ob and other things that trigger from trashing a card).

Closes #6595

### DIFF
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -1874,8 +1874,9 @@
                                        (trash state side card {:cause-card card})
                                        (wait-for
                                          (gain-credits state side 3)
-                                         (draw state side eid 3)
-                                         (system-msg state side (str "uses " (:title card) " to gain 3 [Credits] and draw 3 cards")))))}}}
+                                         (wait-for (draw state side 3)
+                                                   (system-msg state side (str "uses " (:title card) " to gain 3 [Credits] and draw 3 cards"))
+                                                   (effect-completed state side eid)))))}}}
                      card nil))}]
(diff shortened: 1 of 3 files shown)


src/clj/game/cards/assets.clj
```diff
@@ -1867,14 +1867,15 @@
...
                        :yes-ability
                        {:async true
-                        :msg "gain 3 [Credits] and draw 3 cards"
                         :effect (req (wait-for
                                        (trash state side card {:cause-card card})
                                        (wait-for
                                          (gain-credits state side 3)
-                                         (draw state side eid 3))))}}}
+                                         (draw state side eid 3)
+                                         (system-msg state side (str "uses " (:title card) " to gain 3 [Credits] and draw 3 cards")))))}}}
```
NoahTheDuke, line 1878: Either wait-for the draw with a new eid and then call effect-completed on the existing eid, or call system-msg before draw.

PR #6597, opened 2022-09-04
[quality of life] Skip confirming optional abilities

francescopellegrini:
Remove confirmation prompt from optional abilities which can be cancelled to prevent them from resolving anyway.


NoahTheDuke:
I don't think I like this change. I understand the intention and removing clicks is good, but this makes it harder to determine when an optional ability is an actual optional ability and will make it harder to determine "usage" in the future.

I'd be more inclined to remove the "Done" button from the targeting system, but that's a much larger change with implications I haven't thought through.

francescopellegrini:
> I don't think I like this change. I understand the intention and removing clicks is good, but this makes it harder to determine when an optional ability is an actual optional ability and will make it harder to determine "usage" in the future.
> 
> I'd be more inclined to remove the "Done" button from the targeting system, but that's a much larger change with implications I haven't thought through.

That's an interesting twist I didn't consider, but I agree with you on being more explicit on optional abilities.
I'll convert this PR to draft in the meanwhile.

PR #6600, opened 2022-09-04
rework Audacity to work like Moon Pool

francescopellegrini:
i.e. mentions how many counters are left to place

### DIFF
--- a/src/clj/game/cards/operations.clj
+++ b/src/clj/game/cards/operations.clj
@@ -235,18 +235,15 @@
              card nil)))}})
 
 (defcard "Audacity"
-  (letfn [(audacity
-             [x] {:prompt (msg "Choose a card that can be advanced to place advancement counters on (" x " remaining)")
-                  :async true
-                  :choices {:card can-be-advanced?
-                            :all true}
-                  :msg (msg "place 1 advancement counter on " (card-str state target))
-                  :effect (req (wait-for (add-prop state side target :advance-counter 1 {:placed true})
+  (letfn [(audacity [x]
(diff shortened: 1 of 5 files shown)


src/clj/game/cards/operations.clj
```diff
@@ -235,22 +235,26 @@
...
-            (when (< n 2)
-              {:prompt "Choose a card to place advancements tokens on"
-               :async true
-               :choices {:card can-be-advanced?
-                         :all true}
-               :msg (msg "place 1 advancement token on " (card-str state target))
-               :effect (req (add-prop state :corp target :advance-counter 1 {:placed true})
-                            (continue-ability state side (audacity (inc n)) card nil))}))]
+  (letfn [(audacity
+             [x] {:prompt (msg "Choose a card that can be advanced to place advancement counters on (" x " remaining)")
```
NoahTheDuke, line 239: Format the args vector like it was before please.

src/clj/game/cards/operations.clj
```diff
@@ -235,22 +235,26 @@
...
+                  :async true
+                  :choices {:card can-be-advanced?
+                            :all true}
+                  :msg (msg "place 1 advancement counter on " (card-str state target))
+                  :effect (req (wait-for (add-prop state side target :advance-counter 1 {:placed true})
+                                    (if (> x 1)
```
NoahTheDuke, line 245:
```suggestion
                                    (if (pos? x)
```

francescopellegrini: On a second thought, this is not correct: Audacity inner function would be called 3 times while it should only be called twice.

PR #6601, opened 2022-09-04
fix Eli 2.0 and Tapestry draws to be optional

francescopellegrini:
[Eli 2.0](https://netrunnerdb.com/en/card/13034) and [Tapestry](https://netrunnerdb.com/en/card/13037) subroutines say indeed "may draw"


NoahTheDuke (commented):
Add auto resolve?

PR #6602, opened 2022-09-05
Deep dive now plays nice with bacterial and degree mill

### LINKED ISSUE #6580: Deep Dive + Degree Mill results in "Continue to No Current Run"
I got the same issue as  #6324 with "Continue to No Current Run" when I stole a Degree Mill off Deep Dive.

NBKelly:
Alright, a bit of detective work and this one is actually solved.

* Previously, shuffling any card anywhere during an access would mark R&D as shuffled during a run, so paying for degree mill would fill the run register and cause a brick
* This also meant (and nobody appears to have noticed before) that stealing a degree mill during a multi-access on R&D put you back to the top of R&D
* Same goes for shuffling a card back with buffer drive
* I also fixed bacterial programming (again) - this time, it's the `reorder-choices` command ordering R&D. It now checks that a run is occurring before writing to the run register.

Closes #6580


francescopellegrini:
What about Sportsmetal drawing when an agenda is stolen via Deep Dive? I think I remember drawing the same cards set aside in the exact order, but it could have happened quite some time ago..

NBKelly:
At the moment, the cards are set aside and not actually in R&D. 

The card was implemented for ~3 or so months before we got set-aside done, I think it would have been then.

Saintis:
I wonder if there are any other parts of the code that implictly assumes all accesses happen inside a run...

NBKelly:
> I wonder if there are any other parts of the code that implictly assumes all accesses happen inside a run...

Yeah, that's a good thought actually. I'll see if I can go through the code and spot anything else which marks R&D as shuffled.

PR #6609, opened 2022-09-07
fix Apex installing programs facedown when MU is full

### LINKED ISSUE #6606: Apex Installing a Program failing at maximum MU
I used Apex's ID ability to face-down install a Reaver whilst having 0 available. The prompt for trashing existing programs came up, I wasn't sure what to press so I hit the continue button, and then it proceeded to my action phase without having installed the Reaver.

francescopellegrini:
Fixes #6606


NoahTheDuke:
Ugh, good catch.

PR #6610, opened 2022-09-07
Fix trap ice messing up with pass-ice

### LINKED ISSUE #6552: Bravado gives too many credits when interacting with Trap ice
Witnessed in an Ob game: Bravado on special offer, special offer trashes itself, runner gains 7. The ice was not passed, so it should be 6.

francescopellegrini:
Fixes #6552


src/clj/game/cards/ice.clj
```diff
@@ -575,7 +575,6 @@
                                                 (str/join ", " (map :title (take 3 (:deck runner))))
                                                 " from the Runner's Stack"))
                                (wait-for (mill state :corp :runner 3)
-                                         (system-msg state side (str "trashes Aimor"))
```
NoahTheDuke, line 578: Reason for this? Seems relevant for tracking what's happening.

francescopellegrini: You're right, I was misled by the fact Aimor was the only trap ice logging its own trashing.

I found out there might be room for improvement in this area, so I'd like to work on it a bit more

PR #6633, opened 2022-09-17
fix DRM preventing scoring agendas installed afterwards

francescopellegrini:
In #6621 Project Vitruvius was installed via DRM ability, so not actually **after** DRM finished resolving, thus making the test invalid.


francescopellegrini:
cc @NBKelly

NBKelly:
Oh, whoops. My mistake :clown_face:

Issue #6542, opened 2022-08-15
Afterimage lets you bypass using only 1 stealth credit


locks:
The card text says you have to use noly stealth credits to pay for the bypass, so using 1 stealth and 1 regular credit should not allow the runner to bypass.

NBKelly:
@NoahTheDuke this should be fixed by your costs rework (#7346)

NBKelly:
Completed via #7346

PR #6634, opened 2022-09-18
improve Afterimage stealth restriction

francescopellegrini:
Apparently its tests have been wrong ever since.. 🤦 

See also #6542

### DIFF
--- a/src/clj/game/cards/programs.clj
+++ b/src/clj/game/cards/programs.clj
@@ -373,11 +373,14 @@
                         1 "Sentry")]))
 
 (defcard "Afterimage"
-  (auto-icebreaker {:events [{:event :encounter-ice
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
I believe you can implement full stealth restriction and implementation. Look for :stealth-req in other card impls.

francescopellegrini:
> I believe you can implement full stealth restriction and implementation. Look for :stealth-req in other card impls.

The only stealth restrictions I could find were breakers and Mu Safecracker, which I tried to mimic but with no success. 
Can `:cost-req all-stealth` be provided to `can-pay?` function?

NoahTheDuke:
Honestly, I don't remember so I won't worry about it for now. Better to get it fixed immediately.

PR #6635, opened 2022-09-18
always provide card title to can-pay?

francescopellegrini:
Also fix some ice not caring if the runner could afford to pay credits and Otoroshi tax check.

### DIFF
--- a/src/clj/game/cards/events.clj
+++ b/src/clj/game/cards/events.clj
@@ -320,7 +320,7 @@
   {:on-play
    {:req (req (some #(and (ice? %)
                           (rezzed? %)
-                          (can-pay? state side eid card nil
+                          (can-pay? state side eid card (:title card)
                                     [:credit (rez-cost state side %)]))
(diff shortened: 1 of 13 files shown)


NoahTheDuke:
What's the point here?

francescopellegrini:
> What's the point here?

Have the toaster mention the card which is adding a cost to an action (e.g. rez, play)

NoahTheDuke:
The toast only appears when the cost can't be paid and it prints the name of the originating card. We use can-pay? in lots of places where we don't want to print the message multiple times.

For example, Brute Force Hack would now print if you try to play it and can't afford it (which they'll already know when it isn't highlighted in their hand isn't played if they click on it anyway) and then it will print once for every ice you can't target. That's not helpful, just messy.

I would suggest looking over the list of changes you've made and think about which ones will provide helpful and necessary feedback to the player at the time of playing above the feedback we already provide.

PR #6637, opened 2022-09-18
fix Datapike first sub implementation

francescopellegrini:
The ETR was missing

### DIFF
--- a/src/clj/game/cards/ice.clj
+++ b/src/clj/game/cards/ice.clj
@@ -1212,12 +1212,12 @@
                  end-the-run-if-tagged]})
 
 (defcard "Datapike"
-  {:subroutines [{:msg "force the Runner to pay 2 [Credits] if able"
-                  :async true
+  {:subroutines [{:async true
                   :effect (req (wait-for (pay state :runner (make-eid state eid) card :credit 2)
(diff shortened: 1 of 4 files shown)


src/clj/game/cards/ice.clj
```diff
@@ -1212,12 +1212,12 @@
...
 
 (defcard "Datapike"
-  {:subroutines [{:msg "force the Runner to pay 2 [Credits] if able"
-                  :async true
+  {:subroutines [{:async true
                   :effect (req (wait-for (pay state :runner (make-eid state eid) card :credit 2)
-                                         (when-let [payment-str (:msg async-result)]
-                                           (system-msg state :runner payment-str))
-                                         (effect-completed state side eid)))}
+                                         (if (:cost-paid async-result)
```
NoahTheDuke, line 1217: Oh man, we should be using this everywhere instead of checking the msg, good call.

NoahTheDuke (approved):

src/clj/game/cards/ice.clj
francescopellegrini: I applied a similar update to 2 more functions, I hope they are ok

PR #6647, opened 2022-09-27
malapert effect completes

NBKelly:
Malapert now gives the correct number of args to effect-completed.

Closes #6647

### DIFF
--- a/src/clj/game/cards/upgrades.clj
+++ b/src/clj/game/cards/upgrades.clj
@@ -900,7 +900,10 @@
                               (reveal state side target)
                               (shuffle! state side :deck)
                               (move state side target :hand)
-                              (effect-completed state side eid)))}}}]})
+                              (effect-completed state side eid)))}
+              :no-ability
+              {:effect (effect (system-msg "declines to use Malapert Data Vault")
+                               (effect-completed eid))}}}]})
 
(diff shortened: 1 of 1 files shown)


francescopellegrini:
[My fault](https://github.com/mtgred/netrunner/commit/7a59db6939c22be2a1df33caf31b084e1ce565db#diff-602d6cda215cb1685363535900de1f3aedab7857357de1d9c4f82d59bc5ac8bfR905-R908), sorry.. 🙇

NoahTheDuke:
I don't think this needs `effect-completed` at all. Given that the ability map isn't `:async`, it will be completed by `resolve-ability` naturally.

src/clj/game/cards/upgrades.clj
```diff
@@ -900,10 +900,7 @@
                               (reveal state side target)
                               (shuffle! state side :deck)
                               (move state side target :hand)
-                              (effect-completed state side eid)))}
-              :no-ability
-              {:effect (effect (system-msg "declines to use Malapert Data Vault")
-                               (effect-completed eid))}}}]})
```
francescopellegrini: Why do we want not to offer this feature?

NBKelly: oops, I removed too much :joy:

PR #6652, opened 2022-10-02
Meridian ends the encounter (rather than continuing to a nil state)

### LINKED ISSUE #6651: Run Stuck in Loop After Only Ice Removed
Hopefully this isn't a duplicate. I tried to find this in other Issues and didn't see anything, but it's possible I'm just not assembling the right search terms.

This evening, I was playing a game during which I had only [Meridian][meridian] installed protecting R&D. My opponent ran, the subroutine fired and they chose to accept the -1 point and continue the run, which removed the ice from protecting the server and into their score area as an agenda. This seems to have stuck us in a loop where they were continually stuck trying to approach an ice that wasn't there any more. You can see it in action (and our confusion in chat and how we worked around it) at [this timestamp in the replay][playthrough].

[meridian]: https://netrunnerdb.com/en/card/22028
[playthrough]: https://jinteki.net/replay/782b0baf-1293-460a-815a-65e5d8fd3898?n=16&d=0

I hope this is helpful. Let me know if I can answer any other questions. And if I don't respond here in a reasonable time, hit me up on Discord.

NBKelly:
Slightly bizarre thing to see in the code. I assume it worked at some point, since I remember playing Meridian.

Closes #6651

### DIFF
--- a/src/clj/game/cards/ice.clj
+++ b/src/clj/game/cards/ice.clj
@@ -2462,8 +2462,7 @@
                                  (do (system-msg state :runner "adds Meridian to their score area as an agenda worth -1 agenda points")
                                      (as-agenda state :runner card -1)
                                      (when current-ice
-                                       (continue state :corp nil)
-                                       (continue state :runner nil))
+                                       (encounter-ends state side eid))
                                      (effect-completed state side eid))))}]})
 
 (defcard "Merlin"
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
These kinds of effects have gone through many iterations, so I'm unsurprised that at some point they broke. The intention is that when the effect is triggered, the game automatically continues instead of waiting for the players to both hit "continue", as that can require communication between corp and runner. But it's all kind of hacky.

NoahTheDuke:
I suspect the other ICE that do the same (there are a couple) are also broken, but I haven't tried them out.

PR #6653, opened 2022-10-03
add missing waiting prompts

### LINKED ISSUE #6643: No Indicator that Runner is choosing for Fairchild 3.0 subroutine
When the Corp fires Fairchild 3.0 subroutines, there is no indication while the runner is choosing to pay 3 credits or trash an installed card.  It would be helpful to have an indicator that says the Runner is choosing, to avoid the Corp mistakenly firing the subroutines twice.

francescopellegrini:
Fixes #6643

### DIFF
--- a/src/clj/game/cards/agendas.clj
+++ b/src/clj/game/cards/agendas.clj
@@ -1117,6 +1117,7 @@
   {:on-score {:interactive (req true)
               :async true
               :prompt "Choose one"
+              :waiting-prompt "Corp to choose an option"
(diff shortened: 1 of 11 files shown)


francescopellegrini:
> Given how similar these are, should we instead strip them or change them to `:waiting-prompt true` and then if `(and (true? (:waiting-prompt ability)) (:choices ability))`, make the text `(str (other-side side) " to make a decision")` or something? Seems like a lot of effort to keep everything duplicated.

I think this is a great idea!

PR #6654, opened 2022-10-03
Small fixes to messages

francescopellegrini:
Remove multiple redundant "to" from messages, prefer decapitalize over str/lower-case, fix Cerebral Cast message

### DIFF
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -1883,8 +1883,7 @@
                                        (trash state side card {:cause-card card})
                                        (wait-for
                                          (gain-credits state side 3)
-                                         (wait-for (draw state side 3)
-                                                   (effect-completed state side eid)))))}}}
+                                         (draw state side eid 3))))}}}
                      card nil))}]
     {:derezzed-events [corp-rez-toast]
      :flags {:corp-phase-12 (req true)}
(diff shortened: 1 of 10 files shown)


src/clj/game/cards/assets.clj
```diff
@@ -1876,15 +1876,14 @@
...
+                       :msg "gain 3 [Credits] and draw 3 cards"
                        :yes-ability
                        {:async true
                         :effect (req (wait-for
                                        (trash state side card {:cause-card card})
                                        (wait-for
                                          (gain-credits state side 3)
                                          (wait-for (draw state side 3)
-                                                   (system-msg state side (str "uses " (:title card) " to gain 3 [Credits] and draw 3 cards"))
                                                    (effect-completed state side eid)))))}}}
```
NoahTheDuke, line 1887: Given that you're removing the `system-msg`, you can simplify this to not use `wait-for`, something like `(wait-for (gain=-credit state side 3) (draw state side eid 3))`

francescopellegrini: Do you mean like this?

```clojure
(wait-for
  (trash state side card {:cause-card card})
  (wait-for
    (gain-credits state side 3)
    (draw state side 3)
```

NoahTheDuke: Looks like you figured it out!

PR #6657, opened 2022-10-03
fix Economic Warfare credits check

francescopellegrini:
While fixing Machicolation A sub, I noticed Econ Warfare was also using can-pay? but, since they are both worded as "if able", they should not consider hosted credits or similia.


NoahTheDuke:
Huh, I didn't know that was a distinction! This game is too damn complicated, lol.

PR #6658, opened 2022-10-03
rework waiting-prompt as boolean key

francescopellegrini:
As suggested [here](https://github.com/mtgred/netrunner/pull/6653#issuecomment-1265496081).

### DIFF
--- a/src/clj/game/core/prompts.clj
+++ b/src/clj/game/core/prompts.clj
@@ -6,6 +6,7 @@
     [game.core.toasts :refer [toast]]
     [game.macros :refer [when-let*]]
     [game.utils :refer [pluralize]]
+    [jinteki.utils :refer [other-side]]
     [medley.core :refer [find-first]]))
 
 (defn choice-parser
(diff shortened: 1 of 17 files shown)


src/clj/game/core/prompts.clj
```diff
@@ -43,13 +44,13 @@
...
                (pos? (count choices)))
-       (when waiting-prompt
+       (when (true? waiting-prompt)
          (add-to-prompt-queue
            state (if (= :corp side) :runner :corp)
            {:eid (select-keys eid [:eid])
             :card card
             :prompt-type :waiting
-            :msg (str "Waiting for " waiting-prompt)}))
+            :msg (str "Waiting for " (str (other-side side)) " to make a decision")}))
```
NoahTheDuke, line 53: Likewise, here you can say `(str "Waiting for " (if (true? waiting-prompt) (str (other-side side)) " to make a decision" waiting-prompt))` or something.

NoahTheDuke: We could also do something more complex here, if you think it would be helpful. You previously differentiated between "make a decision" and "choose an option" so we could have `(cond (and (true? waiting-prompt) (:choices ability)) " to choose an option" (true? waiting-prompt) " to make a decision" :else waiting-prompt)` or similar.

What do you think? Do we need that level of granularity?

francescopellegrini: I think it would become too hidden in the internal implementation and not very clear to developers when to use which and why.

NoahTheDuke: Cool, good call.

PR #6659, opened 2022-10-03
remove :activatemsg key handling

francescopellegrini:
TIL `:activatemsg` is a thing 😄


francescopellegrini:
> Honestly, do we even need this? Should we just remove it? It's barely used and mostly covered by existing `:msg` functionality.

I'm all in for removing it since only 2 cards make use of it.

PR #6664, opened 2022-10-04
Explicit reveals for jinja city grid

### LINKED ISSUE #6663: :runner-abilities on cards installed in servers gives an invisible/out of frame dialog box
At least this is what it seems like to me.

I think the default "box appears above the card" kind of breaks down in this edge case.

It's relevant because I want to give jinja city grid an option to explicitly reveal cards, but may be relevant for future cards as well.

abilities

NBKelly:
Right now I've solved the "runner abilities don't appear to work" issue by having the dropdowns drop "down" when the ability belongs to a card in the 'corp-top-opponent' area. 

The exact offset number (115%) can probably be tweaked for best aesthetics, I don't really have a good eye for that. 

For now, Closes #6663.

example example3 example2

Next up will be
1) implementing the ability for jinja
2) maybe implementing it for corporate defector as well

### DIFF
--- a/src/clj/game/cards/upgrades.clj
+++ b/src/clj/game/cards/upgrades.clj
@@ -8,7 +8,7 @@
    [game.core.bad-publicity :refer [lose-bad-publicity]]
    [game.core.board :refer [all-active-installed all-installed card->server
                             get-remotes server->zone server-list]]
-   [game.core.card :refer [agenda? asset? can-be-advanced?
+   [game.core.card :refer [agenda? asset? can-be-advanced? card-index
                            corp-installable-type? corp? get-card get-counters get-zone
                            has-subtype? ice? in-discard? in-hand? installed? operation? program? resource? rezzed?
                            runner? upgrade?]]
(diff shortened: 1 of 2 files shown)


NBKelly:
So as far as I can tell, this works.

* There's a `runner-ability` on Jinja City Grid which allows the runner to say yes/no to explicit reveals
* The value they choose is propagated across all active copies of Jinja city grid (no need to click another one if you happen to trash the one you chose)
* it gets copied to a newly rezzed Jinja city grid

What it does:
* If the runner says yes, then you get a little box popping up forcing you to read the cardname whenever the corporation reveals a card

runner-prompt

Why:
* The physical act of revealing/showing a card is a big part of functional memory. Hiding that in a chat log, which is usually overwhelmed with information, strongly effects the power level of these cards (in my opinion).

Getting the runner prompt to not interact with the corp prompt/ability requires a little bit of an ugly hack. If there's a better way to do it, I would be happy to change it.
* It can't complete the eid because that ruins things on the corp side
* Making the prompt async and deliberately failing to complete the eid appears to work

NoahTheDuke:
Still says WIP, so any updates on this?

NBKelly:
oh, this is fully functional, it's more a question of is that ugly hack I did going to cause any damage/other side-effects?

If no, then this is perfectly good to go.

NoahTheDuke:
Okay, I never replied or merged this but I think that it's viable. I wonder if we could generalize this, maybe put it into `reveal`? That way we don't have to mess with eids.

NoahTheDuke:
lmao i can't seem to merge this. you have commit power so you know what? merge at your pleasure. i trust you to get it right.

NBKelly:
I think I'm going to shelve this in favor of writing a direct `'explicit-reveal` layer that just plugs into our reveal interface and uses a setting in the side pane.

Issue #6665, opened 2022-10-04
Palisade has the wrong strength if moved from a central to a remote server


tolaasin:
Palisade installed on archives, rezzed
tao moves it to a remote on steal
remote palisade has 2 strength instead of the 4 expected.

francescopellegrini:
See also #6620

francescopellegrini:
Fixed by #6667

Issue #6412, opened 2022-07-05
Winchester having all the subs even if protecting R&D


PreNic:
Yesterday I was running against a Winchester being installed on R&D and the "End the run" sub was listed when I tried to break the subs. Moreover, the trace of the third sub would start if I did not break it.

francescopellegrini:
Releated to #6247

NoahTheDuke:
I can't recreate this. Any details?

francescopellegrini:
Might have been fixed by #6667

francescopellegrini:
Might as well close this.

PR #6667, opened 2022-10-06
update constant effects when swapping ice

### LINKED ISSUE #6247: Winchester retains extra sub when Tao'd off of HQ
My opponent was Tao and moved my Winchester to a remote server and we still saw the ETR subroutine when encountering the ICE. It also came up in the interactive modals I (the corp) saw when trying to help resolve the issue during the game.

Not sure if this is specific to Tao or if other ways of moving ICE would have this same outcome.

### LINKED ISSUE #6620: When Rime gets moved via Tao, it has 0 strength
Minor bug, that can be easily fixed in game:
After an agenda got stolen, Rime got moved from R&D to Archives. It was then the only ICE on Archives. Rime usually gives itself strength +1, too. But it had 0 strength. It was a bit relevant, since there was an Atman with strength 1 out.
We were able to fix it, via derezzing and rezzing it.

francescopellegrini:
Fixes #6620 and #6665.
It should fix #6247 and #6412 as well, I hope.

### DIFF
--- a/src/clj/game/core/moving.clj
+++ b/src/clj/game/core/moving.clj
@@ -11,7 +11,7 @@
     [game.core.finding :refer [get-scoring-owner]]
     [game.core.flags :refer [can-trash? card-flag? cards-can-prevent? get-prevent-list untrashable-while-resources? untrashable-while-rezzed? zone-locked?]]
     [game.core.hosting :refer [remove-from-host]]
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
Ah perfect, thank you for catching this.

francescopellegrini:
> Ah perfect, thank you for catching this.

Still working on condition counters to work properly.. 😓

PR #6669, opened 2022-10-08
use empty? when checking collection size

francescopellegrini:
Also fix small typo on sabotage tests (`disard`)


NoahTheDuke:
Can also check for (zero? (count ...))

PR #6682, opened 2022-10-14
Add msg where missing

francescopellegrini:
Also use new wording on Omar

### DIFF
--- a/src/clj/game/cards/ice.clj
+++ b/src/clj/game/cards/ice.clj
@@ -2466,15 +2466,16 @@
   {:subroutines [{:label "Gain 4 [Credits] and end the run"
                   :waiting-prompt true
                   :prompt "Choose one"
-                  :choices ["End the run" "Add Meridian to score area"]
+                  :choices ["Corp gains 4 [Credits] and end the run" "Add Meridian to score area"]
+                  :msg (msg (if (str/starts-with? target "Corp")
+                              "gain 4 [Credits] and end the run"
+                              (str "force the Runner to " (decapitalize target))))
                   :player :runner
                   :async true
-                  :effect (req (if (= target "End the run")
-                                 (do (system-msg state :corp "uses Meridian to gain 4 [Credits] and end the run")
-                                     (wait-for (gain-credits state :corp 4)
-                                               (end-run state :runner eid card)))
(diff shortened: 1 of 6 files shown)


src/clj/game/cards/ice.clj
```diff
@@ -2466,15 +2466,16 @@
...
-                                               (end-run state :runner eid card)))
-                                 (do (system-msg state :runner "adds Meridian to their score area as an agenda worth -1 agenda points")
-                                     (as-agenda state :runner card -1)
+                  :effect (req (if (str/starts-with? target "Corp")
+                                 (wait-for (gain-credits state :corp 4)
+                                           (end-run state :runner eid card))
+                                 (do (as-agenda state :runner card -1)
                                      (when current-ice
                                        (encounter-ends state side eid))
                                      (effect-completed state side eid))))}]})
```
NoahTheDuke, line 2481: Can't use the eid and then complete it. Gotta be an if, not when.

src/clj/game/cards/identities.clj
```diff
@@ -1521,12 +1521,12 @@
...
              :req (req (and (get-in card [:special :omar-run])
                             (= :archives (-> run :server first))))
-             :prompt "Treat as a successful run on which server?"
+             :prompt "Choose one"
              :choices ["HQ" "R&D"]
+             :msg (msg "change the attacked server to " target)
              :effect (req (let [target-server (if (= target "HQ") :hq :rd)]
                             (swap! state assoc-in [:run :server] [target-server])
-                            (trigger-event state :corp :no-action)
-                            (system-msg state side (str "uses Omar Keung: Conspiracy Theorist to make a successful run on " target))))}
```
NoahTheDuke, line 1529: Why remove this?

NoahTheDuke (changes requested):

src/clj/game/cards/identities.clj
francescopellegrini: I moved it to `:msg`

src/clj/game/cards/ice.clj
francescopellegrini: cc @NBKelly (see #6652)

francescopellegrini: Indeed I just tested it locally and it creates an infinite loop of encounter and approach prompts

PR #6691, opened 2022-10-17
Add autoresolve to abilities firing from archives

### LINKED ISSUE #6690: Feature request: Allow setting cards like Mavirus in Archives to auto resolve
It would be great if I could set Mavirus to autoresolve when accessed from Archives.

In many matchups, it doesn't matter whether I purge or not because they don't have any Virus cards in their deck. In other matchups, I will gladly purge every chance I get. It would be great to not be shown this prompt when the runner accesses cards like Mavirus from Archives:

Screen Shot 2022-10-16 at 5 02 28 PM

(It gets even more annoying when there are multiple copies of Mavirus in Archives.)

I'm imagining this could work by having some sort of mechanism similar to how cards like Lat and 419 allow you to set "Always / Never / Ask" for their prompt by clicking on them.

Screen Shot 2022-10-16 at 5 02 51 PM

I'm not sure if there's precedent for doing things like this with cards in Archives, so forgive my ignorance of how difficult this feature request may be.

Happy to provide further details upon request.

francescopellegrini:
Fixes #6690

### DIFF
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -2194,10 +2194,12 @@
    :access {:optional
             {:waiting-prompt true
              :prompt "Place 1 advancement token on a card that can be advanced?"
+             :autoresolve (get-autoresolve :auto-fire)
              :yes-ability {:msg (msg "place 1 advancement token on " (card-str state target))
                            :prompt "Choose a card to place an advancement token on"
                            :choices {:card can-be-advanced?}
-                           :effect (effect (add-prop target :advance-counter 1 {:placed true}))}}}})
+                           :effect (effect (add-prop target :advance-counter 1 {:placed true}))}}}
+   :abilities [(set-autoresolve :auto-fire "Space Camp")]})
 
 (defcard "Spin Doctor"
   {:on-rez {:async true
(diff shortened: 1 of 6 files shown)


NoahTheDuke:
You can set autoresolve on cards in archives?

francescopellegrini:
> I didn't know we set up abilities on cards in archives.

Technically it's possible, but I found out the UI doesn't show the prompt 😭

francescopellegrini:
I'm going to create a separate PR for changes to cards non related to Archives

NBKelly:
Neat idea.

Crowdfunding might be a tricky one (since the autoresolve should get cleared whenever it's installed or trashed). Same for subliminal. I wonder if we could check (when the cards get trashed/revived) if the last zone/version of the card had an autoresolve, and force it to persist/copy the key over.

NoahTheDuke:
Any status updates on this?

francescopellegrini:
> Any status updates on this?

I'm sorry, but I feel like I'm not skilled enough to tackle the UI part.. also, as @NBKelly mentioned, keeping track of the user preference on cards moving in and out of a zone could be tricky and I think it could become a usability issue.

PR #6692, opened 2022-10-17
Add autoresolve where missing

francescopellegrini:
and fix some other ones.


src/clj/game/cards/hardware.clj
```diff
@@ -1432,7 +1432,7 @@
              :optional
              {:req (req (first-event? state side :successful-run))
               :player :runner
-              :autoresolve (get-autoresolve :auto-fire)
+              :autoresolve (get-autoresolve :auto-gain-credit-and-scry-stack)
```
NoahTheDuke, line 1435: This doesn't need to change cuz it's not user facing and there aren't multiple auto resolves.

PR #6730, opened 2022-11-27
WIP Card disable overhaul

NBKelly:
Posting this now so it's out there/started.

So far I have:
* Modified all the non-identity-disabling disabling cards to make floating/constant effects (this breaks all of their tests for now)
* On-rez and on-install (supposedly) check to see if a card is disabled. I still need to check these work at all.

There is still a ton to do.


NBKelly:
I think I started this in the wrong way, so I'm going to redo it

PR #6732, opened 2022-11-28
WIP re-implement disabling cards

NBKelly:
Doing this in a less eclectic fashion this time.

I'm picking one card (this time, rumor mill), changing it to a constant/floating effect, then working backwards to get things functional.

So far, the following changes have been made:
* `:on-rez` effects require the card to not be disabled
* Most of the effects that looked for a `:disabled` key now check `:any-effect`. The outliers at the moment are card-init, disable-card, and diffs
* I still need to add a "gather all disabled cards and do changes" helper that goes in each checkpoint and fake-checkpoint

So far, the following cards have been changed:
* Rumor Mill (the fake checkpoint can maybe go away, since it trashes anyway)


NoahTheDuke:
Would love to see this worked on more.

NoahTheDuke:
This has been done in #7400!

Issue #6729, opened 2022-11-27
Clean up all of the 'TODO' statements in code


NBKelly:
There's a few of these that have accumulated over the years

- [ ] `new-game-button` in 'src/cljs/nr/lobby.cljs' - "rewrite this check"
- [ ] `load-replay-button` in 'src/cljs/nr/lobby.cljs' - "rewrite this check"
- [ ] `game-lobby` in 'src/cljs/nr/lobby.cljs' - "make starting a game from the deckbuilder work again"
- [ ] 'ws/event-msg-handler :angel-arena/run-update' - implement this feature
- [ ] 'broadcast-to!` in `src/clj/web/ws.clj` - multiple go blocks could compete, this might be a source of black screens sometimes. Update/fix this.
- [ ] `check-for-inactivity` in `src/clj/web/angel_arena.clj` - maybe make this an option for all games (notify lobby of inactivity before you get booted)
- [ ] `ugly-counter-hack` in `src/clj/came/core/engine.clj` - find a way to no longer need this hack
- [ ] `derez` in `src/clj/game/core/rezzing.clj` - make this asynchronous (and maybe let it complete an eid too)
- [ ] `register-successful-run` in `src/clj/game/core/runs.cljs` - We queue a `pre-successful-run` event in here, and it's implied we instead want these abilities (sneakdoor/omar) to instead prevent/replace the successful run
- [ ] `card-str` in `src/clj/game/core/to_string.clj` - add naming of scoring area of corp/runner
- [ ] `card-flag?` in `src/clj/game/core/flags.clj` - Add a register for mutable state card flags, separate from this
- [ ] `prevent-draw/jack-out/prevent-current/lock-zone/release-zone` in `src/clj/game/core/flags.clj` - look into migrating these to turn-flags and run-flags (or better yet, maybe they can be floating/constant effects?)
- [ ] `find-undefined-locals` in `src/clj/game/macros.clj` - switch to requiring that passage if/when possible, instead of in-lining the function
- [ ] `Trade-In` in `src/clj/game/cards/events.clj` - I'm not sure, but it's demanding attention
- [ ] `Hunting Grounds` in `src/clj/game/cards/resources.clj` - when suppression is fixed (has this happened yet?), should be `(:ice context)`
- [ ] `Jackpot!` in `src/clj/game/cards/resources.clj` - this is async, but it's getting called in move (which is sync) - issue dated oct `20
[...]

PR #6736, opened 2022-12-02
implement name-zone for :scored

francescopellegrini:
See #6729

### DIFF
--- a/src/clj/game/core/to_string.clj
+++ b/src/clj/game/core/to_string.clj
@@ -20,7 +20,6 @@
                          installed-ice " protecting "
                          (is-root? zone) " in the root of "
                          :else " in ")
-                       ;TODO add naming of scoring area of corp/runner
                        (zone->name (or (second zone) zone)) ;; handles [:hand] as well as [:servers :hq]
                        (when installed-ice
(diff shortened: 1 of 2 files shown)


src/clj/game/core/to_string.clj
```diff
@@ -20,7 +20,6 @@
                          installed-ice " protecting "
                          (is-root? zone) " in the root of "
                          :else " in ")
-                       ;TODO add naming of scoring area of corp/runner
```
NoahTheDuke, line 23: `name-zone` isn't called in this function lol. The issue is that a stolen agenda will be in `:scored` (with no indication it's runner's score zone), meaning we have to look for the card in both of the score zones and then we can use that to call `name-zone`. Maybe use `find-cid` on both score areas and whichever it's found in, pass that side keyword to `name-zone`.

francescopellegrini: LOL!
Why not replace `zone->name` with `name-zone` in `card-str`?

NoahTheDuke: I think cuz the current one pulls the remote number out.

Issue #6735, opened 2022-12-02
Start of turn trigger for Lewi Guilherme cannot be triggered before Dadiana Chacon, and also adds weird text to the chat


Planet-Hopper:
Link to bug replay: https://jinteki.net/bug-report/b8f17d0d-9129-46bf-8046-7afa6c09e1f3?b=0

Description:

Having Lewi and Dadiana installed at start of turn with 6 credits in the bank should allow me to:
Trigger Lewi first, losing a credit and taking me to 5 total
Trigger Dadiana, gaining 1 credit because I am now below 6

However, the game does not give me the option to trigger Lewi first; it just triggers Dadiana automatically and then asks if I want to trigger Lewi.

PR #6740, opened 2022-12-03
make Lat ability interactive

### LINKED ISSUE #6715: Chameleon & Lat triggers don't work together
I think Chameleon's "When your discard phase ends" and Lat's "When your turn ends" triggers are the same. This means that the player should be able to choose the order and even get a Lat draw if the corp is at 6 cards and Lat gets to 6 after returning Chameleon to grip. Currently there's no choices being presented (and I think Chameleon is triggered last).

For reference the CR1.6 relevant steps from runner's turn order:

5.7.2. During the Runner’s discard phase, the following steps occur in order:
a. The Runner discards one card at a time from the grip until the number of cards in
the grip is less than or equal to the Runner’s maximum hand size.
b. A paid ability window occurs, in which players may use paid abilities and the Corp
may rez non-ice cards. (P) (R)
c. If the Runner has any unspent [click], the Runner loses those [click].
**d. The Runner’s turn formally ends. Conditions related to the turn or discard phase
ending are met.**
e. The Runner’s discard phase and turn are complete. Play proceeds to the Corp’s
turn.

francescopellegrini:
Fixes #6715


francescopellegrini:
> Change Lat's ability to have the req check in the :effect instead of in the :req.

Wouldn't this change be required for making other abilities interactive too? Like the interaction reported in #6735?

NoahTheDuke:
I don't know about those cards, but Lat's ability says: "When your turn ends, if you have the same number of cards in your grip as the Corp has in HQ, you may draw 1 card."

The "if" clause is the conditional and it is nested inside of the "when", meaning that the whole ability always fires at turn end and then at ability resolution, the total binder of cards in hand is checked.

PR #6765, opened 2022-12-10
Fixes and testing for Issuaq

### LINKED ISSUE #6747: Issuaq Wins on the spot
Link to bug replay: https://jinteki.net/bug-report/f4b933de-7df6-4b40-bbfe-0358699e0ffe?b=0

Description: Issuaq scored a Blood In The Water that triggered the Issuaq power counter ability and won instantly when it shouldn't. Guess some comparator is broken somewhere?

poi2000:
Created what should be a fix for Issuaq Adaptics, along with a few tests. I'm not terribly familiar with either Clojure or the jnet codebase, so the code might not be as up to style as it could be, but testing appeared to work locally.

For reference, I would have modified the agenda point requirement using the starting requirement and the number of power counters on the ID, but the starting requirement is not constant(thanks draft), and doesn't appear to be stored anywhere.

Fixes #6747

### DIFF
--- a/src/clj/game/cards/identities.clj
+++ b/src/clj/game/cards/identities.clj
@@ -842,7 +842,7 @@
              :msg "put 1 charge counter on itself"
              :effect (req (add-counter state side card :power 1)
                           (swap! state assoc-in [:corp :agenda-point-req]
-                                 (get-counters (get-card state card) :power))
+                                 (- (get-in @state [:corp :agenda-point-req]) 1))
                           (check-win-by-agenda state))}]})
 
 (defcard "Jamie \"Bzzz\" Micken: Techno Savant"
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/identities.clj
```diff
@@ -842,7 +842,7 @@
              :msg "put 1 charge counter on itself"
              :effect (req (add-counter state side card :power 1)
                           (swap! state assoc-in [:corp :agenda-point-req]
-                                 (get-counters (get-card state card) :power))
+                                 (- (get-in @state [:corp :agenda-point-req]) 1))
```
NoahTheDuke, line 845: This is better written with `dec`.

NoahTheDuke (changes requested):
This looks great! Thank you! Just one small change please.

PR #6775, opened 2022-12-10
fix World Tree bug when accessing multiple cards

### LINKED ISSUE #6757: World Tree prompt appears out of order
https://jinteki.net/bug-report/0f19f818-9ee6-48d7-a5e4-d61dce438947?b=0

When running on a Server with only World Tree installed, the card gets accessed first, then the World Tree prompt appears.
When multiple cards trigger on successful run, the prompts get very out of order, esp. with DreamNet + The Class Act.

francescopellegrini:
Fixes #6757


NoahTheDuke:
I don't mind this change but i should note that all that's needed in the original code is :async true in the base event ability lol

PR #6779, opened 2022-12-10
superdeep borehole fix and tests

### LINKED ISSUE #6776: Superdeep Borehole causes win to trigger on rez
I was on 6 points, and rezzed a superdeep borehole prior to the start of my turn.  Then the game announced I won and showed the ending stats screen.

### LINKED ISSUE #6778: Game immediately ended with a win when Superdeep borehold was rezzed.
Link to bug replay: https://jinteki.net/bug-report/2b2e5215-5a86-4192-a54b-69ffbebd8962?b=0

Description:

prior to clicking "Start turn", I rezzed an installed Superdeep borehole. The game ended in an immediate win for me.

NBKelly:
Just need to say `(get-card state card)` instead of `card`, it was fetching the pre-rez version with 0 counters on it before.

I also added a test to make sure it doesn't happen, and another test to verify the Light the Fire!/other disabe-card interactions work right.

Closes #6778, Closes #6776

### DIFF
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -2340,7 +2340,7 @@
                           (gain-bad-publicity state :corp eid 1))}
             {:event :counter-added
              :req (req (and (same-card? card target)
-                            (not (pos? (get-counters card :power)))
+                            (not (pos? (get-counters (get-card state card) :power)))
                             (:borehole-valid (:special card))))
              :msg "win the game"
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
Good catch. I wondered why `card` wasn't being updated until i remembered that `target` here IS the updated card, lol. Using `target` could also be a fix, but certainly not worth changing your code for.

NoahTheDuke:
Makes me think we need an :on-counter-added self-event instead of this (same-card) nonsense.

PR #6784, opened 2022-12-11
add missing label to subs

francescopellegrini:
I noticed Vampyronassa and Unsmiling Tsarevna missed the label on an optional sub, so I checked all ice for the same problem.


NoahTheDuke:
Huh, shouldn't this be caught by the label test?

PR #6797, opened 2022-12-11
fix Mr. Hendrik not using install-access-trigger function

francescopellegrini:


src/clj/game/cards/upgrades.clj
```diff
@@ -1034,33 +1034,25 @@
...
-                                                  (if (= target "Take 1 core damage")
-                                                       (damage state :corp eid :brain 1 {:card card})
-                                                       (do (lose-clicks state :runner (:click runner))
-                                                           (effect-completed state side eid))))}
-                                       card nil)))}}}})
+  (installed-access-trigger
+    2
+    {:async true
+     :effect
+     (effect (continue-ability {:player :runner
```
NoahTheDuke, line 1041: There should be no need for the nested ability, right?

francescopellegrini: Unfortunately I think there is: the `install-access-trigger` function adds the ability cost to the ability provided, so the Runner ends up paying the cost instead of the Corp

NoahTheDuke: Ah okay, cool. Bummer that's how i built it.

PR #6800, opened 2022-12-11
Fix missing waiting prompts

francescopellegrini:


NoahTheDuke:
Actually, wanna make a different change here? I don't remember where we check for waiting-prompt, but try changing it to look for async first, and then waiting prompt. All async prompts should wait, I'm thinking.

francescopellegrini:
> Actually, wanna make a different change here? I don't remember where we check for waiting-prompt, but try changing it to look for async first, and then waiting prompt. All async prompts should wait, I'm thinking.

You mean check for `:async true` presence and show the waiting prompt so that each instance can avoid specifying the additional key `:waiting-prompt true`?

NoahTheDuke:
Yeah, I'm wondering if we can get this feature "for free". Having the waiting-prompt key is a way too specific the wording but otherwise everything async will create and destroy it's own waiting prompt.

PR #6806, opened 2022-12-11
rename brain to core damage in prompts

alangfiles:
This PR just renames `brain damage` to `core damage` across the visual strings used. 

It leaves the underlying functions like `do-brain-damage`, etc untouched. 

This PR is just concerned with the visual strings/prompts.

This is similar to the work done in: https://github.com/mtgred/netrunner/pull/6592 except it finds a few more places to update, and this PR doesn't add a `take-core` command. It seemed like that PR was a little stale, so I took a go at just updating the strings alone for this one.


NoahTheDuke (approved):

NoahTheDuke:
Thanks so much for pulling this out! I really appreciate it.

locks:
Thanks for picking this up!

PR #6830, opened 2022-12-15
Fix The Root cannot pay for advancing cards

### LINKED ISSUE #6714: The Root does trigger when advancing
The Root does not trigger to discount the basic action when it still has recurring credits left on it

francescopellegrini:
Fixes #6714

### DIFF
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -2491,7 +2491,7 @@
 
 (defcard "The Root"
   {:recurring 3
-   :interactions {:pay-credits {:req (req (or (some #(= % (:source-type eid)) [:advance :corp-install :rez])
+   :interactions {:pay-credits {:req (req (or (#{:advance :corp-install :rez} (:source-type eid))
                                               (is-basic-advance-action? eid)))
                                 :type :recurring}}})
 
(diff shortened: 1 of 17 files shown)


src/clj/game/cards/basic.clj
```diff
@@ -82,14 +84,13 @@
...
                 :msg (msg "advance " (card-str state target))
-                :req (req (can-advance? state side target))
-                :effect (effect (update-advancement-requirement target)
-                                (add-prop (get-card state target) :advance-counter 1)
-                                (play-sfx "click-advance")
-                                (effect-completed eid))}
+                :req (req (and (can-advance? state side target)
+                               (can-pay? state side
+                                         (assoc eid :source :action :source-type :advance)
+                                         card nil [:credit 1 :click 1])))
```
NoahTheDuke, line 92: Does this cover when there are additional costs on the click ability?

francescopellegrini: Can you make an example of a card creating additional costs on the click ability? I couldn't find any.. :/

NoahTheDuke: [Enhanced Login Protocol](https://netrunnerdb.com/en/card/06022) changes the cost of a basic action:

> As an additional cost to take the basic action to run a server for the first time each turn, the Runner must spend click.

It's true that a card doesn't exist that changes the "advance a card" basic action. But we should still be writing abilities like those cards do exist.

NoahTheDuke (changes requested):

PR #6834, opened 2022-12-15
Fix ice abilities not being a cost

francescopellegrini:
Also net, meat, and core damage not setting `:card` key which caused First Responders not to trigger on Formicary's net damage.

### DIFF
--- a/src/clj/game/cards/ice.clj
+++ b/src/clj/game/cards/ice.clj
@@ -2192,37 +2192,30 @@
    :subroutines [end-the-run]})
 
 (defcard "Kamali 1.0"
-  (letfn [(better-name [kind] (if (= "hardware" kind) "piece of hardware" kind))
-          (runner-trash [kind]
-            {:prompt (str "Choose an installed " (better-name kind) " to trash")
-             :label (str "Trash an installed " (better-name kind))
-             :msg (msg "trash " (:title target))
-             :async true
-             :choices {:card #(and (installed? %)
-                                   (is-type? % (capitalize kind)))}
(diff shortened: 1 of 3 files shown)


src/clj/game/cards/ice.clj
```diff
@@ -2721,28 +2703,26 @@
...
                                 (damage state :runner eid :net net-dmg {:card card}))
-                            (do (system-msg state :corp
-                                            (str "uses Mlinzi to trash "
-                                                 (enumerate-str (map :title (take mill-cnt (:deck runner))))
-                                                 " from the runner's stack"))
-                                (mill state :runner eid :runner mill-cnt))))})]
-    {:subroutines [(net-or-trash 1 2)
-                   (net-or-trash 2 3)
-                   (net-or-trash 3 4)]}))
+                            (wait-for (pay state :runner (make-eid state eid) card [:trash-from-deck mill-cnt])
```
NoahTheDuke, line 2720: > Do 1 net damage unless the Runner trashes the top 2 cards of the stack.

This is a cost?

francescopellegrini: Yes in the sense that if the runner must pay it or else suffer the net damage. As in Turing, Public Trail, or Data Ward the main effect can be avoided by the Runner if they pay an alternative cost, which in the case of Formicary is the 2 net damage. The Runner cannot prevent this damage or else Formicary would end the run, the same that occurs with Obokata Protocol's steal cost.

NoahTheDuke: Cool thanks

Issue #6107, opened 2021-10-27
Daily Business Show & The Class Act won't let you select a card to move to the bottom


jonniemarbles:
As stated in the title, this bug sometimes occurs when attempting to choose a card with DBS or TCA. 

This bug is intermittent and seems to be resolved by:
Undoing the turn
Leaving and rejoining

It is variable whether it is the active player (the one drawing the cards) or their competitor who needs to leave and rejoin. 

I'm sorry I can't offer more detail at this time, I will try to replicate the behaviour in my next few games and provide things like game state etc.

francescopellegrini:
See also #6107

francescopellegrini:
Fixed by #6265

francescopellegrini:
Might as well close this.

PR #6845, opened 2022-12-18
implement The Class Act latest wording

### LINKED ISSUE #6347: Class act leaves a card in the stack when drawing your entire deck (unless it's one card)

francescopellegrini:
Fixes #6347 and #6107


NoahTheDuke:
When did the wording change again??? How many times will we have to fix her? 😭

francescopellegrini:
> When did the wording change again??? How many times will we have to fix her? 😭

Current Official Text:

When your discard phase ends, if you installed this resource this turn, draw 4 cards.
[interrupt] → The first time each turn you would draw any number of cards, look at the
top X cards of your stack. Add 1 of those cards to the bottom of your stack. X is equal
to the number of cards you will draw plus 1.

src/clj/game/core/drawing.clj
```diff
@@ -46,50 +46,50 @@
...
-                   (doseq [c (get-set-aside state side set-aside-eid)]
-                     (move state side c :hand))
-                   (effect-completed state side eid))
-                 (let [draw-event (if (= side :corp) :corp-draw :runner-draw)]
-                   (swap! state update-in [side :register :currently-drawing] conj drawn)
-                   (queue-event state draw-event {:cards drawn
-                                                  :count drawn-count})
-                   (wait-for
-                     (checkpoint state nil (make-eid state eid) nil)
+       (wait-for (trigger-event-sync state side (make-eid state eid) (if (= side :corp) :pre-corp-draw :pre-runner-draw) n)
```
NoahTheDuke, line 49:
```suggestion
       (wait-for (trigger-event-simult state side (make-eid state eid) (if (= side :corp) :pre-corp-draw :pre-runner-draw) n)
```

src/clj/game/core/drawing.clj
```diff
@@ -46,50 +46,50 @@
    (if (zero? n)
      (effect-completed state side eid)
      (do
```
NoahTheDuke, line 48: This should be removed because `wait-for` has an implicit `do` already.

PR #6857, opened 2022-12-31
fix Reaver not triggering when multiple cards are trashed simultaneously

### LINKED ISSUE #6856: Trashing cards to Prey doesn't trigger Wasteland and Reaver (was first time I trashed my cards that turn)
I suspect its a Prey-specific issue and not an issue with trashing multiple cards in general, because when taking more than 1 damage and preventing it with heartbeat Reaver and Wasteland trigger.

francescopellegrini:
Fixes #6856

### DIFF
--- a/src/clj/game/cards/programs.clj
+++ b/src/clj/game/cards/programs.clj
@@ -2393,7 +2393,8 @@
              :interactive (req true)
              :once-per-instance true
              :req (req (and (some #(installed? (:card %)) targets)
-                            (first-event? state side :runner-trash)))
+                            (first-event? state side :runner-trash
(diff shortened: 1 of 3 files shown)


src/clj/game/cards/programs.clj
```diff
@@ -2394,7 +2393,7 @@
              :interactive (req true)
              :once-per-instance true
              :req (req (and (some #(installed? (:card %)) targets)
-                            (first-installed-trash? state side)))
+                            (first-event? state side :runner-trash)))
```
NoahTheDuke, line 2396: Reaver requires an installed card, and this change removes that.

NoahTheDuke (changes requested):

PR #6876, opened 2023-01-08
placeholder impl for hush

NBKelly:
This is a boilerplate "it's at least convenient" non-implementation. You can host it on ice, you can move it to other ice, and magnet doesn't (or shouldn't) silence it. 

There's also reminder text on it so players know how to resolve cards like anansi.

I've put TODOs in so we can remember to fix it later.


NoahTheDuke:
The ability is missing actually performing the host lol

NBKelly:
heh, no clue how I missed that.

PR #6884, opened 2023-01-15
Vanilla docker compose fails with nr.dev required namespace not available.

smmathews-bw-boston:
to mirror namespace prod, move dev to src/cljs/nr

fixes issue #6883


BlindGuyNW:
Is it worth either merging this or removing the Docker bits if they aren't well-supported? I'm not sure what consequences moving files around like this has (very new to the project) but it seems a bit odd. 

I'd like to understand why the underlying issue happens.

PR #6900, opened 2023-01-29
Add initial Japanese translation

csbisa:
Add Japanese translations for most of the entries in `translations.cljs`. We plan on following up with more edits later.


csbisa:
Ping. Would appreciate it if this gets merged.

PR #6902, opened 2023-02-04
Support for translated card data

csbisa:
This adds support for displaying translated card data, namely:
- The card browser will display the translated title, data, and formats if available.
- The deck builder will display transalted card names if available. (The text listing will always be in English.)
- The search functionality (in the card browser and deck builder) will consider both English and the user selected language. For example, with a Japanese language setting, searching for Overclock or オーバークロック will return the same card, always displayed as オーバークロック. There is no behavior change if the selected language is English.
- The card pattern search text search for chat will consider English and the selected language. For example, with a Japanese language setting, typing in Overclock or オーバークロック will parse the text, using the original language of the text while displaying the same card image pop-up. There is no behavior change if the selected language is English.

Localized card data is pulled from raw_data.edn under `:cards-<lc>` (changes there will be going to the other repo) and is exposed via `/cards/lang/<lc>`. The client will fetch localized data for the user-selected language and will merge it into card data under `:localized`, along with language checks in order to re-generate the data on a language change.

~~Note that translated data does not currently exist in the raw data. I'm looking for a contact to follow up with to discuss the best way to get the translated data in netrunner-cards-json to the data used by jnet. In the meantime, I've tested these changes by manually editing the raw data.~~

jnet-jp
jnet-ko

### DIFF
--- a/src/cljs/nr/utils.cljs
+++ b/src/cljs/nr/utils.cljs
@@ -10,6 +10,7 @@
    [goog.string :as gstring]
    [goog.string.format]
    [nr.appstate :refer [app-state]]
+   [nr.translations :refer [tr-data]]
    [reagent.dom :as rd]))
 
 ;; Dot definitions
@@ -194,12 +195,18 @@
   "A sequence of card pattern pairs consisting of a regex, used to match a card
   name in text, and the span fragment that should replace it"
   []
-  (letfn [(span-of [title] [:span {:class "fake-link" :data-card-title title} title])]
-    (->> (:all-cards-and-flips @app-state)
-         (vals)
-         (remove :replaced_by)
-         (map (fn [c] [(:title c) (span-of (:title c))]))
-         (sort-by (comp count str first) >))))
+  (letfn [(span-of [title tr-title] [:span {:class "fake-link" :data-card-title title} tr-title])]
+    (distinct (concat
+     (->> (:all-cards-and-flips @app-state)
+          (vals)
+          (remove :replaced_by)
+          (map (fn [c] [(:title c) (span-of (:title c) (:title c))]))
+          (sort-by (comp count str first) >))
+     (->> (:all-cards-and-flips @app-state)
+          (vals)
+          (remove :replaced_by)
+          (map (fn [c] [(tr-data :title c) (span-of (:title c) (tr-data :title c))]))
+          (sort-by (comp count str first) >))))))
 
 (def card-patterns (memoize card-patterns-impl))
 
@@ -211,7 +218,7 @@
     (->> (:all-cards-and-flips @app-state)
          (vals)
          (filter #(not (:replaced_by %)))
-         (map (fn [k] (regex-escape (:title k))))
+         (map (fn [k] (join "|" (map regex-escape (distinct [(:title k) (tr-data :title k)])))))
          (join "|"))))
 
 (def contains-card-pattern (memoize contains-card-pattern-impl))
--- a/src/clj/web/api.clj
+++ b/src/clj/web/api.clj
@@ -47,7 +47,8 @@
       ["/cards"
        ["" {:get data/cards-handler}]
        ["/version" {:get data/cards-version-handler}]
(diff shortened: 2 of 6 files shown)


csbisa (commented):
The particular format is worth discussion -- please let me know if there's a more preferred way to do this.

src/cljs/nr/utils.cljs
```diff
@@ -190,16 +191,24 @@
...
+     (->> (:all-cards-and-flips @app-state)
+          (vals)
+          (remove :replaced_by)
+          (map (fn [c] [(:title c) (span-of (:title c) (:title c))]))
+          (sort-by (comp count str first) >))
+     (->> (:all-cards-and-flips @app-state)
+          (vals)
+          (remove :replaced_by)
+          (map (fn [c] [(tr-data :title c) (span-of (:title c) (tr-data :title c))]))
+          (sort-by (comp count str first) >)))))
```
csbisa, line 211: This isn't very good but I couldn't come up with anything nicer. Any suggestions would be welcome.

csbisa:
Discussed this with @NoahTheDuke  over slack. We'll need some more work here because we want to avoid loading all translated data upfront due to size limits. I'll follow up with an updated patchset once I make some progress there.

csbisa:
Updated the patch series. Now there's three new patches at the beginning of the series which deals with exposing, fetching, and merging localized data. Now only localized data for the selected language is fetched, which means there's no change in local storage for English language users, and a small increase for users of other languages.

The data generation part will be covered over in netrunner-data.git, as the jnet side automatically loads any new fields from the raw data EDN.

NoahTheDuke:
I think the fetch task needs to be updated as well to actually place the card data into mongo.

csbisa:
> I think the fetch task needs to be updated as well to actually place the card data into mongo. Otherwise, this looks good to me.

Surprisingly, this is actually all automatic since it just imports everything that's in raw_data.edn. Here's the output from a fetch run with this patch series only:
```
$ lein fetch -l `pwd`/data -j
WARNING: boolean? already refers to: #'clojure.core/boolean? in namespace: puppetlabs.kitchensink.core, being replaced by: #'puppetlabs.kitchensink.core/boolean?
WARNING: uuid? already refers to: #'clojure.core/uuid? in namespace: puppetlabs.kitchensink.core, being replaced by: #'puppetlabs.kitchensink.core/uuid?
Wrote data/cards.edn to disk
Wrote data/cards-ja.edn to disk
Wrote data/cards-ko.edn to disk
Wrote data/cycles.edn to disk
Wrote data/formats.edn to disk
Wrote data/mwls.edn to disk
Wrote data/sets.edn to disk
Imported cards into database
Imported cards-ja into database
Imported cards-ko into database
Imported cycles into database
Imported formats into database
Imported mwls into database
Imported sets into database
Removing old images from db cards
Added 2188 images to :en :default :stock
Added 356 images to :ja :default :stock
Adding override images...
Indexes successfully created.
25 cycles imported
68 sets imported
15 MWL versions imported
1776 cards imported
```

The edn file contains `:cards-jp` and `:cards-ko` from my in-progress changes in netrunner-data.git.

NoahTheDuke:
Oh yeah, duh. We just call `replace-collection` lol. Thanks for verifying.

NoahTheDuke:
To double check, this is good to merge?

csbisa:
Yep, the remaining work is over at netrunner-dat.git and this is ready for merge. If you prefer to wait until the netrunner-data.git changes are PR'd to ensure everything's working end-to-end, that's fine with me.

csbisa:
The other side of the changes have been posted at https://github.com/NoahTheDuke/netrunner-data/pull/35.

csbisa:
With the completion of https://github.com/NoahTheDuke/netrunner-data/pull/35 I've confirmed that System Gateway decks display their respective languages' card titles in all of jinteki.net's supported languages, with the exception of both Chinese languages which I'll follow up with in https://github.com/mtgred/netrunner/issues/6960.

PR #6913, opened 2023-03-01
fix Friday Chip log

francescopellegrini:
It was previously printing some kind of noise like this:
image


NoahTheDuke:
lol that's the string version of a function. Nice find

Issue #6940, opened 2023-05-20
Not able to trigger Odore's second routine


locks:
Reported on GLC:

> When you have 3 virtual resources installed, jnet does not offer the option of using the 2: break any number of subs option on Odore.
Had a game where I wanted to charge the Twinning using that, but only was allowed to break for 0.

Card: [Odore](https://netrunnerdb.com/en/card/26071)

francescopellegrini:
@NoahTheDuke fixed by #7220

PR #6941, opened 2023-05-22
Fix for Odore's ability

Tamijo-Netrunner:
Odore's first ability `2c: Break all subs` can be used while 3 Virtuals are installed. (Relevant interactions exist where a player wants to do this over the `0c: Break a sub` ability.)

### DIFF
--- a/src/clj/game/cards/programs.clj
+++ b/src/clj/game/cards/programs.clj
@@ -2086,9 +2086,7 @@
    :abilities [(set-autoresolve :auto-fire "Nyashia")]})
 
 (defcard "Odore"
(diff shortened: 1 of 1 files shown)


NoahTheDuke (changes requested):
Gotta update the tests to account for this. And i think it was written this way to make auto pumping and breaking work. Might need to disable that if you can't get it to work together.

NBKelly:
Closes #6940

NoahTheDuke:
hey @Tamijo-Netrunner, wanna take another swing at this?

NoahTheDuke:
Okay, turns out it's slightly trickier than it seems. I'm working on a version of this that should fix the additional issues it raised. Thanks for the initial work!

PR #6947, opened 2023-06-19
fix trash-hosted-cards log when no cards were trashed

francescopellegrini:
I noticed that rezzing Tithonium with no hosted cards logged the following message: "trashes because Tithonium cannot host cards".


francescopellegrini:
Is there anything against merging this PR?

PR #6948, opened 2023-06-19
added aria live to in-game log, and added aria label to icon pattern …

utati:
…pairs. This would make it so a screen reader would read out the events of a game/replay as it was happening.

https://www.youtube.com/watch?v=XxEfvgoZd3Y Here's a video showing it in action.

I'm not sure about the change for icons. I noticed there's a tooltip that appears when you hover over the icons, the code that creates that tooltip may be a better place to add the aria-label.

References:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions

### DIFF
--- a/src/cljs/nr/utils.cljs
+++ b/src/cljs/nr/utils.cljs
@@ -151,7 +151,7 @@
 (def icon-patterns
   "A sequence of icon pattern pairs consisting of an regex, used to match icon
   codes, and the span fragment that should replace it"
-  (letfn [(span-of [icon] [:span {:class (str "anr-icon " icon) :title (str " " icon) :aria-label (str icon) :role (str "img") }])
+  (letfn [(span-of [icon] [:span {:class (str "anr-icon " icon) :title (str " " icon) :aria-label (str icon) :role ("img") }])
           (regex-of [icon-code] (re-pattern (str "(?i)" (regex-escape icon-code))))]
     (->> {"[credit]" "credit"
           "[credits]" "credit"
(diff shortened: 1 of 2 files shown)


src/cljs/nr/utils.cljs
```diff
@@ -151,7 +151,7 @@
 (def icon-patterns
   "A sequence of icon pattern pairs consisting of an regex, used to match icon
   codes, and the span fragment that should replace it"
-  (letfn [(span-of [icon] [:span {:class (str "anr-icon " icon) :title (str " " icon)}])
+  (letfn [(span-of [icon] [:span {:class (str "anr-icon " icon) :title (str " " icon) :aria-label (str icon) :role (str "img") }])
```
NoahTheDuke, line 154: Is this right? :role (str "img")?

utati: I'm not sure if it's the tidiest clojure code, but it should apply the image aria role to the icons' elements. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/img_role, I got stuck in a rabbit hole of trying to get the whole sentence to play when you hover over it (it currently stops at an icon if there is one) and that was one of the things I tried to fix it.

NoahTheDuke: Okay cool. You can just write the string "img" with the str call.

utati: Right, I'll go tidy that up now. I'm still geting most of my clojure knowledge from looking at the nearest piece of code haha

PR #6951, opened 2023-06-22
fix Fairchild N.0 subs when runner cannot pay either costs

### LINKED ISSUE #6942: Runners being given a choice when they shouldn't get one
When resolving subroutines where the runner has a choice, JNet allows the runner to choose an effect they are unable to resolve. In the case of Fairchild 3.0 and similar effects, the runner most choose one they can resolve. 

image

francescopellegrini:
Fixes #6942


francescopellegrini:
Is there anything against merging this PR?

PR #6954, opened 2023-06-28
fetch: Fetch translated card images in supported languages

csbisa:
jinteki.net has supported translated card images for a few years but no proper data import process existed, which this change attempts to rectify.

Initially, only Japanese images will be hosted on nrdbassets and supported-img-langs can be expanded once more languages are added.

To avoid 404ing on a whole lot of card images that will never exist, we restrict non-English image downloads to NSG-era cards. Note that we'll still get 404s on some data sets (alt arts, Salvaged Memories, System Core 2019). I can extend the filtering further if we care enough, but it might be best to live with 404s as we'll otherwise have to dance around newly released cards since translated cards aren't available for a few days to weeks after the English release.

### DIFF
--- a/src/clj/tasks/nrdb.clj
+++ b/src/clj/tasks/nrdb.clj
@@ -16,7 +16,8 @@
     [tasks.setup :refer [connect disconnect]]))
 
 (def ^:const edn-base-url "https://raw.githubusercontent.com/NoahTheDuke/netrunner-data/master/edn/raw_data.edn")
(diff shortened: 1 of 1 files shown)


csbisa:
Note this shouldn't be merged until I give the all-clear -- mainly looking for any feedback in the meantime.

NoahTheDuke:
Any updates on this?

csbisa:
plural never got back to me about uploading the additional images to NRDB. Let me ping him again.

PR #6957, opened 2023-07-12
Adding NSG card back and translation

tblackwe:
Card backs and faction glyphs, and adding NSG to translation files


tblackwe:
I'm a bit concerned about file size. but I assume the images get cached after first download

PR #6986, opened 2023-08-03
maw fix

NBKelly:
We don't have any events for the maw thing, so it's just checking that at the end of the access, the number of cards not trashed/scored = 1, and that the card is not in the trash.
This is obviously not logically complete. 
We're now checking it's not in the score area either, which is still not perfect, but should fix all the issues.

If it doesn't, then a more complete (but much less terse) fix is to actually raise an event instead of just incrementing a register.

Actually, that's on my todo list. I'll do it in the next month or so.


NoahTheDuke:
I don't think firing an event works because this needs to be checked after accessing the card which is when :post-access-card fires. How do you envision it would work?

NBKelly:
Fair.

That's a problem for me to figure out in a month or two. This patch *should* fix all the issues with maw currently.

PR #6991, opened 2023-08-05
Adding support for the Neo format

tblackwe:
Removing Classic format and adding Neo
Updating git ignore


tblackwe:
Requires https://github.com/NoahTheDuke/netrunner-data/pull/38 to be merged

PR #6995, opened 2023-08-07
Fix Hermes and Leela being able to bounce Public agendas

butzopower:
I ended up doing this in a game and realizing it was probably not allowed.  Looking into new printing of Oaktown Renovation and it's clear neither of these should be able to bounce it.

image

EDIT: Searched through and found one more that should follow this rule which is Analog Dreamers, very rotated but maybe it gets brought back in some System Update.


butzopower:
Fixes #7013

Issue #7000, opened 2023-08-08
Tucana not firing even though it's persistent


Joseph7695:
Link to bug replay: https://jinteki.net/bug-report/cf809213-a936-45c5-90c5-588ec5a8ccda?b=0

Description:

Tucana not firing even though it's persistent

PR #7002, opened 2023-08-08
Fix Persistent issue with Tucana

butzopower:
Fixes issue where Tucana would not trigger after being trashed - I think its previous implementation expected a context from a run-end event (i.e. AMAZE) instead of agenda-stolen.

Also make sure Tucana shuffles R&D after firing.

### DIFF
--- a/src/clj/game/cards/upgrades.clj
+++ b/src/clj/game/cards/upgrades.clj
@@ -1681,8 +1681,10 @@
 (defcard "Tucana"
   (let [ability {:async true
                  :prompt "Choose a piece of ice to install and rez"
-                 :choices (req (cancellable (filter ice? (:deck corp))))
-                 :effect (req (corp-install state side eid target nil {:install-state :rezzed :combined-credit-discount 3}))
+                 :choices (req (cancellable (filter ice? (:deck corp)) true))
+                 :effect (req (do (corp-install state side eid target nil {:install-state :rezzed :combined-credit-discount 3})
+                                  (shuffle! state :corp :deck)
+                                  (system-msg state side (str "shuffles R&D"))))
                  :cancel-effect (effect (system-msg  "declines to use Tucana to install a card")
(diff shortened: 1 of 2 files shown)


butzopower:
Aha! Someone else noticed this too #7000

NBKelly:
Closes #7000

src/clj/game/cards/upgrades.clj
```diff
@@ -1681,8 +1681,10 @@
 (defcard "Tucana"
   (let [ability {:async true
                  :prompt "Choose a piece of ice to install and rez"
-                 :choices (req (cancellable (filter ice? (:deck corp))))
-                 :effect (req (corp-install state side eid target nil {:install-state :rezzed :combined-credit-discount 3}))
+                 :choices (req (cancellable (filter ice? (:deck corp)) true))
+                 :effect (req (do (corp-install state side eid target nil {:install-state :rezzed :combined-credit-discount 3})
+                                  (shuffle! state :corp :deck)
+                                  (system-msg state side (str "shuffles R&D"))))
```
NoahTheDuke, line 1687: The way our "async" stuff works is kind of awkward. You have to use the macro `wait-for` and then explicitly "complete" the callback associated with the `eid` at the end:

```clojure
(req (wait-for (corp-install state side (make-eid state eid) target nil {:install-state :rezzed :combined-credit-discount 3})
               (shuffle! state :corp :deck)
               (system-msg state side "shuffles R&D")
```

Something like that. Make sure it compiles but it should be good to go otherwise.

butzopower: I ran into some issues with the tests complaining about an already used eid.  I modified this to add `make-eid` for the corp install, not sure if that was correct.

butzopower: Just reread your suggested change and you actually had `make-eid` in there, I just forgot to copy it!

PR #7004, opened 2023-08-08
Use effects for unrunnable servers

NBKelly:
Another step in moving away from the flags system entirely.

There's a little bit of cleanup that can be done - notably, `can-run-server` is duplicated in `actions` and `runs` due to dependency issues - I can probably move it entirely to runs.

This means that RP and front company and off the grid and marathon can all work together in the same game (This is still a nightmare scenario, but only for the player now).

Though I can hardly believe it, it all just works.

### DIFF
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -27,7 +27,7 @@
    [game.core.engine :refer [pay register-events resolve-ability]]
    [game.core.events :refer [first-event? no-event? turn-events]]
    [game.core.expose :refer [expose-prevent]]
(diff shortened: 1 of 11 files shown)


NoahTheDuke:
I'm almost mad at how simple this is. This looks great. The cyclic dependency stuff is annoying; can you refer to the new version from `runs` in `actions` to get around it?

Maybe see if you can add a test to check that nightmare scenario, just to verify everything's chill.

Once you feel confident about it all working, deleting the old code and let's merge it!

NBKelly:
I think I can do that after work on Friday or Saturday.

PR #7014, opened 2023-08-10
Fix issue with card preview links disappearing

butzopower:
This happens very frequently to me where the preview links just stop working after a refresh or accidental horizontal scroll turns into a browser back.  This is especially difficult for new players who as runners are only able to see an installed unrezzed card via the preview link when trashing.

I think I've narrowed the issue down to a race condition where the card data isn't loaded yet when the checking functions are memoized and so further invocations don't contain the card data - I was able to reproduce this very consistently by deleting the cards from local storage and throttling the browser requests.

The attached video shows the very simple fix of passing the card loaded state to the memoized function to ensure memoization remains but allows for refresh when cards are loaded.

https://github.com/mtgred/netrunner/assets/5345/a6f7de76-af22-4289-b58e-5f8c1709d7e7

### DIFF
--- a/src/cljs/nr/utils.cljs
+++ b/src/cljs/nr/utils.cljs
@@ -208,7 +208,8 @@
           (map (fn [c] [(tr-data :title c) (span-of (:title c) (tr-data :title c))]))
           (sort-by (comp count str first) >))))))
 
-(def card-patterns (memoize card-patterns-impl))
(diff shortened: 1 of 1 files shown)


butzopower:
Also happy to move fetching the card loaded state closer to the memoization/impl itself I just couldn't figure out how to do it in clojure with def and defn.

PR #7020, opened 2023-08-10
rework Daniela Jorge Inácio and add test

francescopellegrini:


src/clj/game/cards/upgrades.clj
```diff
@@ -483,29 +483,45 @@
...
-                               (doseq [r (take 2 (shuffle (:hand runner)))] (move state :runner r :deck)))
-                             (effect-completed state side eid))}}))
+                :req (req (and run (= :runner side)))
+                :msg "force the Runner to add 2 random cards from the grip to the bottom of the stack as additional cost to steal agendas from this server or its root"
+                :effect
+                (req (when (can-pay?
+                             state :runner
+                             (assoc eid :source card :source-type :ability)
+                             card nil [:add-random-from-hand-to-bottom-of-deck 2])
+                       (wait-for (pay state :runner (make-eid state eid) card :add-random-from-hand-to-bottom-of-deck 2)
```
NoahTheDuke, line 514: This can't be used in a `when`. If you're worried about code duplication, make a helper function. All code branches have to end in an async function or effect-completed.

NoahTheDuke (changes requested):

src/clj/game/cards/upgrades.clj
francescopellegrini: My fault, the `when` part shouldn't go all the way to the end

PR #7031, opened 2023-08-10
add Virtual Service Agent test

francescopellegrini:


src/clj/game/cards/ice.clj
```diff
@@ -4042,13 +4042,8 @@
    :implementation "Might be incorrect if decoder is uninstalled"
```
NoahTheDuke, line 4042: Is this still true?

francescopellegrini: Yes, I didn't change the `req` logic

PR #7056, opened 2023-08-10
add Hermes test

francescopellegrini:


butzopower:
I put in a fix for Hermes in a separate PR since it lets you bounce public agendas: https://github.com/mtgred/netrunner/pull/6995 - don't think these tests will conflict though since I didn't really write a happy path test for it

francescopellegrini:
> I put in a fix for Hermes in a separate PR since it lets you bounce public agendas: #6995 - don't think these tests will conflict though since I didn't really write a happy path test for it

Cool, thank you!

PR #7058, opened 2023-08-10
add Epiphany Analytica test

francescopellegrini:


src/clj/game/cards/identities.clj
```diff
@@ -1,5 +1,6 @@
 (ns game.cards.identities
   (:require
+   [clojure.pprint :as pprint]
```
NoahTheDuke, line 3: I don't think this is used.

francescopellegrini: It's referenced at L605

NoahTheDuke: Thank you

Issue #6996, opened 2023-08-07
Starlit Knight in Acme


mcgreenhalgh:
If I'm understanding the wording/rules correctly, in Acme, once threat=4, Starlit Knight should gain an ETR sub.

NBKelly:
That makes sense to me, and looking at it in the code is looks right, so I feel like figuring it out in engine is going to be a nightmare.

Poor ACME.

francescopellegrini:
Fixed by #7073

PR #7073, opened 2023-08-13
Fix Starlit Knight interaction with ACME Consulting

butzopower:
Spent a bit of time trawling through backtraces to understand a bit why the issue reported here: #6996 happens.

From what I can tell from the `continue` action will process `on-encounter` (including any ice on-encounter effects) and then do a fake-checkpoint to sort of realign all the different values from what's been processed. Since the fake-checkpoint is the earliest that the tags effects are being reevaluated this leads to the total tags not yet set by the time Starlit Knight's on-encounter happens.  I guess this is generally just a problem with on-encounter effects is that they've run before some constant effects have been reevaluated (maybe this is just limited to tag total?).

I've updated Starlit Knight to just calculate the tag effects directly rather than relying on the stored total, but I wasn't sure if this was risky or too much breaking of encapsulation (since this function isn't used anywhere else currently).

### DIFF
--- a/src/clj/game/cards/ice.clj
+++ b/src/clj/game/cards/ice.clj
@@ -57,7 +57,7 @@
                               target-server zone->name]]
    [game.core.shuffling :refer [shuffle!]]
    [game.core.subtypes :refer [update-all-subtypes]]
-   [game.core.tags :refer [gain-tags lose-tags]]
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
There's a better way to do this, but I can't think of it right now so let's at least fix this.

PR #7081, opened 2023-08-16
Clarify cost to place advancement counter for Vasilisa on-encounter effect

jtrawinski:
First PR here :)

On encounter, Vasilisa would prompt: 

> Place 1 advancement counter on a card that can be advanced?

I found this a bit misleading, as it's not clear from the prompt that it actually costs 1 credit. This PR changes the prompt to:

> Pay 1 [Credits] to place 1 advancement counter on a card that can be advanced?

which makes the cost clear to the user.

### DIFF
--- a/src/clj/game/cards/ice.clj
+++ b/src/clj/game/cards/ice.clj
@@ -3976,7 +3976,7 @@
 
 (defcard "Vasilisa"
   {:on-encounter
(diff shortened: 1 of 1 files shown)


NoahTheDuke:
Thanks for the PR! Welcome to jnet development. This looks great. If you're interested in doing more work, holler at me at NoahTheDuke on Stimslack, and I'll add you to our dev slack server.

PR #7098, opened 2023-08-19
Explicit no-op handler for sente pong from client -> server

butzopower:
It looks like sente now sends pong acks on ping.  The server is currently logging these pongs as unknown message so I've just copied to same way we "handle" the pings.

### DIFF
--- a/src/clj/web/ws.clj
+++ b/src/clj/web/ws.clj
@@ -64,6 +64,7 @@
     (?reply-fn {:msg "Unhandled event"})))
 
 (defmethod -msg-handler :chsk/ws-ping [_])
+(defmethod -msg-handler :chsk/ws-pong [_])
 ;; NOTE - :chsk/uidport-close is handled in game.clj
 (defmethod -msg-handler :chsk/uidport-open
   [{uid :uid


NoahTheDuke:
Are the ping pongs meaningful? Should we be using them in some way?

butzopower:
I don't think so, they are just being passed on from sente.  It looks like the latest version of sente has both the client and the server send ack (pong) responses when receiving a ping - I think this is to enable optional broken connection detection as introduced here (along with the pong): https://github.com/taoensso/sente/commit/0dc8a12e20656ce22c667cb1142a5e6b60d15a65

I'm not sure there's much use in providing a logging handler or something for these messages, it looks like sente actually already logs these we'd just have to set the log level up on timbre/sente if we actually wanted to see them.

NoahTheDuke:
from that commit:

```
              ;; So we're rolling this new feature out in 2 steps:
              ;;   1. Update clients to respond to pings (with pongs)
              ;;   2. Update servers to regard lack of pong as broken conn
```

maybe we shouldn't be just ignoring them?

butzopower:
I think if there's concern around this specifically: https://github.com/mtgred/netrunner/pull/7101#issuecomment-1685144474 then this is currently happening very explicitly with this not being handled (whereas I think the sente logs at least obscure a little bit of the client_id to just 6 chars).

butzopower:
Also I don't think sente is ignoring the pings and pongs messages (which is why the pongs are sent at all), it's just then forwarding them to the user provided router as well if we wanted to do any additional handling.

butzopower:
Should note that sente's own example project doesn't provide a handler for these (and they get logged as unhandled in their example application message handler) - so not sure it's necessarily expected or required that consumers of sente handle these.

NoahTheDuke:
Cool

PR #7102, opened 2023-08-19
Don't allow Cybersand Harvester trash ability when no credits

butzopower:
image

### DIFF
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -703,6 +703,7 @@
              :effect (effect (add-counter :corp card :credit 2))}]
    :abilities [{:label "Take all hosted credits"
                 :cost [:trash-can]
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/assets.clj
```diff
@@ -703,6 +703,7 @@
              :effect (effect (add-counter :corp card :credit 2))}]
    :abilities [{:label "Take all hosted credits"
                 :cost [:trash-can]
+                :req (req (> (get-counters card :credit) 0))
```
NoahTheDuke, line 706: should use `pos?` here

NoahTheDuke (changes requested):

src/clj/game/cards/assets.clj
butzopower: 👍🏻

PR #7106, opened 2023-08-21
Allow Under the Bus to be played without Connection

### LINKED ISSUE #7097: Under the bus should be playable just for bad pub
You shouldn't need to see a connection on the table

butzopower:
Fixes #7097

### DIFF
--- a/src/clj/game/cards/operations.clj
+++ b/src/clj/game/cards/operations.clj
@@ -2994,11 +2994,7 @@
 
 (defcard "Under the Bus"
   {:on-play
-   {:req (req (and (last-turn? state :runner :accessed-cards)
-                   (not-empty (filter
-                                #(and (resource? %)
-                                      (has-subtype? % "Connection"))
-                                (all-active-installed state :runner)))))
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/operations.clj
```diff
@@ -2994,20 +2994,18 @@
...
     :prompt "Choose a connection to trash"
     :choices {:card #(and (runner? %)
                           (resource? %)
                           (has-subtype? % "Connection")
                           (installed? %))}
     :msg (msg "trash " (:title target) " and take 1 bad publicity")
     :async true
-    :effect (req (wait-for (trash state side target {:cause-card card})
-                           (gain-bad-publicity state :corp eid 1)))}})
+    :effect (req (if target (wait-for (trash state side target {:cause-card card})
```
NoahTheDuke, line 3005: When would target not exist?

butzopower: Woops, I left this in from a different implementation - should be unchanged from original effect now.

PR #7112, opened 2023-08-26
Fix self references and runner zones in messages

francescopellegrini:
- All mentions of cards in messages have been replaced with `(:title card)` when possible.
- Grip, stack, and heap are referenced using "the" instead of "Runner's" or "your".
- Replaced some uses of "token" with "counter".
- Replaced some uses of "pieces of installed ice" with "installed pieces of ice".
- Used `enumerate-str` when possible.


NoahTheDuke:
This is great work.

Now that it's done, I wonder if we should add format string support: `"... {card-title} ..."` evaluates to `(str "..." (:title card) "...")` and similar for grip etc. what do you think? Worth trying or nah?

francescopellegrini:
> This is great work.
> 
> Now that it's done, I wonder if we should add format string support: `"... {card-title} ..."` evaluates to `(str "..." (:title card) "...")` and similar for grip etc. what do you think? Worth trying or nah?

I think it's not worth doing, especially because in certain scenarios messages need more accuracy so they use `(card-str ...)` instead. I might consider doing another round of search&replace in the future.. ;)

PR #7129, opened 2023-09-05
Refactor storing lobby-updates as separate key in app-state

butzopower:
Refactored the lobby update tracking a bit so that it's managed under it's own key, don't think this will necessarily fix anything but might make it easier if we want to further extract it.

Would be good to have @lostgeek opinion on this change.

### DIFF
--- a/src/clj/web/app_state.clj
+++ b/src/clj/web/app_state.clj
@@ -9,8 +9,9 @@
 
 (defn register-user
   [app-state uid user]
-  (merge app-state {:users (assoc (:users app-state) uid (assoc user :uid uid))
-                    :lobby-updates (assoc (:lobby-updates app-state) uid true)}))
+  (-> app-state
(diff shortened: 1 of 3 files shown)


src/clj/web/app_state.clj
```diff
@@ -4,11 +4,13 @@
 
 (defonce app-state
   (atom {:lobbies {}
+         :lobby-updates {}
          :users {}}))
 
 (defn register-user
   [app-state uid user]
-  (assoc-in app-state [:users uid] (assoc user :uid uid :lobby-updates true)))
+  (merge app-state {:users (assoc (:users app-state) uid (assoc user :uid uid))
```
NoahTheDuke, line 12: Instead of merge, let's do a thread-first and two `assoc-in`s, as `merge` isn't a deep merge.

butzopower: 👍🏻

PR #7132, opened 2023-09-06
Remove Gachapon prompt

francescopellegrini:
adding a log with all set-aside cards.

Also add check on whether or not the Runner can actually install a card


NoahTheDuke:
Printing to the log seems smart so both sides can see. But it also seems like displaying the set aside cards prominently helps the runner know what's happening. Is removing the display prompt the right move?

francescopellegrini:
I was replicating Deep Dive's behavior which has no prompt showing the list of cards. If you think otherwise I can do the opposite, adding a prompt to Deep Dive :)

francescopellegrini:
Superseded by #7142

PR #7140, opened 2023-09-09
Tracking changes to the game state to make diffs quicker

lostgeek:
WIP for tracking changes to the game state.

All instances of `swap!` were changed to `swap!*`, which is defined in `utils.clj`. This function checks if the function for `swap!` was assoc(-in), dissoc(-in) or update(-in) and if so, tracks the keys that were changed in a hash-set `(:unsent-changes @state)`.

If all calls to `swap!*` between two calls of `diffs/public-states` were only such trackable changes, the differ is only fed the subset of the state-map that could've been changed (using a `utils/deep-select-keys`)

If there is a call to `swap!*` with some other function, the flag `(:unsent-changes-send-all @state)` is set to true and the next diff will act on the entire state as usual.

### DIFF
--- a/src/clj/game/utils.clj
+++ b/src/clj/game/utils.clj
@@ -1,5 +1,6 @@
 (ns game.utils
   (:require
+    [game.core.state]
     [jinteki.cards :refer [all-cards]]
     [clojure.string :as str]
     [clj-uuid :as uuid]))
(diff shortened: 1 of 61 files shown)


src/clj/game/utils.clj
```diff
@@ -1,5 +1,6 @@
 (ns game.utils
   (:require
+    [game.core.state]
```
NoahTheDuke, line 3: Don't do this. This namespace is meant to be fully independent from the engine. Given that we're only referring to swap!* in engine or cards functions, there's no need to check if the given atom is an instant of State.

src/clj/game/utils.clj
```diff
@@ -136,3 +137,68 @@
...
+  [a f & args]
+  (when (instance? game.core.state.State @a)
+    (if (and
+          (not (:unsent-changes-send-all @a))
+          (or (= assoc f)
+              (= assoc-in f)
+              (= update f)
+              (= update-in f)
+              (= dissoc f)
+              (= dissoc-in f)))
```
NoahTheDuke, line 152: This can be a set: (if (#{a b c} f) ...)

src/clj/game/utils.clj
```diff
@@ -136,3 +137,68 @@
...
+  (apply swap! a f args))
+
+(defn deep-select-keys
+  [m in-keys]
+  (loop [ks in-keys
+         res {}]
+    (if (empty? ks)
+      res
+      (if (keyword? (first ks))
+        (recur (rest ks) (assoc res (first ks) (get m (first ks))))
```
NoahTheDuke, line 166: Make ks [ks (seq in-keys)], then check (if ks ...) then use (next ks) to call seq on ks when recurring.

NoahTheDuke (changes requested):

NoahTheDuke:
Gonna close this for now. If you feel like resurrecting it, try running some performance comparisons so we can make a valid judgement call about whether to use this or not.

PR #7154, opened 2023-09-20
Fix issues with trojans losing mu cost when hosted ice moves

### LINKED ISSUE #7151: MU of Botulus hosted on Formicary is no longer counted after Formicary moves
When an unrezzed Formicary has a Botulus hosted on it, if the corp rezzes and moves the Formicary using its ability, then the MU of the Botulus is no longer counted against the runner's MU. This can observed in click 51 of this replay:
https://www.jinteki.net/replay/7cc3a0ed-6c1b-4580-ad3c-5207299bc9ce
The Formicary with this same Botulus is later derezzed, and then moved again, throughout that entire time the botulus still does not take MU.

Potentially related issue: https://github.com/mtgred/netrunner/issues/6917

butzopower:
Not the most elegant fix - there seems like there should be a refactor here somewhere.

I think it's currently safe to ignore the no-mu option in these two cases as 
- `swap-installed` is only used with ICE
- `get-moved-card` only refreshes hosted cards if the host stays in the same zone - and I'm not sure there's any way for this to happen on runner's side?

Perhaps the no-mu should be a property or something of the host card rather than the hosting/install actions - and then every program can just always reset its memory effect when moved or avoid if the host ignores mu cost?

Fixes #7151

### DIFF
--- a/src/clj/game/core/moving.clj
+++ b/src/clj/game/core/moving.clj
@@ -3,9 +3,9 @@
     [clojure.string :as string]
     [game.core.agendas :refer [update-all-agenda-points]]
     [game.core.board :refer [all-active-installed]]
-    [game.core.card :refer [active? card-index condition-counter? convert-to-agenda corp? facedown? fake-identity? get-card get-title get-zone has-subtype? ice? in-hand? in-play-area? installed? resource? rezzed? runner?]]
+    [game.core.card :refer [active? card-index condition-counter? convert-to-agenda corp? facedown? fake-identity? get-card get-title get-zone has-subtype? ice? in-hand? in-play-area? installed? program? resource? rezzed? runner?]]
     [game.core.card-defs :refer [card-def]]
-    [game.core.effects :refer [register-static-abilities unregister-static-abilities]]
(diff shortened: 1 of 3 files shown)


NoahTheDuke:
Gotta remove the old lingering effect, right? Otherwise, this looks good.

butzopower:
In both these instances `unregister-static-abilities` is called before which I think is what is clearing out the original MU abilities in the first place (and why they have to be re-registered).  Or maybe theres still other clean-up that needs to happen for the original abilities (assumed they were cleaned up because they are duration `:while-active`)?

butzopower:
@NoahTheDuke ^^ see above - not sure there's an old lingering effect to remove?

NoahTheDuke:
If you don't think so, then I'm fine with this. If it blows up, we'll find out pretty quickly lol.

This business does make me think we should revisit MU calculations. The lingering effects system is solid but doesn't conform to the CR and has a fair amount of fiddliness. I don't have a solid plan yet but I think we could do something better that requires less deliberate effect building.

PR #7157, opened 2023-09-22
Fix MU issue with Magnet 

### LINKED ISSUE #6917: Magnet disables the MU cost of hosted programs
I think this is because we unregister events and effects, one of which is [update used MU](https://github.com/mtgred/netrunner/blob/220a4dd5f26fda0923105f11ec726403c6ff4b0b/src/clj/game/core/hosting.clj#L65-L69).

### LINKED ISSUE #7121: Incorrect MU changes after installs
I just had this happen to me in two separate jnet games, after installing a program with Arissana's ability, I was gifted with extra MU. The only common factors between both occurrences seem to be that I performed the install mid-run, and that I was prompted to trash another program to make room (which I did so).

Game 1, turn 12, click 2. Running HQ, installed a Slap Vandal onto a Magnet, trashed a Kyuban on another server to make room, but the newly installed SV didn't seem to increase my used MU: 

https://www.jinteki.net/replay/dd1a3000-4466-4441-8c20-432c66581286?n=107&d=0

Game 2, turn 6, click 3. Running R&D, installed a Kyuban onto a Thimblerig, trashing a Kyuban elsewhere, and after passing the ice (and Thimblerig moves) I suddenly have 2MU free:

https://www.jinteki.net/replay/39e0281c-7ad7-4359-aad8-6fe57aba7e30?n=53&d=0

Additionally, my MU in the second game eventually fixes itself, I "regain" an MU when returning the Kyuban that was on the Thimble with Art Vernissage (runner turn 8), and then a second time when Thimble moves whilst hosting another Kyuban (corp turn 12).

In conclusion, Thimblerig is a hell of a card, huh.

butzopower:
Just do the same re-apply MU code to the Magnet effect as everywhere else.  

Refactored init / re-apply MU code to helper function.

Fixes #6917
Fixes #7121

### DIFF
--- a/src/clj/game/cards/ice.clj
+++ b/src/clj/game/cards/ice.clj
@@ -38,7 +38,7 @@
    [game.core.initializing :refer [card-init]]
    [game.core.installing :refer [corp-install corp-install-list
                                  corp-install-msg]]
-   [game.core.memory :refer [available-mu update-mu]]
+   [game.core.memory :refer [available-mu init-mu-cost]]
    [game.core.moving :refer [as-agenda mill move swap-cards swap-cards-async
                              swap-ice swap-installed trash
                              trash-cards]]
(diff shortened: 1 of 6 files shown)


butzopower:
Actually need to make sure this doesn't double up MU cost if Magnet is derezzed

PR #7172, opened 2023-10-04
Fix Tatu-Bola not registering derezzed events

### LINKED ISSUE #7168: Installing Formicary with Tatu-Bola with Asa Group vs 419 seems to break the "when approach" trigger on Formicary.
I was playing Asa Group into 419 with a rezzed Tatu-Bola that was passed. This set of actions happened:

1. Tatu-Bola is passed on R&D.
2. Tatu-Bola gives me an option to swap with a piece of ice in HQ, I choose Formicary in hand (was my only ice in HQ at the time, not sure if important though).
3. Asa Group gives me a trigger to install a non-agenda card, I choose the Tatu-Bola that was returned to HQ from the swap.
4. 419 trigger asks if I want to expose the Formicary just installed, I pay 1 credit to prevent the expose.
5. The run continues.

After this point, the Formicary message to rez and move Formicary while the runner was approaching a server would not show up, even when trying to change the behavior of Formicary by clicking on the card. I'm not sure which trigger or group of triggers causes this issue (I haven't actually run into a situation yet with less then the above triggers, but I'll update if I run into it again in a different scenario).

butzopower:
The swap method used by Tatu-Bola wasn't registering derezzed effects (think only Formicary uses this right now). Sort of a bespoke engine fix for this problem but possibly covers other future derezzed ice effects.

Fixes #7168

### DIFF
--- a/src/clj/game/core/moving.clj
+++ b/src/clj/game/core/moving.clj
@@ -539,12 +539,14 @@
     ;; todo - we might need behaviour for runner swap installs down the line, depending on future cards
     ;; that's a problem for another day
     (if (and install-event (= :corp side))
-      (do (queue-event
-            state :corp-install
-            {:card (get-card state (if (installed? moved-a) moved-a moved-b))
-             :install-state (:install-state (card-def (if (installed? moved-a) moved-a moved-b)))})
-          (wait-for (checkpoint state nil (make-eid state eid))
-                    (complete-with-result state side eid async-result)))
+      (let [installed-card (if (installed? moved-a) moved-a moved-b)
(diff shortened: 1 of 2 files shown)


butzopower:
Digging into this issue in the engine a bit more and it looks like most of the swapping code doesn't actually account for `:derezzed-events` and only really installing considers it. This PR fixes probably the currently most common occurrence of the problem but I will look into a more broader fix for where I suspect this issue occurs with things like Project Yagi-Uda and ancient cards like Tenma Line.

NoahTheDuke:
I'll merge this fix and you can look into deeper fixes in the next MR.

PR #7174, opened 2023-10-06
Only send toast to the side the toast is for

butzopower:
High ranking for best commit message of the year so far.

Here's a screenshot of what the toast diffs look on each side with this change:

image


NBKelly:
Have you noticed the toasts on the corp side sometimes being delayed in testing?

PR #7176, opened 2023-10-08
fix Adept, Sage, and Savant breaking all subs for 2 credits

### LINKED ISSUE #7137: Sage, Savant, and Adept don't cost credits per usage
When you use one of the three ICEBreakers, Sage, Savant, or Adept to break more than once on a single piece of ICE, it only costs 2 credits total, no matter how many subroutines you break.

francescopellegrini:
Fixes #7137


NoahTheDuke (commented):
I don't want to split the abilities. They're a single ability  on the cards, we should be able to fix it without that.

I suspect you just gotta remove the :all (< 1 qty) parts

PR #7180, opened 2023-10-10
fix Adept, Sage, and Savant breaking all subs for 2 credits

francescopellegrini:
Supersedes #7176 

Courtesy of @NoahTheDuke ❤️


NoahTheDuke:
Thanks for writing a good demonstrative test!

PR #7185, opened 2023-10-17
Fix interaction of Verbal Plasticity and The Class Act

### LINKED ISSUE #7182: Verbal Plasticity with Class Act
If you have both Verbal Plasticity and the Class Act installed it will only show two cards as choices to bottom with TCA when it should show three cards.

butzopower:
Created a new bonus key specifically for drawing via basic action - as the two runner cards that used `draw-bonus` are both scoped to just basic draw.  This bonus is then applied as part of the basic action rather than within the draw function.

Fixes #7182

### DIFF
--- a/src/clj/game/cards/basic.clj
+++ b/src/clj/game/cards/basic.clj
@@ -5,7 +5,7 @@
    [game.core.card :refer [agenda? asset? event? get-card hardware? ice?
                            in-hand? operation? program? resource? upgrade?]]
    [game.core.def-helpers :refer [defcard]]
-   [game.core.drawing :refer [draw]]
+   [game.core.drawing :refer [draw use-bonus-click-draws!]]
    [game.core.eid :refer [effect-completed]]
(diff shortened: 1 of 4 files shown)


src/clj/game/cards/basic.clj
```diff
@@ -125,7 +125,9 @@
                 :effect (req (trigger-event state side :runner-click-draw (-> @state side :deck (nth 0)))
                              (swap! state update-in [:stats side :click :draw] (fnil inc 0))
                              (play-sfx state side "click-card")
-                             (draw state side eid 1))}
+                             (let [bonus-click-draws (get-in @state [:bonus :click-draw] 0)]
+                               (swap! state update :bonus dissoc :click-draw);; clear bonus draws
```
NoahTheDuke, line 129: If this needs to be cleared, write a function for it please. That way we only have one place to change if the key changes.

NoahTheDuke:
This is more clever than I had intended, nicely done.

PR #7186, opened 2023-10-19
Sneakdoor Prime A & B

butzopower:
Quite a late addition to implement some campaign cards from 2016 but the new Chimera format that builds random singleton eternal decks is sourcing cards from the entire pool, so thought I'd implement some of the missing ones so they can more easily be used in that format.

### DIFF
--- a/src/clj/game/cards/programs.clj
+++ b/src/clj/game/cards/programs.clj
@@ -53,8 +53,8 @@
                            update-current-encounter]]
    [game.core.sabotage :refer [sabotage-ability]]
    [game.core.say :refer [system-msg]]
-   [game.core.servers :refer [is-central? is-remote? protecting-same-server?
-                              target-server zone->name]]
+   [game.core.servers :refer [central->name is-central? is-remote? protecting-same-server?
+                              target-server unknown->kw zone->name]]
    [game.core.shuffling :refer [shuffle!]]
    [game.core.tags :refer [gain-tags lose-tags]]
    [game.core.to-string :refer [card-str]]
@@ -2784,6 +2784,29 @@
                                     :effect (req (swap! state assoc-in [:run :server] [:hq])
                                                  (trigger-event state :corp :no-action))}])
                                 (make-run eid :archives (get-card state card)))}]})
+
+(defcard "Sneakdoor Prime B"
+  {:abilities [{:cost [:click 2]
+                :prompt "Choose a server"
+                :choices (req (cancellable
+                                (->> runnable-servers
+                                     (map unknown->kw)
+                                     (filter is-central?)
+                                     (map central->name))))
+                :msg "make a run on central server"
+                :makes-run true
+                :async true
+                :effect (effect (register-events
+                                  card
+                                  [{:event :pre-successful-run
+                                    :duration :end-of-run
+                                    :unregister-once-resolved true
+                                    :prompt "Choose a server"
+                                    :choices (req (cancellable remotes))
+                                    :msg (msg "change the attacked server to " target)
+                                    :effect (req (swap! state assoc-in [:run :server] [(unknown->kw target)])
+                                                 (trigger-event state :corp :no-action))}])
+                          (make-run eid target card))}]})
 
 (defcard "Snitch"
   {:events [{:event :approach-ice
--- a/src/clj/game/cards/programs.clj
+++ b/src/clj/game/cards/programs.clj
@@ -54,7 +54,7 @@
    [game.core.sabotage :refer [sabotage-ability]]
    [game.core.say :refer [system-msg]]
    [game.core.servers :refer [central->name is-central? is-remote? protecting-same-server?
-                              target-server unknown->kw zone->name]]
+                              remote->name target-server unknown->kw zone->name]]
    [game.core.shuffling :refer [shuffle!]]
    [game.core.tags :refer [gain-tags lose-tags]]
    [game.core.to-string :refer [card-str]]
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/programs.clj
```diff
@@ -2785,6 +2785,55 @@
                                                  (trigger-event state :corp :no-action))}])
                                 (make-run eid :archives (get-card state card)))}]})
 
+(defcard "Sneakdoor Prime A"
+         {:abilities [{:cost [:click 2]
```
NoahTheDuke, line 2789: Indentation

src/clj/game/cards/programs.clj
```diff
@@ -2785,6 +2785,55 @@
...
+                       :effect (effect (register-events
+                                         card
+                                         [{:event :pre-successful-run
+                                           :duration :end-of-run
+                                           :unregister-once-resolved true
+                                           :prompt "Choose a server"
+                                           :choices (req (->> servers
+                                                              (map unknown->kw)
+                                                              (filter is-central?)
+                                                              (map central->name)))
```
NoahTheDuke, line 2808: This can just be `["Archives" "R&D" "HQ"]`.

src/clj/game/cards/programs.clj
```diff
@@ -2785,6 +2785,55 @@
...
+                                           :msg (msg "change the attacked server to " target)
+                                           :effect (req (swap! state assoc-in [:run :server] [(unknown->kw target)])
+                                                        (trigger-event state :corp :no-action))}])
```
NoahTheDuke, line 2811: Is this necessary?

src/clj/game/cards/programs.clj
```diff
@@ -2785,6 +2785,55 @@
...
+                                       (->> runnable-servers
+                                            (map unknown->kw)
+                                            (filter is-remote?)
+                                            (map remote->name))))
+                       :msg "make a run on a remote server"
+                       :makes-run true
+                       :async true
+                       :effect (effect (register-events
+                                         card
+                                         [{:event :pre-successful-run
```
NoahTheDuke, line 2801: It's ambiguous if this needs to check if it matches the target server or not. What do you think?

NoahTheDuke (changes requested):

src/clj/game/cards/programs.clj
butzopower: Do you mean in case the runner was switched to another server? It's hard to tell from the card text:

> Make a run on a remote server. If successful, instead treat it as a successful run on a central server.

I think Sneakdoor Beta is checking archives but I wasn't sure if that was actually a ruling that the run had to be successful on archives.

src/clj/game/cards/programs.clj
butzopower: really need to figure out how to get IntelliJ to indent clj properly

src/clj/game/cards/programs.clj
butzopower: maybe not - I copied it from Sneakdoor Beta

src/clj/game/cards/programs.clj
NoahTheDuke: There should be some way to mark `defcard` as `def`, so it'll just inherit those indentation settings.

src/clj/game/cards/programs.clj
NoahTheDuke: I know it's a dumb card, but maybe ask in one of the rules channels, see what people think?

src/clj/game/cards/programs.clj
NoahTheDuke: https://cursive-ide.com/userguide/formatting.html#code-style-settings look there

src/clj/game/cards/programs.clj
butzopower: Wild - checking with GLC and it looks like these cards as written because they have 

> If successful,

Then the do need to match the target server for the effect to be made according to:

> 6.7.4. Many abilities that initiate a run contain a conditional ability with the trigger condition “If successful”. This means, “After the run created this way becomes successful”. 
a. If an effect that initiates a run specifies that that run must be made against a specific server or servers, any “If successful” abilities associated with that effect are tied to that server or servers. If the attacked server has changed since the run was initiated, and no longer is a server that could have been chosen for this run, the “If successful” abilities do not meet their trigger conditions.

But funnily enough `Sneakdoor Beta` has actually had it's own text changed to:

> :CLICK Run Archives. If that run would be declared successful, change the attacked server to HQ for the remainder of that run.

So technically these should check that the run succeeds on the target server but Sneakdoor Beta should be updated to no longer care about that restriction!

src/clj/game/cards/programs.clj
butzopower: Ah awesome, that totally did the trick! Classic JetBrains "command+return" gets you 99% of the way to anything.

src/clj/game/cards/programs.clj
butzopower: Okay there's a big discussion on GLC about what's right here for current Sneakdoor Beta - but I definitely think it's clear these cards should have the restriction so I'll add them.

NoahTheDuke: Cool, thank you.

src/clj/game/cards/programs.clj
butzopower: 👍🏻

src/clj/game/cards/programs.clj
butzopower: Removed as I don't think this was necessary

PR #7190, opened 2023-10-29
Ensure Kimberlite Field uses printed rez cost

### LINKED ISSUE #7189: Kimberlite Field uses actual rez cost for trashing rather than printed rez cost
Link to bug replay: https://jinteki.net/bug-report/d0554f87-561d-4d0b-a3d6-c5da08532266?b=0

Description:

1. Create a remote server containing Tucana, Vovo Ozetti, and Kimberlite Field
2. Score the Kimberlite Field
3. Use Tucana to install Magnet on the same server, rezzing it for net 1 credit (because of Vovo)
4. Use Kimberlite Field prompt to trash Magnet
5. Resulting prompt for trashing runner card reads "Trash a card costing up to 1" instead of "trash a card costing up to 3"

butzopower:
Fixes #7189 

> When you score this agenda, you may trash 1 of your rezzed cards. If you do, trash 1 installed Runner card with a printed install cost equal to or less than the printed rez cost of the Corp card you trashed.

### DIFF
--- a/src/clj/game/cards/agendas.clj
+++ b/src/clj/game/cards/agendas.clj
@@ -1048,7 +1048,7 @@
     :choices {:card #(rezzed? %)}
     :cancel-effect (effect (system-msg (str "declines to use " (:title card)))
                            (effect-completed eid))
-    :effect (req (let [target-cost (rez-cost state :corp target)]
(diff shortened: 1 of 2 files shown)


butzopower:
Actually - this should use printed runner cost as well, so I'll fix that too

PR #7192, opened 2023-10-30
Fix Yakov firing on all trashed cards when included in trash

butzopower:
This fixes this issue mentioned on GLC.

image

The existing issue is because `:once-per-instance false` makes it way to the actual `on-trash` effect resolving, and therefore it resolves for ever single card being trashed at the same time by MAD (or whatever else) regardless of any criteria.  This means any `:req` added is really just to check if the event should fire at all (only in the context of Yakov) and then the effect fires for every instance of card being trashed at the same time.  I don't think this is the correct keyword for this situation.

In an ideal engine scenario: when everything is trashed by MAD, Yakov would fire the `:on-trash` event on itself being trashed and still be able to trigger events off its `:corp-trash` event to catch the other cards being trashed and determine if they should gain credits or not.  Unfortunately it's not able to do that when it itself is part of the list of cards being trashed as I believe it's already gone by the time the actual `:corp-trash` event comes around to resolve.

The implementation for this to actually work is a total hack and I'm not even sure there's precedent for it.  I noticed through all this debugging that the `:on-trash` event actually receives the entire list of cards being trashed in `targets`.  So using that list of trashing cards check to see if this any of them meet the Yakov criteria (including Yakov itself - which of course succeeds since it was in the same server as itself) against the card triggering the `:on-trash` event whose reference has actually been moved so need to check its `:previous-zone` (again it checks itself as well).

I'm not sure how to really untangle a better solution out of the engine, I feel like there's maybe an exception to be made for certain kinds of events that should also be able to trigger on cards in the middle of being trashed but it will be quite complicated least of all because the engine will still need to trash not firing that event for a card against itself (it should prefer `:on-trash` instead of `:xxx-trash` in that scenario I think).

### DIFF
--- a/src/clj/game/cards/upgrades.clj
+++ b/src/clj/game/cards/upgrades.clj
@@ -47,7 +47,7 @@
    [game.core.say :refer [system-msg]]
    [game.core.servers :refer [central->zone from-same-server? in-same-server?
                               is-central? protecting-same-server? same-server?
-                              target-server unknown->kw zone->name]]
+                              same-zone? target-server unknown->kw zone->name]]
    [game.core.shuffling :refer [shuffle!]]
    [game.core.tags :refer [gain-tags]]
    [game.core.threat :refer [threat-level]]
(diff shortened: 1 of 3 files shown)


src/clj/game/cards/upgrades.clj
```diff
@@ -1838,10 +1838,14 @@
                  (corp? (:card target))
                  (installed? (:card target))))]
     {:on-trash {:async true
-                :once-per-instance false
                 :interactive (req true)
-                :msg "gain 2 [Credits]"
-                :effect (effect (gain-credits eid 2))}
+                :effect (req (doseq [t targets]
```
NoahTheDuke, line 1842: Can't call `doseq` on async functions. You need to either use recursion to gain the credits one at a time or do something like `(gain-credits state side eid (filter #(and ...) (map :card targets)))`.

butzopower: I wonder if this ability even needs to be async anymore (or in the first place)? I might have just inherited it from its old implementation? Maybe `:interactive` is also unnecessary?

If still required, do you have an example of where we do this with recursion? I don't know if the `args` route will work with gain-credits since I want to provide a message for each credit gain as well.

NoahTheDuke:
I don't think this is the right solution. The issue is that we're checking the `:on-trash` effect's req too soon. It needs to be unconditionally passed to the checkpoint, and be checked there.

butzopower:
Superseded by much nicer implementation in #7193

PR #7209, opened 2023-11-11
Fix indexes on printed subs when ice loses all variables subs

### LINKED ISSUE #6584: Envelopment gains second "Trash this ice" Subroutine
image
Envelop seems to be gaining a second "Trash this ice" subroutine when it loses its last power counter.

### LINKED ISSUE #7201: Bankhar incorrectly interact with Envelopment
Link to bug replay: https://jinteki.net/bug-report/c1176d41-2534-4fa2-82d1-947e9369d17a?b=0

Description:
Running a remote server with Bankhar on an empty Envelopment.
Should be just 1 damage? 

But log shows 

Tankston trashes Lago Paranoá Shelter due to net damage.
Tankston uses Tsakhia "Bankhar" Gantulga to force the Corp to resolve " Do 1 net damage".
yaozheng1996 uses Envelopment to do 1 net damage.
Tankston has the option to prevent damage.
Tankston will not prevent damage.
Tankston trashes due to net damage.
Tankston is flatlined.
yaozheng1996 wins the game.
yaozheng1996 resolves **2** unbroken subroutines on Envelopment (**" Trash this ice" and " Trash this ice"**).

butzopower:
Fixes #7201 
Fixes #6584 

The underlying issue for both these is that the subroutine indexes were only being updated via `add-sub` function, and so when there were 0 calls to `add-sub`, then the original subroutines stick around but never had their indexes reset.  This would cause an issue further in the game during ice encounter in `reset-sub` which relies on `:index` being correct to `assoc` the subroutines back into the list properly after removing broken/fired/resolved status - effectively cloning any subroutines whose indexes were previously greater than or equal to the current number of subs.

This would have caused issues I think only with ice that appends variable subs to the front, so this hopefully fixes issues with Blockchain, Envelopment, and Loki.

### DIFF
--- a/src/clj/game/cards/ice.clj
+++ b/src/clj/game/cards/ice.clj
@@ -78,6 +78,7 @@
          new-subs (->> (range total)
                        (reduce (fn [ice _] (add-sub ice sub (:cid ice) args)) new-card)
                        :subroutines
+                       (map-indexed (fn [idx sub] (assoc sub :index idx)))
                        (into []))
          new-card (assoc new-card :subroutines new-subs)]
      (update! state :corp new-card)
@@ -90,6 +91,7 @@
          new-subs (->> (range total)
                        (reduce (fn [ice _] (add-sub ice sub (:cid ice) args)) card)
                        :subroutines
+                       (map-indexed (fn [idx sub] (assoc sub :index idx)))
                        (into []))
          new-card (assoc card :subroutines new-subs)]
      (update! state :corp new-card)
--- a/src/clj/game/cards/ice.clj
+++ b/src/clj/game/cards/ice.clj
@@ -75,11 +75,12 @@
                                 (:variable %))
                           (:subroutines card))
          new-card (assoc card :subroutines old-subs)
-         new-subs (->> (range total)
-                       (reduce (fn [ice _] (add-sub ice sub (:cid ice) args)) new-card)
-                       :subroutines
-                       (map-indexed (fn [idx sub] (assoc sub :index idx)))
-                       (into []))
+         new-subs (if (pos? total)
+                    (->> (range total)
+                         (reduce (fn [ice _] (add-sub ice sub (:cid ice) args)) new-card)
+                         :subroutines
+                         (into []))
+                    (into [] (map-indexed (fn [idx sub] (assoc sub :index idx)) (:subroutines new-card))))
          new-card (assoc new-card :subroutines new-subs)]
      (update! state :corp new-card)
      (trigger-event state side :subroutines-changed (get-card state new-card)))))
@@ -91,7 +92,6 @@
          new-subs (->> (range total)
                        (reduce (fn [ice _] (add-sub ice sub (:cid ice) args)) card)
                        :subroutines
-                       (map-indexed (fn [idx sub] (assoc sub :index idx)))
                        (into []))
          new-card (assoc card :subroutines new-subs)]
      (update! state :corp new-card)
--- a/src/clj/game/core/ice.clj
+++ b/src/clj/game/core/ice.clj
@@ -218,14 +218,11 @@
   [state ice]
   (update! state :corp (dont-resolve-all-subroutines ice)))
 
-(defn reset-sub
-  [sub]
-  (dissoc sub :broken :fired :resolve))
-
 (defn reset-all-subs
   "Mark all broken/fired subroutines as unbroken/unfired"
   [ice]
-  (update ice :subroutines #(into [] (map reset-sub %))))
+  (letfn [(reset-sub [sub] (dissoc sub :broken :fired :resolve))]
+    (update ice :subroutines #(into [] (map reset-sub %)))))
 
 (defn reset-all-subs!
   "Marks all broken subroutines as unbroken, update!s state"
--- a/src/clj/game/core/ice.clj
+++ b/src/clj/game/core/ice.clj
@@ -219,17 +219,13 @@
   (update! state :corp (dont-resolve-all-subroutines ice)))
 
 (defn reset-sub
-  [ice sub]
-  (assoc ice :subroutines (assoc (:subroutines ice) (:index sub) (dissoc sub :broken :fired :resolve))))
-
-(defn reset-sub!
-  [state ice sub]
-  (update! state :corp (reset-sub ice sub)))
+  [sub]
+  (dissoc sub :broken :fired :resolve))
 
 (defn reset-all-subs
   "Mark all broken/fired subroutines as unbroken/unfired"
   [ice]
-  (reduce reset-sub ice (:subroutines ice)))
+  (update ice :subroutines #(into [] (map reset-sub %))))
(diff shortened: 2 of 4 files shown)


NoahTheDuke:
Maybe the test should be more explicit about what it's checking? Checking the total number doesn't seem comprehensive.

butzopower:
@NoahTheDuke The original test was more showing how the issue manifested (e.g. only after doing an encounter would the subroutines increase). I've updated the test to check the `:index` property specifically as that was the issue with the implementation which potentially affects more than just the `reset-subs` function.  I sort of feel it leaks into the test a bit of  what should be private engine info (or perhaps we can refactor `:index` out entirely? it's a bit weird to have to track an items position in a list) but it's definitely more explicit and scoped in terms of what was previously incorrect with the data.

butzopower:
@NoahTheDuke I've also added a refactor to `reset-all-subs` that I think does the same thing without worrying about the `:index` of a sub.  I've also removed `reset-sub!` as it seemed like it was unused.

NoahTheDuke:
I checked the originating PR for `:index`, and it was to allow for calling `get-in` to get a specific ice without doing a full scan, O(1) vs O(n). This mattered when we're building the log message to sort by the order of the subroutines on the ice.

I think we use it in other ways too now but I don't remember. I would prefer to keep it, especially given that you've caught the couple places I missed initially.

src/clj/game/cards/ice.clj
```diff
@@ -78,6 +78,7 @@
          new-subs (->> (range total)
                        (reduce (fn [ice _] (add-sub ice sub (:cid ice) args)) new-card)
                        :subroutines
+                       (map-indexed (fn [idx sub] (assoc sub :index idx)))
```
NoahTheDuke, line 81: This shouldn't be necessary. It's already done by `add-ice`. Are you sure this is the source of the problem?

src/clj/game/cards/ice.clj
```diff
@@ -90,6 +91,7 @@
          new-subs (->> (range total)
                        (reduce (fn [ice _] (add-sub ice sub (:cid ice) args)) card)
                        :subroutines
+                       (map-indexed (fn [idx sub] (assoc sub :index idx)))
```
NoahTheDuke, line 94: Same as above.

butzopower:
@NoahTheDuke I wonder if that's `:index` for ICE rather than the `:index` used on the subs? Or maybe it was for both - it's actually hard to determine from the code unfortunately because they use the same name :/

I think it's fine to keep it around, but I think if we can not rely on it for things like "reset everything in this list" that's probably safer just in case it gets stale.

src/clj/game/core/ice.clj
```diff
@@ -219,17 +219,13 @@
...
 
 (defn reset-sub
-  [ice sub]
-  (assoc ice :subroutines (assoc (:subroutines ice) (:index sub) (dissoc sub :broken :fired :resolve))))
-
-(defn reset-sub!
-  [state ice sub]
-  (update! state :corp (reset-sub ice sub)))
+  [sub]
+  (dissoc sub :broken :fired :resolve))
```
NoahTheDuke, line 223: I would prefer to keep the api the same as before. Every sub manipulation function takes ice and sub, so this one being different isn't great.

Given that `:index` isn't being touched by this function, why shouldn't we updating it on the ice directly? When will resetting a sub change its indexes?

(Same question applies to reset-all-subs too.)

NoahTheDuke (changes requested):

src/clj/game/cards/ice.clj
butzopower: Yes - the problem is when Envelopment reaches 0 counters, then `reset-variable-subs` gets called with `total` equal to 0 which means the `(range 0)` reduces with no calls to `add-ice`, so no subroutines get added but also the previous printed subroutines don't get their indexes reset.

If you remove this line you'll see the test fail with the only sub in Envelopment having an `:index` equal to `1` (instead of `0`).

src/clj/game/core/ice.clj
butzopower: The problem arose here when you ended up with something like: 

```
{:subroutines ({:index 1 ...})}
```

With the associating directly into the subroutines list via list index, this actually did a copy rather than an update, hence why the `Trash this ICE` sub was being duplicated.

To me the top level function looks like it just wants to clear the resolved/broken/fired flags on all subroutines for the ice, this seems completely like something that should just apply to everything in the list and feels more like a straight up `map` of something rather than something that has to be concerned about the stored indexes being correct.

butzopower: I would actually be fine with just inlining and deleting `reset-sub` as well since it's currently only used in `reset-all-subs`.

src/clj/game/cards/ice.clj
NoahTheDuke: Ah yeah, range + reduce can be silly. This is a bit nit picky, but I'd prefer to only call map-indexed if total is 0. Add-sub encompasses everything needed to correctly add a sub, so it's not clear why this call is necessary.

src/clj/game/core/ice.clj
NoahTheDuke: Deleting the function is fine with me.

src/clj/game/cards/ice.clj
butzopower: I wonder if like an `add-subs` function that takes a list of subs to add which then resets the indexes would be better or something.  Like the current way actually just recalculates the indexes n times which also seems pretty weird.  I wasn't sure all the places we are using `add-sub` where it's not in bulk already 🤔

NoahTheDuke:
> I wonder if that's `:index` for ICE rather than the `:index` used on the subs?

`break-subroutine-msg` uses the `:index` of the subroutine.

src/clj/game/cards/ice.clj
NoahTheDuke: Yeah, recalculating the index n times is annoying but I couldn't figure out a way to only update the indexes from new-index onward (which is an easy thing to do in javascript lol).

Add-subs is hard because each sub might have a different index to be placed in which will influence the position of later subs, and subs are added one at a time according to the CR.

src/clj/game/core/ice.clj
butzopower: Cool - I still declared it separately via `leftfn` which names it a bit (it was hard for me in the old code to figure out what was happening) but also because my clojure-fu is weak and I wasn't sure how to do the map function inside the inline function

src/clj/game/cards/ice.clj
butzopower: Removed this re-indexing as I don't think it should be required for this one since I think it's append only.

src/clj/game/core/ice.clj
NoahTheDuke: Perfect use of letfn.

PR #7215, opened 2023-11-14
fix Oppo Research prompt showing when the Corp has fewer than 5 credits

francescopellegrini:
Also moved Threat check to continue-ability's `:req`

### DIFF
--- a/src/clj/game/cards/operations.clj
+++ b/src/clj/game/cards/operations.clj
@@ -1777,18 +1777,18 @@
                            (last-turn? state :runner :stole-agenda)))
              :effect (req
                        (wait-for (gain-tags state :corp (make-eid state eid) 2)
-                                 (if (threat-level 3 state)
-                                   (continue-ability
+                                 (continue-ability
                                      state side
                                      {:optional
                                       {:prompt "Pay 5 [Credit] to give the Runner 2 tags?"
+                                       :req (req (and (threat-level 3 state)
+                                                      (can-pay? state :corp eid card nil [:credit 5])))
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/operations.clj
```diff
@@ -1779,18 +1779,18 @@
...
              :effect (req
                        (wait-for (gain-tags state :corp (make-eid state eid) 2)
-                                 (if (threat-level 3 state)
-                                   (continue-ability
+                                 (continue-ability
                                      state side
                                      {:optional
                                       {:prompt "Pay 5 [Credit] to give the Runner 2 tags?"
+                                       :req (req (and (threat-level 3 state)
+                                                      (can-pay? state :corp eid card nil [:credit 5])))
```
NoahTheDuke, line 1787: weird that this isn't being checked by the resolve-ability system.

francescopellegrini: That's what I thought too

NoahTheDuke: Okay, on further reflection, this line isn't needed. The players should click "No" when it's not available.

francescopellegrini: I'm sure that the ability won't resolve if players select Ok, so are you fine with the prompt showing up?

NoahTheDuke: Yeah, the instructions resolve, the player just has to choose "No". With my other PR merged, this will work correctly by not showing "Yes".

PR #7216, opened 2023-11-14
Fix additional credit costs not being ignored by :ignore-costs :rez-costs

### LINKED ISSUE #6824: Ob ability is not ignoring additional credit costs from tread lightly (possibly others?)
The Ob ability should ignore all credit costs including those from tread lightly but it is currently making the player pay them.

### LINKED ISSUE #7214: Ob doesn't ignore Tread Lightly cost
Link to bug replay: https://jinteki.net/bug-report/b0f533ab-1486-42be-aa16-1ebc77eff123?b=0

Description:

Reina Roja ran against a server with border control using Tread Lightly. Rezzed the BC, paying the extra 3. Let the runner past 2 more unrezzed ICE before firing BC and getting Sandstone. Was forced to pay the extra 3 from Tread Lightly while Ob should ignore credit costs to rez.

butzopower:
Fixes #7214
Fixes #6824

### DIFF
--- a/src/clj/game/core/cost_fns.clj
+++ b/src/clj/game/core/cost_fns.clj
@@ -41,11 +41,13 @@
           (max 0)))))
 
 (defn rez-additional-cost-bonus
-  [state side card]
-  (merge-costs
-    (concat (:additional-cost card)
-            (:additional-cost (card-def card))
-            (get-effects state side :rez-additional-cost card))))
+  ([state side card] (rez-additional-cost-bonus state side card nil))
+  ([state side card pred]
+   (let [costs (merge-costs
+                 (concat (:additional-cost card)
+                         (:additional-cost (card-def card))
+                         (get-effects state side :rez-additional-cost card)))]
+     (if pred (filterv pred costs) costs))))
 
 (defn score-additional-cost-bonus
   [state side card]
--- a/src/clj/game/core/rezzing.clj
+++ b/src/clj/game/core/rezzing.clj
@@ -26,7 +26,8 @@
       (= :all-costs ignore-cost) [:credit 0]
       alternative-cost alternative-cost
       :else (let [cost (rez-cost state side card {:cost-bonus cost-bonus})
-                  additional-costs (rez-additional-cost-bonus state side card)]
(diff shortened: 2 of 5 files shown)


butzopower:
Not sure if this is better with the refactor or was better with the `remove` filtering out the credit costs.

src/clj/game/core/cost_fns.clj
```diff
@@ -41,11 +41,12 @@
...
 
 (defn rez-additional-cost-bonus
-  [state side card]
-  (merge-costs
-    (concat (:additional-cost card)
-            (:additional-cost (card-def card))
-            (get-effects state side :rez-additional-cost card))))
+  ([state side card] (rez-additional-cost-bonus state side card (constantly true)))
+  ([state side card pred]
+   (filter pred (merge-costs
```
NoahTheDuke, line 46: I'd prefer `(let [costs (merge-costs ...)] (if pred (filterv pred costs) costs))`. No need to do extra work.

src/clj/game/core/rezzing.clj
```diff
@@ -26,7 +26,8 @@
       (= :all-costs ignore-cost) [:credit 0]
       alternative-cost alternative-cost
       :else (let [cost (rez-cost state side card {:cost-bonus cost-bonus})
-                  additional-costs (rez-additional-cost-bonus state side card)]
+                  additional-costs (rez-additional-cost-bonus state side card #(or (nil? ignore-cost)
```
NoahTheDuke, line 29: I don't like this here. Instead, `(when ignore-cost #(= :credit ...))`. Along with the change above, no need to check every step.

src/clj/game/core/cost_fns.clj
butzopower: happy with either, kinda took the first one to be more functional / less having to declare extra state but also think the `if` does read more straightforward

src/clj/game/core/rezzing.clj
butzopower: Ah ya this is much better and makes sense then for the above to use `nil` instead of a true function

PR #7222, opened 2023-11-16
modifies engine to accept ability-title as an input parameter for events

NBKelly:
This is probably not the best way to do it, but tbh that entire system seems a bit confusing to me.

This should allow multiple events on the same card with the same timing to be named and arbitrarily fired in any order. It also means you can name some of the triggers, though I'm not sure we need it right now.

I'll figure out a cleaner way to write this a bit later, just had to put it up while I remembered.

### DIFF
--- a/src/clj/game/core/engine.clj
+++ b/src/clj/game/core/engine.clj
@@ -717,7 +717,8 @@
                                           (not (and silent-fn
                                                     (silent-fn state side (make-eid state) card event-targets))))
                                        handlers)
-                    titles (map :card non-silent)
+                    titles (map #(assoc (:card %) :title (or (:ability-name (:ability %))
+                                                             (:title (:card %)))) non-silent)
                     interactive (filter #(let [interactive-fn (:interactive (:ability %))
                                                card (card-for-ability state %)]
                                            (and interactive-fn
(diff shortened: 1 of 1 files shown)


NoahTheDuke:
Hard to review without a demonstration

NBKelly:
No worries.

Here's a simple example where the (1996) card 'Butcher Boy' registers a handler against the runner ID (it needs to persist indefinitely, even if the card vanishes). Since it's registered under the ID, it's competing with MaxX to trigger first.

Prior to this change, you wouldn't be able to distinguish the triggers or pick which trigger to resolve first.

demo

demo1

demo2

I probably still need to rewrite this to be a bit less hacky.

NoahTheDuke:
This is good but after thinking about it a bit, best to hold off until we have a need in jnet proper.

NoahTheDuke:
I think we need this now? Or did you end up adding it in another PR that we've since merged? Either way, status?

NBKelly:
Oh, I don't think we need it yet. Sooner or later we might, so it's fine to let this one sit.

NoahTheDuke:
This was done on the other pr, right?

NBKelly:
Yeah, I'm going to close this one

PR #7226, opened 2023-11-18
Add settings to allow sorting archives and heap popups

butzopower:
image

Screen Shot 2023-11-18 at 13 15 32

Screen Shot 2023-11-18 at 13 15 43

Intentionally didn't apply this to the card that shows at the top of the archives/heap so you can always know what the last card you played was.

### DIFF
--- a/src/cljs/nr/gameboard/board.cljs
+++ b/src/cljs/nr/gameboard/board.cljs
@@ -289,6 +289,18 @@
      [tag {:src (str "/img/" card-back "-" s ".png")
            :alt alt}])))
 
+(defn sort-archives
+  [cards] (->> cards (sort-by get-title) (sort-by #(not (faceup? %)))))
+
+(defn sort-heap
+  [cards] (sort-by get-title cards))
+
+(defn sort-archives?
+  [] (get-in @app-state [:options :archives-sorted] false))
+
+(defn sort-heap?
+  [] (get-in @app-state [:options :heap-sorted] false))
+
 (defn card-img
   "Build an image of the card (is always face-up). Only shows the zoomed card image, does not do any interaction."
   [{:keys [code] :as card}]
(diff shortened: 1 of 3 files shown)


NoahTheDuke:
Great idea!

src/cljs/nr/gameboard/board.cljs
```diff
@@ -289,6 +289,18 @@
...
            :alt alt}])))
 
+(defn sort-archives
+  [cards] (->> cards (sort-by get-title) (sort-by #(not (faceup? %)))))
+
+(defn sort-heap
+  [cards] (sort-by get-title cards))
+
+(defn sort-archives?
+  [] (get-in @app-state [:options :archives-sorted] false))
```
NoahTheDuke, line 299: Is app-state an ratom? If so, should these be cursors?

butzopower: app-state is an ratom but it didn't seem like we had a cursor for any of the other options in the panel either.  I did some local testing and toggling these on and off from the settings panel automatically updated any open discard popup windows - so maybe something else is triggering a render cycle.  Maybe we should move all of options into a cursor, but currently the following options are not: `:card-back` `:language` `:card-resolution` `:show-alt-art` `:log-width` `:card-zoom` `:pin-zoom` and lots of instances of `:stacked-cards` options.

butzopower: Happy to put up a separate PR to move all them into a cursor and pass them through to the different components.

NoahTheDuke: > so maybe something else is triggering a render cycle.

Updating an ratom starts a render cycle for everything that derefs it. So clicking the checkbox will update every usage, such as the discard here.

I don't think it's worth worrying about right now.

PR #7230, opened 2023-11-22
Adding settings for showing card labels for face-up and/or unrezzed cards

butzopower:
I've ported over the labelling from the Cyberfeeder Firefox plugin (with permission from Rhahi) as these seem like they'd be very generally useful, especially for people who are streaming and having to constantly hover over cards for viewers.

Below are some screenshots - I think I've accounted for most scenarios but I'm sure there's some odd edge-case missing:

**Runner with labels off**
runner-no-labels

**Runner with labels on**
runner-with-labels

**Corp with labels off**
corp-no-labels

**Corp with just unrezzed labels on**
corp-unrezzed-labels

**Corp with all labels on**
corp-all-labels

There are definitely issues with overlapping, currently hovering over a card will make sure it's label is brought to the forefront.  I think many of these issues already exist without labels and I don't think they labels really make anything worse when it gets cluttered (and it's pretty easy to just turn them off).

### DIFF
--- a/src/clj/web/auth.clj
+++ b/src/clj/web/auth.clj
@@ -161,8 +161,9 @@
 
 (defn profile-keys []
   [:background :pronouns :language :default-format :show-alt-art :blocked-users
-   :alt-arts :card-resolution :deckstats :gamestats :card-zoom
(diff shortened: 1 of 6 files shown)


NoahTheDuke:
I've never seen stylus functions before, that's cool

PR #7233, opened 2023-11-24
automated program tester + example use with buzzsaw

NBKelly:
Basically just a way to test out boost/break functionality of programs in 1-2 easy lines.

How it works:
* Mother Goddess gains the subtype we choose (we have a dummy ice on archives)
* we pump mogo by some amount between 7 and 14
* we give it some random number of subroutines
* We assert that we can correctly count to the strength of mogo by boosting the breaker
* We assert that we can correctly count to the subroutine count of mogo by breaking (this takes awkward break counts into account too)
* We assert that we've broken all the subroutines on the ice

The function can be called like:
`(basic-program-test card-name base-strength boost break )`

The `base-strength` strength field can probably be removed, since that's not actually part of the implementation, but I chose to leave it in for now.

where:
Boost is defined as `{:ab a :amount b :cost c}` (or, when I implement it, nil for fixed strength breakers)
Break is defined `{:ab a :amount b :cost c :type d}` where d is an ice subtype, or it can even be a specific ice.

Intent:
Basic tests should be easy to implement - the overhead of writing (most) basic tests shouldn't be significantly higher than the overhead of implementing a card. This can help out when we implement cards in the future by easily verifying they do what they say.


NBKelly:
I updated this to "just work" with fixed strength breakers - you can supply nil as the argument for those purposes. Additionally, there was no unit test for Num. Now there is.

NBKelly:
I've updated this a bit more with what I'm doing on playtest, I'll apply those here at a later date (~1 month), then it should be roughly complete.

NoahTheDuke:
Conflicts

NBKelly:
Alright, I updated this one to align with the current engine.

Sooner or later I'll port over the automated subroutine testing for ice too.

PR #7235, opened 2023-11-24
Rez helper should fail on no rez

butzopower:
The first commit refactors the `rez` test helper to stop passing through arbitrary arguments to the rez action.  This seemed to be used in very random places and almost always out of necessity or just not setting enough credits up at the start of the test.

I then repuprosed these args to allow passing an `:expect-no-rez` option.  I considered inverting this with `:expect-rez false` as the override but I wasn't sure about having the default be a map with values rather than just nil.  I think the consequence is mostly the same and perhaps this should just be made into a more explicit test helper though I'm not exactly sure what because as seen in the tests that change there's quite a few cases where you would expect the rez action to either fail or be deferred from either additional or alternative costs.


test/clj/game/core_test.clj
```diff
@@ -548,17 +548,20 @@
...
 (defn rez-impl
-  [state card & args]
-  (let [card (get-card state card)]
-    (is' (installed? card) (str (:title card) " is installed"))
-    (is' (not (rezzed? card)) (str (:title card) " is unrezzed"))
-    (when (and (installed? card)
-               (not (rezzed? card)))
-      (core/process-action "rez" state :corp (merge {:card card} (first args))))))
+  ([state card] (rez-impl state card nil))
+  ([state card {:keys [expect-no-rez]}]
```
NoahTheDuke, line 552: I would prefer `:expect-rez false`.

```suggestion
  ([state card] (rez-impl state card nil))
  ([state card {:keys [expect-rez] :or {expect-rez true}}]
```

Generally, positive options are more easily understood than negative options.

NoahTheDuke (changes requested):

NoahTheDuke:
This is a great change, thank you for catching it.

test/clj/game/core_test.clj
```diff
@@ -548,17 +548,20 @@
...
+   (let [card (get-card state card)]
+     (is' (installed? card) (str (:title card) " is installed"))
+     (is' (not (rezzed? card)) (str (:title card) " is unrezzed"))
+     (when (and (installed? card)
+                (not (rezzed? card)))
+       (core/process-action "rez" state :corp {:card card})
+       (if expect-no-rez (is' (not (rezzed? (get-card state card))) (str (:title card) " is still unrezzed"))
+                         (is' (rezzed? (get-card state card)) (str (:title card) " is rezzed")))))))
```
NoahTheDuke, line 560: I prefer to put both branches on their own lines.

butzopower: 👍🏻

PR #7241, opened 2023-12-03
Implement BMI Buffer and BMI Buffer 2

butzopower:
Two more campaign cards from TD which have shown up in Chimera

image
image

### DIFF
--- a/src/clj/game/cards/hardware.clj
+++ b/src/clj/game/cards/hardware.clj
@@ -27,7 +27,7 @@
    [game.core.events :refer [event-count first-event? first-trash? no-event?
                              run-events]]
    [game.core.expose :refer [expose]]
-   [game.core.finding :refer [find-card]]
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/hardware.clj
```diff
@@ -234,6 +234,52 @@
...
+(defcard "BMI Buffer"
+  (let [grip-program-trash?
+        (fn [card]
+          (and (runner? card)
+               (program? card)
+               (in-discard? card)
+               (= (first (:previous-zone card)) :hand)))
+        triggered-ability
+        {:async true
+         :effect (req (doseq [c (filter grip-program-trash? (mapv #(find-latest state (:card %)) targets))]
```
butzopower, line 246: Doing a `find-latest` here because the card may have moved since the trash event (for example Buffer Drive) - in the case that it's not gone to heap then it shouldn't host on BMI Buffer.

PR #7242, opened 2023-12-04
Fix Arissana-Jua interaction (and more)

### LINKED ISSUE #7240: Infinite Loop with Jua and Arissana
Link to bug replay: https://jinteki.net/bug-report/75e60b6a-da85-4d44-8bf8-96f34fcc1f1b?b=0

Description:
In this bug, my opponent had a Jua installed in archives and I had multiple programs and a simulchip on the table.  After using Arissana's ability to install Slap Vandal from the hand, the warnings trigger indefinitely about not being able to install my heap programs which blocks the chat console.
image

francescopellegrini:
Fixes #7240.

Also adds heap-locked check to Katorga Breakout and The Price

### DIFF
--- a/src/clj/game/cards/events.clj
+++ b/src/clj/game/cards/events.clj
@@ -2041,7 +2041,8 @@
              :choices (req runnable-servers)
              :effect (effect (make-run eid target card))}
    :events [{:event :successful-run
-             :req (req this-card-run)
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
Thank you for catching these. There's gotta be a better way lol.

PR #7249, opened 2023-12-20
Highlight player name in logs to more easily determine side

butzopower:
image

Adds an additional message render helper to put the player names in different spans when they are found in the message.  This makes it possible to add styles to these spans to make it more obvious if the player is Corp or Runner.  This should also make it much easier to do tricks like redacting the player names in replays, etc.

Additionally I made sure this processing runs before the `render-cards` helper which means that player names will no longer get turned into cards (example is "Imposter" above which normally would get a card link for "Imp").

### DIFF
--- a/src/cljs/nr/utils.cljs
+++ b/src/cljs/nr/utils.cljs
@@ -304,17 +304,22 @@
   [input]
   (render-specials (render-icons (render-cards input))))
 
-(defn highlight-patterns-impl [corp runner]
+(defn- player-highlight-patterns-impl [corp runner]
   (letfn [(regex-of [player-name] (re-pattern (str "(?i)" (regex-escape player-name))))]
     (->> {corp [:span.corp-username corp]
           runner [:span.runner-username runner]}
          (filter (fn [[k _]] (not-empty k)))
          (map (fn [[k v]] [(regex-of k) v]))
          (sort-by (comp count str first) >))))
-(def highlight-patterns (memoize highlight-patterns-impl))
+(def player-highlight-patterns (memoize player-highlight-patterns-impl))
 
-(defn highlight-side [message corp runner]
-  (render-input message (highlight-patterns corp runner)))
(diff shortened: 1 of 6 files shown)


src/cljs/nr/utils.cljs
```diff
@@ -292,12 +292,7 @@
 (defn render-cards
   "Render all cards in a given text or HTML fragment input"
   [input]
-  (cond
```
butzopower, line 295: this function was only ever called by `render-message` - I'm not really sure there's a need to check contains-card-pattern anymore as the card-patterns are now memoized.  Additionally `re-find` will error against non-strings so some of the branches in this condition didn't even make sense.

src/cljs/nr/utils.cljs
```diff
@@ -307,9 +302,19 @@
 (defn render-message
   "Render icons, cards and special codes in a message"
   [input]
-  (if (string? input)
```
butzopower, line 310: simplified all these `render-` functions to just wrap `render-input` which can take a string or a vector of elements, so this guard isn't necessary anymore (I think it originally was for `render-cards`)

src/cljs/nr/utils.cljs
```diff
@@ -307,9 +302,19 @@
...
-    (render-specials (render-icons (render-cards input)))
-    input))
+  (render-specials (render-icons (render-cards input))))
+
+(defn highlight-patterns-impl [corp runner]
+  (letfn [(regex-of [player-name] (re-pattern (str "(?i)" (regex-escape player-name))))]
+    (->> {corp [:span.corp-username corp]
+          runner [:span.runner-username runner]}
+         (filter (fn [[k _]] (not-empty k)))
```
butzopower, line 311: filter is important here if either of the names becomes `""` or nil this will end up wrapping each character with a span

butzopower:
Now can be toggled off in player settings

image

PR #7262, opened 2024-01-10
Add changed? macro

NoahTheDuke:
As discussed on slack.

Includes a wrapper around play-dynamic-ability.


NoahTheDuke:
Ah crap, thanks for double checking that. I thought I had it right. :sob:

PR #7271, opened 2024-01-26
fixed facedown-cards selection to only be able to select each card once

### LINKED ISSUE #7135: We need to handle face-down cards in archives better
When the runner has to randomly access face-downs in archives, we can get weird effects like them making more accesses than there are face-downs, or hitting the same face-down multiple times, etc.

I'm not sure exactly why this is, but it does make our friendly eternal IG players a bit more powerful than they should be.

NBKelly:
Took a while to figure out.

The function (*facedown-cards*) expected a function (*already-accessed-fn*) as input, and we were making the input function, but just not using it and instead passing *already-accessed* as input.

I genuinely don't know how to write a test for this, and would like input for that so we can't recreate this mistake later.

Closes #7135


NoahTheDuke:
You could add a `:pre` condition to `facedown-cards` to require that the provided function is indeed a function with [fn?](https://clojuredocs.org/clojure.core/fn_q). This will throw an assertion error if anything is passed in that's not a function.

```clojure
(defn facedown-cards
  [state already-accessed-fn]
  {:pre [(fn? already-accessed-fn)]}
  ...
```

Holy shit at these function and variable names tho, lmao. It's genuinely hard to keep track of which is which. I have no doubt the majority are my fault. :grimacing:

NoahTheDuke:
As for writing a test, why can't it be as simple as writing a test that `click-card`s the same card in archives twice? If the helpers don't allow it, then that's a mistake and we need to instead fix it.

NBKelly:
I decided that just counting the accesses in archives was correct enough (we can't do click card, since they're all random accesses). In the theoretical case where it's incorrect (accessing the same card 7 times), the test isn't correct, but idk how realistic it is to rig the random accesses for that.

I also updated the procedure for accessing face-down cards from archives so that the runner is given a prompt (default is just no action) - there should be some form of visual queue for all accesses, and there wasn't for explicitly these ones before.

Additionally, I updated the cards with :access-abilities to explicitly not function in archives. Those rules should be managed by the cards themselves, and (potentially) future cards will have access abilities applicable in archives. Consider that a tiny bit of future-proofing.

test/clj/game/cards/agendas_test.clj
```diff
@@ -448,6 +448,32 @@
...
+    (new-game {:corp {:discard ["Hedge Fund" "Bacterial Programming"]
+                      :hand ["Bellona"]
+                      :deck ["IPO" "NGO Front" "Rashida Jaheem" "Tithe"
+                             "Ice Wall" "Fire Wall" "Enigma"]}})
+    (take-credits state :corp)
+    (run-empty-server state :archives)
+    (click-prompt state :runner "Bacterial Programming")
+    (click-prompt state :runner "Steal")
+    (click-prompt state :corp "Yes")
+    (doseq [_ (range 7)
```
NoahTheDuke, line 461: this is `dotimes [_ 7]`

NoahTheDuke:
Nice find and fixes.

test/clj/game/cards/agendas_test.clj
NBKelly: ah, true. 
I just mindlessly copied the other test :joy: 

Updated it on both, keeping the `TODO` in because it's something that should (eventually) be done if it hasn't already.

PR #7276, opened 2024-02-06
show player choice on Hyoubu Precog Manifold, Targeted Marketing, and Saisentan

francescopellegrini:
Similar to what we already do on Security Testing, Tsakhia "Bankhar" Gantulga, etc.


NoahTheDuke:
Thanks for this, will be helpful. Maybe we're at the point of a helper function to tag these things?

PR #7280, opened 2024-02-14
Prevent format filter dropdown from closing when toggling formats

butzopower:


src/cljs/nr/lobby.cljs
```diff
@@ -164,7 +164,9 @@
                               :type "checkbox"
                               :on-change #(on-change-format-visibility slug %)
                               :checked (format-visible? slug)}]
-     [:label {:for id} (-> slug slug->format tr-format)]]))
+     [:label {:for id
+              :on-click #((.stopPropagation %))}
```
NoahTheDuke, line 168:
```suggestion
              :on-click #(.stopPropagation %)}
```
I don't think you meant to call the return value of `stopPropogation`, right?

butzopower: I did not! Fixed.

PR #7281, opened 2024-02-16
fix threat function causing an infinite loop

francescopellegrini:


francescopellegrini:
@NoahTheDuke merge please? 🙏

PR #7282, opened 2024-02-17
Allow /undo-click to go back multiple clicks

butzopower:


butzopower:
Do we have much metrics on the current memory usage of the server - are even large board states taking up that much memory comparatively? I guess right now we are generally tracking two sets of states (current state and last click), this will potentially on average roughly double the memory footprint of games given it clears out the previous clicks at start of turn (which doesn't happen right now either so there's actually a small window where this takes up less memory).

Maybe I can sort out the load testing again (I should try and actually merge the fixes I had for it back over the summer :P) and see the impact.

butzopower:
Actually, I wonder if in a very surprising twist this diff might actually decrease the memory footprint 😮 

We append to the state's `:history` pretty much every time there's any diff calculated - this is always growing over the course of the game, even after 2 turns if you print out the history to a string it's about 56k characters.

The current implementation supporting undo-click is actually _**not**_ doing a `dissoc` on the `:history` key when putting the last click into the state, so this is actually being copied twice into the state.  But even more, the whole state (including this last click state which includes the whole history) is then copied again when the turn starts to support `undo-turn`.

So I think currently at most times in the game you actually have roughly the whole game history stored in the state atom 4 times.

--

The current implementation fixes the issue with the history being retained when storing off previous clicks (didn't really mean to make this as an improvement but realizing it now), and also resets the tracked previous clicks when the turn restarts.

It could definitely be a nice improvement to just fix the above issues and not make it worse, but I guess I'm trying to say that I think this feature at least keeps the memory usage no worse than it was before (though I don't have the profile facts to back this up!)

butzopower:
Okay I did some profiling!

The following is the memory snapshot of a game just finishing the 13th turn of the Runner (so they've taken 4 clicks).  Of the 1.2MB taken up by the lobby state, the click states are contributing 27kb:

Screenshot 2024-02-18 at 17 41 00

(this is a dictionary so the second to last part of the tree is the symbol/key and then the `clojure.lang.PersistentVector` is the value.

After finishing the turn and the Corp starting their next turn we can see that this vector is cleared out:

Screenshot 2024-02-18 at 17 40 18

So I believe the memory footprint here should be relatively small.

NBKelly:
neat

PR #7298, opened 2024-03-14
Ack toasts explicitly and prevent toast cycles

butzopower:
There's currently an issue where toasts seem to be able to cause a cycle where the existing toast isn't cleared but updates the client state to think there is a new toast which leads to a new diff which then cycles infinitely popping up a toast over and over.

The current way of dismissing the toasts is currently not consistent as toasts are just dismissed FIFO regardless of what triggered.  This change adds a uuid to each toast to allow them to be explicitly ack'd.

Further I've memoized the ack function on the client side to ensure it only sends a message once to the server for that client's session to prevent an infinite loop if the message fails to ack on the server for whatever reason.

### DIFF
--- a/src/cljs/nr/gameboard/actions.cljs
+++ b/src/cljs/nr/gameboard/actions.cljs
@@ -108,18 +108,19 @@
       (build-report-url error)
       "');\">Report on GitHub</button></div>")))
 
+(defn ack-toast ([id] (send-command "toast" {:id id})))
+
 (defn toast
   "Display a toast warning with the specified message.
   Sends a command to clear any server side toasts."
   [msg toast-type options]
   (set! (.-options js/toastr) (toastr-options options))
   (let [f (aget js/toastr (if (= "exception" toast-type) "error" toast-type))]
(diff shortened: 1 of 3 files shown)


src/cljs/nr/gameboard/actions.cljs
```diff
@@ -108,18 +108,21 @@
       (build-report-url error)
       "');\">Report on GitHub</button></div>")))
 
+(defn ack-toast-impl ([id] (send-command "toast" {:id id})))
+(def ack-toast (memoize ack-toast-impl))
```
NoahTheDuke, line 112: This might grow big. Should we try to limit it?

butzopower: I spent a bit of time looking into trying to set up a fifo cache memoization strategy with clojure.core.memoize only to realize it's not cljs 👎🏻 - I actually think this memoization maybe isn't necessary/helpful - or if it would help it means the problem is different than what would be solved by giving a message with IDs (e.g. it'd mean unique toasts are being generated each message rather than the same message getting stuck).

PR #7314, opened 2024-03-19
RWR bugfixes

### LINKED ISSUE #7308: Sissyphus protocol forced encounter lets the runner to jack out
Link to bug replay: https://jinteki.net/bug-report/e571d166-12c4-4534-8819-3ba5f601eee7?b=0

Description:

Scored Sissyphus protocol foes the ice encounter again but it lets the runner jack out. I believe it shouldnt since it forces to encounter the ice again.

### LINKED ISSUE #7309: Valentina Ferreira Carvalho triggers when corp removes tag
Card text says whenever "you" remove a tag, but Valentina gives a credit when the corp removes a tag with B-1001
Chat log:
image

### LINKED ISSUE #7310: Juli Moreira Lee triggers multiple times a turn
This happens if you use multiple different cards with click abillities in the same turn. Should only work once.

### LINKED ISSUE #7311: Privileged Access ability doesn't fire
From slack:

> Priveleged access gives you the prompt bt then the effect doesnt fire when you click a legal card
>
> it burns the click and full price

### LINKED ISSUE #7312: Descent doesn't charge the corp a credit when expending the card
It just doesn't cost a credit as well as the click for the expend effect

### LINKED ISSUE #7313: End the Line Opens an Emanuensis dialog for the corp
Runner had Emanuensis out with a counter on it.  I played end the line which removed a tag, and I the corp got a dialog offering to remove a power counter and draw 2 cards.

### LINKED ISSUE #7315: Window of oppotunity doesn't have "no" option on the "may install program or hardware"
When played, the only options available are the ones in hand, when with a may you should be able to decline

### LINKED ISSUE #7316: Kingmaking forces draw 3
As noted, there is no option to draw "up to" 3 as the card should provide.

### LINKED ISSUE #7317: Privileged Access triggers on tags taken from other sources
Privileged Access triggered on a tag taken from Arruaceiras Crew during the run on archive.
It the correctly triggered again after the successful run, but prompted to breach archives after the trigger.

Link to bug replay: https://jinteki.net/replay/a9d1b5c4-dc6f-48d5-8897-1d2e2b8536ca

### LINKED ISSUE #7319: Isaac Liberdade leaves prompt open for opponent
Link to bug replay: https://jinteki.net/bug-report/ac0e40a7-a82b-453a-bafb-11909aad2f63?b=0

Description:

when using Isaac to place an adv counter it apparently left a prompt open for my opponent that made them think I was still acting.

### LINKED ISSUE #7320: Server error when trying to use boi-tata
Link to bug replay: https://jinteki.net/bug-report/35ac9864-0927-43b6-a62f-b1959d69341b?b=0

Description:

After trashing one of my cards, I am unable to break ice with boi-tata

### LINKED ISSUE #7321: Cloud Eater after encounter "trash" option not implemented correctly
When cloud eater fires the "after encounter" text there are two bugs with the trash option.

The first is that if you have things installed and choose trash, it gives the runner the choice of what to trash instead of the corp

The second is that if you have nothing to trash it doesn't let you choose that option even though it should.

### LINKED ISSUE #7322: Pressure Spike "break all" option bugged at threat 4
Once you hit threat 4 and pressure spike has 2 boost options, the match and break all subs option no longer works. This is even the case if the strength is already boosted and it just needs to break subs. I'd suggest either removing it at that point or giving separate break all buttons for if the runner wants the +3 or +9 boost option

### LINKED ISSUE #7324: Lightning Laboratory demanding derez every turn after agenda counter is used
When I used Lightning Laboratory last night, the turn I used the agenda counter, I was rightly required to derez the ice on the relevant server, but going forward for the rest of the game, it continued to require me to derez those ice on that server at the end of every turn.

### LINKED ISSUE #7325: Kingmaking offers no option on how many cards to draw
Link to bug replay: https://jinteki.net/bug-report/70bc7226-e5b4-451d-b4b7-941389f559dd?b=0

Description:

When kingmaking was scored 3 cards were automatically drawn, rather than a choice of 0, 1, 2 or 3 cards.

### LINKED ISSUE #7326: Cards hosted on cupellation aren't disabled
yet another thing that will need to be changed when we finally rejig the engine, BUT at the very least, this one can be implemented

### LINKED ISSUE #7327: Nuvem SA timing issues
Eminent domain doesn't trigger (probably missing async)
Descent triggers before shuffle (probably ending eid early)

NBKelly:
descent charges a credit when expended (closes #7312) - not testing
Juli moreira once per turn (closes #7310)
valentina only when runner removes tags (closes #7309) - unit tests
amaneunsis only works when runner removes tags (closes #7313) - unit tests
cloud eater trash program ability (closes #7321)
lightning lab demanding use every turn (closes #7324) - unit tests
kingmaking draw-up-to (and async on decline) (closes #7316, closes #7325) - unit tests
cupellation disables cards when hosting them (closes #7326) - unit tests
fixed async issues on descent and eminent domain (closes #7327 ) - no tests
fixed isaac lingering prompt (closes #7319) - unit tests
fixed Window of Op optional (closes #7315) - no tests
removed auto-break from pressure spike and boi-tata until we figure those out (closes #7322, closes #7320)
Updated force-ice-encounter to not offer jack-out when used during a run (but before access) (Closes #7308)

===============

Praccess actually works they way it's meant to now. The changes are mostly sane, needing to store an ephemeral state while resolving each install is so that Jarogniew Mercs taking a tag doesn't load a second copy of the same event (which is now resolving in a different window), which would then resolve first (before allowing you to resolve, say, your trigger from seb), before returning you to the first window with an empty button to press (because it's already resolved).
Reading that paragraph may induce a schizm. It is what it is. I'm not probably not writing a unit test for this case, because it's cumbersome to actually reproduce the bad behaviour.
If we get more complaints later then I'll work that out.
Closes #7317 and also Closes #7311 (I'm getting all three installs, before if you ordered wrong you missed one)

### DIFF
--- a/src/clj/game/cards/agendas.clj
+++ b/src/clj/game/cards/agendas.clj
@@ -1922,33 +1922,32 @@
               :effect (effect (damage eid :meat 2 {:card card}))}})
 
 (defcard "Sisyphus Protocol"
-  {:events [{:event :pass-ice
-             :req (req (and (rezzed? (:ice context))
-                            (or (has-subtype? (:ice context) "Code Gate")
-                                (has-subtype? (:ice context) "Sentry"))
-                            (first-event? state side :pass-ice
-                                          (fn [targets]
-                                            (let [context (first targets)]
-                                              (and (rezzed? (:ice context))
-                                                   (or (has-subtype? (:ice context) "Code Gate")
-                                                       (has-subtype? (:ice context) "Sentry"))))))))
-             :prompt (msg "Make the runner encounter " (:title (:ice context)) " again?")
-             :choices (req [(when (can-pay? state :corp (assoc eid :source card :source-type :ability) card nil [:credit 1]) "Pay 1 [Credit]")
-                            (when (can-pay? state :corp (assoc eid :source card :source-type :ability) card nil [:trash-from-hand 1]) "Trash 1 card from HQ")
-                            "Done"])
-             :async true
-             :effect (req (if (= target "Done")
-                            (effect-completed state side eid)
-                            (let [enc-ice current-ice]
-                              (continue-ability
-                                state side
-                                (assoc {:msg (msg "make the runner encounter " (card-str state enc-ice) " again")
-                                        :async true
-                                        :effect (req (force-ice-encounter state side eid enc-ice))}
-                                        :cost (if (= target "Pay 1 [Credit]")
-                                                [:credit 1]
-                                                [:trash-from-hand 1]))
-                                card nil))))}]})
+  (letfn [(rezzed-gate-or-sentry [context]
+            (and (rezzed? (:ice context))
+                 (or (has-subtype? (:ice context) "Code Gate")
+                     (has-subtype? (:ice context) "Sentry"))))]
+    {:events [{:event :pass-ice
+               :req (req (and (rezzed-gate-or-sentry context)
+                              (first-event? state side :pass-ice
+                                            #(rezzed-gate-or-sentry (first %)))))
+               :prompt (msg "Make the runner encounter " (:title (:ice context)) " again?")
+               :choices (req [(when (can-pay? state :corp (assoc eid :source card :source-type :ability) card nil [:credit 1]) "Pay 1 [Credit]")
+                              (when (can-pay? state :corp (assoc eid :source card :source-type :ability) card nil [:trash-from-hand 1]) "Trash 1 card from HQ")
+                              "Done"])
+               :async true
+               :effect (req (if (= target "Done")
+                              (effect-completed state side eid)
+                              (let [enc-ice current-ice]
+                                (continue-ability
+                                  state side
+                                  (assoc {:msg (msg "make the runner encounter " (card-str state enc-ice) " again")
+                                          :async true
+                                          :effect (req
+                                                    (force-ice-encounter state side eid enc-ice))}
+                                         :cost (if (= target "Pay 1 [Credit]")
+                                                 [:credit 1]
+                                                 [:trash-from-hand 1]))
+                                  card nil))))}]}))
 
 (defcard "Slash and Burn Agriculture"
   {:expend {:req (req (some #(can-be-advanced? %) (all-installed state :corp)))
--- a/src/clj/game/cards/agendas.clj
+++ b/src/clj/game/cards/agendas.clj
@@ -732,7 +732,7 @@
                                           (corp-installable-type? %))}
                     :msg "install and rez 1 card from HQ, paying 5 [Credits] less"
                     :async true
-                    :effect (req (corp-install state side (make-eid state eid) target nil
+                    :effect (req (corp-install state side eid target nil
                                                {:install-state :rezzed
                                                 :combined-credit-discount 5}))}
         score-abi {:interactive (req true)
--- a/src/clj/game/cards/agendas.clj
+++ b/src/clj/game/cards/agendas.clj
@@ -1943,7 +1943,7 @@
                                   (assoc {:msg (msg "make the runner encounter " (card-str state enc-ice) " again")
                                           :async true
                                           :effect (req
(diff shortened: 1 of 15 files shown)


src/clj/game/cards/resources.clj
```diff
@@ -1832,7 +1832,12 @@
   {:data {:counter {:power 4}}
    :events [(trash-on-empty :power)
             {:event :runner-spent-click
+             :once :per-turn
              :req (req (let [all-cards (get-all-cards state)]
+                         ;; TODO - assert this works when (e.g.) using the last click on lib-acc
+                         ;; and having it trashed (all-cards wont find it)
```
NoahTheDuke, line 1838: Why won't all-cards find it?

NoahTheDuke:
I don't wanna get in the way of bug fixes, but could you add some test asserts for some/all of these?

NBKelly:
> I don't wanna get in the way of bug fixes, but could you add some test asserts for some/all of these?

No worries, that was my plan anyway. Just gotta leave it for the weekend since I'm crunched on week days.

src/clj/game/cards/resources.clj
NBKelly: If the card gets trashed through use (ie last click on liberated accounts). I need to test this and check the behaviour, since I think things like that are where the issue actually is, and why some people aren't reproducing the bug and some are - and first-event is checking for cids using the current state.

src/clj/game/cards/agendas.clj
```diff
@@ -1163,6 +1164,8 @@
             {:event :runner-turn-ends
              :req (req (seq (filter #(= (:zone %) [:servers zone :ices])
                                     (all-active-installed state :corp))))
+             :duration :end-of-turn
+             :unregister-once-resolved true
```
NoahTheDuke, line 1168: Only needed if it's expected to fire and before the duration happens. Given that this is a conditional ability, it'll be cleared at turn end.

test/clj/game/cards/agendas_test.clj
```diff
@@ -2132,12 +2132,26 @@
...
         "Corp drew 3 cards (2 of which moved to the score area)")
     (is (zero? (get-counters (get-scored state :runner 1) :agenda)) "House of Knives should have 0 agenda counters")))
 
+(deftest kingmaking-draw-n
+  (dotimes [n 4]
+    (do-game
+      (new-game {:corp {:hand ["Kingmaking" "Hedge Fund"]
+                        :deck ["House of Knives" "Project Atlas" "Hedge Fund"]}})
+      (is (changed? [(count (:deck (get-corp))) (- n)
+                     (count (:hand (get-corp))) (- n 1)]
```
NoahTheDuke, line 2148: How does this go negative?

NBKelly: scoring a card from hand (-1), and drawing 0

NBKelly:
That's all my fixes, I'll add more tests after work on Saturday and we can hopefully push this before the AMT on Sunday

NBKelly:
OK now it's DONE done. The only outstanding issue is Juli - she's still not CORRECT, but she's more correct that she was before. That's a bandaid, and I'll investigate making it correct correct next weekend.

src/clj/game/core/runs.clj
```diff
@@ -350,12 +350,23 @@
...
-            (if (and (not (:run @state))
-                     (empty? (:encounters @state)))
-              (forced-encounter-cleanup state :runner eid)
-              (effect-completed state side eid))))
+  ;; do we need to mess with the run state
+  (let [old-state (get-in @state [:run :phase])
+        new-state (if (= :movement old-state)
+                    :encounter-ice
+                    old-state)]
+    (system-msg state side old-state)
```
NoahTheDuke, line 358: Are these supposed to be here?

NBKelly: nope, but I bet it would have taken weeks for any player to notice :)

PR #7340, opened 2024-03-27
first five mode

NBKelly:
A mode allowing for starting hands to be pulled from the deck (non-dupliates only).

Intended for april fools, but I believe it's a keepable feature.

Why yes, every single corp list does include one copy of Crisium Grid, how could you tell?


NoahTheDuke:
I think we should hold off on this, I don't want to get too far away from what's legal, for fear of confusing new players.

PR #7346, opened 2024-03-28
Rewrite costs to use maps instead of vectors

NoahTheDuke:
Includes a bunch of fixes (Boi-Tata, stealth credits), probably some new bugs, and a general unification/clean up of all cost-related business. There's further work to be done, but this is big enough lol.

Instead of a vector of type and quantity (with stealth credits being in the third spot), it's now a namespaced map: `{:cost/type TYPE :cost/amount N :cost/additional false :cost/stealth X :cost/args {}}`. Using namespaces will make searching and usage more understandable than `:type` and getting like 1000 results. It's also significantly better than using `first` and `second` which we've been doing everywhere to mean `:cost/type` and `:cost/amount`. `value` will still be around for usage when necessary.

One sad issue here is that it's inconsistent whether a given function will return a map or vector, so we have to be careful, making sure to `(concat [cost] ...)` and then flattening, or `(if (map? cost) ...)`. That's kind of annoying but a small price to pay.

Stealth credits are now super duper easy: either a number or `:all-stealth`. This can be set when creating a cost: `(->c :credit 2 {:stealth 1})`. That will be used to know the minimum (or equivalent) number of stealth credits required.

`:cost/additional` isn't used yet but after this is merged, I'm going to wire up support for asking ahead of time about additional costs and exiting early if they're unpaid. Then we can tear out all of the dumb ad-hoc "additional payment" prompts.

`:cost/args` is arbitrary data, currently used for `:source` and `:source-type`. This is only useful in steal-costs but it should probably be made generic for all costs. I don't have a good solution for that yet.

EDIT: did some digging, lol at https://github.com/mtgred/netrunner/pull/5258. I should have done this 4 years ago, and yet here we are. Cleaning up past-Noah's mistakes yet again.


NoahTheDuke:
I'm fuckin blown that all of the tests pass.

PR #7347, opened 2024-03-29
Unhinged/April Fools Set

NBKelly:
This is an unofficial set with cards drafted up by the playtesters during some downtime.

Don't merge this (or do, it doesn't do any harm and the cards can be removed from the build with a single line if desired).

There's only three engine changes:
1) Before trashing consoles, check if there's any effect allowing multiple consoles
2) Forced encounters work on non-ice cards, so long as they have a `:strength` field
3) The interface names the score area if you happen to access a card from a score area (this sounds unhinged because it is)

I added a new format (standard plus), which is the "default format". It's easy enough to change, and if people don't like it, then it's 1 second when making a game to change.

I think 2-3 weeks is probably a good duration for this.

Here are all of the cards I implemented. The dreamer one with rng turns is actually nuts, and I'm shocked I made it work. I'll make a corresponding PR to netrunner-data soon.

### DIFF
--- a/src/clj/game/cards/unhinged.clj
+++ b/src/clj/game/cards/unhinged.clj
@@ -177,9 +177,11 @@
 
 (defcard "You have three cards"
   ;;
-  {:events [
-            {:event :post-corp-draw
-             :req (req (> (count (:hand corp)) 3))
+  {:events [{:event :post-corp-draw
+             :req (req (and (> (count (:hand corp)) 3)
+                            (not (get-in card [:special :trying-to-mulligan]))
+                            (:turn @state)
+                            (pos? (:turn @state))))
              :async true
              :effect (req (let [to-destroy (- (count (:hand corp)) 3)]
                             (continue-ability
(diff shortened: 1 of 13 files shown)


src/cljs/nr/new_game.cljs
```diff
@@ -196,7 +196,7 @@
 
 (defn create-new-game [lobby-state user]
   (r/with-let [state (r/atom {:flash-message ""
-                              :format (or (get-in @app-state [:options :default-format]) "standard")
+                              :format (or (get-in @app-state [:options :default-format]) "standard-plus")
```
NoahTheDuke, line 199: It shouldn't be the default, if we're going to let this run for a while.

src/clj/game/cards/unhinged.clj
```diff
@@ -0,0 +1,617 @@
...
+                  ;; draw a card
+                  :draw {:effect (req
+                                   (if (seq (:deck runner))
+                                     (resolve-bac-abi state 1 nil card eid rem)
+                                     (continue-ability
+                                       state side
+                                       (choose-an-action rem)
+                                       card nil)))
+                         :async true}))))]
+    {;:data {:counter {:credit 2}}
```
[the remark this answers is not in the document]
NBKelly: nah, I took it out. I found that when there were recurring credits, they didn't behave well when resolving stuff at the start of the turn

NBKelly:
I added the weyland cards I forgot to add before.

src/cljs/nr/new_game.cljs
NBKelly: true - fixed

PR #7350, opened 2024-03-29
icons for april fools

NBKelly:
icons2_000
icons2
icons

This will probably only be funny for a week, but it's a really simple one to do/undo (just rebuild css - the server doesn't even need to restart I think)


NBKelly:
Now it's an optional feature.

feature

PR #7351, opened 2024-03-31
Unhinged updated costs

NBKelly:
This takes into account the updated costs
See #7347

### DIFF
--- a/src/clj/game/cards/unhinged.clj
+++ b/src/clj/game/cards/unhinged.clj
@@ -51,7 +51,7 @@
    [game.core.moving :refer [as-agenda flip-faceup forfeit mill move swap-ice swap-installed
                              remove-from-currently-drawing trash trash-cards
                              trash-prevent]]
-   [game.core.payment :refer [can-pay?]]
+   [game.core.payment :refer [can-pay? ->c]]
    [game.core.play-instants :refer [can-play-instant? play-instant]]
    [game.core.prompts :refer [cancellable clear-wait-prompt clear-run-prompts]]
    [game.core.props :refer [add-counter add-icon add-prop remove-icon]]
@@ -153,7 +153,7 @@
                                           card nil)
                                         ;; installable
                                         (if (can-pay? state side (assoc eid :source card :source-type :runner-install) top-card nil
-                                                      [:credit (install-cost state side top-card)])
+                                                      (install-cost state side top-card))
                                           (continue-ability
                                             state side
                                             {:optional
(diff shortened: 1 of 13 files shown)


src/clj/game/cards/unhinged.clj
```diff
@@ -0,0 +1,663 @@
...
+                                                                                   (continue-ability
+                                                                                     state side
+                                                                                     (reveal-fn)
+                                                                                     card nil)
+                                                                                   (effect-completed state side eid))))}
+                                            :no-ability {:msg (msg "decline to play " (:title top-card))}}}
+                                          card nil)
+                                        ;; installable
+                                        (if (can-pay? state side (assoc eid :source card :source-type :runner-install) top-card nil
+                                                      (install-cost state side top-card))
```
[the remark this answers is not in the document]
NBKelly: (install-cost ...) already returns a (->c) function, right?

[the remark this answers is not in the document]
NBKelly: oh no, that's runner-install-cost (which I should be using)

NBKelly:
I updated this to take the standard+ format out of visibility. We can merge this, or scrap the whole thing, I don't mind either way (but if we scrap it, we should remove the cards from netrunner-data).

src/clj/game/core/engine.clj
```diff
@@ -1060,10 +1060,12 @@
   [state _ eid]
   (let [corp-uniques (get-old-uniques state :corp)
         runner-uniques (get-old-uniques state :runner)
-        consoles (->> (get-in @state [:runner :rig :hardware])
+	consoles (->> (get-in @state [:runner :rig :hardware])
```
NoahTheDuke, line 1063: spacing

src/clj/game/core/ice.clj
```diff
@@ -320,6 +320,7 @@
 (defn get-strength
   [card]
   (when (or (ice? card)
+            (:strength card)
```
NoahTheDuke, line 323: how does this apply?

src/clj/web/lobby.clj
```diff
@@ -44,7 +44,7 @@
      :room room
      :save-replay save-replay
      :spectatorhands spectatorhands
-     :singleton (when (some #{format} `("standard" "startup" "casual" "eternal")) singleton)
+     :singleton (when (some #{format} `("standard" "startup" "standard-plus" "casual" "eternal")) singleton)
```
NoahTheDuke, line 47: remove this?

NoahTheDuke: and the rest of the references?

NoahTheDuke (changes requested):

src/clj/game/core/ice.clj
NBKelly: It means we can do forced encounters with non-ice cards, so long as they have a listed strength.
There's an upgrade that pretends to be an ice on access, and it's surprising that this is the *only* change needed to make it work.

src/clj/game/core/engine.clj
```diff
@@ -1064,6 +1064,8 @@
                       (filter console?))
         consoles (when (< 1 (count consoles))
                    (butlast consoles))
+	consoles (when-not (any-effects state :runner :enable-multiple-consoles)
```
NBKelly, line 1067: I *should* make a command to enable this too (a job for later)

src/clj/web/lobby.clj
NBKelly: I *think* that's everything

src/clj/game/core/ice.clj
NoahTheDuke: in that case, can you make it more specific? either by name or by "upgrade and :strength"?

src/clj/web/lobby.clj
NoahTheDuke: thanks homie

src/clj/game/core/ice.clj
NBKelly: I added a comment which explains the usage. We can get the (printed) strength of a card if it's ice, a program, or has a printed strength.

PR #7378, opened 2024-04-07
fix Working Prototype needing a resource

### LINKED ISSUE #7342: Working Prototype should let you click for 6 credits when there are no resources installed
Currently, Working Prototype's second click ability will not fire if there are no installed resources (you can select it, and a prompt to choose a resource appears, but if you just click done nothing happens (no credits, no click taken, no power counters removed)).  This behavior is correct when there are resources on the board (the resource added to the top of stack is mandatory) but is not correct when there are not resources on the board (execute abilities as much as possible).

francescopellegrini:
Fixes #7342

### DIFF
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -9,7 +9,7 @@
                               update-all-agenda-points]]
    [game.core.bad-publicity :refer [bad-publicity-prevent gain-bad-publicity
                                     lose-bad-publicity]]
-   [game.core.board :refer [all-active-installed all-installed get-remotes
+   [game.core.board :refer [all-active-installed all-installed all-installed-runner-type get-remotes
                             installable-servers]]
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/assets.clj
```diff
@@ -3140,16 +3140,21 @@
...
+                (req (wait-for (gain-credits state side 6)
+                               (continue-ability
+                                 state side
+                                 {:prompt "Choose a resource"
+                                  :req (req (seq (all-installed-runner-type state :resource)))
+                                  :choices {:card #(resource? %)}
+                                  :msg (msg (str "add " (:title target) " to the top of the stack"))
+                                  :effect (req (move state :runner target :deck {:front true}))}
+                                 card nil)
+                               (effect-completed state side eid)))}]})
```
NoahTheDuke, line 3157: This isn't necessary but it shouldn't break anything.

PR #7400, opened 2024-04-18
Card disable system experimental overhaul

### LINKED ISSUE #6627: Magnet should disable ALL hosted cards
Noticed while working on #6626

Whenever Magnet is rezzed or the runner installs a card directly on Magnet, the `disable-hosted` function is called.

But there's another way Magnet can host cards: Take a look at Bishop.

When the runner uses Bishop's ability to host it on Magnet, Bishop does not get disabled via `disable-hosted`.

There's no `:event` for hosting, but a `:hosted-gained` key that is used in the `host` function:

```
       (when-let [hosted-gained (:hosted-gained cdef)]
         (hosted-gained state side (make-eid state) (get-card state card) [target]))
```

AFAIK, this is only used by Baba Yaga.

Maybe this should be an `:event`? 🤔

Also, the NISEI Magnet talks about _moving_ the program, while the original FFG Magnet talked about _hosting_ it. Is this a distinction we have to respect?

### LINKED ISSUE #6728: We need to handle disabling cards more gracefully
Just a summary of conversation on slack.

Right now, disabling cards is close to becoming unmaintainable with how it is managed. Here's a rough outline of what we think can be done:

- [ ] Modify all cards that disable cards to use a constant/floating effects system (this will break these cards temporarily)
- [ ] Change installs to check if cards are disabled before executing on-install
- [ ] Same for on-rez
- [ ] Change do-action to check before executing the ability, and change diffs to mark abilities as disabled
- [ ] Change constant effects to need the source card to not be disabled
- [ ] Cards that register keys in the state on rez/install and deregister them on disable (like starlight crusade funding) should instead register then on `:enter-play` (new key) and deregister them on `:leave-play` (key already exists)
- [ ] Check for disabled/enabled cards at the start of each checkpoint, and do what needs to be done?

There may be a few more I forgot. I don't know how we will handle cards that disable other cards, and I pray that it's not possible to make a disable-card triangle.

For reference, the only static (non floating) disable card abilities are Magnet and Malia.

### LINKED ISSUE #6872: Hush doesn't install onto ICE, and can't be moved
Link to bug replay: https://jinteki.net/bug-report/ba68b3b0-820b-44e2-8091-a4ac5eae5d7d?b=0

Description:

Hush doesn't target ICE when it's installed, and its ability can't be used when it's installed to host it / move it on to ICE.
Thought this was a bug with Prognostic Q-Loop at first, but it seems like it happens even on install from hand as well.

### LINKED ISSUE #6909: Hush + Stavka + other ICE's
Link to bug replay: https://jinteki.net/bug-report/354c7f49-4c46-47ad-a292-fd104d582f82?b=0

Description:

Jinteki allows to USE stavka with Hush + there is an ICE which strength is 7 without any fracter and with Hush it's still 7 should be 0

### LINKED ISSUE #6916: Hush didn't counter the printed words on Tollbooth
Link to bug replay: https://jinteki.net/bug-report/6101253f-2b9a-49c5-8730-4b07b67a9a4c?b=0

Description:

Make a run on Tollbooth with Hush installed on Tollbooth using Overclock and see that 3 credits are still removed from Overclock.   Hush should nullify this effect.

### LINKED ISSUE #7247: Slash Disable Command doesn't seem to be disabling the 7 wonder Ice's 3-advancement abilities
Link to bug replay: https://jinteki.net/bug-report/4187e717-289c-448a-ae61-36cc0d39dbf5?b=0

Description:
Read on discord that a Slap Vandal couldn't break a Hushed Hortum. Decided to test and found that a hushed akhet doesn't allow slap vandal+vial to break the ice either.

### LINKED ISSUE #7301: Hush Not Stopping Funhouse Encounter
Just played a game where Hush was installed on Funhouse and failed twice to prevent the encounter trigger.

NBKelly:
Basically an overhaul of how we enable/disable cards. I've applied it in most cases.
It still doesn't correctly axe static-abilities of cards. 
I'm currently trying to think of a way to correctly apply this without having `any-effects?` recursively call itself with every call. 
I believe it may require updating checkpoint to cache a list of disabled cards in state, and the effect's functions to check
1) if a an effect belongs to a static ability
2) if the owner of that effect is in the disabled cards list
and then you see if the list of disabled cards has changed, and if you do, you resolve another checkpoint (yuck, I know, but that resolves chains with malia/light the fire)

A consequence of this should be that when the above is resolved, **hush** should be *at least* 99% correct.

Closes #6872
Closes #7301 ~~(I'll write a test for this case)~~
Closes #6916 ~~(I'll write a test for this case)~~
Closes #6909 ~~(I'll write a test for this case)~~

I'm going to tackle this too:
Closes #7247 (I'll be writing tests for this)

Some older issues this will close:
Closes #6627
Closes #6728

Ice that I'm adding tests for right now. This seems like a large list, but I added a few helpers to make it not as horrible to do:
- [x] Afshar
- [x] Akhet
- [x] Anansi
- [x] Attini
- [x] Blockchain
- [x] Echo
- [x] Envelopment
- [x] Funhouse
- [x] Hive
- [x] Hortum
- [x] Information Overload
- [x] Masvingo
- [x] Mausolus
- [x] Cloud Eater
- [x] NEXT Bronze
- [x] NEXT Gold
- [x] NEXT Sapphire
- [x] NEXT Silver
- [x] Nebula
- [x] Orion
- [x] Wormhole
- [x] Asteroid Belt
- [x] Saisentan
- [x] Salvage
- [x] Seraph
- [x] Searchlight
- [x] Stavka
- [x] Surveyor
- [x] Swarm
- [x] Thoth
- [x] Tithonium
- [x] Tollbooth
- [x] Tour Guide
- [x] Turing
- [x] Tyr
- [x] Tyrant
- [x] Woodcutter
- [x] Wraparound

### DIFF
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -11,7 +11,7 @@
                                     lose-bad-publicity]]
    [game.core.board :refer [all-active-installed all-installed all-installed-runner-type get-remotes
                             installable-servers]]
-   [game.core.card :refer [agenda? asset? can-be-advanced? corp? event? corp-installable-type?
+   [game.core.card :refer [agenda? asset? corp? event? corp-installable-type?
                            faceup? fake-identity? get-advancement-requirement
                            get-agenda-points get-card get-counters get-title get-zone hardware? has-subtype? ice?
                            identity? in-deck? in-discard? in-hand? in-server? installed? is-type?
@@ -28,7 +28,7 @@
    [game.core.engine :refer [pay register-events resolve-ability trigger-event]]
    [game.core.events :refer [first-event? no-event? turn-events event-count]]
    [game.core.expose :refer [expose-prevent]]
-   [game.core.flags :refer [lock-zone prevent-current
+   [game.core.flags :refer [can-really-be-advanced? lock-zone prevent-current
                             prevent-draw
                             register-turn-flag! release-zone]]
    [game.core.gaining :refer [gain gain-clicks gain-credits lose lose-clicks
@@ -684,7 +684,7 @@
                                    :title)]
                   (as-> (all-installed state :corp) it
                     (filter ice? it)
-                    (filter can-be-advanced? it)
+                    (filter #(can-really-be-advanced? state %) it)
                     (remove empty? it)
                     (map :title it)
                     (split-with (partial not= a-token) it)
@@ -700,9 +700,9 @@
                           (continue-ability
                             (let [from-ice target]
                               {:prompt "Choose a piece of ice that can be advanced"
-                               :choices {:card #(and (ice? %)
-                                                     (not (same-card? from-ice %))
-                                                     (can-be-advanced? %))}
+                               :choices {:req (req (and (ice? target)
+                                                        (not (same-card? from-ice target))
+                                                        (can-really-be-advanced? state target)))}
                                :msg (msg "move an advancement token from "
                                          (card-str state from-ice)
                                          " to "
@@ -920,14 +920,14 @@
 
 (defcard "Early Premiere"
   {:derezzed-events [corp-rez-toast]
-   :flags {:corp-phase-12 (req (some #(and (can-be-advanced? %)
+   :flags {:corp-phase-12 (req (some #(and (can-really-be-advanced? state %)
                                            (in-server? %))
                                      (all-installed state :corp)))}
    :abilities [{:cost [(->c :credit 1)]
                 :label "Place 1 advancement token on a card that can be advanced in a server"
-                :choices {:card #(and (can-be-advanced? %)
-                                      (installed? %)
-                                      (in-server? %))} ; should be *in* a server
+                :choices {:req (req (and (can-really-be-advanced? state target)
+                                         (installed? target)
+                                         (in-server? target)))} ; should be *in* a server
                 :once :per-turn
                 :msg (msg "place 1 advancement token on " (card-str state target))
                 :effect (effect (add-prop target :advance-counter 1 {:placed true}))}]})
@@ -1244,8 +1244,8 @@
 (defcard "Hearts and Minds"
   (let [political {:req (req unprotected)
                    :prompt "Choose a card you can advance to place 1 advancement counter on"
-                   :choices {:card #(and (can-be-advanced? %)
-                                         (installed? %))}
+                   :choices {:req (req (and (can-really-be-advanced? state target)
+                                            (installed? target)))}
                    :msg (msg "place 1 advancement counter on " (card-str state target))
                    :effect (effect (add-prop target :advance-counter 1 {:placed true}))}
         ability {:req (req (:corp-phase-12 @state))
@@ -1260,9 +1260,9 @@
                            (continue-ability
                              (let [from-ice target]
                                {:prompt "Choose an installed card you can advance"
-                                :choices {:card #(and (installed? %)
-                                                      (can-be-advanced? %)
-                                                      (not (same-card? from-ice %)))}
+                                :choices {:req (req (and (installed? target)
+                                                         (can-really-be-advanced? state target)
+                                                         (not (same-card? from-ice target))))}
                                 :msg (msg "move 1 hosted advancement counter from "
                                           (card-str state from-ice)
                                           " to "
@@ -2333,7 +2333,7 @@
                                  state side
                                  {:async true
                                   :prompt "Choose a card that can be advanced"
-                                  :choices {:card can-be-advanced?}
+                                  :choices {:req (req (can-really-be-advanced? state target))}
                                   :effect (effect (add-counter target :advancement num-counters {:placed true})
                                                   (system-msg (str "uses " (:title card) " to move " (quantify num-counters "hosted advancement token") " to " (card-str state target)))
                                                   (effect-completed eid))}
@@ -2614,7 +2614,7 @@
                 :prompt "Place 1 advancement token on a card that can be advanced?"
                 :yes-ability {:msg (msg "place 1 advancement token on " (card-str state target))
                               :prompt "Choose a card to place an advancement token on"
-                              :choices {:card can-be-advanced?}
+                              :choices {:req (req (can-really-be-advanced? state target))}
                               :effect (effect (add-prop target :advance-counter 1 {:placed true}))}}}})
 
 (defcard "Spin Doctor"
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -11,7 +11,7 @@
                                     lose-bad-publicity]]
    [game.core.board :refer [all-active-installed all-installed all-installed-runner-type get-remotes
                             installable-servers]]
-   [game.core.card :refer [agenda? asset? corp? event? corp-installable-type?
+   [game.core.card :refer [agenda? asset? can-be-advanced? corp? event? corp-installable-type?
                            faceup? fake-identity? get-advancement-requirement
                            get-agenda-points get-card get-counters get-title get-zone hardware? has-subtype? ice?
                            identity? in-deck? in-discard? in-hand? in-server? installed? is-type?
@@ -28,7 +28,7 @@
    [game.core.engine :refer [pay register-events resolve-ability trigger-event]]
    [game.core.events :refer [first-event? no-event? turn-events event-count]]
    [game.core.expose :refer [expose-prevent]]
-   [game.core.flags :refer [can-really-be-advanced? lock-zone prevent-current
+   [game.core.flags :refer [lock-zone prevent-current
                             prevent-draw
                             register-turn-flag! release-zone]]
    [game.core.gaining :refer [gain gain-clicks gain-credits lose lose-clicks
@@ -684,7 +684,7 @@
                                    :title)]
                   (as-> (all-installed state :corp) it
                     (filter ice? it)
-                    (filter #(can-really-be-advanced? state %) it)
+                    (filter #(can-be-advanced? state %) it)
                     (remove empty? it)
                     (map :title it)
                     (split-with (partial not= a-token) it)
@@ -702,7 +702,7 @@
                               {:prompt "Choose a piece of ice that can be advanced"
                                :choices {:req (req (and (ice? target)
                                                         (not (same-card? from-ice target))
-                                                        (can-really-be-advanced? state target)))}
+                                                        (can-be-advanced? state target)))}
                                :msg (msg "move an advancement token from "
                                          (card-str state from-ice)
                                          " to "
@@ -920,12 +920,12 @@
 
 (defcard "Early Premiere"
   {:derezzed-events [corp-rez-toast]
-   :flags {:corp-phase-12 (req (some #(and (can-really-be-advanced? state %)
+   :flags {:corp-phase-12 (req (some #(and (can-be-advanced? state %)
                                            (in-server? %))
                                      (all-installed state :corp)))}
    :abilities [{:cost [(->c :credit 1)]
                 :label "Place 1 advancement token on a card that can be advanced in a server"
-                :choices {:req (req (and (can-really-be-advanced? state target)
+                :choices {:req (req (and (can-be-advanced? state target)
                                          (installed? target)
                                          (in-server? target)))} ; should be *in* a server
                 :once :per-turn
@@ -1244,7 +1244,7 @@
 (defcard "Hearts and Minds"
   (let [political {:req (req unprotected)
                    :prompt "Choose a card you can advance to place 1 advancement counter on"
-                   :choices {:req (req (and (can-really-be-advanced? state target)
+                   :choices {:req (req (and (can-be-advanced? state target)
                                             (installed? target)))}
                    :msg (msg "place 1 advancement counter on " (card-str state target))
                    :effect (effect (add-prop target :advance-counter 1 {:placed true}))}
@@ -1261,7 +1261,7 @@
                              (let [from-ice target]
                                {:prompt "Choose an installed card you can advance"
                                 :choices {:req (req (and (installed? target)
-                                                         (can-really-be-advanced? state target)
+                                                         (can-be-advanced? state target)
                                                          (not (same-card? from-ice target))))}
                                 :msg (msg "move 1 hosted advancement counter from "
                                           (card-str state from-ice)
@@ -2333,7 +2333,7 @@
                                  state side
                                  {:async true
                                   :prompt "Choose a card that can be advanced"
-                                  :choices {:req (req (can-really-be-advanced? state target))}
+                                  :choices {:req (req (can-be-advanced? state target))}
                                   :effect (effect (add-counter target :advancement num-counters {:placed true})
                                                   (system-msg (str "uses " (:title card) " to move " (quantify num-counters "hosted advancement token") " to " (card-str state target)))
                                                   (effect-completed eid))}
@@ -2614,7 +2614,7 @@
                 :prompt "Place 1 advancement token on a card that can be advanced?"
                 :yes-ability {:msg (msg "place 1 advancement token on " (card-str state target))
                               :prompt "Choose a card to place an advancement token on"
-                              :choices {:req (req (can-really-be-advanced? state target))}
+                              :choices {:req (req (can-be-advanced? state target))}
                               :effect (effect (add-prop target :advance-counter 1 {:placed true}))}}}})
 
 (defcard "Spin Doctor"
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -17,6 +17,7 @@
                            identity? in-deck? in-discard? in-hand? in-server? installed? is-type?
                            operation? program? resource? rezzed? runner? upgrade?]]
    [game.core.card-defs :refer [card-def]]
+   [game.core.checkpoint :refer [fake-checkpoint]]
    [game.core.damage :refer [damage damage-prevent]]
    [game.core.def-helpers :refer [corp-recur corp-rez-toast defcard
                                   reorder-choice trash-on-empty get-x-fn]]
@@ -1594,8 +1595,8 @@
 
 (defcard "Malia Z0L0K4"
   (let [unmark
-        (req (when-let [malia-target (:malia-target card)]
-               (update! state side (assoc (get-card state card) :malia-target nil))
+        (req (when-let [malia-target (get-in card [:special :malia-target])]
+               (update! state side (assoc-in (get-card state card) [:special :malia-target] nil))
                (remove-icon state :runner card (get-card state malia-target)))
              ;; I'm not sure why the side is nil here
              ;; but the old impl had it, so 🤷
@@ -1606,15 +1607,18 @@
                                     (installed? %)
                                     (resource? %)
                                     (not (has-subtype? % "Virtual")))}
-              :effect (effect (add-icon card target "MZ" (faction-label card))
-                              (update! (assoc (get-card state card) :malia-target target)))}
+              :effect (req (add-icon state side card target "MZ" (faction-label card))
+                           (update! state side (assoc-in (get-card state card) [:special :malia-target] target))
+                           (fake-checkpoint state))}
      :leave-play unmark
      :move-zone unmark
      :static-abilities [{:type :disable-card
-                         :req (req (and
-                                     (get-card state card)
-                                     (rezzed? (get-card state card))
-                                     (same-card? target (get-in card [:malia-target]))))
+                         :req (req
+                                (let [malia-target (get-in (get-card state card) [:special :malia-target])]
+                                  (or (same-card? target malia-target)
+                                      (and (same-card? (:host target) malia-target)
+                                           (= (:title malia-target) "DJ Fenris")
+                                           (= (:type target) "Fake-Identity")))))
                          :value true}]}))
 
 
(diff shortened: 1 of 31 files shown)


NBKelly:
ok more updates on this:
* Employee strike works perfectly (I think)
* Same for Direct Access (I think)
* Static abilities are correctly disabled
* Paid abilities are disabled at the diffs level, but still show up on cards
* Hushed ice cannot gain subroutines
* We cache disabled cards at each fake and real checkpoint
* If you're PE and you score an agenda while employee striked, you do no damage (I added a test for this)
* If you're PU and you score a String! while striked, you still trash the top card (the current test passes)

Still todo:
All the errors
Make recurring credits and credit sources work properly
At least one test has to be rewritten for magnet
Tests for hush

src/clj/game/core/effects.clj
```diff
@@ -3,8 +3,13 @@
             [game.core.card :refer [get-card]]
             [game.core.card-defs :refer [card-def]]
             [game.core.eid :refer [make-eid]]
+            [game.core.board :refer [get-all-cards]]
             [game.utils :refer [same-card? to-keyword]]))
 
+(defn is-disabled-reg?
+  [state card]
+  (some #(same-card? % card) (:disabled-card-reg @state)))
+
```
NoahTheDuke, line 12: index by `:cid`:

```clojure
(defn is-disabled-reg?
  [state card]
  (get (:disabled-card-reg @state) (:cid card))
```

src/clj/game/core/effects.clj
```diff
@@ -124,3 +133,19 @@
...
+  "Check if a card is disabled"
+  ([state side target]
+   (any-effects state side :disable-card true? target)))
+
+(defn all-disabled-cards
+  "Gets all cards currently disabled"
+  [state]
+  (let [all-cards (get-all-cards state)
+        disabled-cards (filter #(is-disabled? state nil %) all-cards)]
+    (into [] disabled-cards)))
```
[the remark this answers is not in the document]
NBKelly: Sick. I'll add that in at the end - for now it's super useful to be able to print out the registry between checkpoints for debugging

NBKelly:
Now we can't spend from disabled cards, and we can't activate menu-less abilities (like taking credits from miss bones) when the ability isn't playable (which has somehow never been an issue before?)

NBKelly:
Zona Sul and Beckman have been slightly reworked with a :trash-on-tagged keyword. It only kind of works, because we don't actually call checkpoint that often.
I could put it in fake checkpoint? I'm not sure if that's smart, but it's an option I guess.

NBKelly:
I added an `:enforce-condition` key and engine trigger for making sure earth station and a teia work right with the new disable system.
It appears to work for me, and it passes the old tests.

NoahTheDuke:
soon as the tests are green, i'm good to merge this. this is a phenomenal effort, i'm quite grateful that you've put in the work to figure it out.

NBKelly:
I managed to get saci working correctly, which currently is correct until there are other corp effects that can disable programs in the game (and as long as the runner does not use dr. lovegood to disable their own saci).
All tests pass now.

src/clj/game/core/tags.clj
```diff
@@ -33,7 +33,11 @@
        (trigger-event state :runner :tags-changed {:new-total new-total
                                                    :old-total old-total
                                                    :is-tagged is-tagged?}))
-     changed?)))
+     (if is-tagged?
+       (wait-for
+         (trash-on-tag state nil)
```
NoahTheDuke, line 38: this doesn't work because the calling context isn't async. it will return nil to fake-checkpoint which will loop. better to just call the function with a new eid and not worry about the async context, if this really is the best place for this check.

NBKelly: Fair. I'm not particularly sure where else we can throw it - the idea is this meta-function gets called every fake checkpoint (up to 10 times), so this is guaranteed to enforce the trash condition on those cards.
That did work, but only because we just threw away the result of trash-on-tag (and wait-for) anyway and returned the changed? fn we had before.

Seems like we're still on all tests passing though :100:

NBKelly:
ok, I *think* this is good to go again

src/clj/game/cards/assets.clj
```diff
@@ -1593,28 +1594,33 @@
...
-                 (effect-completed state nil eid))
-               (effect-completed state nil eid)))]
+  (let [unmark
+        (req (when-let [malia-target (get-in card [:special :malia-target])]
+               (update! state side (assoc-in (get-card state card) [:special :malia-target] nil))
+               (remove-icon state :runner card (get-card state malia-target)))
+             ;; I'm not sure why the side is nil here
+             ;; but the old impl had it, so 🤷
+             ;; --nbk, Apr '24
+             (effect-completed state nil eid))]
```
NoahTheDuke, line 1604: for historical reasons, `effect-completed` takes 3 args, but it doesn't use the middle arg, `side`. so as i've been working on the engine, i've been slowly changing that over to be `nil`. i hope to someday completely remove it, but that will also involve removing the `effect` macro, which i don't wanna do yet lol.

src/clj/game/cards/identities.clj
```diff
@@ -1081,9 +1078,9 @@
 (defcard "Jinteki: Potential Unleashed"
   {:events [{:async true
              :event :pre-resolve-damage
-             :req (req (and (-> @state :corp :disable-id not)
-                            (= target :net)
-                            (pos? (last targets))))
+             :req (req (and ;;(-> @state :corp :disable-id not)
```
NoahTheDuke, line 1081: is this the only place that uses `disable-id`? i can't find it elsewhere. we should remove `:disable-id` from all of the cdefs as it's just cruft.

src/clj/game/cards/programs.clj
```diff
@@ -2925,16 +2979,19 @@
...
-  {:hosting {:card #(and (ice? %)
-                         (can-host? %))}
+  {:hosting {:req (req (and (ice? target)
+                            (installed? target)
+                            (can-host? state target)))}
    :events [{:event :rez
-             :req (req (and (same-card? (:card context) (:host card))
-                            (not= (:title (:card context)) "Magnet")))
+             :req (req (same-card? (:card context) (:host card)))
+             ;;(not= (:title (:card context)) "Magnet")))
```
NoahTheDuke, line 2987: Don't need this.

src/clj/game/cards/resources.clj
```diff
@@ -2003,68 +2007,29 @@
...
+                  :cost [(->c :click 1) (->c :trash-can) (->c :brain 1)]
+                  :prompt "Choose a remote server"
+                  :choices (req (cancellable (filter #(can-run-server? state %) remotes)))
+                  :msg (msg "make a run on " target " during which cards in the root of the attacked server lose all abilities")
+                  :makes-run true
+                  :async true
+                  :effect (effect
+                            (register-events card [successful-run-trigger])
+                            (make-run eid target card)
+                            (register-lingering-effect
```
NoahTheDuke, line 2026: these have to be in the other order. we don't remove `:end-of-run` static abilities before initiating new runs, so you can add the lingering-effect before calling it. if there's some ordering nonsense here that needs this, then we gotta change how we write these things, because passing the `eid` to an async function _must_ be in tail position.

src/clj/game/core/checkpoint.clj
```diff
@@ -3,11 +3,13 @@
    [game.core.agendas :refer [update-all-advancement-requirements update-all-agenda-points]]
    [game.core.actions :refer [generate-runnable-zones]]
    [game.core.board :refer [get-remotes clear-empty-remotes]]
+   [game.core.effects :refer [update-disabled-cards]]
    [game.core.ice :refer [update-all-ice update-all-icebreakers]]
    [game.core.hand-size :refer [update-hand-size]]
    [game.core.initializing :refer [update-all-card-labels]]
    [game.core.link :refer [update-link]]
    [game.core.memory :refer [update-mu]]
+   [game.core.say :refer [system-msg]]
```
NoahTheDuke, line 12: Unneeded.

src/clj/game/core/cost_fns.clj
```diff
@@ -44,8 +45,13 @@
   ([state side card] (rez-additional-cost-bonus state side card nil))
   ([state side card pred]
    (let [costs (merge-costs
-                 [(:additional-cost card)
-                  (:additional-cost (card-def card))
+                 ;; idk what the difference is between the additional cost for the card,
+                 ;; and the additional cost for the cdef, but we should probably actually
+                 ;; have a note here explaining it (somewhere in this file)
```
NoahTheDuke, line 50: There's no difference, just two different places they're tracked. I don't remember why/how `:additional-cost` can be attached to a card. A quick search makes me think we don't do that, so you could try removing it and seeing what breaks.

src/clj/game/core/effects.clj
```diff
@@ -124,3 +133,18 @@
...
+  ([state side target]
+   (any-effects state side :disable-card true? target)))
+
+(defn all-disabled-cards
+  "Gets all cards currently disabled"
+  [state]
+  (let [all-cards (get-all-cards state)
+        disabled-cards (filter #(is-disabled? state nil %) all-cards)]
+    (into {} (map (juxt :cid identity)) disabled-cards)))
+(defn update-disabled-cards [state]
```
NoahTheDuke, line 148: Spacing

src/clj/game/core/engine.clj
```diff
@@ -1039,6 +1050,7 @@
                       (filter console?))
         consoles (when (< 1 (count consoles))
                    (butlast consoles))
+
```
NoahTheDuke, line 1053: I know i'm weird about this, but plz no spaces in functions just cuz.

src/clj/game/core/engine.clj
```diff
@@ -1053,6 +1065,46 @@
...
+                  (system-say state (to-keyword (:side card))
+                              (str "trashes " (card-str state card) " for being tagged")))
+                (effect-completed state nil eid))
+      (effect-completed state nil eid))))
+
+(defn- enforce-conditions-impl
+  [state _ eid cards]
+  (if (seq cards)
+    (let [fc (first cards)
+          rest (rest cards)]
```
NoahTheDuke, line 1087: Please don't shadow core function names in variables. you could even not include this here and just write `(enforce-conditions-impl state nil eid (rest cards))` below.

src/clj/game/core/engine.clj
```diff
@@ -1053,6 +1065,46 @@
...
+            (:enforce-conditions fc))
+          fc nil)
+        (enforce-conditions-impl state nil eid rest)))
+    (effect-completed state nil eid)))
+
+(defn enforce-conditions
+  [state _ eid]
+  (wait-for
+    (trash-on-tag state nil (make-eid state eid))
+    (if-let [cards (filter
```
NoahTheDuke, line 1102: This will always be true because even an empty sequence is considered truthy. You gotta call `seq` to coerce to `nil` when empty: `(->> [(get-in ...)] (filter ...) (seq))`.

src/clj/game/core/engine.clj
```diff
@@ -1083,10 +1135,12 @@
   ([state _ eid] (checkpoint state nil eid nil))
   ([state _ eid {:keys [duration durations] :as args}]
    ;; a: Any ability that has met its condition creates the appropriate instances of itself and marks them as pending
-   (let [{:keys [handlers context-maps]} (mark-pending-abilities state eid args)]
+   (let [{:keys [handlers context-maps]} (mark-pending-abilities state eid nil)]
```
NoahTheDuke, line 1138: Why are you not passing `args` to `mark-pending-abilities`?

test/clj/game/cards/ice_test.clj
```diff
@@ -6266,7 +6266,7 @@
       (click-prompt state :runner "0")
       (is (= 1 (count-tags state)) "Trace succeeds with 1 advancement"))))
 
-(deftest searchlight-x-fn
+(deftest ^:kaocha/pending searchlight-x-fn
```
NoahTheDuke, line 6269: Why is this one now broken?

test/clj/game/cards/programs_test.clj
```diff
@@ -3541,6 +3544,479 @@
       (is (= 3 (get-counters (refresh sg) :power)) "Has 3 power counters")
       (is (= 3 (get-strength (refresh sg))) "3 strength"))))
 
+(deftest hush-has-caused-me-to-suffer
```
NoahTheDuke, line 3547: lmaooooooo

test/clj/game/cards/programs_test.clj
```diff
@@ -3618,9 +4094,10 @@
           (rez state :corp magnet)
           (click-card state :corp ika)
           (is (zero?(count (:hosted (refresh enigma)))) "Ika was removed from Enigma")
-          (is (= 1 (count (:hosted (refresh magnet)))) "Ika was hosted onto Magnet")
-          (let [ika (first (:hosted (refresh magnet)))]
-            (is (zero?(count (:abilities ika))) "Ika was blanked"))))))
+          ;; note that when cards are disabled now, the abilities still exist
+          ;; we just don't allow them to be used
+          ;; since card-ability bypasses :playable, that wont work in the test
+          (is (= 1 (count (:hosted (refresh magnet)))) "Ika was hosted onto Magnet")))))
```
NoahTheDuke, line 4100: You could check `:playable` here as the test.

src/clj/game/core/effects.clj
```diff
@@ -124,3 +133,18 @@
...
+  (swap! state assoc-in [:disabled-card-reg] (all-disabled-cards state))
+  (get-in @state [:disabled-card-reg]))
```
NoahTheDuke, line 150:
```suggestion
  (swap! state assoc :disabled-card-reg (all-disabled-cards state))
  (:disabled-card-reg @state))
```

src/clj/game/core/diffs.clj
```diff
@@ -54,6 +55,13 @@
              :source-info {:ability-idx ability-idx}}]
     (if (and (or (active? card)
                  (:autoresolve ability))
+             ;; using the reg instead of any-effect here because this function gets called
+             ;; a LOT, and this is very close to linear time and should (I think)
+             ;; always be correct when this fn is called.
+             ;; If that's ever an issue, we can switch this to:
+             ;; (is-disabled? state side card)
+             ;; --n kelly, apr 2024
+             (not (is-disabled-reg? state card))
```
NoahTheDuke, line 64: Diffs are generated after `fake-checkpoint` is called, so it'll be up-to-date. You can leave the comment if you want, but you don't need to change it to `is-disabled?`.

src/clj/game/cards/resources.clj
```diff
@@ -110,15 +111,8 @@
...
-  (let [trash-effect {:async true
-                      :effect (req (if tagged
-                                     (do (system-msg state :runner (str "trashes " card-name " for being tagged"))
-                                         (trash state :runner eid card {:unpreventable true :cause-card card}))
-                                     (effect-completed state side eid)))}]
-    (-> definition
-        (update :events conj (assoc trash-effect :event :tags-changed))
-        (assoc :reactivate trash-effect))))
+  [definition]
+  (assoc definition :trash-when-tagged true))
```
NoahTheDuke, line 115: This can be entirely removed as it's now unused.

src/clj/game/cards/programs.clj
```diff
@@ -74,6 +73,20 @@
...
+                                (:printed %))
+                          (:subroutines card))
+         new-card (assoc card :subroutines old-subs)
+         new-subs (->> (range total)
+                       (reduce (fn [ice _] (add-sub ice sub (:cid ice) args)) new-card)
+                       :subroutines)
+         new-card (assoc new-card :subroutines new-subs)]
+     (update! state :corp new-card)
+     (trigger-event state side :subroutines-changed (get-card state new-card)))))
+
```
NoahTheDuke, line 89: This is unused and should be removed.

src/clj/game/cards/programs.clj
```diff
@@ -1760,18 +1774,52 @@
...
+                subs (vec (remove #(or (:variable %)
+                                       (not (:printed %)))
+                                  (:subroutines card)))
+                ;; special case for hive, because it unprints subroutines...
+                subs (if (= "Hive" (:title card))
+                       (subroutines-init card (card-def card))
+                       subs)
+                new-card (assoc card :subroutines subs)]
+            (update! state :corp new-card)
+            (trigger-event state side :subroutines-changed (get-card state new-card))))
```
NoahTheDuke, line 1789: If you're worried about loops, only call this when the old card `:subroutines` don't match the new card `:subroutines`. That way, you won't get into loops when the same card triggers over and over again.

src/clj/game/cards/programs.clj
```diff
@@ -2925,16 +2979,19 @@
...
-                            (not= (:title (:card context)) "Magnet")))
+             :req (req (same-card? (:card context) (:host card)))
+             ;;(not= (:title (:card context)) "Magnet")))
              :msg "gain 3 [Credits]"
              :async true
              :effect (effect (gain-credits :runner eid 3))}
             {:event :derez
              :req (req (same-card? (:card context) (:host card)))
+             ;; special cheat for working with magnet
+             :while-disabled true
```
NoahTheDuke, line 2994: Looking at the code again, I don't think this is necessary? It should be possible to update disabled status in `derez` and then this will just "magically" work.

test/clj/game/cards/ice_test.clj
NoahTheDuke: i ran the code and i don't see any reason why this would fail, so i'm confused what's happening. i'd prefer that the underlying issue is fixed and not merely disabling the test.

src/clj/game/cards/programs.clj
NoahTheDuke: Yeah, just add `(update-disabled-cards state)` before the `trigger-event` call in `derez` and you can drop `:while-disabled` (including the other uses in `engine`).

src/clj/game/cards/resources.clj
```diff
@@ -1166,21 +1161,30 @@
              :prompt "Choose an installed card to make its text box blank for the remainder of the turn"
              :once :per-turn
              :interactive (req true)
-             :async true
+             ;;:async true
```
NoahTheDuke, line 1164: Remove please

src/clj/game/cards/resources.clj
```diff
@@ -1166,21 +1161,30 @@
...
+                           state side card
+                           {:type :disable-card
+                            :duration :runner-turn-ends
+                            :req (req (same-card? c target))
+                            :value (req true)})
+                         (fake-checkpoint state)))}]})
+
+                            ;;
+                            ;; (disable-card state side (get-card state target))
+                            ;; (effect-completed state side eid)))}]})
```
NoahTheDuke, line 1187: Remove please.

test/clj/game/cards/ice_test.clj
NBKelly: oh, the test itself calls `disable-card` directly (the old bad one), it's on my list of things to fix

NoahTheDuke: oh i see. yeah, with that changed, it'll be good to go.

src/clj/game/cards/identities.clj
NBKelly: I think it might be have been the only one, so I guess I can remove `:disable-id true` from direct access, employee strike and cerebral static and see if anything breaks.

src/clj/game/cards/resources.clj
NBKelly: oh yeah, I think we did that as a hack the first time around. I'll fix it to just apply an effect and not be as hacky

src/clj/game/core/engine.clj
NBKelly: oops :joy: (it doesn't actually use them for anything ever I think - for some reason I thought it was nil before I tried a million different changes)

src/clj/game/core/cost_fns.clj
NBKelly: ok this one does actually break the `get-rez-cost-test` (`rezzing_test.clj:24,26`), both of them mentioning `:trash-can` as one of the costs.
My base thought is this has something to do with abilities that have a fake trash can cost or something of the like, idk.

src/clj/game/cards/programs.clj
NBKelly: whoops :joy:

src/clj/game/cards/programs.clj
NBKelly: I *think* it's not possible - so long as we don't ever make effects like warden fatuma actually work as written, and constantly apply an ability.
In that case, I think your example would cause the same problem.
I realized I can just refactor this part to card-init the subs though, so I did that.

src/clj/game/cards/programs.clj
NBKelly: neat

src/clj/game/core/diffs.clj
NBKelly: yeah, I think that comment is probably worth keeping for now. No clue what changes we'll make in the next decade :joy:.

test/clj/game/cards/ice_test.clj
NBKelly: ok, refactored it to use hush

src/clj/game/cards/assets.clj
NBKelly: yeah, I really dislike the effect macro. I feel like it makes the code just that little bit harder to maintain (and it also means you can't throw lets into the code).

src/clj/game/core/cost_fns.clj
NoahTheDuke: yeah, that's just a test checking the functionality works. that part of the test can be deleted and this removed.

Issue #6994, opened 2023-08-06
Show game time in lobby


NBKelly:
We're already sorting the lobby based on the time created/time started, it would be cool to put the current game time in the lobby. 

This would help you decide if a game is worth watching.

Wikignometry:
Hi I'm working on this issue right now. Does it make sense to refresh the times every minute or would that cause performance issues? Any advice you have would be appeciated!

NBKelly:
Closed as completed via https://github.com/mtgred/netrunner/pull/7407

PR #7407, opened 2024-04-29
Show game time in lobby

Wikignometry:
This PR addresses issue #6994 around showing the game time in lobby. The idea is that it would help users figure out if a game is worth watching or not. 

As shown in the screenshot below, the expected behavior is that it:

-  displays the number of minutes a game has taken in the bottom right corner of the game row
- does not display a number if the game has not started

Screenshot:
image

### DIFF
--- a/src/cljs/nr/game_row.cljs
+++ b/src/cljs/nr/game_row.cljs
@@ -1,6 +1,9 @@
 (ns nr.game-row
   (:require
    [jinteki.utils :refer [superuser?]]
+   [cljc.java-time.instant :as inst]
+   [cljc.java-time.duration :as duration]
+   [cljc.java-time.temporal.chrono-unit :as chrono]
    [nr.appstate :refer [app-state]]
    [nr.auth :refer [authenticated] :as auth]
    [nr.player-view :refer [player-view]]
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
This is cool! Could you make it hover the start timestamp? Some people (myself included) prefer to view the raw time.

NBKelly:
I'm finding this is actually causing me errors (black screen in the lobby whenever there is an active game). 

I can force this to work if I make `create-new-lobby` in `lobby.clj` to create a new `instant` for the date (it's fallback behavior). Otherwise I think this will work for some people (if their browser doesn't support whatever date function we use? I'm not sure what the conditions are to not send a `:date` through ws), and will cause black screens if a `:date` key is passed through the ws connection.

With a tiny bit of deconstruction, I'm seeing that the `:date` we get from `game` is some sort of native code not compatible with the `instant` type.

check

so:
* type of `inst/now` as called in the lobby:  correct
* date object that my browser is spitting at me (as a string): Mon Jul 01 2024 19:18:02 GMT+1200 (New Zealand Standard Time)
* type of that object: `function Date() { [native code] }`

Issue #6205, opened 2022-01-07
Run Amok prevented by Ansel 1.0


verumvenari:
If the runner makes a run using the Run Amok event during which Ansel 1.0 was rezzed, and allows the final subroutine on Ansel to fire, the runner will be prevented from trashing Ansel with Run Amok's effect. The final subroutine on Ansel prevents the runner from trashing or stealing "for the remainder of this run," where Run Amok happens "when this run ends."

francescopellegrini:
Allegedly fixed by #7415

PR #7415, opened 2024-05-03
Fix Ansel 1.0 by clearing run register before checkpoint

### LINKED ISSUE #7343: Trick Shot Preventing Trash after Ansel Fire on R&D Run
Link to bug replay: https://jinteki.net/bug-report/aae7fb74-e09e-4e09-92b1-c0cde7be1063?b=0

Description:

If runner uses trick shot and third Ansel sub fired on R&D ("Runner cannot steal or trash cards") that text is still applying on the remote run.  Seems like Jnet is treating the remote run as a redirect instead of a new run.

NoahTheDuke:
Closes #7343


NBKelly:
Could also close #6205 -  it might be worth adding a test for that

NoahTheDuke:
I don't have time to write tests for those at the moment, but you're welcome to!

PR #7422, opened 2024-05-18
Highlight active cards in discard

NBKelly:
This is mainly a usability feature (since we can't physically see the cards on a table, a little bit of hinting is nice).

Highlight active cards (agendas, poison, cards like subliminal/crowdfunding/conspiracy breakers) in discard piles when they are otherwise visible.

The idea is that this should help boardstates be comprehensible to both players. And I've seen one too many games ruined by "I access archives for value *whoops* I forgot it was poisoned"

Additionally, I reworked *Heliamphora* to look for a `:poison` key in a card, rather than working on a list of names. This way, we can maintain the card in the long term without needing to rework it around new cards.

demo_000
demo_001

### DIFF
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -422,6 +422,7 @@
 
 (defcard "Breached Dome"
   {:flags {:rd-reveal (req true)}
+   :poison true
    :on-access {:async true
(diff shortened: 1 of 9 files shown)


NoahTheDuke:
I don't like `:poison`, it's not thematic or rules appropriate, but I'm not gonna bikeshed the name right now.

PR #7431, opened 2024-05-24
no longer try to open menus if a card is disabled but only had one ability

### LINKED ISSUE #7425: Dr. Lovegood breaks Always Be Running
Link to bug replay: https://jinteki.net/bug-report/2ba7ed6b-8671-4cc0-96a5-27514df3db86?b=0

Description:
[edit] After using Dr. Lovegood on Always Be Running, Always Be Running's "break subroutine" ability does not trigger when used and the click buttons become unresponsive.

NBKelly:
Fixes #7425

### DIFF
--- a/src/clj/game/cards/resources.clj
+++ b/src/clj/game/cards/resources.clj
@@ -1172,7 +1172,7 @@
                          (register-lingering-effect
                            state side card
                            {:type :disable-card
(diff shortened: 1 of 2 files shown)


NBKelly:
I also forgot to make the timing a real one when I edited this, so that's what the second PR is

PR #7463, opened 2024-06-21
add option to have ghost version of hosted programs appear in the rig

### LINKED ISSUE #6645: Suggestion - Hosted programs have outline in runner's program row
Suggestion: Programs hosted on Ice have an outline in the Runner program row, maybe a kind of ghosty image or something to remind that MU is being used on Corp's side of the board. I very often install a program only to be surprised by the pop-up that MU is over and I have to trash a program, then look up and see a Botulus or something. Would be a nice feature!

NBKelly:
Closes #6645

An option to make the board-state easier to parse for either side (ie counting mu).

option

This will make ghost versions of hosted programs appear in your rig. This includes anything hosted on those programs.

ghosted-2

You can click on the ghost cards and use them in the same way as if they were the regular ones, and you can target them with abilities in the exact same way.

menu

I'm not a massive fan of the doubled-up menus, but I can't figure out a way to fix that behaviour. I might need some help for that one :joy:.

This also works with card stacking. The ghost cards are always on the right.
stacked-ghosts

And it works from both perspectives:
ghosted-3


NBKelly:
ok I figured the menu thing out.

[hosted-trojan-ghosts.webm](https://github.com/mtgred/netrunner/assets/9095245/88326903-8403-4548-a11b-9e50ef2b4b41)

Maybe there's a cleaner way to do it (ie passing the whole card in and figuring it out in open-card-menu), but I'm not sure

NoahTheDuke:
This is very cool

NBKelly:
It now saves like the other options with an entry on the settings page.
I also added a grayscale filter (80%) to the ghost tag as well, which I *think* makes it look nicer and clearer.

src/cljs/nr/gameboard/board.cljs
```diff
@@ -120,21 +122,22 @@
             (and (corp? card)
                  (not (faceup? card))))
         (do (when (= side card-side)
-              (if (= (:cid card) (:source @card-menu))
+              (if (and (= (:cid card) (:source @card-menu))
+                       (= (:ghost card) (:ghost @card-menu)))
                 (close-card-menu)
-                (open-card-menu (:cid card))))
+                (open-card-menu (:cid card) (:ghost card))))
```
NoahTheDuke, line 128: Do you mean `(:ghost @card-menu)`? When is `:ghost` set on a card object?

NoahTheDuke: Below, `ghost-call` sets `:ghost` to be `true`, not a card object.

src/cljs/nr/gameboard/board.cljs
```diff
@@ -1142,7 +1147,22 @@
                     :central-view [discard-view-corp player-side discard]
                     :run (when (= server-type "archives") @run)}]]]))
 
-(defn board-view-runner [player-side identity deck deck-count hand hand-count discard rig run]
+(defn- ghost-card
```
NBKelly, line 1150: `ghost-card` fn sets `ghost` key here for ghosted cards

src/cljs/nr/gameboard/board.cljs
NBKelly: no, this is correct. We're passing whether it's the `:ghost` version of the card or the regular one into the open-card-menu function, so it can open the menu on the correct one (since they have the same cid, since they're the same card).

The `ghost-card` fn below adds that key to the ghosts

PR #7470, opened 2024-06-25
saisentant and bankhar play nice

### LINKED ISSUE #6013: Saisentan doesn't always trigger the additional effect
Boomerang on saisentan, corp names event, runner breaks two with boomerang, fires the remaining sub, hits an event, additional net damage does not fire.

### LINKED ISSUE #6875: Bankhar + Saisentan interaction
This one probably needs input from the rules team.

I use bankar to force the corp to resolve "Do 1 net damage" on saisentan instead of the printed subroutine, so it shouldn't do additional net damage (I think).

Additionally, the same is true with marcus batty (we have confirmation on that one, but it works differently).

I genuinely haven't tested how the cards work together implementation wise, so we probably need to confirm it works right then add a test. (and then I can implement tracker too...)

### LINKED ISSUE #7466: I think Saisentan + Bankhar might be bugged
Link to bug replay: https://jinteki.net/bug-report/e9995224-22f0-400b-a57a-2ec6b5b8c8b5?b=0

Description:

Bankhar target server, Saisentan is rezzed, Corp names event.  Subroutines fire.  Net damage hits event.

Expected behaviour is that the Saisentan ability fires, and a 2nd net damage is done.

Observed behaviour is that the Saisentan ability does not fire.

NBKelly:
I reworked saisentan to be an event that triggers on damage, and specifies that:
* The damage comes from a subroutine
* Is net damage
* Comes from saisentan
Ontop of the card being the right type.

This also theoretically works if effects are later introduces which boost net damage or something. I left a comment to that effect.

I modified the damage event to add a `:cause` key to it's context, which can be passed in the args when calling `(damage`. It's only really touched by saisentan (and set by saisentan and bankhar), but it's there if we ever need it for stuff in the future too.

Closes #6875
Closes #7466

Because this divorces the event from the subroutine, this also Closes #6013 (but I'm not writing a test for this, since it's an illegal game state anyway)

### DIFF
--- a/src/clj/game/cards/ice.clj
+++ b/src/clj/game/cards/ice.clj
@@ -3546,20 +3546,33 @@
   (let [sub {:label "Do 1 net damage"
              :async true
              :msg "do 1 net damage"
-             :effect (req (wait-for (damage state side :net 1 {:card card})
-                                    (let [choice (:card-target card)
-                                          cards async-result
-                                          dmg (some #(when (= (:type %) choice) %) cards)]
-                                      (if (and dmg (not (is-disabled-reg? state card)))
-                                        (do (system-msg state :corp (str "uses " (:title card) " to deal 1 additional net damage"))
-                                            (damage state side eid :net 1 {:card card}))
-                                        (effect-completed state side eid)))))}]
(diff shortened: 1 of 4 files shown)


NoahTheDuke:
ooo this is a good fix

PR #7473, opened 2024-06-26
swapping installed cards no longer fizzles pending events.

### LINKED ISSUE #5591: Rover on Thimblerig does not trigger
Hello.

After installing Rover Algorithm on Thimblerig, it does not trigger when the runner pass Thimblerig and I move Thimblerig to the next position, so that the runner must pass it again (which would then boost its strength).

Is this normal?

Thanks.
Diogene

### LINKED ISSUE #6915: Passing ICE bug that is swapped
When an ICE is passed and is swapped by Inversificator, it does not seem to be treated as passed.

This bug can be identified in the following situations:

- Inversificator went through the Bioroid ICE and swapped it; Haas-Bioroid: Architects of Tomorrow's ability does not trigger.
- Successful in the run with Inversificator, exchanging multiple ices, but the Into the Depth ability counted fewer ices that passed through.

### LINKED ISSUE #7269: Swapping multiple Thimblerigs loses a prompt
Start your turn, have 2 Thimblerigs rezzed.
See prompt for swapping Thimblerig №1. Swap it with Thimblerig №2. (for _reasons_)
No prompt for swapping Thimblerig №2 occurs in this case.

### LINKED ISSUE #7421: Thimblerig + inversificator: the corp's trigger only works if the runner swaps it onto the same server
Per ruling 3 on thimblerig:

> If the runner passes a piece of ice after fully breaking it, and uses [Inversificator](https://netrunnerdb.com/en/card/12048) to swap it to another position, can "when the runner passes this ice" abilities on that ice still resolve?
> Yes.

...but it only seems to give the corp a trigger if the thimblerig is still on the same server after the runner's swap. I believe it should work regardless of position.

NBKelly:
This is our accommodation for the rules reversal NSG did on this/our engine handling of cards being swapped.

I added tests for the outstanding issues I could find.

Closes #7421
Closes #7269
Closes #6915

Also Closes #5591 in a more conclusive way than the interactive tag did.

### DIFF
--- a/src/clj/game/cards/events.clj
+++ b/src/clj/game/cards/events.clj
@@ -2086,6 +2086,7 @@
                :effect (effect (register-events
                                  card
                                  [{:event :pass-ice
+                                   :silent (req true)
                                    :duration :end-of-run
                                    :effect (effect (update! (update-in (get-card state card) [:special :how-deep-are-we] (fnil inc 0))))}])
                                (make-run eid target card))}
--- a/src/clj/game/core/engine.clj
+++ b/src/clj/game/core/engine.clj
@@ -3,12 +3,13 @@
     [clj-uuid :as uuid]
     [clojure.stacktrace :refer [print-stack-trace]]
     [cond-plus.core :refer [cond+]]
-    [game.core.board :refer [clear-empty-remotes get-all-cards all-installed-runner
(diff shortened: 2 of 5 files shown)


NoahTheDuke:
Good catch.

NoahTheDuke:
Maybe `get-card` should search by `cid` instead of location?

PR #7477, opened 2024-06-26
marilyn no longer interrupts regenisis working

### LINKED ISSUE #6895: Marilyn Campaign prevents Regenesis from triggering
When Marilyn Campaign is trashed, it is added to Archives while the prompt to shuffle into R&D is being resolved even though it says "instead of adding to Archives" in the interrupt text. This prevents Regenesis from meeting its trigger conduition.

### LINKED ISSUE #7614: Marilyn recycling considered "adding a card to archives" for Regenesis
Link to bug replay: https://jinteki.net/bug-report/34212528-7176-4ea6-bad0-29ef63e7e848?b=0

Description: Marilyn Campagin recycle disables Regenesis

1. Have Holo Man on Marilyn Campaign, facedown agenda in archives, and Regenesis in HQ 
2. Install Regenesis on Holo Man, overwriting the Marilyn, but shuffle Marilyn back into R&D instead of trashing it (according to card text, this is instead of adding it to Archives)
3. FA the Regenesis 

Desired behavior: should be prompted to add facedown agenda to score area

Observed behavior: no prompt

NBKelly:
Added the ability to selectively remove things from turn-events. It sounds cursed, and it is a bit of a hack, but the alternative is completely rewriting how cards are trashed again to insert another hack, and this one at least has the potential to help us fix interactions later.

Really, every solution I can think of for this is cursed.

Closes #6895 
Closes #7614

### DIFF
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -26,7 +26,7 @@
    [game.core.effects :refer [register-lingering-effect]]
    [game.core.eid :refer [complete-with-result effect-completed is-basic-advance-action? make-eid get-ability-targets]]
    [game.core.engine :refer [pay register-events resolve-ability]]
-   [game.core.events :refer [first-event? no-event? turn-events event-count]]
+   [game.core.events :refer [first-event? no-event? truncate-turn-event turn-events event-count]]
    [game.core.expose :refer [expose-prevent]]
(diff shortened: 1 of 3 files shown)


NoahTheDuke:
hmmmm. this works but you're right that it's incredibly hacky. you've nerd sniped me, so gimme a bit to think about alternatives. maybe it is time to implement fully general interrupts...

PR #7479, opened 2024-06-26
nightmare archive vs sabotage

### LINKED ISSUE #6890: [EDITME] Please give a short description of your bug here
Link to bug replay: https://jinteki.net/bug-report/fd70b971-daf6-44b0-ab40-0a554310f5c7?b=0

Description:

It looks like after the runner (playing Esa Afontiva sp) accesses Nightmare Archives in HQ, zer (sp) chooses to take the core damage and then sabotage, but Nightmare Archives is offered as a choice for me to trash from HQ, when it should be removed from game already, and in the game replay I just trash it, putting it in archives. I only realize the mistake later (after winning a couple of turns later) and then we discuss it briefly in chat

NBKelly:
I just swapped the order of these. It has no other in-game implications and we don't actually do instructions the way the CR does.

Closes #6890

### DIFF
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -1984,9 +1984,9 @@
                :prompt "Choose one"
                :choices ["Suffer 1 core damage" "Add Nightmare Archive to score area"]
                :effect (req (if (= target "Suffer 1 core damage")
-                              (wait-for (damage state :corp :brain 1 {:card card})
-                                        (move state :corp card :rfg)
-                                        (effect-completed state side eid))
+                              (do (move state :corp card :rfg)
+                                  (wait-for (damage state :corp :brain 1 {:card card})
+                                            (effect-completed state side eid)))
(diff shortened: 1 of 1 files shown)


src/clj/game/cards/assets.clj
```diff
@@ -1984,9 +1984,9 @@
                :prompt "Choose one"
                :choices ["Suffer 1 core damage" "Add Nightmare Archive to score area"]
                :effect (req (if (= target "Suffer 1 core damage")
-                              (wait-for (damage state :corp :brain 1 {:card card})
-                                        (move state :corp card :rfg)
-                                        (effect-completed state side eid))
+                              (do (move state :corp card :rfg)
+                                  (wait-for (damage state :corp :brain 1 {:card card})
+                                            (effect-completed state side eid)))
```
NoahTheDuke, line 1989:
```suggestion
                                  (damage state :corp eid :brain 1 {:card card}))
```

PR #7489, opened 2024-06-27
use pronouns in the game where appropriate

### LINKED ISSUE #5280: Pronouns in chat windows.
Hi,

now that we have pronouns, it should use correct pronoun in chat.
As it is now, it say "Zombie keeps their hand"
My correct pronoun is "he/him" as configured in settings :)

NBKelly:
Added a `pronoun` fn in jinteki.utils which checks the users options and selects the correct pronoun if possible, or falls back to `their` otherwise. 

And then I replaced (almost) every occurance of `their` in code with the pronoun function, targeting the correct side when necessary.

I only put mappings in the following, because I don't know what's correct for everything else:
* he/him -> his
* she/her -> her
* it/its -> its

Closes #5280

pronouns
pronouns_000
pronouns_001
pronouns_002

### DIFF
--- a/src/clj/game/cards/agendas.clj
+++ b/src/clj/game/cards/agendas.clj
@@ -1132,7 +1132,7 @@
                                (in-hand? %)
                                (>= 1 (:agendapoints %)))}
          :waiting-prompt true
-         :msg (msg "add " (:title target) " from HQ to " (pronoun state side) " score area")
+         :msg (msg "add " (:title target) " from HQ to [their] score area")
          :effect (req
                    (let [c (move state :corp target :scored)]
                         (card-init state :corp c {:resolve-effect false
@@ -1324,7 +1324,7 @@
 
 (defcard "Midnight-3 Arcology"
   {:on-score {:async true
-              :msg (msg "draw 3 cards and skip " (pronoun state side) " discard step this turn")
+              :msg (msg "draw 3 cards and skip [their] discard step this turn")
               :effect (effect
                         (register-lingering-effect
                           card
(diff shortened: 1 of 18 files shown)


NoahTheDuke:
Oh this is interesting. I realize you've already made the change across the board, but what if it was a template, like `{{pronoun}}` or `{{runner-pronoun}}` or `{{corp-pronoun}}` which is then replaced in `system-msg`? Something like: `(-> msg (str/replace "{{runner-pronoun}}" (pronoun state :runner)) (str/replace "{{corp-pronoun}}" (pronoun state :corp)) (str/replace "{{pronoun}}" (pronoun state side)))`. This would free you from having to use the function everywhere, keeping the strings as 'strings' instead of `msg` or `req` calls.

NBKelly:
> Oh this is fantastic.
> 
> I realize you've already made the change across the board, but what if it was a template, like `{{pronoun}}` or `{{runner-pronoun}}` or `{{corp-pronoun}}` which is then replaced in `system-msg`? Something like: `(-> msg (str/replace "{{runner-pronoun}}" (pronoun state :runner)) (str/replace "{{corp-pronoun}}" (pronoun state :corp)) (str/replace "{{pronoun}}" (pronoun state side)))`. This would free you from having to use the function everywhere, keeping the strings as 'strings' instead of `msg` or `req` calls.

Yeah, wish I thought of that at the start haha.
String replacements are `[their]|[pronoun]` for speaking side, and `[corp-pronoun]|[runner-pronoun]` for target side.

PR #7491, opened 2024-06-28
Install runner cards directly onto hosts

### LINKED ISSUE #4632: Baba Yaga & installed abilities
Baba can't use any abilites of installed breakers on it.

### LINKED ISSUE #6607: Bug Report on Sunny Breaker MU Problems and No-chat glitch.
Several incidents of the chat box disappearing. Log itself is available bug there is no text box and therefore no way to write in the chat in-game. Occasionally but not consistently fixed by opening new windows. 

New MU system is appreciated, but has caused a bug with Sunny's breakers when trying to install them at max MU, it refuses an install even when they would cost none.

### LINKED ISSUE #6693: Cloud Breakers cost MU to install, even if you meet Link requirement
Jnet will prompt you to trash a program to install any cloud breaker (GS Sherman M3, GS Striker M1, GS Shrike M2) even if you have 2 link. The programs are 0 MU at 2 link, even when not installed, so this seems like an error for installing them on Jnet. Once installed, they cost 0 MU correctly, however you can't install them without trashing a program on Jnet.

NBKelly:
It would be neat if this was automated, so I'm having a crack.

Part of this involves fixing the cloud breaker MU problem.
Closes #6693, Closes #6607

This issue is ancient, but I added a test for it anyway: Closes #4632

The following cards are effected:
- [x] Off-Campus Apartment (enforce connection)
- [x] Baba Yaga (enforce non-icebreaker program)
- [x] Scheherazade (enforce program)
- [x] Djinn (enforce non-icebreaker programs, no more than 3 mu)
- [x] Dhegdheer (enforce single program, -1 install cost, no mu)
- [x] Progenitor (enforce virus program, ignore mu, enforce single card)
- [x] Leprechaun (enforce program, enforce no more than 2 cards)
- [x] Omni-Drive (enforce single program one mu or less, no mu)
- [x] NetChip (enforce the netchip conditions)
- [x] Flame-Out (enforce single program)
- [x] Dinosaurus (enforce single non-ai icebreaker program, no mu)

And also:
- [x] Enforce-condition on omni-drive
- [x] Enforce-condition on netchip

Those are the only two cards where it's possible to break the hosting restrictions through game actions (manipulating puffer for omni-drive, trashing another netchip)


NBKelly:
Current Issues:
* everything which supports Y mu needs an enforce-condition feature (primarily netchip and puffer that make this an issue, but technically cloud breakers too)
* You can install things on dheghdeer when you couldn't normally pay for them. I'm not sure it's worth ripping apart the runner-install framework for that one special case.

I removed the manual abilities for hosting, since they wont be needed anymore.

I expect my current commit to have half a thousand errors.

NBKelly:
And it's done. That was a marathon :)

src/clj/game/core/installing.clj
```diff
@@ -420,13 +420,14 @@
...
-  [state side eid card {:keys [facedown] :as args}]
+  [state side eid card {:keys [no-mu facedown] :as args}]
   (let [costs (runner-install-cost state side (assoc card :facedown facedown) (dissoc args :cached-costs))]
     (if-not (runner-can-pay-and-install? state side eid card (assoc args :cached-costs costs))
       (effect-completed state side eid)
       (if (and (program? card)
                (not facedown)
-               (not (sufficient-mu? state card)))
+               (not (sufficient-mu? state card))
+               (not no-mu))
```
NoahTheDuke, line 430: Should this be wrapped in an `or`? EITHER `no-mu` OR not sufficient mu?

src/clj/game/core/memory.clj
```diff
@@ -137,14 +139,29 @@
...
+        abreq (:req ab)
+        abval (:value ab)]
+    (if (and ab (or (nil? abreq) (abreq state :runner nil card nil)))
+      (abval state :runner nil card nil)
+      0)))
+
+(defn expected-mu
+  [state card]
+  (when (program? card)
+    (+ (:memoryunits card) (some-mu-effect? state card))))
```
NoahTheDuke, line 154: This returns `nil` if it's not a program, so you either need to guard against nils everywhere it's used or maybe default to 0 if not a program. If guarding, use `keep` instead of map: `(reduce + 0 (keep expected-mu some-coll))`.

NoahTheDuke:
Killer work. Instead of a top-level `:can-host`, what do you think about this being a `:constant-effect`? I've been trying to move us away from having bespoke top-level entries on card-defs, as they're harder to work with in a consistent way. It shouldn't change much about the implementation except to do a `(find-first #(= :can-host (:type %)) (:effects @state))` or whatever. I don't think it's currently possible to have multiple abilities that can allow hosting, so we don't have to worry about that at the moment 😅

NBKelly:
Yeah, I could make that change.

PR #7494, opened 2024-06-28
fixed broken tests

NBKelly:
just changed `:pass-ice` to `:movement`. That should fix the broken tests in master, which is also the broken tests in #7233


NBKelly:
This one's trivial and it's making the other PR tests fail, so I'm going to merge it myself

PR #7497, opened 2024-06-30
[wip] prevent players playing cards outside of expected times

### LINKED ISSUE #5009: Enforce rules in more cases and add "Fix board state" mode
Since we're more and more approaching full rules coverage in the jnet interface, we could start thinking about removing the last bits and pieces of "freedom" outside the rules. There are cases like installing programs mid-run from hand, which we still allow to make fixing the board state easier once something does go wrong.

We could now think about a fix board state mode (indicated by `(:fix-board-state @state)`), which is indicated clearly in the UI and releases some actions that were previously prevented by the rules (jacking out during a run / installing things from hand / using the +- buttons on clicks, credits, etc.).

### LINKED ISSUE #5776: Suggestion:  Stopping the Game while the Corp is acting
Not actually sure if this is a suggestion per se, but for some cards in the game, the Runner gets no indication that the Corp is doing something (this may be the case for some Runner cards as well but I'm not sure).  For example, while the Corp is making a decision with Climactic Showdown, the runner can keep doing things.  My intuition is that the game should pause while the Corp chooses, then allow the players to proceed, since this is a required effect.

Similarly, I have noticed that using Spin Doctor doesn't alert the Runner that the Corp is deciding on cards.  For example, firing Spin Doctor after discarding allows the runner to carry on their turn while the Corp chooses.

Not huge deals, but just wondering if it could be implemented that when the Corp is making a decision on these cards, the Runner is required to wait.  Thanks!

### LINKED ISSUE #5870: Feature request: Prevent runners from installing programs from grip during runs
Was teaching a friend the game earlier this week and I was surprised to see the platform allowed them to do this. I explained the rule and we undo-turn'ed of course but it seems like it could be a good pedagogy moment if jnet could pop up a warning saying why that's not allowed instead.

### LINKED ISSUE #6708: Overlapping "Select N cards" prompts from spin dr + chastushka makes the second one get stuck
Link to bug replay: https://jinteki.net/bug-report/c1cd4b40-7b09-4260-8820-4df1b7b06a70?b=0

Description:

Lately I've been noticing the spin dr prompt can get preempted if the runner makes another run before you finish choosing cards in archives. Normally this isn't a big deal since you can finish it later but in this case what happened was

- runner runs on spin dr; i begin choosing what to shuffle
- runner plays chastushka, which preempts spin dr prompt with the run dialog
- i press through to successful run
- spin dr prompt comes back, i finish shuffling
- game gets stuck on "current phase: success" without giving me the next prompt to choose cards for chastushka

### LINKED ISSUE #6718: If Esa does 2 actions that require sabotage too quickly for the corp to respond to the first, the first prompt gets eaten
Link to bug replay: https://jinteki.net/bug-report/3c4e180d-81cf-4f63-a35a-fe2fdea66590?b=0

Description:

...and moreover, it seems like any "select N cards" prompt for the entire rest of the game gets messed up too. I was unable to make correct decisions for subsequent sabotage effects, unable to trash a resource with above the law, and I had to resolve spin dr shuffles manually.

I think this is probably related to https://github.com/mtgred/netrunner/issues/6708 but it's worse since there seems to be some global corruption that can't be resolved with undo-click.

### LINKED ISSUE #6930: [BUG] (Minor) Internal server error with unexpected actions from the player
Link to bug replay: https://jinteki.net/bug-report/b78b3df8-4a29-414f-955e-55b28914bb5b?b=0

Description:
Doing Paperclip install from Overclock while the Corp is doing Ob install from firing border control

### LINKED ISSUE #6944: Clicking on Simulchip to install while first Masterwork prompt is open, then trying to undo and redo locked run
Link to bug replay: https://jinteki.net/bug-report/2dba09a0-08c0-48eb-b386-5c41200c58db?b=1

Description:

- Had Masterwork (v37) and Prognostic Q-Loop installed
- Started a run
- Masterwork prompt "Install? yes/no" appeared.
- Clicked on Simulchip before clicking yes on Masterwork prompt, spending a click and underpaying
- /undo-click and put Simulchip back in hand
- Clicked yes in Masterwork prompt and installed Simulchip appropriately
- Q-Loop prompt showing next 2 cards appeared, but would not close when clicking "ok"
- Refreshed the browser window.  Q-loop prompt is gone, but "Current phase: Movement" window is now up and will not respond.
- Fixed by doing /jack-out and resetting to beginning

### LINKED ISSUE #7150: Lock Grip/HQ during card selection prompts
We should be able to find a way to make it so during prompts, you cannot click to install/play cards from these zones.

I feel like there is a technical solution somewhere we can find.

### LINKED ISSUE #7170: HQ cards playable while mandatory draw is pending
Link to bug replay: https://jinteki.net/bug-report/b8b4511b-a423-43e7-9a18-d217be32679f?b=0

Description:

After Start Turn, while Mandatory Draw was pending, I was able to install cards from hand and play an operation.  When Mandatory Draw is pending, HQ cards should not be playable.

### LINKED ISSUE #7198: "bblum uses Stegodon MK IV to shuffle Spin Doctor and 1 unseen card into R&D."
Link to bug replay: https://jinteki.net/bug-report/549a9087-06ef-47b5-945a-698fee7a1adf?b=0

Description:

I believe this is similar to other multiple-overlapping-prompt issues, like #6727 and #6718, but this time it manifested in the game log in this amusing and also surprisingly self-explanatory way, so I thought I'd share.

Additional detail: After resolving the first shuffle-2 ability, I got another prompt to shuffle 2 with spin dr, and _no_ stegodon prompt (which I just resolved manually) -- as if spin's ability had been copied onto the stegodon prompt.

NBKelly:
As per the title, I essentially want to do the foolproofing our engine has needed for quite a while.

This one is going to take a while to finish/figure out completely/test, so I'll chip away at it over the next week or so.

Here's the basic checklist of things that need to happen:
- [x] (front-end only) Disable interacting with cards in hand except when there are no prompts open, or there is a selection prompt open. This is enough to make it so you can't accidentally play programs during runs, or install while your mandatory draw is pending, etc  - but doesn't protect against both players quickly doing stuff at the same time (network race conditions)
- [x] (backend) check if there are any open prompts when receiving the `play` fn, and short-circuit it if there are - this will protect against network race conditions

This will actually tackle a few issues. I'll need to dig deeper to find out exactly which ones, but for a start, it will hit these: 
Closes #7170
Closes #7150
Closes #5870
Closes #5009 (in combination with the install commands)

- [x] Make cards and abilities unplayable (in-engine) when there is an active `:waiting` type prompt
- [x] Make cards and abilities unplayable (frontend) when there is an active `:waiting` type prompt
- [ ] Also gray out the buttons on the front-end when this is the case

Means we can also clear the following, by disallowing scenarios where they can happen:
Closes #6708
Closes #6718

- [ ] Make abilities on cards unplayable if there is any prompt open other a `run` type prompt, unless there's an active prevention window going on

Closes #6944 (by preventing the scenario where it happens)
Closes #6930 (by preventing the scenario where it happens)

- [x] Add a /install command. This will help fix gamestates, and will charge full cost. Can target cards in hand or discard.
- [x] Add a /install-free command. This will help fix gamestates, and will install for free. Can target cards in hand or discard.
- [x] Prevent cards in hand being highlighted while a prompt is open/they're not actually playable.

Additionally, if we do the following:
- [x] Prevent players from using basic actions while there is a prompt open (this already happens by the prompt overlapping the button, but it can't hurt )
- [x] Add a waiting prompt to Spin Doctor

Then this also closes #5776 and closes  #7198 (cancel, or shuffle 0, on spin doc left an open eid)

Then if we also do:
- [ ] Add `(req (not (:run @state)))` to all of the cards in the game with a `:click` cost

we can prevent runners bricking a turn by clicking one of these mid-run

### DIFF
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -2630,7 +2630,6 @@
             :effect (effect (draw eid 2))}
    :abilities [{:label "Shuffle up to 2 cards from Archives into R&D"
                 :cost [(->c :remove-from-game)]
-                :waiting-prompt true
                 :async true
                 :effect (effect (shuffle-into-rd-effect eid card 2))}]})
 
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -2630,6 +2630,7 @@
             :effect (effect (draw eid 2))}
    :abilities [{:label "Shuffle up to 2 cards from Archives into R&D"
                 :cost [(->c :remove-from-game)]
+                :waiting-prompt true
                 :async true
                 :effect (effect (shuffle-into-rd-effect eid card 2))}]})
 
--- a/src/clj/game/core/actions.clj
+++ b/src/clj/game/core/actions.clj
@@ -53,6 +53,10 @@
       (update-click-state state ability)
       (resolve-ability state side eid ability card targets))))
 
+(defn- not-waiting
+  [state side]
+  (not= (get-in @state [side :prompt-state :prompt-type]) :waiting))
+
 (defn play-ability
   "Triggers a card's ability using its zero-based index into the card's card-def :abilities vector."
   ([state side args] (play-ability state side nil args))
@@ -61,12 +65,13 @@
          args (assoc args :card card)
          ability (nth (:abilities card) ability-idx)
          cannot-play (or (:disabled card)
+                         (not (not-waiting state side))
                          (any-effects state side :prevent-paid-ability true? card [ability ability-idx]))]
      (when-not cannot-play
        (do-play-ability state side eid (assoc args :ability-idx ability-idx :ability ability))))))
 
 (defn expend-ability
-  "Called when the player clicks a card from hand."
+  "Called when the player expends a card from hand."
   [state side {:keys [card]}]
   (let [card (get-card state card)
         eid (make-eid state {:source card :source-type :ability})
@@ -76,7 +81,8 @@
 (defn play
   "Called when the player clicks a card from hand."
   [state side {:keys [card] :as context}]
-  (when-let [card (get-card state card)]
+  ;; if there is ANY open prompt, you should not be able to click-play a card from hand
+  (when-let [card (and (not (get-in @state [:side :prompt-state])) (get-card state card))]
     (let [context (assoc context :card card)]
       (case (:type card)
         ("Event" "Operation")
@@ -91,14 +97,16 @@
 (defn click-draw
   "Click to draw."
   [state side _]
-  (play-ability state side {:card (get-in @state [side :basic-action-card])
-                            :ability 1}))
+  (when (not-waiting state side)
+    (play-ability state side {:card (get-in @state [side :basic-action-card])
+                              :ability 1})))
 
 (defn click-credit
   "Click to gain 1 credit."
   [state side _]
-  (play-ability state side {:card (get-in @state [side :basic-action-card])
-                            :ability 0}))
+  (when (not-waiting state side)
+    (play-ability state side {:card (get-in @state [side :basic-action-card])
+                              :ability 0})))
 
 (defn move-card
   "Called when the user drags a card from one zone to another."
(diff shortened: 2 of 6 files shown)


NBKelly:
On second thought, I think this is too much for one PR, so I'm going to tackle this piecemeal over the next few weeks.

PR #7498, opened 2024-07-01
Prevent playing cards while prompts are open or spending clicks when it's not legal

### LINKED ISSUE #5870: Feature request: Prevent runners from installing programs from grip during runs

### LINKED ISSUE #6708: Overlapping "Select N cards" prompts from spin dr + chastushka makes the second one get stuck

### LINKED ISSUE #6718: If Esa does 2 actions that require sabotage too quickly for the corp to respond to the first, the first prompt gets eaten

### LINKED ISSUE #6944: Clicking on Simulchip to install while first Masterwork prompt is open, then trying to undo and redo locked run

### LINKED ISSUE #7150: Lock Grip/HQ during card selection prompts

### LINKED ISSUE #7500: Prevent bricking gamestate during runs by restricting click actions
We should add safeguards to every card with a `:click` cost, to assert that you can't play them while any effects are pending.

NBKelly:
Simple as the title. This is done on both the front-end and the back-end.

While there is any prompt open, you cannot click to play cards from hand.
The card highlight for cards in hand will be disabled while a prompt is open.

I also added a waiting prompt to spin doctor, to stop people initiating runs while the corp resolves the spin doctor effect.

This caught out a few of our unit tests that installed cards from HQ during runs, so I fixed those too.

This will close the following issues:
Closes #7150
Closes #5870
Closes #6708
Closes #6718

~~Additionally, I made it so `:click` costs are not permissible during runs unless they're passed with the `{:allowed-during-run ...}` key.~~

Additionally, I marked all of our actions abilities with the `:action true` key (I think, if I missed one or two then people can make those issues later if they ever discover them) which makes them unplayable during runs. 

This makes it so you can't brick a run by clicking the wrong card by mistake (while also bricking the undo-click function in the process).

Closes #7500
Closes #6944


src/clj/game/core/payment.clj
```diff
@@ -46,6 +49,9 @@
     (->c (:cost/type cur)
          (+ (:cost/amount acc 0) (:cost/amount cur 0))
          (conj {:additional (:cost/additional cur)
+                ;; :allowed-during-run (or (:cost/allowed-during-run acc)
+                ;;                         (:cost/allowed-during-run cur))
+
```
NoahTheDuke, line 54: This should be fixed or removed.

src/cljs/nr/gameboard/board.cljs
```diff
@@ -52,6 +52,10 @@
 
 (defonce button-channel (chan))
 
+(def any-prompt-open?
+  (let [side (:side @game-state)]
+    (get-in @game-state [side :prompt :prompt-state])))
+
```
NoahTheDuke, line 58: This will cause re-renders when game-state changes, so you should figure out how to pass a cursor in for the given side.

NoahTheDuke:
Great idea. Does `{:allowed-during-run ...}` map to anything in the rules? If so, can the name be changed to better align with those rules?

src/clj/game/core/payment.clj
NBKelly: oops

NBKelly:
> Great idea. Does `{:allowed-during-run ...}` map to anything in the rules? If so, can the name be changed to better align with those rules?

Hmm. I think the terminology they use is `action` for a click cost only usable as an action (so those technically aren't paid abilities at all). I'm not really sure of a great way to label that though.

`:action` which defaults to `:true` and needs to be set to false?

Or maybe it's time to *actually* separate abilities from actions in the game logic.

NoahTheDuke:
I think you're right that distinguishing between actions and abilities would be helpful, but I think more generally, we don't track the "location" of costs. So an ability has a `[Click]: Do thing", `[Click]` is a paid ability cost, whereas in "As an additional cost to steal this agenda, the Runner must spend [click] and 2[credit].", `[click] 2[credit]` is an additional cost. Maybe just tracking that would be good here? We don't distinguish between top-level ability maps and nested ability maps (because we treat them as a single continuous ability), which is another hurdle.

Hmmm...

Maybe we need a `:paid-ab-cost`? `:paid-ab-cost` is everything that's before the colon, and then `:cost` would be everything that's not that? We track additional costs on the cost itself, which allows for merging correctly, so maybe we need to also track when something's a paid ability cost? Idk, that seems awkward.

Maybe we should add `:type` to ability maps, so `:abilities` abilities can have `:type :paid-ability` or `:type :conditional` or something. All of the abilities in `:static-abilities` would have it set to `:type :static`? I'm not entirely sure, but something in this vein. Cuz right now, as you've noticed, this sucks lol

NBKelly:
Yeah. I'll put some more thought into it over the coming weeks.

NBKelly:
We do already track if the costs are additional, but unfortunately that doesn't work due to a few cursed cards (Jaguarundi, troll, etc).

So I've put a tiny bit more thought into this, and I think what we could do going forward is put `:type action` in all of the `actions` in the game.

This is enough to do several things:
1) prevent enough information to actions during runs
2) provide enough information to distinguish actions from paid abilities on the front-end
3) give enough of a hook for when we eventually enforce the turn structure (PAW -> action -> repeat)

src/cljs/nr/gameboard/board.cljs
NBKelly: not going to lie, I don't understand any part of that sentence that comes after the comma :joy:

NoahTheDuke:
Yeah, sadly, I think you're right. I hate having to add `:type :action` to one thousand card abilities, but there's really no good way of automating it given how our ability system works with nested abilities 😭

NBKelly:
I don't quite understand how to do that cursor thing, but everything else is ready to go

src/clj/game/core/actions.clj
```diff
@@ -77,7 +79,7 @@
 (defn play
   "Called when the player clicks a card from hand."
   [state side {:keys [card] :as context}]
-  (when-let [card (get-card state card)]
+  (when-let [card (and (nil? (get-in @state [side :prompt-state :prompt-type])) (get-card state card))]
```
NoahTheDuke, line 82:
```suggestion
  (when-let [card (when-not (get-in @state [side :prompt-state :prompt-type]) (get-card state card))]
```

src/clj/game/core/diffs.clj
```diff
@@ -62,6 +64,8 @@
              ;; (is-disabled? state side card)
              ;; --n kelly, apr 2024
              (not (is-disabled-reg? state card))
+             ;; actions cannot be used during runs
+             (or (not (:action ability)) (not (:run @state)))
```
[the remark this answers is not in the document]
NoahTheDuke: Apply De Morgan's so it's clear we're checking an action during a run.

src/cljs/nr/gameboard/board.cljs
```diff
@@ -171,13 +180,15 @@
         ;; Runner clicking on a runner card
         (and (= side :runner)
              (= "Runner" (:side card))
+             (not (any-prompt-open? side))
```
NoahTheDuke, line 183: Isn't this checked by `playable?`? `:playable` is set by the diffs, which are already checking this, right?

test/clj/game/cards/hardware_test.clj
```diff
@@ -2930,11 +2936,15 @@
...
+       (run-continue-until state :movement)
+       (run-jack-out state)
        (play-from-hand state :runner "Lucky Charm")
+       (run-on state "Server 1")
        (card-ability state :corp mb 0)
        (click-prompt state :corp "1 [Credits]")
        (click-prompt state :runner "0 [Credits]")
-       (card-subroutine state :corp (refresh iw) 0)
+       (click-card state :corp iw)
+        (click-prompt state :corp "End the run")
```
NoahTheDuke, line 2947: spacing?

src/cljs/nr/gameboard/board.cljs
NBKelly: So (as far as I can tell) `playable?` is only actually used by the frontend.

*I think* it would be possible to initiate a run, then play a card before the server responds with the state changes that a run is happening, and have the game-state break because of that.

In terms of unit tests, they bypass the `playable` key on the card.

NoahTheDuke: cool

NBKelly:
So, turns out the basic action cards already have `:action true`.
IDK how that got there though :joy: .

I'll check out the other part of that message though.

NBKelly:
Looks like setting the undo-click index to be exclusively on `:action true` didn't break anything.

NoahTheDuke:
let er rip

PR #7499, opened 2024-07-01
install and install-free commands

### LINKED ISSUE #5009: Enforce rules in more cases and add "Fix board state" mode

NBKelly:
If we're going to disallow installing from hand while prompts are up, we should allow a way to fix gamestates if they're broke - so these commands will do.

Closes #5009, in concert with #7498

### DIFF
--- a/src/clj/game/core/commands.clj
+++ b/src/clj/game/core/commands.clj
@@ -209,6 +209,28 @@
     (swap! state dissoc-in [side :selected])
     (effect-completed state side (:eid prompt))))
 
+(defn command-install
+  ([state side] (command-install state side nil))
+  ([state side {:keys [ignore-all-cost] :as args}]
+   (resolve-ability
+     state side
+     (if (= side :corp)
+       {:prompt (str "Choose a card to install" (when ignore-all-cost " (ignoring all costs)"))
+        :choices {:card #(and (corp? %)
+                              (not (installed? %)))}
+        :async true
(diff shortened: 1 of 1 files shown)


src/clj/game/core/commands.clj
```diff
@@ -209,6 +209,28 @@
...
+        :choices {:card #(and (corp? %)
+                              (not (installed? %)))}
+        :async true
+        :effect (req (corp-install state side eid target nil {:ignore-all-cost ignore-all-cost}))}
+       {:prompt (str "Choose a card to install" (when ignore-all-cost " (ignoring all costs)"))
+        :choices {:card #(and (runner? %)
+                              (not (installed? %)))}
+        :async true
+        :effect (req (runner-install state side eid target {:ignore-all-cost ignore-all-cost}))})
+     nil nil)))
```
NoahTheDuke, line 228:
```suggestion
     (make-card {:title (str "/install" (when ignore-all-cost "-free") " command")}) nil)))
```

PR #7511, opened 2024-07-06
Check types match at runtime for time comparison

### LINKED ISSUE #7508: black screen when clicking on play due to game time computation error
There is a situation when clicking on play triggers a black screen. This issue can be reproduced with a recent version (commit 14486e) on Chromium and Firefox as follows:

- With userA create and start a new game
- With userB click on Play, the screen goes black and the following error fires

```
TypeError: startInclusive.until is not a function
    at Duration.between (js-joda.js:514:35)
    at Function.eval [as cljs$core$IFn$_invoke$arity$variadic] (core.cljs:542:30)
    at Object.cljc$java_time$duration$between [as between] (duration.cljs:34:242)
    at Object.nr$game_row$time_since [as time_since] (game_row.cljs:145:15)
    at cmp.nr$game_row$game_time [as reagentRender] (game_row.cljs:152:25)
    at eval (component.cljs:87:28)
    at Object.reagent$impl$component$wrap_render [as wrap_render] (component.cljs:91:31)
    at Object.reagent$impl$component$do_render [as do_render] (component.cljs:117:6)
    at eval (component.cljs:260:64)
    at Object.reagent$ratom$in_context [as in_context] (ratom.cljs:44:6)
```

NBKelly:
Again, I think this issue is really bizarre.

From what I can tell now, it only occurs when running locally via REPL, and will disappear when web/lobby.clj is reloaded. I can only guess that it's either some sort of weird dependency order issue in our project, or it's a genuine repl bug of some sort. I genuinely got gaslit by the issue the first time around.

Also funny enough, we can't run instance? .. time because the native code that the time sometimes is isn't interoperable, so we're stuck with the `type` comparison, which looks uglier :joy: 

We should still merge #7509 because that fix ended up being invalid.

Closes #7508

### DIFF
--- a/src/cljs/nr/game_row.cljs
+++ b/src/cljs/nr/game_row.cljs
@@ -147,8 +147,13 @@
     minutes))
 
 (defn game-time [game]
-(when (:started game)
-  [:div.game-time (str (time-since (:date game)) "m")]))
+  ;; NOTE: while running locally (repl), the :date field ends up being
+  ;; native code, rather than Instant type. I don't understand this,
+  ;; but when running via uberjar (or after reloading web/lobby.clj)
+  ;; it is of the correct type. IDK how to fix the problem, but this
+  ;; is a workable temporary fix - NBKelly, Jul 2024
+  (when (and (:started game) (= (type (:date game)) (type (inst/now))))
+    [:div.game-time (str (time-since (:date game)) "m")]))
 
 (defn players-row [{players :players :as game}]
   (into


acollign:
Thanks for this `workable temporary fix` as you said.

NoahTheDuke:
What are the two types? I don't quite understand.

NBKelly:
> What are the two types? I don't quite understand.

So essentially, the `:date` field coming from a lobby is expected to be of the (java) type `Java.Time.Instant`, and when you either 1) run jinteki via a jar or 2) reload `web/lobby.clj` at any point, it is.

But if you run it via REPL, the `:date` field of every lobby will be of the type `function Date() { [native code] }`, which presents as a formatted date/time string instead. This is actually highly confusing because as you attempt to bugfix the issue, it actually dissapears (because you invariably ending up reloading `web/lobby.clj`)

PR #7524, opened 2024-07-10
Automatically pay trivial costs

### LINKED ISSUE #7505: Automate paying costs when there are no choices
Costs like:
* `trash [x] cards from your hand` when you have exactly x cards in hand
* Forfeit your only agenda automatically when you (choose to) rez archer
* other costs where there is zero possibility to make a choice (ie trash your only installed program with simulchip)

Could potentially be a setting if it's jarring.

I just think it would lead to a smoother game experience.

NBKelly:
This is a little usability/convenience feature I'm working on.

If there's only one possible set of targets for an action (in this case, paying costs), it should be possible to automate it. I've made it optional so that you're not forced into the feature (people who play physical as well can get their mechanical practice in)

Additionally, this adds functionality to update *in-game* settings, which will be important later when I add explicit reveals into the game, or options for the runner to pre-trash programs regardless of MU, or the corp to pre-trash ice.

Closes #7505

* If you need to forfeit exactly as many agendas as there are in your score area, this can be done automatically

* If you need to rfg exactly as many programs as you have installed programs, this can be done automatically (applies to Rosetta 2.0 and nothing else)

* If you need to trash exactly as many cards as you have installed (or exactly as many "other" cards), this can be done automatically

[auto-trash-installed.webm](https://github.com/mtgred/netrunner/assets/9095245/78403603-c571-490c-a40d-2600417da2b1)

* If you need to trash exactly as much hardware/programs/resources/connections as you have installed

* If you need to derez exactly as many harmonics as you have installed (bloop enjoyers will appreciate this)
* If you need to trash exactly as many cards in hand as you have in hand (ie playing moshing at 4 cards in hand)

### DIFF
--- a/src/clj/game/core/costs.clj
+++ b/src/clj/game/core/costs.clj
@@ -254,29 +254,35 @@
   (<= 0 (- (count (get-in @state [side :scored])) (value cost))))
 (defmethod handler :forfeit
   [cost state side eid card]
-  (continue-ability
-    state side
-    {:prompt (str "Choose " (quantify (value cost) "Agenda") " to forfeit")
-     :async true
-     :choices {:max (value cost)
-               :all true
-               :card #(is-scored? state side %)}
-     :effect (req (doseq [agenda targets]
-                    ;; We don't have to await this because we're suppressing the
-                    ;; checkpoint and forfeit makes all of the trashing unpreventable,
-                    ;; meaning that there will be no potential for input. Once
-                    ;; everything is queued, then we perform the actual checkpoint.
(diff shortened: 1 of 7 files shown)


NBKelly:
I think I hit all the cost types that are every likely to see play.

NoahTheDuke:
I worry that this will make the game harder to understand even for seasoned veterans. Paying costs is important and skipping past it will potentially hurt that understanding

NBKelly:
Fair. This is explicitly an opt-in feature, and you can turn it off/on at will during a game.

Issue #7518, opened 2024-07-09
Implement a Web Push Notifications for when someone joins your room.


Larrea-Mon:
Web Push notifications are the kinds of notifications that Discord or many websites send, and show up at the bottom right of your screen on Windows.  I would love to be sent a gentle reminder that someone has joined my jnet room. This would also help cut down on the amount of abandoned rooms in jnet.

NBKelly:
That sounds like a neat idea

acollign:
> That sounds like a neat idea

Agreed!

I gave it a shot and came up with a rather simple solution. The patch below fires a notification whenever someone enters the lobby. This is triggered right before the "ting" sound is played.

```
diff --git a/src/cljs/nr/lobby.cljs b/src/cljs/nr/lobby.cljs
index a13ef4068..3112bd7c1 100644
--- a/src/cljs/nr/lobby.cljs
+++ b/src/cljs/nr/lobby.cljs
@@ -20,6 +20,10 @@
    [reagent.core :as r]
    [taoensso.sente :as sente]))
 
+(defn notify-presence []
+  (.requestPermission js/Notification)
+  (js/Notification. "Someone has entered the lobby" #js {:body "Good luck and have fun!" :icon "/img/icons/jinteki_167.png"}))
+
 (defmethod ws/event-msg-handler :lobby/list [{data :?data}]
   (swap! app-state assoc :games data))
 
@@ -30,6 +34,7 @@
       (ws/ws-send! [:game/resync {:gameid (:gameid data)}]))))
 
 (defmethod ws/event-msg-handler :lobby/notification [{data :?data}]
+  (notify-presence)
   (play-sound data))
 
 (defmethod ws/event-msg-handler :lobby/toast [{{:keys [message type]} :?data}]
```

As you can see, the `notify-presence` function does two things a) ask for permission to notify b) create the notification which will fire if that action was granted. Asking the permission will open the browser pop-up which gives the user the option to either allow or always block. Subsequent calls to the the permission function is a nil operation ensuring that the user is not disturbed once she or he made her or his choice. It is then possible to clear out the permission using the browser UI.

It is important to note that this notification is sent to the person who created the lobby and the person joining it like the "ting" sound is played. For this reason, both the title and the body are generic.

@NBKelly would you say that this approach is worth a PR? I will be happy to contribute one and make the necessary changes.

References:
- https://developer.mozilla.org/en-US/docs/Web/API/Notifications_API/Using_the_Notifications_API
[...]

PR #7525, opened 2024-07-10
 feat: web push notification when someone enters the lobby 

acollign:
This PR implements #7518 which consists of notifying players when someone enters the lobby.

### DIFF
--- a/src/cljs/nr/lobby.cljs
+++ b/src/cljs/nr/lobby.cljs
@@ -20,6 +20,10 @@
    [reagent.core :as r]
    [taoensso.sente :as sente]))
 
+(defn notify-presence []
+  (.requestPermission js/Notification)
+  (js/Notification. "Someone has entered the lobby" #js {:body "Good luck and have fun!" :icon "/img/icons/jinteki_167.png"}))
+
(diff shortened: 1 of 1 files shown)


NoahTheDuke (changes requested):
This isn't correct, I'm sorry to say:

1) `.requestPermission` is a promise, so calling it without a `.then` won't actually wait to get permission.
2) This doesn't handle if the browser doesn't support notifications, so it'll fail and not play any sounds in that case.
3) It doesn't support if the user rejects the permissions request.
4) In Safari and Firefox, `.requestPermissions` can only be called in a user event, like clicking. So you'd need to move the permissions check to something like the `create-game` and/or `join-game` callbacks.
5) This will pop a notification if someone joins your game or if you join someone else's game because we fire `:lobby/notification` in both cases. I don't think it's good UX to get a notification when deliberately clicking on a game. You'll need to figure out a way to only alert the user when someone else joins the game.

acollign:
> This isn't correct, I'm sorry to say:

No problem @NoahTheDuke. Thank you very much for your feedback. I will work on a different solution.

PR #7526, opened 2024-07-10
update take/gain templating on a few cards

NBKelly:
These were the only cards that still said "take credits" on them.

Since we currently use `take` exclusively to refer to illegal game actions (manually fixing the boardstate with cards that host credits), I updated this. They're now consistent with all our other cards.


jayson687:
Always getting weekly 30,000 US $ from this hacker and their bank transfers are safe, fast and same day delivery contact via: d a r k w e b p r o g r a m m e r @ g m a i l . c o m

PR #7533, opened 2024-07-12
negative hand size vs end of turn test

### LINKED ISSUE #5893: Taking brain damage didn't cause runner loss when it should have
Link to bug replay: https://jinteki.net/bug-report/3f632c5f-913a-45e4-ad12-70606b56eb01?b=0

Description:

Took brain damage from a Brainstorm to a total of 8 brain damage. Runner's max hand size was -3, but the game didn't register a victory for the corp.

NBKelly:
Surprised we had no unit tests for this case. Now we do.

Closes #5893


NoahTheDuke:
I have no doubt there are card cases that check this, but having a dedicated unit test is great.

PR #7534, opened 2024-07-12
subroutines-broken is a context map, hippo works right now

### LINKED ISSUE #5879: Hippo gives you the option to use it on a Spooned run
Link to bug replay: https://jinteki.net/bug-report/9ff20644-3cce-4dce-8f99-6e8d47f52fcd?b=0

Description:

During a Spooned run on HQ, there were two Code Gates (an Engram and DNA Tracker). I broke all subs on the Engram and trashed it with Spooned, but after breaking subs on DNA Tracker it gave me the option to use Hippo (which shouldn't happen as far as I know!).

### LINKED ISSUE #6170: Hippo Is able to work Multiple times in 1 turn
"The first time each turn you fully break the outermost piece of ice protecting the attacked server during a run, you may remove this hardware from the game to trash that ice."

Currently the runner can install and use two hippos on the same turn. This has never been allowed, and despite the awkward wording, it still can only trigger once per turn

NBKelly:
Added some of the info we can't wind back the clock to check (if the ice was outermost at the time) into the context map.

Closes #6170
Closes #5879

### DIFF
--- a/src/clj/game/cards/events.clj
+++ b/src/clj/game/cards/events.clj
@@ -84,13 +84,13 @@
              :effect (effect (make-run eid target card))}
    :events [{:event :subroutines-broken
              :async true
-             :req (req (and (has-subtype? (:card target) subtype)
-                            (:all-subs-broken target)
-                            (let [pred #(and (has-subtype? (:card (first %)) subtype)
-                                             (all-subs-broken? (:card (first %))))]
-                              (first-run-event? state side :subroutines-broken pred))))
-             :msg (msg "trash " (card-str state (:card target)))
-             :effect (effect (trash eid (:card target) {:cause-card card}))}]})
+             :req (req (let [pred (every-pred :all-subs-broken #(has-subtype? (:ice %) subtype))]
+                         (and (pred context)
+                              ;; don't need to show messages if the ice is already trashed!
+                              (get-card state (:ice context))
+                              (first-run-event? state side :subroutines-broken #(pred (first %))))))
+             :msg (msg "trash " (card-str state (:ice context)))
+             :effect (effect (trash eid (:ice context) {:cause-card card}))}]})
 
 ;; Card definitions
 
--- a/src/clj/game/cards/events.clj
+++ b/src/clj/game/cards/events.clj
@@ -84,13 +84,13 @@
              :effect (effect (make-run eid target card))}
    :events [{:event :subroutines-broken
              :async true
-             :req (req (and (has-subtype? target subtype)
-                            (all-subs-broken? target)
-                            (let [pred #(and (has-subtype? (first %) subtype)
-                                             (all-subs-broken? (first %)))]
+             :req (req (and (has-subtype? (:card target) subtype)
+                            (:all-subs-broken target)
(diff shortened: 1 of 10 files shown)


src/clj/game/cards/events.clj
```diff
@@ -84,13 +84,13 @@
              :effect (effect (make-run eid target card))}
    :events [{:event :subroutines-broken
              :async true
-             :req (req (and (has-subtype? target subtype)
-                            (all-subs-broken? target)
-                            (let [pred #(and (has-subtype? (first %) subtype)
-                                             (all-subs-broken? (first %)))]
+             :req (req (and (has-subtype? (:card target) subtype)
```
NoahTheDuke, line 87: If it's a context map, please change the name to `context`. It's the only way to keep this mess from getting messier lmao.

src/clj/game/cards/events.clj
```diff
@@ -84,13 +84,13 @@
...
+                            (:all-subs-broken target)
```
NoahTheDuke, line 88: Why not use `(all-subs-broken? (:card context))`?

src/clj/game/core/ice.clj
```diff
@@ -534,6 +534,16 @@
               (map :label (sort-by :index broken-subs)))
         "\")")))
 
+(defn break-subs-event-context
+  [state _ ice broken-subs breaker]
+  {;; pass a few more pieces of information relevant to other cards
+   :outermost (when-let [server-ice (:ices (card->server state ice))] (same-card? ice (last server-ice)))
+   :during-run (if (:run @state) true nil)
```
NoahTheDuke, line 541:
```suggestion
   :during-run (some? (:run @state))
```

src/clj/game/cards/events.clj
NoahTheDuke: nvm, read the context builder, that's a good idea.

src/clj/game/cards/events.clj
NoahTheDuke: same goes for all of the other changes too.

src/clj/game/core/ice.clj
```diff
@@ -534,6 +534,16 @@
...
+   :all-subs-broken (all-subs-broken? ice)
+   :broken-subs broken-subs
+   :breaker (:cid breaker)
+   :card ice})
```
NoahTheDuke, line 545: This could be `:ice` to make it really clear what the context is. `:card` is fairly generic so that's what I've defaulted to in the past, but I think checking `(:ice context)` and `(:breaker context)` are more clear.

src/clj/game/core/ice.clj
```diff
@@ -534,6 +534,16 @@
               (map :label (sort-by :index broken-subs)))
         "\")")))
 
+(defn break-subs-event-context
+  [state _ ice broken-subs breaker]
```
NoahTheDuke, line 538: Given that this isn't an async function, you can just drop the `side` arg. No need to carry that baggage around if you're not going to write `(wait-for (break-subs-event-context ...)`

NBKelly:
ok I think I hit all of those points

NoahTheDuke:
Hell yeah, thank you for this.

Issue #5171, opened 2020-06-24
Navi Mumbai Grid Did Not Allow No One Home To Fire


rongydoge:
I asked #general and since they obviously know everything they said No One Home isn't a paid ability, so I was CHEATED, CHEATED I SAY out of those two doofs which DEFINITELY would've turned the game around considering I died to accessing a Fetal AI. 

Anyway I think Navi Mumbai Grid is a little overzealous to what it considers a paid ability on non-icebreaker cards. 

TYTYTYTYTYTYTYTYTY!

rongydoge:
Sorry I should actually submit a bug report. 

Me: 419
Villain: Tennin

Me: Boomerang an unrezzed chiyashi, run the server. 
Them: rez navi mumbai, laugh. 

Me: try to avoid the first sub of 2 net dmg with NoH. 
Navi: GTFO

Me: sad

exeunt

NoahTheDuke:
Lmao thanks for this report, it’s a pleasure to read. It doesn’t work because we have made all prevention abilities paid abilities to make the user experience consistent. Looks like we’ll have to revise that now that we’re enforcing Navi Mumbai Grid. Thank you!

PR #7536, opened 2024-07-12
prana once per damage instance!

### LINKED ISSUE #5765: Prana condenser allows preventing all damage
Runner steals a Sting against PE. Sting deals 2 damage and PE deals 1. Corp should only be able to prevent one of the two damage from Sting, but prompt appears to prevent both.

### LINKED ISSUE #5861: Players are not always asked which order to resolve their triggers in
Link to bug replay: https://jinteki.net/bug-report/9bae9250-767e-4b56-a3ae-5b2baeb1de6e?b=0

Description:

jinteki.net seems inconsistent in whether it asks which order to resolve triggers in. In this game, on the last click of the game, the Runner stole two copies of Sting!; that creates three triggers (the Runner's Pantograph trigger, the Corp's Sting! trigger, and the Corp's Jinteki: Personal Evolution trigger). Neither steal asked the Corp which order to resolve the triggers in (even though it normally does ask even when both triggers do 1 damage, and it could matter if, e.g., the Runner has I've Had Worse in hand). Expected behaviour is for the Pantograph trigger to resolve first, and then the Corp to choose which of the Corp's triggers to resolve first.

I'm not sure whether the Pantograph trigger caused this bug to happen, or whether it was irrelevant and the cause is something else (e.g. a buggy attempt to not ask when there are multiple identical triggers); as a consequence, I've had to leave the title fairly vague.

NBKelly:
Instead of just swapping the type of damage into the state damage prevent thing, we swap in a map with `{:type type :eid eid}`.

I updated all *checks notes* one other things that used this area of the engine to match (we really should enforce this a little more :joy:)

Closes #5765

Also made both halves of PE interactive, instead of just the on-score

Closes #5861 (the player wanted pe and sting to interact on steal)

### DIFF
--- a/src/clj/game/cards/identities.clj
+++ b/src/clj/game/cards/identities.clj
@@ -1072,10 +1072,9 @@
   (let [ability {:async true
                  :req (req (not (:winner @state)))
                  :msg "do 1 net damage"
+                 :interactive (req true)
                  :effect (effect (damage eid :net 1 {:card card}))}]
-    {:events [(assoc ability
-                     :event :agenda-scored
-                     :interactive (req true))
+    {:events [(assoc ability :event :agenda-scored)
               (assoc ability :event :agenda-stolen)]}))
 
 (defcard "Jinteki: Potential Unleashed"
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -2130,8 +2130,11 @@
    :abilities [{:label "Prevent 1 net damage to place power counter on Prāna Condenser"
                 :msg "prevent 1 net damage, place 1 power counter, and gain 3 [Credits]"
                 :async true
-                :req (req true)
+                :req (req (and (= (get-in @state [:prevent :current :type]) :net)
+                               (not= (get-in card [:special :last-prevented-eid])
+                                     (get-in @state [:prevent :current :eid :eid]))))
                 :effect (req (add-counter state side card :power 1)
+                             (update! state side (assoc-in (get-card state card) [:special :last-prevented-eid] (get-in @state [:prevent :current :eid :eid])))
                              (gain-credits state :corp eid 3)
                              (damage-prevent state :corp :net 1))}
                {:msg (msg "deal " (get-counters card :power) " net damage")
--- a/src/clj/game/cards/resources.clj
+++ b/src/clj/game/cards/resources.clj
@@ -2225,7 +2225,7 @@
                                             run
                                             (= :corp (:active-player @state))
                                             (#{:psi :trace} (:source-type eid))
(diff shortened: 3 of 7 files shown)


src/clj/game/cards/identities.clj
```diff
@@ -1072,10 +1072,9 @@
   (let [ability {:async true
                  :req (req (not (:winner @state)))
                  :msg "do 1 net damage"
+                 :interactive (req true)
                  :effect (effect (damage eid :net 1 {:card card}))}]
-    {:events [(assoc ability
-                     :event :agenda-scored
-                     :interactive (req true))
```
NoahTheDuke, line 1078: Why this change?

NoahTheDuke: jk missed that you just moved it a little

NBKelly:
Fair enough. I think this one is trivial enough that we can keep them consistent (most of them literally aren't being used), and then (sooner or later) I do have plans to actually set up an interface for prevention effects.

NoahTheDuke:
i guess for the things that we already have makeshift "prevent" systems, keeping them in sync isn't so bad. but like, netrunner itself allows for [arbitrary prevention](https://netrunnerdb.com/find/?q=x%3A%22prevent+%22&sort=name&view=text&_locale=en) (Airblade: prevent a "when encountered", alexa belsky: prevent shuffling a card in hq into r&d lol, credit crash: prevent runner trashing a card, dummy box: prevent corp trashing a card, etc etc etc).

we handle almost all of these directly in the cards but they could be done in an automated way if we had the prevention system set up to be modular. just hard to figure out how to design it with what we have.

NBKelly:
Yeah. My idea for prevention is essentially going to mimic how I did traces in ONR:
* An event fires (let's say, damage)
* The (let's say, for now) `prevention manager` picks out all the interrupts which can prevent/interact with that damage on the active side, and lists them as a set of options on a prompt so long as they're payable
* You can click on whatever interrupts you want and resolve them in any order, or click done when you're done
* Same thing happens with inactive player
* Proceed with damage as normal

Which means we can probably actually define the abilities in terms of:
```Clojure
:interrupts [{:type :prevent-meat
              :label "Prevent 3 meat damage"
              :effect ...
              :msg ...
              :req (req whatever)
              :cost [(->c :trash-can 1)]]
```
which would be distinct from the paid abilities on a card (and would also let us dodge issues like this: #5171)

NoahTheDuke:
That's killer. Very similar to how I've envisioned, but not so fleshed out.

NBKelly:
Thinking about this a tiny bit more, I think this is also something I want to wrap through my choose-one helper once that gets fully fleshed out - I think I'll port that piece of the puzzle over in a few months.

PR #7539, opened 2024-07-13
pending test on attini + attini/prana test + RSVP

### LINKED ISSUE #1074: RSVP + Gagarin
There doesn't seem to be an option for me to make RSVP (NBN ice) fire its subroutine. Normally, I can just tell the player in chat, but when I run Gagarin, that means that I can't let the run be successful without the runner auto-accessing a card (even though with the RSVP sub, they shouuldn't be able to). Am I maybe doing something wrong, or is that just still an unsupported feature?

### LINKED ISSUE #7277: Attini + Prana Condenser
Firing Attini subroutines don't give me the prompt to prevent the damage with prana if the runner choose to not pay the 2

https://netrunnerdb.com/en/card/26107
https://netrunnerdb.com/en/card/34042

NBKelly:
Attini now actually has a cut-in which makes the runner unable to pay for credits during subroutine resolution. The pending test we had in that regard does just work now.

This means one less use of the flags system as well, so we're slowly getting there with that - and I updated guru davinder to match (and also take into account core/meat damage).

I also added in a test for prana threat + corp prevent with prana condenser - Closes #7277

And just to follow this system to it's logical conclusion, I implemented RSVP as well and added a unit test for that, which I guess would close  #1074, but it's already closed!

### DIFF
--- a/src/clj/game/cards/resources.clj
+++ b/src/clj/game/cards/resources.clj
@@ -1537,11 +1537,11 @@
 
 (defcard "Guru Davinder"
   {:static-abilities [{:type :cannot-pay-net
-                       :value (constantly true)}
+                       :value true}
                       {:type :cannot-pay-meat
-                       :value (constantly true)}
+                       :value true}
                       {:type :cannot-pay-brain
-                       :value (constantly true)}]
+                       :value true}]
    :events [{:event :pre-damage
              :req (req (and (#{:meat :net} (:type context))
                             (pos? (:amount context))))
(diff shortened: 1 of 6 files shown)


src/clj/game/cards/resources.clj
```diff
@@ -1536,7 +1536,12 @@
                             (continue-ability state side ab card targets)))}]})
 
 (defcard "Guru Davinder"
-  {:flags {:cannot-pay-net true}
+  {:static-abilities [{:type :cannot-pay-net
+                       :value (constantly true)}
```
NoahTheDuke, line 1540: it has to be a function? why not a literal true?

NBKelly: honestly, I don't know why I did that :joy:

PR #7543, opened 2024-07-13
dont display empty prompt buttons

### LINKED ISSUE #7424: Overseer Matrix' UI shows additionalt blank options
When multiple Overseer Matrix trigger for the second time on the same run the UI shows 2 blank buttons each above the options to select the Overseer Matrix trigger. It looks similar to the weird UI bugs that showed up in the past with vampironassa and bacterial programming.

NBKelly:
This isn't a great fix, because I still have no clue how empty options end up in the prompt state, but at the very least this is a stopgap.

Honestly, if somebody else can spy a reason for there to be some extra entries in the prompt state, they're better at inspecting code than I am :joy: 

It adds a test, but that test was just to determine that the issue was actually on the frontend and not the backend (still, doesn't hurt).

Closes #7424

### DIFF
--- a/src/cljs/nr/gameboard/board.cljs
+++ b/src/cljs/nr/gameboard/board.cljs
@@ -1652,7 +1652,7 @@
        ;; otherwise choice of all present choices
        :else
        (doall (for [{:keys [idx uuid value]} choices]
-                (when (not= value "Hide")
+                (when (and (seq value) (not= value "Hide"))
                   [:button {:key idx
                             :on-click #(do (send-command "choice" {:choice {:uuid uuid}})
                                            (card-highlight-mouse-out % value button-channel))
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
yeah something fucky is happening here. i suspect it's in `trigger-queued-event-player`:

```
          choices-map (map #(vector (or (:ability-name (:ability (:handler %)))
                                        (card-for-ability state (:handler %)))
                                    %) cards-with-titles)
          choices-titles (map first choices-map)
```

this is where a `nil` value will leak in. i'm not sure why, but maybe we should use `keep`? that'll remove any `cards-with-titles` that aren't allowed (idk why `card-for-ability` fails only here and not earlier).

NoahTheDuke:
jk i figured it out, it's a real dumb issue lmao, it's tied to #7564.

NoahTheDuke:
Covered by #7565

PR #7544, opened 2024-07-13
added a system to condense the install messages a bit

### LINKED ISSUE #3542: Chat log sequencing for Team Sponsorship installs is not clear.
I'm frequently confused by the sequencing of the chat log for team sponsorship reinstalls/installs, such that it's hard to work out immediately which card is installed into which server. We've just chatted about it on slack and others feel the same way. Could this be looked at?

### LINKED ISSUE #7545: Allow passing counters in as part of our install system
It's kind of awkward to write cards like Arella, which install something and then place counters on it.
We could probably condense this a bit, to something like:
`{:counter {:advancement 1}}`
which should be straightforward, cover (I think) all the usual cases, and (I think) let us condense our chat log a little.

And by what I mean, Arella currently works like this:
```Clojure
:effect (req (wait-for (corp-install state :corp target nil {:ignore-all-cost true :display-message false})
                                (let [inst-target (find-latest state target)]
                                  (add-prop state :corp inst-target :advance-counter 1 {:placed true})
                                  (system-msg state :corp
                                              (str "uses " (:title card) " to install and place a counter on "
                                                   (card-str state inst-target) ", ignoring all costs"))
                                  (effect-completed state side eid))))}]
```
which gives us a message like this:
`uses Arella Salvatore to install and place a counter on ice protecting Server 1 at position 0, ignoring all costs.`
Which looks ok, but only because we had to deliberately craft the install message - and the templating wont be consistent accross similar cards (ie saraswati, mushin-no-shin, etc)

Where it could probably be written like this:
```Clojure
:effect (req (corp-install state :corp eid target nil {:ignore-all-cost true 
                                                       :display-message false
                                                       :counters {:advancement 1}
                                                       :msg-keys {:install-source card
                                                                  :display-origin true}}}))

```

Which would (ideally) give us a message like this for free:
`uses Arella Salvatore to install a piece of ice from HQ protecting Server 1, ignoring all costs, and place 1 Advancement counter on it`

### LINKED ISSUE #7549: Architect Deployment Test - and maybe others - install from the deck chat log issues.
With the latest update, installing from the deck with Architect Deployment Test shows this interesting message.
image
This behavior seems expected (until today, architect deployment test has been on a "the runner does not know which card is being installed" setting). However I asked Cephalopod Wizard about this and it seems like it might be getting an update that reveals the position of the card installed. 
**
image
**

Now I'm not saying that you should beat the rules team to the punch with implementing this but maybe in a few weeks you can circle back to this and see if it has been codified into the rules and implement it then.

NBKelly:
Teamspo is the test case here.

I'm going to run through all the corp installs in the game and clarify this wherever it seems appropriate.

We can do the following:
* Say the source of the card (archives, hq, etc)
* Say the source of the install ("uses team sponsorship to ...")
* Where relevant (ie Architect deployment test), say the index the card came from
* When the card is known, say the name of the card (ie when the corp is instructed to rez the card, or it's a faceup card in archives, or we pass a key in because the card has already been revealed)

Previous case would be:
`nealpro uses Team Sponsorship to install an unseen card from Archives.`
`nealpro installs a card in Server 3 (new remote).`
Closes #3542

A few example cases:
  * Mumbad City Hall mch
  * Team Sponsorship teamspo-cleanup
  * Epiphany Analytica epiphany_000
  * ADT adt
  * Psychokinesis psycho

While I was at it, I also made it so we could condense abilities which install a card and place an advancement counter on it - so these now are easier to program, and also have messages which are actually useful.

A few example cases:

mushin
mitosis
kakurenbo

Closes #7545

Closes #7549

### DIFF
--- a/src/clj/game/core/installing.clj
+++ b/src/clj/game/core/installing.clj
@@ -126,6 +126,9 @@
 (defn- corp-install-message
   "Prints the correct install message."
   [state side card server install-state cost-str args]
+  (let [display-origin (get-in args [:msg-keys :display-origin])
+        source (get-in args [:msg-keys :install-source :title])
+        origin-index (get-in args [:msg-keys :index])]
   (when (:display-message args true)
     (let [card-name (if (or (#{:rezzed-no-cost :face-up} install-state)
                             (rezzed? card))
@@ -133,12 +136,24 @@
                       (if (ice? card) "ice" "a card"))
           server-name (if (= server "New remote")
                         (str (remote-num->name (dec (:rid @state))) " (new remote)")
-                        server)]
-      (system-msg state side (str (build-spend-msg cost-str "install") card-name
+                        server)
+          origin (if display-origin
+                   (str " from "
+                        (when origin-index (str " position " (inc origin-index) " of "))
+                        (name-zone :corp (:zone card)))
+                   "")
+          install (str (when source (str "use " source " to ")) "install")]
+      (system-msg state side (str (build-spend-msg cost-str install) card-name origin
                                   (if (ice? card) " protecting " " in ") server-name))
       (when (and (= :face-up install-state)
                  (agenda? card))
-        (implementation-msg state card)))))
+        (implementation-msg state card))))))
+
+;; Unused in the corp install system, necessary for card definitions
+(defn corp-install-msg
+  "Gets a message describing where a card has been installed from. Example: Interns."
+  [card]
+  (str "install " (if (:seen card) (:title card) "an unseen card") " from " (name-zone :corp (:zone card))))
 
 (defn corp-install-list
   "Returns a list of targets for where a given card can be installed."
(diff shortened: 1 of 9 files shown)


src/clj/game/core/installing.clj
```diff
@@ -104,6 +105,14 @@
                     (effect-completed state :corp eid)))}
     nil nil))
 
+(defn- corp-install-place-counters
+  "Places counters on a card via installation"
+  [state side eid target-card {:keys [counters] :as args}]
+  ;; for now, only advancement counters are checked
+  (if-not (:advance-counter counters)
```
NoahTheDuke, line 112: Style nit: avoid if-not when the branches are equally "normal".

src/clj/game/core/installing.clj
```diff
@@ -123,22 +132,51 @@
...
-                        (str (remote-num->name (dec (:rid @state))) " (new remote)")
-                        server)]
-      (system-msg state side (str (build-spend-msg cost-str "install") card-name
-                                  (if (ice? card) " protecting " " in ") server-name))
-      (when (and (= :face-up install-state)
-                 (agenda? card))
-        (implementation-msg state card)))))
+  (let [display-origin (get-in args [:msg-keys :display-origin])
+        source (get-in args [:msg-keys :install-source :title])
+        origin-index (get-in args [:msg-keys :index])]
```
NoahTheDuke, line 147: Style nit: can you condense this to a destructure?

NoahTheDuke (approved):

NoahTheDuke:
These are great changes, thank you.

PR #7560, opened 2024-07-16
make unbeatable and beat-trace buttons

### LINKED ISSUE #5602: Add "Make unbeatable" and "beat trace" buttons to traces
Oftentimes traces come down to the same two options: pay nothing or pay enough to beat the trace / make it unbeatable. Might as well add buttons for those options (pay nothing is already the default)

NBKelly:
Adds these buttons into the trace prompt, where relevant:

make-unbeatable

beat-trace

The buttons only show up if it's possible to guarantee a victory, and have a minimum value of 0 (will show up if zero is a winning bid still).

The button doesn't confirm, it just adjusts the interior selector (I figured this would be less jarring for new players).

I'm not going to lie, it took an embarrassing amount of time to figure out how to hook the select box up to be controllable. I'm not sure I chose the best key (or even atom) to use, so if you suggest something better I'll change it.

Closes #5602

### DIFF
--- a/src/cljs/nr/gameboard/board.cljs
+++ b/src/cljs/nr/gameboard/board.cljs
@@ -1526,7 +1526,7 @@
     [runner-run-div run encounters]))
 
 (defn trace-div
-  [{:keys [base strength player link bonus choices corp-credits runner-credits]}]
+  [{:keys [base strength player link bonus choices corp-credits runner-credits unbeatable beat-trace] :as prompt-state}]
   [:div
    (when base
      ;; This is the initial trace prompt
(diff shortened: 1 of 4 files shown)


src/cljs/nr/gameboard/board.cljs
```diff
@@ -1554,13 +1554,23 @@
           [:span link " " [:span {:class "anr-icon link"}] (str " + " )]
           (let [strength (if bonus (+ base bonus) base)]
             [:span (str strength " + ")]))))
-    [:select#credit {:onKeyUp #(when (= "Enter" (.-key %))
+    [:select#credit {:value (or (get-in @app-state [:select :credits]) 0)
```
NoahTheDuke, line 1557: This will cause a re-render when `app-state` changes. If you need a local bit of state, you can use a form-2 or `r/let` over an ratom.

PR #7561, opened 2024-07-16
Media blitz just works + scored keys in tests

### LINKED ISSUE #4933: Media Blitz becomes targeted agenda
If you play Media Blitz and target Jumon, then drag Media Blitz out of the current area and into your grip/Archives, it BECOMES the agenda it was targeting:
image

As you see, I have 5 copies of Jumon in play.  2 of them, including one of the ones the Runner "scored" is actually a copy of Media Blitz (the Jumon in HQ was also a Media Blitz I played for further testing).

### LINKED ISSUE #5028: Media Blitz + Ad Blitz + runner scored Rebranding team not working
Ad Blitz + Rebranding Team is working correctly.
When Media Blitz is copying a runner scored Rebranding Team the Ad Blitz fails to allow assets with printed Advertisement to be chosen.

-- Also this is my first bug report. Just wanted to say, you guys rock. Jinteki.net is such a work of art to be up and running.

NBKelly:
Media blitz was fixed in 98e5f3a, but I added a few new unit tests to close some old issues.

Closes #4933
Closes #5028

I also added a feature that I wanted ages ago and forgot about, which is setting `:scored` keys for the runner and corp in the test framework. I just think it makes a whole lot of tests easier and cleaner to implement.

### DIFF
--- a/src/clj/game/cards/operations.clj
+++ b/src/clj/game/cards/operations.clj
@@ -1638,17 +1638,13 @@
 (defcard "Media Blitz"
   {:on-play
    {:async true
-    :effect
-    (effect
-      (continue-ability
-        {:prompt "Choose an agenda in the runner's score area"
-         :choices {:card #(and (agenda? %)
-                               (is-scored? state :runner %))}
-         :change-in-game-state (req (seq (:scored runner)))
-         :effect (req (update! state side (assoc card :title (:title target) :abilities (ability-init (card-def target))))
-                      (card-init state side (get-card state card) {:resolve-effect false :init-data true})
-                      (update! state side (assoc (get-card state card) :title "Media Blitz")))}
-        card nil))}})
+    :prompt "Choose an agenda in the runner's score area"
+    :choices {:req (req (and (agenda? target)
+                             (is-scored? state :runner target)))}
+    :change-in-game-state (req (seq (:scored runner)))
+    :effect (req (update! state side (assoc card :title (:title target) :abilities (ability-init (card-def target))))
+                 (card-init state side (get-card state card) {:resolve-effect false :init-data true})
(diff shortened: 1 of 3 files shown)


NBKelly:
turns out it's easy for tests to pass when you run the wrong test file :)

NoahTheDuke (changes requested):
Can't have a `:scored` part of the test set up because too many things rely on the complex game-state changes that come from cards being scored. You gotta use the `play-and-score` helper function.

NBKelly:
hmm - so my intention was specifically that these wouldn't be treated like scores/steals, just as shortcuts for things like this, or (for example) testing exchange of information (essentially, I just want it to be as easy as possible to write tests).

The moving part is exactly what happens in `steal` in `access.clj`, and there are checkpoints which occur between then and the game starting, so for everything except resolving on-scored and on-stolen events it *should* be fine.

Perhaps if I renamed them `:score-area`, that would be clear and reasonable?

NBKelly:
Just did that, and I simplified the basic EoI test using that (just to assert that the corp side works).

If it's too much, I can still just scrap it.

NoahTheDuke:
i think it'll be okay. :score-area is more clear so that's fine. Why are you moving cards into the hand? You're moving the scored cards from deck into the score area.

NBKelly:
That one's copying how the `starting-hand` fn works - since these are "post mulligan phase", with the hand being a random mix of what you specify in your hand, what you specify in the deck, and what's specified in the score areas, the hand has to go back to the deck before doling out the scored cards, and then that many cards have to go back in the hand.

I would be hesitant to put it before the mull phase.

NoahTheDuke:
Hmm I see. Cool, thanks.

Issue #6497, opened 2022-07-30
Engine shortcomings


NoahTheDuke:
Alongside the continual match of bug fixes and general enhancements, I have a desire to make the engine strictly enforce the rules without players having to know the technical minutiae and fix mistakes.

To that end, I would like to compile a list of all of the current known areas where jnet does not properly enforce the rules. To avoid arguing about what counts and what doesn't, you can post bugs here as well. I'll sort out what doesn't exist yet and what are bugs in existing code.

NoahTheDuke:
Priority is completely absent.

nicemoreoften:
overrezzing/installing uniques/consoles must be done manually on jnet but is handled "automatically" by "the game" when you install things in the cr

jaypumpkin:
Some of these would be a disaster to implement or might have been fixed since last check but in the spirit of the question:
1. Hosting restrictions (e.g. hosting a program that had already been installed)
2. That you have the option to trash any cards in the server when installing
3. That you have the option to trash any programs when installing
4. NCIGS
5. Probably redundant to the priority comment but if I don't know to say "action" or have a nice opponent who asks, they can score out even when i have access to clot or similar.
6. Conditional paid abilities that don't prompt for use (e.g. I may not realize i needed to click TTW to get extra access until after i've seen hidden info in the form of the first card).

locks:
As a runner, you can go over the memory limit for programs.

postinternetsyndrome:
The ability windows at end of turn and beginning of turn are fudged, with jnet apparently having a space between turns (which does not exist in the rules) where things done in those windows are typically handled. I belive I was once told that actions taken in this limbo does count towards the upcoming player's turn for the purposes of "once per turn" restrictions and triggers, but if that's indeed the case, it might be better to represent it as what it actually is rather than obfuscate it with the "start turn" simplification that will immediately run triggers and draw mandatory.

The full sequence when switching turns is:
Player A:
- Last action
- Discard
- PAW
- Turn end triggers
Player B:
- Gain clicks
- PAW
- Recurring credits refill (I have no idea why this happens on the other side of the PAW compared to gaining clicks lol)
- Turn begins triggers
- Mandatory draw (if corp)
- PAW
- First Action

tolaasin:
> As a runner, you can install over the memory limit for programs.

and losing a console doesn't enforce memory usage after.

tolaasin:
[...]

PR #7562, opened 2024-07-16
updated formatting of the charge ability (so it no longer silently fails when nothing happens)

NBKelly:
Addresses one of the comments in #6497.


NoahTheDuke:
How exactly does this address the issue? Can you show it in a test?

NBKelly:
> How exactly does this address the issue? Can you show it in a test?

Specific issue: 
> `locks uses Daeg, First Net-Cat to charge.` showed up in the game log although there was nothing able to charge.

I just removed the `:msg "charge"` key from daeg. I thought that bit was a little bit too trivial for a test.

The other half of the PR:
> When there were no charge targets for these cards, we were essentially calling:
> `continue-ability state side (effect-completed state side eid) card nil`
> and when there were targets, two of the variables passed in (eid, card) were just being redefined in the ability map anyway, so were useless

Tests not failing is a sign that one works (though I did let one sneak through) - I'm updating charge-test.clj, because the first time through it looks like I actually had a handful of no-ops instead of executing the abilities I thought I was :joy: (so that test covers the changes on that side)

NoahTheDuke:
okay cool, just wanna make sure i wasn't missing something important. looks good

PR #7564, opened 2024-07-17
Make gameboard use edn instead of json

NoahTheDuke:
All of my fears about ruining history is incorrect, as we're already storing the history as json in the database. If replays are broken, we can fix them. Otherwise, I think we should pull the trigger.


NoahTheDuke:
I should have merged this lol. I'll update and merge this week

NBKelly:
hmm, maybe we can try to merge this after the elevation release?

NoahTheDuke:
Yeah I gotta dust this off and bite the bullet. I mostly didn't do it because I don't have control over deployments so didn't want to do something I can't undo fast/give jdeng a bunch of extra work, but I think it'll be worthwhile in the long run.

NoahTheDuke:
lol okay THIS TIME I'LL DO IT

PR #7565, opened 2024-07-17
Fix differ issues from json->edn conversion

### LINKED ISSUE #7424: Overseer Matrix' UI shows additionalt blank options

NoahTheDuke:
Closes #7424.

Would be handled by #7564 but that's a mess and would break all game histories, so that's on pause.

Differ handles [sequential types](https://github.com/robinheghan/differ?tab=readme-ov-file#sequential-types) in a special way, with `:+` meaning "insert at the end". Our reliance on json means that it was always being treated as `"+"` and thus differ would ignore it. I believe this has infected every portion of the entire frontend as long as we've used differ, but it's only surfaced here cuz we tend to ignore nils everywhere lol.

The result of this bug would be that the diff from the waiting prompt to the choices prompt would have `:choices ["+" {:value {:title "Overseer Matrix" ...} ...} "+" {:value {:title "Overseer Matrix" ...} ...} "+" {:value {:title "Overseer Matrix" ...} ...} "+" {:value {:title "Yakov Erikovich Avdakov" ...} ...}]`, and thus be rendered incorrectly because you can't destructure a string. With this fix, it'll now just be the correct values merged in to their right place.

Having written all of this out, I have also updated the `prompt-state` differ stuff to be smarter, borrowing code from #7564.

### DIFF
--- a/src/cljs/nr/gameboard/board.cljs
+++ b/src/cljs/nr/gameboard/board.cljs
@@ -1606,16 +1606,16 @@
 
        ;; otherwise choice of all present choices
        :else
-       (doall (for [{:keys [idx uuid value]} choices]
-                (when (not= value "Hide")
-                  [:button {:key idx
-                            :on-click #(do (send-command "choice" {:choice {:uuid uuid}})
-                                           (card-highlight-mouse-out % value button-channel))
-                            :on-mouse-over
-                            #(card-highlight-mouse-over % value button-channel)
-                            :on-mouse-out
-                            #(card-highlight-mouse-out % value button-channel)}
-                   (render-message (or (not-empty (get-title value)) value))]))))]))
+       (doall (for [{:keys [idx uuid value]} choices
+                    :when (not= value "Hide")]
+                [:button {:key idx
+                          :on-click #(do (send-command "choice" {:choice {:uuid uuid}})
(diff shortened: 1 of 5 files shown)


src/cljs/nr/gameboard/board.cljs
```diff
@@ -1606,16 +1606,16 @@
...
-                  [:button {:key idx
-                            :on-click #(do (send-command "choice" {:choice {:uuid uuid}})
-                                           (card-highlight-mouse-out % value button-channel))
-                            :on-mouse-over
-                            #(card-highlight-mouse-over % value button-channel)
-                            :on-mouse-out
-                            #(card-highlight-mouse-out % value button-channel)}
-                   (render-message (or (not-empty (get-title value)) value))]))))]))
+       (doall (for [{:keys [idx uuid value]} choices
+                    :when (not= value "Hide")]
```
NoahTheDuke, line 1610: Unrelated, but we shouldn't return `nil` anyway.

PR #7567, opened 2024-07-17
test for jinteki rp vs sand storm

### LINKED ISSUE #4851: Sandstorm combination with Jinteki: replicating perfection
Sandstorm was on R&D. First run of runner rezzed sandstorm. Deflected at remote server. Both runner and corp had no action possibilities. With replicating perfection, the corp cannot run on a remote server before running on a central one first.

NBKelly:
Closes #4851


NoahTheDuke:
Should there be an assertion at the end of this test? i know click-prompt has asserts for the prompt itself, but is that all you're checking?

PR #7572, opened 2024-07-18
Notify users 30 seconds before lobby times out

### LINKED ISSUE #6485: Suggestion: Warning before game closes for inactivity.
G'day,

Would just like to suggest that there be some kind of warning (preferably with a sound?) happen say, a few minutes before a game gets closed for inactivity.

Cheers.

NBKelly:
Everything is in place except the noise. I don't know what's actually compatible with our license.

Closes #6485


NoahTheDuke:
The codebase is MIT. As long as we're attributing when using a sound, we should be free to use "royalty-free audio" from whichever website.

NBKelly:
I'm going to let this one sit until after I've figured out #7575, since they're going to clash

NBKelly:
ok, this one should be good to go now

PR #7573, opened 2024-07-19
Cards arent playable during mandatory draw/start of turn

### LINKED ISSUE #4391: After scoring agenda against Leela, corp gets another turn
This is a bit weird.
After spending my turn scoring an SSL against Leela, and after she bounced my IP Block from HQ i got the "mandatory draw" instead of End Turn (see attached).
As i did not have an "end turn" command (that i know of) I had to click "mandatory draw" which first prompted SSL money and then gave me a card.
I then had to put card back on top and shuffle, so that runner could take his turn.
Not sure why this happened. Note that Thimblerig was also installed.
image

### LINKED ISSUE #7170: HQ cards playable while mandatory draw is pending

NBKelly:
Makes it so you can't play cards from your hand while waiting on a mandatory draw, or while start-of-turn events are pending (phase 1.2).

This also caught out a handful of tests that just left phase 1.2 pending and did other things, so I fixed those.

I made Ibrahim Salems phase 1.2 flag check he wasn't disabled as well. I *think* he's the only unique asset that fires on that phase.

Closes #7170
Also closes #4391, by preventing the scenario where it could happen (playing and scoring an agenda during phase 1.2)


test/clj/game/cards/assets_test.clj
```diff
@@ -2496,6 +2496,7 @@
       (take-credits state :corp)
       (take-credits state :runner)
       (take-credits state :corp)
+      (end-phase-12 state :runner)
```
NoahTheDuke, line 2499: why is this needed?

NBKelly: Two sources of draw are triggering at the same time (both crowdfundings), and they're both flagged with runner-phase-12-draw.

NoahTheDuke: Should you choose one of the triggers instead?

NBKelly: true, didn't consider that.

From my testing, it seems that they do both automatically fire when phase-1.2 is ended, but I didn't know that and it wasn't apparent when looking at the test - so I added in lines to that effect.

NoahTheDuke: oh yeah, that's right, it pauses if you want to manually choose. lmao our start of turn stuff is dumb.

PR #7575, opened 2024-07-19
Allow bespoke game sounds

NBKelly:
I'm 50/50 on whether this is even a good idea or not, but I've put a PR up anyway and I don't mind if it gets tossed or kept.

This builds on the changes in #7571.

I thought it would be neat if we could set bespoke sounds for certain ice or cards. Like some noises for the harmonic ice, or a gunshot sound for end of the line.

I kind of want the OCTGN archer sound as well, but I would have to figure out if that was ok - that's a problem for later.

All the sounds I used are attributed in attributions.txt, which will be under our public resources. They're all CC or Attribution 3/4, so (I think) they're fine to use.

See the attached video for examples:

[bespoke-sounds.webm](https://github.com/user-attachments/assets/c8127fd6-bb41-4658-b8b5-277866c704e7)


NoahTheDuke:
I think this is fun and cute. Given that it's new, should the bespoke sounds have like an option somewhere? "I only want archer, not proco"?

NBKelly:
Yeah, I think that's reasonable.

I'll think a little bit over the week as to how to make that work.

NBKelly:
Alright, so I got back around to this.

Essentially, we have a map of optional sounds like
`{name1 {:group ... :default ...}
  name2 ...
  }`

And when we want to play a sound, if it's an optional one, we check if the group (ie harmonics) is enabled, and if it's not, we play the default.

You can selectively enable the sounds through your settings page, and you get a preview whenever you tick the box (it plays a random sound from that group).

For now the groups are `archer, end of the line, harmonics suite`. 

One more thing I want to do is have that specific option group be foldable (like an accordion layout or something, idk what the terminology is), but that might be a job for later. I just imagine if we end up with 30 or so options there one day, it will be a little in the way.

The grouping is nice because I can insert one group for, say, `april fools 2025` for a few dozen cards.

bespoke

NoahTheDuke:
Sorry I never merged this. Good to go you think?

NBKelly:
Yeah this one is good to go

PR #7579, opened 2024-07-19
Enforce first-time where possible

### LINKED ISSUE #4701: Clean up differences between "once per X" and "first time X"
As it stands, we use `:once :per-turn` as an easy means of checking "The first time each turn X happens", when that's _not_ what it means and conflating the two muddies the code base. This no doubt leads to subtle bugs and misimplementations.

NBKelly:
I went through all the cards with `first-time` triggers, and in (almost) every case ensured they were actually checking it was the first time.

Ones where the check was correct, I removed irrelevant once keys. Cards like dbs that merge triggers need to keep those keys though.

Closes #4701

### DIFF
--- a/src/clj/game/cards/identities.clj
+++ b/src/clj/game/cards/identities.clj
@@ -20,8 +20,8 @@
    [game.core.effects :refer [register-lingering-effect is-disabled?]]
    [game.core.eid :refer [effect-completed get-ability-targets is-basic-advance-action? make-eid]]
    [game.core.engine :refer [not-used-once? pay register-events register-once resolve-ability trigger-event]]
-   [game.core.events :refer [event-count first-event? first-run-event?
-                             first-successful-run-on-server? no-event? not-last-turn? run-events turn-events]]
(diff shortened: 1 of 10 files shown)


src/clj/game/cards/identities.clj
```diff
@@ -445,15 +445,13 @@
                             (not (:facedown context))))
              :async true
              :effect (effect (gain-credits :corp eid 2))
-             :once :per-turn
              :msg (msg "gain 2 [Credits] from " (:card-target card))}
             {:event :play-event
              :req (req (and (:card-target card)
                             (first-event? state :runner :play-event)
                             (is-type? (:card context) (:card-target card))))
-             :async true
```
NoahTheDuke, line 454: This is needed

NBKelly: whoops :joy:

NoahTheDuke:
This is great, the difference between `:once :per-x` and `first-event?` have never been properly handled, so I appreciate you doing that here.

PR #7580, opened 2024-07-19
Restructure ganked to encounter ice after the card is done being accessed

### LINKED ISSUE #6155: Acme tag remains after Ganked! pushes you back to the first ice
While tagless, I access 2 cards in Acme's HQ. The first is a Ganked! that sends me to the first ice over HQ. I fully break this ice. There are no indicators that I'm still tagged with Acme's ability (ie no blue exclamation mark). The second card I access is a Quantum Predictive Model, which ends up in the Corp's score area even though I am not tagged.

### LINKED ISSUE #6248: Aeneas Informant does not give a credit if Ganked! is accessed, and is fired by the runner
I think it should!

### LINKED ISSUE #6943: Ganked into Fairchild 3.0 hang the game instead of ETR
Link to bug replay: https://jinteki.net/bug-report/0daead37-fee3-4c63-99d5-868d9b209c18?b=0

Description:

With Architect of Tomorrow, after using Ganked! on a Fairchild 3.0 and choosing to End the run, the game hang at the "Success" state and does not let the command /close-prompt, /end-run or /jack-out work.

Thanks.
Diogene

NBKelly:
Essentially we're firing the post-access event after the trash and before the encounter, which according to the rules team, is how the card works.

This lets us make maw and aeneas informant correctly with with ganked (Closes #6248).

Additionally, it prevents the waiting prompt on FC3.0 breaking the forced encounter (Closes #6943).

And then I added a test for ganked! vs. acme too (Closes #6155).

### DIFF
--- a/src/clj/game/cards/ice.clj
+++ b/src/clj/game/cards/ice.clj
@@ -1722,15 +1722,21 @@
              :waiting-prompt true
              :choices (req [(when (can-pay? state :runner eid card nil [(->c :credit 3)])
                               "Pay 3 [Credits]")
-                            (when (or (not (can-pay? state :runner eid card nil [(->c :credit 3)]))
-                                      (can-pay? state :runner eid card nil [(->c :trash-installed 1)]))
-                              "Trash an installed card")])
+                            (when (can-pay? state :runner eid card nil [(->c :trash-installed 1)])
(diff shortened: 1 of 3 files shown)


NoahTheDuke:
Nice find!

PR #7582, opened 2024-07-21
Implemented pawn properly

### LINKED ISSUE #5326: Pawn implementation should be improved.
Currently Pawn does not verify its location nor does it automate the successful run check. We can improve this.

NBKelly:
Little but too late to see play, but it is what it is.

Closes #5326

### DIFF
--- a/src/clj/game/cards/programs.clj
+++ b/src/clj/game/cards/programs.clj
@@ -2535,7 +2535,42 @@
                 (strength-pump 2 2)))
 
 (defcard "Pawn"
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/programs.clj
```diff
@@ -2535,42 +2535,53 @@
                 (strength-pump 2 2)))
 
 (defcard "Pawn"
-  {:implementation "[Erratum] Program: Caïssa - Trojan. Developer Note: All abilities are manual"
```
NoahTheDuke, line 2538: We should keep the errata implementation note because people seem to like those.

PR #7583, opened 2024-07-22
maybe deregister users when they disconnect from the game or close their websocket connections

NBKelly:
I'm not sure if this was the source of the lag or not, but I think it's worth a try.

This should stop us having 1600 cached users in the app state I think.


NoahTheDuke:
Good find! Let's try it out.

PR #7585, opened 2024-07-23
remove users from the lobby update list when they time out

NBKelly:
Our server looks like this right now:
`2024-07-23T07:58:10.906Z ns521372 INFO [web.telemetry:25] - stats - lobbies: 3 cached-users: 403 lobby-subs: 64 lobby-update-uids: 18 | websockets - :ajax {  uid: 0 conn: 0 } :ws {  uid: 54 conn: 108 }`

We're caching a silly amount of users (I think I fixed that early), and we're sending updates to a bunch of people that probably don't even exist. This is maybe one of the reasons for lag? idk

In any case, here's an attempt.

Instead of just true/false, we're putting  a timestamp in lobby-subscriptions. We only send out updates to people who are active within the last two hours.

That timeframe is actually super generous, any time you click out and into the page, tab in, enter or exit a game, or change to the play tab, that gets reset. Realistically, it could be 15 minutes and it would be fine. But right now this is going to be no worse than what we have, and probably somewhat better.

I also updated the telemetry to show us the average subscription lifetime, and the oldest subscription.

These little bits will be inserted into the telemetry:
`lobby-update-uids: 1 average-lobby-subs-lifetime: 4m oldest-lobby-sub: 4m`

Thanks to @butzopower for the leads to follow on this one.

### DIFF
--- a/src/clj/web/app_state.clj
+++ b/src/clj/web/app_state.clj
@@ -1,17 +1,21 @@
 (ns web.app-state
   (:require
-    [medley.core :refer [find-first]]))
+   [cljc.java-time.temporal.chrono-unit :as chrono]
+   [cljc.java-time.instant :as inst]
+   [medley.core :refer [find-first]]))
 
 (defonce app-state
   (atom {:lobbies {}
          :lobby-updates {}
          :users {}}))
 
+(defonce lobby-subs-timeout-hours 6)
+
 (defn register-user
   [app-state uid user]
   (-> app-state
       (assoc-in [:users uid] (assoc user :uid uid))
-      (assoc-in [:lobby-updates uid] true)))
+      (assoc-in [:lobby-updates uid] (inst/now))))
 
 (defn uid->lobby
   ([uid] (uid->lobby (:lobbies @app-state) uid))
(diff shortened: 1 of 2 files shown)


src/clj/web/app_state.clj
```diff
@@ -52,21 +56,28 @@
...
-  (let [users (:users @app-state)
-        new-users (dissoc users uid)]
-    (swap! app-state #(assoc %1 :users new-users))))
 
 (defn pause-lobby-updates
   [uid]
-  (swap! app-state assoc-in [:lobby-updates uid] false))
+  (let [lobby-update (:lobby-updates @app-state)
+        new-lobby-updates (dissoc lobby-update uid)]
+    (swap! app-state #(assoc %1 :lobby-updates new-lobby-updates))))
```
NoahTheDuke, line 64:
```suggestion
    (swap! app-state assoc :lobby-updates new-lobby-updates)))
```

NoahTheDuke: jk nvm

src/clj/web/app_state.clj
```diff
@@ -52,21 +56,28 @@
...
+    (do (pause-lobby-updates uid)
+        nil)))
+
+(defn deregister-user!
+  "Remove user from app-state. Mutates."
+  [uid]
+  (let [users (:users @app-state)
+        new-users (dissoc users uid)]
+    (pause-lobby-updates uid)
+    (swap! app-state #(assoc %1 :users new-users))))
```
NoahTheDuke, line 83:
```suggestion
(defn deregister-user!
  "Remove user from app-state. Mutates."
  [uid]
  (pause-lobby-updates uid)
  (swap! app-state dissoc-in [:users uid]))
```

Same as above.

PR #7591, opened 2024-07-27
use noto emoji font in gamerow and replay bar

### LINKED ISSUE #7531: Add a little clock next to the new room time indicator
What's this, 4 meters? 
image

haha, just kidding. I think a clock emoji next to the minutes counter would add a bit of visual weight to the feature.

### LINKED ISSUE #7586: Enhancement: better emoji replay bar icons
image
image

Here's something really quick and easy: make the replay icons use [the black-and-white Noto Emoji font](https://fonts.google.com/noto/specimen/Noto+Emoji?preview.text=%F0%9F%8F%83). I recently changed my Firefox to use it as the default font, and these icons match with the existing click icons and look 600% better than leaving it to the color system font.

If you don't want to include the entire TTF font just for a few glyphs, you can remove all but the useful ones using [fonttools](https://github.com/fonttools/fonttools) or something.

NBKelly:
The noto emoji font uses the Open Font License, so it's basically free game for us (no acknowledgement needed).

Closes #7531
cute-clock

I'm not sure the siren is the best thing for a black and white purge, but it's the closest B&W thing I could find.

Closes #7586
emoji-annotations


NoahTheDuke:
Do we want to ship 850kb just for 3 images? can these be converted to svgs and inlined in some way?

NBKelly:
I'm not sure how to do that in a smart way, but I did min the font down the exactly the 4 glyphs we're using (it's still 4kb though, which seems slightly excessive, but it's at least way better)

NoahTheDuke:
That's much better.

PR #7603, opened 2024-08-01
Fix some missing translations and update Chinese in translations.cljs

klingeling:


NoahTheDuke:
This is killer, thank you.

PR #7605, opened 2024-08-01
sunset angel arena

### LINKED ISSUE #6026: [[Angel Arena]] Win/Loss oponents records show rebirthed IDs rather than original ID
Example, a game where a MaxX rebithered into Omar was recorded as a match vs Omar. 

Should be a fairly straight forward bug, as I don't think anything particularly odd happened that match. 

image

### LINKED ISSUE #6027: [[Angel Arena]] Dangling matchmaking search, leading to matching with self.
I was able to accidentally queue up a game vs myself due to what seems like a dangling search that I thought was cancelled still existing in the system. 

Series of actions. 

1. Queued up for a game as corp and waited for awhile without receiving a match
2. Clicked the watch button on another game of Angel Arena that had been going on between 2 other players. 
3. Returned back to the AA lobby page after the watched game was done. No indication that my search from 1, was still in progress in the UI, and queued up for another game this time on the runner side. 

Resulted in dropping straight into a game vs myself in a broken ui state. Eventually the game was detected as being inactive and a cancel dialog allowed for aborting the match without any negative consequences (which is an awesome quiality of life feature. Awesome job thinking of that preemptively btw). 

What I expected to happen:

When navigating into a game as a spectator, I'd expected that to have either been blocked with a queued match search was in progress, or had it cancel the search. Alternatively, allow for spectating while searching, but handle being interrupted by a match being found, and the ui state to correctly keep track that a search was in progress when navigating back to the AA page.

### LINKED ISSUE #6029: Angel Arena Self Play Bug
Link to bug replay: https://jinteki.net/bug-report/9fec2b48-417b-4cc1-8369-2bdd1d8a15b1?b=0

Description:

Queued as one side on Angel Arena- I thought I cancelled the queue, but definitely went into a casual lobby game, then queued up for the other side and played myself!

### LINKED ISSUE #6034: [Suggestion] Audio alert when Angel Arena game starts
Had a few games recently where it looks like people have queued for a game then forgotten about it.

NBKelly:
Since angel arena is discontinued/no longer supported, we might as well remove it from the UI.

I've only commented out the functional code, so it's in place to be re-assembled if we ever decide we want to do that.

Closes #6034
Closes #6029
Closes #6027
Closes #6026

Additionally, because it already touches the code anyway, I made it so that bug report replays are automatically shared. That way, we're not getting useless replays with no context littering our issues.

### DIFF
--- a/src/clj/web/system.clj
+++ b/src/clj/web/system.clj
@@ -83,8 +83,8 @@
 
 (defmethod ig/init-key :web/lobby [_ {:keys [interval mongo time-inactive]}]
   (let [db (:db mongo)]
-    [;;(tick #(angel-arena/check-for-inactivity db) interval)
-     (tick #(lobby/clear-inactive-lobbies db time-inactive) interval)]))
+    [(tick #(lobby/clear-inactive-lobbies db time-inactive) interval)
+     #_(tick #(angel-arena/check-for-inactivity db) interval)]))
 
 (defmethod ig/halt-key! :web/lobby [_ futures]
   (run! future-cancel futures))
--- a/src/clj/web/system.clj
+++ b/src/clj/web/system.clj
@@ -24,7 +24,7 @@
    [taoensso.sente :as sente]
    [time-literals.data-readers]
    [time-literals.read-write :as read-write]
-   [web.angel-arena :as angel-arena]
+   ;; [web.angel-arena :as angel-arena]
    [web.api :refer [make-app make-dev-app]]
    [web.app-state :as app-state]
    [web.game]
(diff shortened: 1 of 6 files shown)


src/clj/web/system.clj
```diff
@@ -83,8 +83,8 @@
 
 (defmethod ig/init-key :web/lobby [_ {:keys [interval mongo time-inactive]}]
   (let [db (:db mongo)]
-    [(tick #(lobby/clear-inactive-lobbies db time-inactive) interval)
-     (tick #(angel-arena/check-for-inactivity db) interval)]))
+    [;;(tick #(angel-arena/check-for-inactivity db) interval)
+     (tick #(lobby/clear-inactive-lobbies db time-inactive) interval)]))
```
NoahTheDuke, line 87:
```suggestion
    [(tick #(lobby/clear-inactive-lobbies db time-inactive) interval)
     #_(tick #(angel-arena/check-for-inactivity db) interval)]))
```

PR #7607, opened 2024-08-01
allow spectating by side + only compute diff if it's relevant

### LINKED ISSUE #6097: (Feature Request) Player perspective option for spectators in open-hand games
When spectating a game with open information, there could be a little drop down (beside the game-log and settings tabs on the chat box would be convenient) allowing you to view: 1) all hidden information 2) only hidden information for the runner 3) only hidden information for the corp 4) no hidden information. 

I have no clue about how difficult the implementation would be for this one, but I think this could be a nice teaching/practice tool.

### LINKED ISSUE #6351: [feature request] Player perspective when spectating games.
I would like to see a player perspective option on spectated games, like we have in replays.

I think games with public information available should start with all cards unknown to spectators.

* Each player has a dropdown at the top of the screen which they can use to set their information to public/private.
* Each spectator has a dropdown at the top of the screen, allowing them to set their desired perspective (no-open-info, corp, runner, both)
* A spectator can only see the private info if they've opted in to that side (or both), and if that player has set their info to public
* Whenever a player selects a perspective, or a player changes info availability, it's printed in the log
* A player could potentially select only specific spectators to share information with (this would be a more formal 'tournament friendly' option)

I think this could be a reasonable way to enable team tournaments or pair gaming without players needing to rely on third party screen sharing tools, and it could be a useful teaching tool for GLC players.

NBKelly:
This should cut like 1/3rd of the differ load for most games. Why compute stuff for people that don't exist?

Additionally, I added that option in public info games to spectate only the perspective of a given side. I also made sure this works in password protected games, since I think this is really good for private testing and that helps the teams stay honest with themselves.

Closes #6351 and also Closes #6097 (wow, I requested the same feature twice :sweat:)


NBKelly:
Video demonstration here:
[spectate as side.webm](https://github.com/user-attachments/assets/35a76d0c-638f-4a0f-bbed-9f5e791b7104)

src/cljs/nr/game_row.cljs
```diff
@@ -46,16 +46,70 @@
...
-                               (resume-sound))}
-       (tr [:lobby.watch "Watch"])]
-      [:button {:on-click #(if (:password game)
-                             (authenticated
-                               (fn [_]
-                                 (swap! lobby-state assoc :password-game {:game game :action "watch"})))
-                             (do (join-game lobby-state game "watch")
-                                 (resume-sound)))}
-       (tr [:lobby.watch "Watch"])])))
+      (if-not (:spectatorhands game)
```
NoahTheDuke, line 49: This is too long and should be split into separate functions.

src/cljs/nr/game_row.cljs
```diff
@@ -43,19 +43,64 @@
...
-       (tr [:lobby.watch "Watch"])]
-      [:button {:on-click #(if (:password game)
-                             (authenticated
-                               (fn [_]
-                                 (swap! lobby-state assoc :password-game {:game game :action "watch"})))
-                             (do (join-game lobby-state game "watch")
-                                 (resume-sound)))}
-       (tr [:lobby.watch "Watch"])])))
+      (watch-game-button (:spectatorhands game) lobby-state game)
+      (watch-protected-game-button (:spectatorhands game) lobby-state game))))
```
NoahTheDuke, line 103:
```suggestion
      [watch-game-button (:spectatorhands game) lobby-state game]
      [watch-protected-game-button (:spectatorhands game) lobby-state game])))
```

Functions that return hiccup need to be wrapped in `[]` so reagent rerenders them correctly.

NoahTheDuke:
Why not remove `:spectators` entirely and just have `:corp-spectators` & `:runner-spectators`?

NBKelly:
> Why not remove `:spectators` entirely and just have `:corp-spectators` & `:runner-spectators`?

Oh, by spectating a side I don't mean the board orientation (we don't need to change the diffs at all for that) - it's only sharing the hidden information for one of the players at a time. So you can see the hidden information associated with just the corp, just the runner, or both.

Although maybe that doesn't need to go through diffs at all :thinking:

NBKelly:
Now that I think about it, maybe it's possible to do this without actually touching the diffs at all. I'll think about this a little more.

NBKelly:
Ok I changed my mind again, I think this is both safe and fine. If it's tied to diffs, then we can't accidentally leak info, and I like it being tied to lobby state.

PR #7610, opened 2024-08-04
deck builder: Filter ID choices based on selected format

butzopower:
Limit the list of available ID options to IDs that are legal in the selected format. If the currently selected ID is no longer in the chosen format - switch to the first ID in the list.

### DIFF
--- a/src/cljs/nr/deckbuilder.cljs
+++ b/src/cljs/nr/deckbuilder.cljs
@@ -881,7 +881,7 @@
 (defn- change-format
   [s new-format]
   (swap! s assoc-in [:deck :format] new-format)
-  (if-not (legal-in-format (get-in @s [:deck :identity]) new-format)
+  (when-not (legal-in-format (get-in @s [:deck :identity]) new-format)
     (let [side (get-in @s [:deck :identity :side])
           new-id (first (sort-by :title (side-identities side new-format)))]
       (when new-id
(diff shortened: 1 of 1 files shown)


butzopower:
> I'm not a fan of the automated swapping (just because standard changes over time), but other than that this is pretty neat.

I didn't really want to do the automatic swapping but the issue is with the form controls - if the currently selected option disappears then a select control will by default just pick the first option - if we don't then update the deck ID this gets weird where you have an ID that's no longer an option selected, and to actually switch to the option now in the select (say A Teia if you went to start up) you actually have to select a different ID and then select the first ID again to reset it.

NBKelly:
ah, yeah that makes a lot of sense.

src/cljs/nr/deckbuilder.cljs
```diff
@@ -872,12 +878,21 @@
      :value (get-in @s [:deck :name])
      :on-change #(swap! s assoc-in [:deck :name] (.. % -target -value))}]])
 
+(defn- change-format
+  [s new-format]
+  (swap! s assoc-in [:deck :format] new-format)
+  (if-not (legal-in-format (get-in @s [:deck :identity]) new-format)
```
NoahTheDuke, line 884:
```suggestion
  (when-not (legal-in-format (get-in @s [:deck :identity]) new-format)
```

There's no else-branch so `when` is preferred.

NoahTheDuke:
Good change, this will be helpful.

Issue #6960, opened 2023-07-27
Support for Chinese language card data is sort of broken


csbisa:
With https://github.com/mtgred/netrunner/pull/6902 we have a problem with the Chinese language code that will prevent card titles, images, etc from being not loaded for both Simplified and Traditional Chinese. The problem comes from jnet using `zh-simp` and `zh-trad` for the internal setting while external sources use `zh-hans` and `zh-hant`. We have a few options:

1) Change the internal codes from simp/trad to hans/hant. Users with those languages selected will get reverted to English when the update rolls out and they will have to update their language once again.
2) Have a mapping between those two codes when doing conversion (done somewhere in the API layer).

NBKelly:
Closed as completed via  #7616

PR #7616, opened 2024-08-07
Add Set name's translations in cardbrowser; Fix Chinese language card data broken

klingeling:
Add Set name's translations in cardbrowser;
Add some action translations in game play;
Show translated card name in game log;
Fix Chinese language card data broken.


klingeling:
#6960

NoahTheDuke:
Hell yeah. I'll rebase #7615 on top of this after merging so you don't have to do extra work.

PR #7621, opened 2024-08-10
respirocytes, reeducation, djupstad interact nicely now

### LINKED ISSUE #7619: Respirocytes needs to trigger between reeducation and djupstad
So we can probably just add a checkpoint there and it will just work.

NBKelly:
Closes #7619

### DIFF
--- a/src/clj/game/cards/agendas.clj
+++ b/src/clj/game/cards/agendas.clj
@@ -21,7 +21,7 @@
    [game.core.drawing :refer [draw draw-up-to]]
    [game.core.effects :refer [register-lingering-effect]]
    [game.core.eid :refer [effect-completed make-eid]]
-   [game.core.engine :refer [pay register-events resolve-ability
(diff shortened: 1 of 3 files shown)


src/clj/game/cards/agendas.clj
```diff
@@ -430,7 +430,10 @@
                                                      " from the grip"
                                                      " to the bottom of the stack."
                                                      " The Runner draws 1 card"))
-                         (draw state :runner eid 1)))
+                         (queue-event state :runner-hand-changed?)
+                         (wait-for
+                           (checkpoint state side)
```
NoahTheDuke, line 435: Is the checkpoint here tied to a timing structure?

NBKelly: Yeah, there's a checkpoint where there's a comma in brain rewiring. It's more obvious in the rules text than the printed text.

` The Runner adds X cards from the grip at random to the bottom of the stack, then draws 1 card.`

PR #7625, opened 2024-08-10
assert swaps are legal before swapping

### LINKED ISSUE #7482: The corporation should not be able to swap regions into servers that already have regions
I think we clean it up in the checkpoint if it happens, but it's not even legal to attempt to do this apparently.

https://access.nullsignal.games/CompRules/Null_Signal_Games_Netrunner_Comprehensive_Rules_v24.03.pdf#section*.100

3.6.5.e

In terms of implementation, I think we could have a `swap-legal?` function in moving which can determine if swaps are legal (for installed cards):
* The region thing should be considered
* Wrong types (ice->not ice) be considered
* two assets in a server
* cards with a server restriction, like ZATO (not install restriction, like off the grid)

(I got confirmation from rules that these are all invalid targets for swaps)

NBKelly:
Closes #7482

### DIFF
--- a/src/clj/game/core/moving.clj
+++ b/src/clj/game/core/moving.clj
@@ -3,7 +3,7 @@
     [clojure.string :as string]
     [game.core.agendas :refer [update-all-agenda-points]]
     [game.core.board :refer [all-active-installed]]
-    [game.core.card :refer [active? card-index condition-counter? convert-to-agenda corp? facedown? fake-identity? get-card get-title get-zone has-subtype? ice? in-hand? in-play-area? installed? program? resource? rezzed? runner?]]
+    [game.core.card :refer [active? agenda? asset? card-index condition-counter? convert-to-agenda corp? facedown? fake-identity? get-card get-title get-zone has-subtype? ice? in-hand? in-play-area? installed? program? resource? rezzed? runner?]]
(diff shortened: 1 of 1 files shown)


src/clj/game/core/moving.clj
```diff
@@ -457,12 +457,41 @@
   (let [cards (take n (shuffle (get-in @state [to-side :hand])))]
     (trash-cards state from-side eid cards {:unpreventable true})))
 
+(defn swap-legal?
+  "checks that swapping two installed corp cards is legal"
+  [state side a b]
+  (let [pred? (every-pred corp? installed?)
+        xor? (fn [a b f]
+               (or (and (f a) (not (f b)))
+                   (and (f b) (not (f a)))))]
```
NoahTheDuke, line 466: Maybe move the two `f` calls into a let: `(let [ret-a (f a) ret-b (f b)] (or (and ret-a (not ret-b)) (and (not ret-a) ret-b)))`? i'm not sure that's right, but it's close.

NBKelly: Perhaps a less ugly approach would be something like:
`(= 1 (count (filter identity (apply f [a b]))))`

NoahTheDuke: That's significantly less readable, please no.

PR #7626, opened 2024-08-10
Score only during your turn

### LINKED ISSUE #6888: [EDITME] Runner tries to steal scored card
Link to bug replay: https://jinteki.net/bug-report/ea980a40-de94-4ec1-87cb-2cd083d99bfe?b=0

Description:

Runner successfully accessed server with agenda missing one advancement.  NBN Pravdiost agenda then allowed to place one advacement, and score the agenda (should not be allowed).  Game reported "server error" missing the card to steal.

NBKelly:
Closes #6888

I also added a command to manually score agendas so players have a way to fix mistakes or repair the gamestate without needing to rewind the turn.

### DIFF
--- a/src/cljs/nr/gameboard/board.cljs
+++ b/src/cljs/nr/gameboard/board.cljs
@@ -72,8 +72,8 @@
 (defn action-list
   [{:keys [type zone rezzed advanceable
            advancementcost current-advancement-requirement] :as card}]
-  (let [active-player (r/cursor game-state [:active-player])
-        side (r/cursor game-state [:side])]
+  (r/with-let [active-player (r/cursor game-state [:active-player])
+               side (r/cursor game-state [:side])]
     (cond->> []
       ;; advance
       (or (and (= type "Agenda")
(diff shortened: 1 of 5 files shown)


src/cljs/nr/gameboard/board.cljs
```diff
@@ -72,33 +72,36 @@
...
-    ;; rez
-    (and (#{"Asset" "ICE" "Upgrade"} type)
-         (not rezzed))
-    (cons "rez")
-    ;; derez
-    (and (#{"Asset" "ICE" "Upgrade"} type)
-         rezzed)
-    (cons "derez")))
+  (let [active-player (r/cursor game-state [:active-player])
+        side (r/cursor game-state [:side])]
```
NoahTheDuke, line 76: Let's make this a `r/with-let` so the cursors aren't recreated every call.

NoahTheDuke:
Should this check exist in the `can-score?` function? Or should it exist in the basic action? Are there corp abilities that can't score on runner's turn?

NBKelly:
I think the `can-score?` function is a good place for the check, just for future-proofing. The only eligibility info we send in diffs is the advancement requirement though, so the turn check does need to be replicated on the front-end (but that's solely so the 'score' button goes away, and we don't get bug reports about it not working :joy:).

There's only three abilities in the game that score cards:
* Big Deal (which already needs to ignore the terminal requirement)
* Plan B (which needs to ignore the turn requirement)
* Research Grant (which has no special rules I think)

So I *think* this is reasonable.

PR #7633, opened 2024-08-10
cider indent hint for `changed?` macro

NBKelly:
I'm *pretty sure* I got this right. This should let cider read the indent rules for `changed?` to be like a letfn form.


NoahTheDuke:
I don't use cider's indentation system but this is fine.

PR #7635, opened 2024-08-10
Runner install msg cleanup

### LINKED ISSUE #7629: Failing to find a muse target will shuffle the stack, regardless of the location
If you pick heap or grip, it will still shuffle the stack :joy: whoops :joy:

### LINKED ISSUE #7630: We can probably condense the runner install messages the same way we did the corp install messages
Things like displaying where a card was installed from and the source card can potentially all be squeezed into the same log message.

So something like `X uses muse to install Y from the stack, paying Z` would like nice.

NBKelly:
My attempt to make the install messages (marginally) nicer, and make it so we don't have to write these `:msg` keys ourselves.

I updated `engine` to keep track of the last relevant payment-str on an eid, so we can use it when cards have nested costs, instead of needing to hotwire stuff down the line.

Here's a sample of the changed cards, because I know it's really hard to guage what's changed just from reading that runner-install-msg fn:

## Programs ##

Conspiracy breakers all share the same code.

paperclip

Customized Secretary
customized-secretary

DaVinci
davinci

Muse demostrates origins in the text:
muse_000
muse
muse_001
Also Closes #7629

I realized when I did SMC that the listing of costs was kind of ambiguous with nested costs, so I changed it to look like this. I'm not 100% on if this is the best presentation, but it makes sense to me.

smc

## Identities ##

arissana
hayley
wu_000
seb

Here I have to decide if it's worth including the discount in the install message. I think players are responsible enough to glance at the source card if they're confused.

I also added an option to hide if the credit cost happens to be 0. I feel like the cost being 0 is junk information, but I can make this default to false instead of true if desired.

## Hardware ##
So many of these apply discounts or modify costs that I had to add it to the install messages (I did this pretty late, so I've only got pictures of it for simulchip and wizard's chest).

Clone Chip
cc

Gachapon (note that it now displays the trash cost, I can do this because of the eid change)
gachapon_000
Pantograph
pantograph
One too many rabbit holes
rabbit-hole
Wizard's Chest
wiz
Simulchip
simulchip_002

## Events ##

cf

compile

praxis

spark

price

## Resources ##

crowd
hg
paule

Closes #7630

### DIFF
--- a/src/clj/game/cards/events.clj
+++ b/src/clj/game/cards/events.clj
@@ -228,7 +228,9 @@
                                                           (can-pay? state side (assoc eid :source card :source-type :runner-install) target nil
                                                                     [(->c :credit (install-cost state side target {:cost-bonus -1}))])))}
                                  :async true
-                                 :effect (effect (runner-install (assoc eid :source card :source-type :runner-install) target {:cost-bonus -1}))}
+                                 :effect (effect (runner-install (assoc eid :source card :source-type :runner-install) target {:cost-bonus -1
+                                                                                                                               :msg-keys {:install-source card
+                                                                                                                                          :display-origin true}}))}
                                 card nil))}
              {:msg "remove 1 tag"
               :async true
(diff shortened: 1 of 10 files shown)


NBKelly:
Basically the only thing to resolve is if we should bother displaying 0 credit costs. I'm not really sure what the answer is to that.

NBKelly:
ok, current ruling is that it's a 0 cost payment.

I'm keeping the facedown installs as not displaying a cost because there is no cost to pay for this.

NoahTheDuke:
This is cool! I think the pre message and then current message are a little hard to parse ("Runner pays 1 credits, then pays 2 credits, to use x to install y") because which payment is for which action? But overall I think this is really killer and i'm not sure how to fix that. Let's refine later.

NBKelly:
Yeah, I'll admit that one isn't quite optimal. I might see if I can just combine the costs together (even though they aren't technically the same cost, I think it beats the alternatives).

PR #7636, opened 2024-08-11
tied actions to clicks being spend, and the card type, so juli can be correct

### LINKED ISSUE #7330: Juli is still not perfect, but now she's more perfect than before
Writing this so I don't forget. I think if your action trashed a resource, then you install juli and use a resource, you'll still get a click. I'll investigate and fix this later.

### LINKED ISSUE #7387: Juli Moreira Lee adds a click when using Aesop's Pawnshop
This issue was reported on discord so I don't have a replay but apparently Juli used a power counter and added a click when the runner started their turn and used Pawnshop

NBKelly:
Since we can't look back in time at the game state, we need to add a little bit of information. It is what it is.

Closes #7330
And I'm not sure what the exact issue was here, but this impl change should make it impossible as well, so this Closes #7387 too

### DIFF
--- a/src/clj/game/core/costs.clj
+++ b/src/clj/game/core/costs.clj
@@ -36,8 +36,11 @@
   [cost state side eid _card]
   (let [a (:action eid)
         idx (get-in eid [:source-info :ability-idx])
-        is-game-action? (when (= :ability (:source-type eid))
-                          (:action (nth (get-in eid [:source :abilities]) idx nil)))
+        source-abilities (get-in eid [:source :abilities])
+        is-game-action? (when (and (= :ability (:source-type eid))
+                                   (number? idx)
+                                   (seq source-abilities))
+                          (:action (nth source-abilities idx {})))
         source-type (get-in eid [:source :type])]
     (swap! state update-in [:stats side :lose :click] (fnil + 0) (value cost))
     (deduct state side [:click (value cost)])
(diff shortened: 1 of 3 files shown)


src/clj/game/core/costs.clj
```diff
@@ -34,12 +34,21 @@
...
+                                   (seq source-abilities))
+                          (:action (nth source-abilities idx {})))
+        source-type (get-in eid [:source :type])]
     (swap! state update-in [:stats side :lose :click] (fnil + 0) (value cost))
     (deduct state side [:click (value cost)])
     (wait-for (trigger-event-sync state side (make-eid state eid)
                                   (if (= side :corp) :corp-spent-click :runner-spent-click)
                                   {:action a
+                                   :is-game-action? is-game-action?
+                                   :source-type source-type
```
NoahTheDuke, line 51: Why not just put the source here? Then you can check with `(resource? (:source context))`

NBKelly: with how often we duplicate eids, I thought it might bloat the gamestate. Happy to change it though.

NoahTheDuke: hmm maybe! i just don't want to get back into the habit of checking for a string when we mean a type. with the function, the impl can change to a keyword instead, or another type of object. "when would this happen though?" i've considered moving types to keywords lol, it's possible!

NBKelly: hmm. What if we just strip the hell out of the card. Something like `(select-keys c [:type :title :cid])`, and just pass it as `stripped-card`?

Maybe that's worse from a maintenance perspective, I'm not sure.

NoahTheDuke: That's probably the right way to go. Then keys can be added as abilities require.

PR #7640, opened 2024-08-12
Prune duplicate commands

### LINKED ISSUE #5213: Lag when clicking the "draw card" button can lead to ruined games
Sometimes when playing, it is possible that small moments of lag (at either end, or anywhere inbetween) when performing the draw card (or fire subroutines, if it happens to be damage) actions may lead to the actions being taken twice. Alternatively, it may seem that the interface is unresponsive (I clicked draw 10 seconds ago and it didn't draw, so I'll click it again) leading to mistaken double or triple draws. 

I can think of two cases that may cause issues:'
* Multiple 'draw card' or 'fire subroutine' prompts are sent within a small amount of time
* Multiple 'draw card' or 'fire subroutine' prompts are received by the server within a small amount of time

I think that both of these cases could (maybe) be detected serverside, and it may be possible to correct them (ie send prompts instead of executing duplicated actions).

NBKelly:
Essentially, if some packets are slowed down in transit, or there's a holdup, and we end up clicking draw again, sometimes after 3-4 seconds we suddenly draw 2 cards. What this does it asserts, when receiving a game action, that we haven't made another game action within the last 30ms. That should solve *most* of these issues (where the issue is the packet making it to the server, rather than the packet making it back), and should also solve double-clicking mice. 

We can do most of the work to fix lag, but there will always be random path issues for random users at random times, or users with spotty connections, so this will always be relevant.

From my testing, this works fine, but right now it makes every single unit test in the game fail, so there's clearly something significant I need to change :joy: 

Closes  #5213

### DIFF
--- a/src/clj/game/core/process_actions.clj
+++ b/src/clj/game/core/process_actions.clj
@@ -1,6 +1,9 @@
 (ns game.core.process-actions
   (:require
    [clojure.string :as str]
+   [cljc.java-time.instant :as inst]
+   [cljc.java-time.duration :as duration]
+   [cljc.java-time.temporal.chrono-unit :as chrono]
    [game.core.actions :refer [click-advance click-credit click-draw click-run
                               close-deck do-purge generate-install-list
                               generate-runnable-zones move-card expend-ability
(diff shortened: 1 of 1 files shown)


NoahTheDuke:
This is debouncing, which should probably be done on the front end. We're supposed to be checking action ids with each other to avoid some of this too. Maybe we need an ":ack" sent from the server to the frontend to let the frontend send more messages?

PR #7642, opened 2024-08-12
alarm clock enforces first encounter

### LINKED ISSUE #7362: Alarm clock first-run-event seems to not work
Sample case: malandragem bypasses the first ice, then alarm clock bypasses the second

NBKelly:
We just put the req in the wrong spot. That's always going to bug me for `optionals` :weary:.

Closes #7362


NoahTheDuke:
> That's always going to bug me for optionals 😩.

tbh, the whole way optionals are written bothers me lol. Maybe we need to revisit it? Or at least add an assert to check that the :req is in the right spot.

PR #7644, opened 2024-08-12
Slee vs Ankusa return to hand test

### LINKED ISSUE #5834: Chief Slee still getting counters when broken by Ankusa

NBKelly:
Essentially, we need to update the current encounter before we end it (by bouncing the card), because slee looks at what the current encounter was when the encounter ends.

Closes #5834

### DIFF
--- a/src/clj/game/cards/programs.clj
+++ b/src/clj/game/cards/programs.clj
@@ -493,8 +493,9 @@
                                              (all-subs-broken-by-card? (:ice context) card)))
                               :msg (msg "add " (:title (:ice context))
                                         " to HQ after breaking all its subroutines")
-                              :effect (effect (move :corp (:ice context) :hand nil)
-                                              (continue :runner nil))}]}))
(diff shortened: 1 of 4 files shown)


src/clj/game/cards/programs.clj
```diff
@@ -493,8 +493,9 @@
                                              (all-subs-broken-by-card? (:ice context) card)))
                               :msg (msg "add " (:title (:ice context))
                                         " to HQ after breaking all its subroutines")
-                              :effect (effect (move :corp (:ice context) :hand nil)
-                                              (continue :runner nil))}]}))
+                              :effect (req (set-current-ice state (:ice context))
```
NoahTheDuke, line 496: How is the context ice not already the current ice? That seems to be the real source of the bug.

NBKelly: Fair. I changed `runs.clj` to update the current ice after subs get broken, and it appears to have not broken anything.

PR #7646, opened 2024-08-12
dont offer new remote as an option when at remote limit

### LINKED ISSUE #4723: Earth Station + FullImmersion Rec Studio or World Plaza
When hosting cards on other cards in the single remote server, you get an error saying you can't create new servers.

### LINKED ISSUE #4969: Vertically challenged
Hi jnet team. I was playing as Earth Station and I tried to stack something on a full immersion recstudio that was on top of a world's plaza, and the system didn't let me but lost me the click anyway. There was an error message saying I could only have 1 remote. I tried to do it the opposite way, and it also didn't let me.

### LINKED ISSUE #7503: A Teia and Earth Station should not allow users to create server beyond their limit
Asked a judge during EMEA and I got confirmation that both these ID should not allow the Corp to install a card in a new remote server if they have reached the max.

Currently there's an option to install in a new server and trash an existing one, if I remember correctly

NBKelly:
First, this makes it so "New remote" isn't offered as an option when you're at your remote server limit (earth station, a teia).

server-limit

Closes #7503

Then, I added a hosting entry to worlds plaza so it's  valid install location.

Then I also moved the selection of corp host locations out of `corp-install` and into `board`, so I could send them to the frontend as well for the menu when you click on a card in your hand. So now you can install on these without needing to interact with any manual abilities.

[2024-08-12 19-16-30.webm](https://github.com/user-attachments/assets/3e75a918-19a0-4f8a-ac1e-3accaeb3c17e)

Closes #4969
Closes #4723 (we already have tests for stacking these cards in earth station).


NBKelly:
On another look, turns out worlds plaza must host cards using the ability (I just looked at the impl and thought we were sloppy), so I updated that one back.

src/clj/game/core/installing.clj
```diff
@@ -179,11 +179,7 @@
 (defn corp-install-list
```
NoahTheDuke, line 179: Is this function needed anymore? I don't think mere aliases are worthwhile.

PR #7657, opened 2024-08-15
Ru translation v2

xiaat:
Updated some previously translated parts for formatting or clarity, and added more lines per the latest revision of the base en file. Also learned how quote marks work in Clojure


xiaat:
Ok yeah no idea why is it unhappy now, can someone please look at it

PR #7665, opened 2024-08-24
Fixed edge case overinstall unique card issue

### LINKED ISSUE #6722: Cannot overinstall unique card
Link to bug replay: https://jinteki.net/bug-report/c9499755-e2c7-4451-9289-434542ab15be?b=0

Description:

[Boomerang will not install 3rd copy]

2nd copy of boomerang will not trash while trying to install 3rd copy.

### LINKED ISSUE #7205: Using Paladin Poemu to pay for Unique cards
Link to bug replay: https://jinteki.net/bug-report/c04438e1-56ad-4ac8-abd2-0a55be92be3a?b=0

Description:

Figured this would just trash the other unique version of the card being played but paid by Paladin Poemu

### LINKED ISSUE #7464: Couldn't install a unique resource over an existing one
Link to bug replay: https://jinteki.net/bug-report/62186fcd-a2ae-47c1-b42e-caad51da4a80?b=0

Description: Attempting to install a third class act over the second one.

### LINKED ISSUE #7662: Overinstall Boomerang caused an error
Link to bug replay: https://jinteki.net/bug-report/322e261e-0e2b-4db8-bb10-fb02cdc67226?b=0

Description:

Tried to overinstall a (stranded) Boomerang with a second one against Pradivost on a HQ ICE whilst taking a credit from Paladin Poemu when the error message appeared in the top right and it would not take the credit or install the Boomrang. I tried undoing and not taking the credit from Poemu but the error message still appeared. Using /trash to manually trash the Boomerang let me install the new one.

NBKelly:
I was having a lot of trouble actually recreating the bug on any simplified version of the gamestate, so I recreated the whole game from #7662 as a unit test (praying I never need to do this again), and when I did that, I could recreate the issue.

After a bit of fiddling, I think I actually tracked it down :eyes:. In `moving.clj`, like 427, we would sometimes be doing `(swap! state assoc-in [nil :register :trashed-card] true)` when overinstalling (why this is only sometimes happening, I'm not sure). I changed it to only update trash-lists if the side is not nil, and now this case works.

I still don't know why this would only happen *sometimes*, and not *other times*, but at the very least this hits the symptoms of the issue.

Useful things I added to our test framework:
* do-trash-prompt (I was sick of guess-and-checking what the trash button actually looked like)
* `is-hand? [state side expected-hand]` (Is the current hand the expected hand)

Tackles the following:
Closes #7662
Closes #7464
Closes #7205
Closes #6722

### DIFF
--- a/src/clj/game/core/moving.clj
+++ b/src/clj/game/core/moving.clj
@@ -423,7 +423,7 @@
                                    []
                                    trashlist)]
                  (swap! state update-in [:trash :trash-list] dissoc eid)
-                 (when (seq (remove #{side} (map #(to-keyword (:side %)) trashlist)))
+                 (when (and side (seq (remove #{side} (map #(to-keyword (:side %)) trashlist))))
                    (swap! state assoc-in [side :register :trashed-card] true))
                  ;; Pseudo-shuffle archives. Keeps seen cards in play order and shuffles unseen cards.
                  (swap! state assoc-in [:corp :discard]
--- a/test/clj/game/core/scenarios_test.clj
+++ b/test/clj/game/core/scenarios_test.clj
@@ -42,6 +42,347 @@
         (is (no-prompt? state :runner) "No lingering prompt")
         (is (= 1 (count (:discard (get-runner)))))))))
 
+(deftest masterwork-overinstall-boomerang-complex-case-full-game
+  (testing "for issue #7662"
+    (do-game
+      (new-game {:corp {:hand [(qty "Tree Line" 2) "Rashida Jaheem"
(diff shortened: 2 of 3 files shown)


NoahTheDuke:
To find out at side is nil, do something like (when (nil? side) (println (ex-info "" {}))) at that same line and just trace it back to the point of origin. That should be the underlying problem. We shouldn't be moving cards without a side.

NBKelly:
In this case, the card is trashed by a game action (the checkpoint cleanup) and not a player, so the side being nil is correct.

NoahTheDuke:
Glad you found the issue.

PR #7669, opened 2024-08-24
Don't process abilities while a prompt is open

### LINKED ISSUE #5705: Feature idea: Do not allow cards to resolve abilities while a prompt of that card is open
There often are situations, where players open a second prompt, while still resolving the first on the same card (e.g. trying to break a subroutine, while already being in the prompt for breaking subroutines with the same card). We could disable resolving card abilities on cards, while a prompt of that card is still open.

### LINKED ISSUE #5776: Suggestion:  Stopping the Game while the Corp is acting

### LINKED ISSUE #6923: [BUG] Broke interface
Link to bug replay: https://jinteki.net/bug-report/8f219553-006d-4239-a97f-b4f95b364b3c?b=0

Description:

Somehow broke the interface. Only "strange move" / unexpected actions is boosting black orchestra during access using trickster taka around turn 8 (not remembering the exact turn)

### LINKED ISSUE #6930: [BUG] (Minor) Internal server error with unexpected actions from the player

### LINKED ISSUE #7198: "bblum uses Stegodon MK IV to shuffle Spin Doctor and 1 unseen card into R&D."

NBKelly:
Also tags prevention prompts so we can check against them being open.

Prevents players from using card abilities, firing subroutines, expending, or playing dynamic abilities while a prompt is open (another ability is resolving). Actions already block, so there wasn't a need to touch those.

I fixed the unit tests that deliberately played the game wrong, since it's no longer possible to do that.

Closes #5705
Closes #6930 (no longer possible)
Closes #6923 (prevents the scenario where it would happen)
Closes #7198 (prevents overlapping prompts)
Closes #5776 (prevents acting while a prompt is open)


src/clj/game/core/actions.clj
```diff
@@ -37,6 +37,19 @@
...
       (swap! state assoc :click-states click-states))))
 
+(defn- valid-prompt-state?
+  [state side]
+  (let [prompt-type (get-in @state [side :prompt-state :prompt-type])]
+    (or (= nil prompt-type)
+        (= :run prompt-type)
+        (= :prevent prompt-type))))
+
+(defn- no-prompt?
```
NoahTheDuke, line 47: This implementation does not check  that there's no prompt, so the name should change or the impl should change.

src/clj/game/core/actions.clj
```diff
@@ -73,7 +88,8 @@
   (let [card (get-card state card)
         eid (make-eid state {:source card :source-type :ability})
         expend-ab (expend (:expend card))]
-    (resolve-ability state side eid expend-ab card nil)))
+    (when (no-prompt? state side)
```
NoahTheDuke, line 91: Move this before the let.

PR #7673, opened 2024-08-25
Choose one helper function

### LINKED ISSUE #7404: Business As Usual did not resolve both modes when threat was active
Link to bug replay: https://jinteki.net/bug-report/b3da9467-ca76-4836-9474-2f14e4d7b8d6?b=1

Description:

[Business As Usual](https://netrunnerdb.com/en/card/34128) still prompted only a single choice, even through threat was active.

### LINKED ISSUE #7671: [Log] Wildcat Strike should maybe say that the Corp player makes a decision
image
IMHO, this should have an extra line between these two saying "\<Corp Player\> chooses to let the runner [gain / draw]" so that it's clear the other player was involved in resolving this effect (compare other interactive effects like psi games). Could be a help when scanning logs to track each players' decisions.

NBKelly:
I imported the choose-one helper that I used as the backbones for a lot of playtesting cards.

First, I made the following engine changes:
* You can pass a `:display-side` key in as part of an ability. This is the side that is shown as the source in `display-message` in engine
* Additionally, you can simply set the `:msg` key to `:cost`, when the only purpose of the message is that a cost has been paid. This displays a message like: `X pays Y to satisfy [card name]`.
public-trail_000

How does the helper work?
* Takes (optionally) a set of arguments (more later on this), and then a list of options
* Creates a contextual prompt based on the given options, handling the following things for free:
  *  Only display options that are payable and pass their req test
  * Allow selecting multiple options (cards like deuces wild), with or without pruning (can you pick the same option more than once? ie fullyop)
  * Automatically match chosen options to resulting functions (so no more bugs due to us mis-spelling the option in the effect)
  * Option to automatically terminate if the only possible choice is "Done" (don't show irrelevant prompts)
  * Automatically display a count of the remaining choices when multiple are selected
  * Has a default prompt ("choose one"), so we don't need that redundant line on every other card
  * Can pass in a 5-fn as the count and have it "just work"

I changed the following cards. I could change a whole lot more, but I don't want to fix what's not broken, so I'll only touch stuff if there are issues or bugs:
* Business As Usual (the modal runs nicely through the helper, and isn't cobbled together anymore) - Closes #7404
* Public Trail - just a proof of concept
* Wildcat Strike logs the corp involvement - Closes #7671


src/clj/game/core/def_helpers.clj
```diff
@@ -255,6 +256,94 @@
...
+           ;; converts options to choices
+           choices-fn (fn [x state side eid card targets]
+                        (when (payable? x state side eid card targets)
+                          (if-not (:req x)
+                            (costed-str x)
+                            (when ((:req x) state side eid card targets)
+                              (costed-str x)))))
+           ;; this lets us selectively skip the prompt if 'done' is the only choice
+           meaningful-req? (when require-meaningful-choice
+                             (req (and (let [cs (seq (remove nil? (map #(choices-fn
```
NoahTheDuke, line 300: (remove nil? (map ...)) is just (keep ...)

src/clj/game/core/def_helpers.clj
```diff
@@ -255,6 +256,94 @@
...
+                                                                          % state side
+                                                                          eid card targets) xs)))]
+                                         (and (not= cs '("Done"))
```
NoahTheDuke, line 303:
```suggestion
                                         (and (not= cs ["Done"])
```

NoahTheDuke:
i gotta look at this on my computer but those stood out to me

NoahTheDuke:
Having thought about it some more, can `:does` be `:ability`? I want to keep some consistency with our DSLs. `successful-run-replace-breach` uses `:ability`, and it should be obvious that this key is an ability map, not a `req`-style 5-fn.

PR #7677, opened 2024-08-25
prevent accessing only card to access if it wasn't already a candidate

### LINKED ISSUE #7675: [Bug] Adrian Seis interacts poorly with breach access replacement effects
Replay link here: https://jinteki.net/replay/614918ac-cd67-44d1-86fa-f69d578d7f42

Was chatting with CephalopodWizard about this on GLC. When Adrian is protecting a server (like Archives) and a run is initiated on it using a breach replacement effect like Pinhole, Sneakdoor, or Eru, then depending on the results of the psi game played with Adrian, the runner may not be legally able to access any other cards. This would effectively cause Eru/Sneakdoor/Pinhole to fizzle out. 

I tested this online, and found that the behavior is not properly implemented

Pinhole Threading: When a run on a server protected by Adrian is successful, the psi game results when differing should prevent the runner from effectively accessing anything since Adrian is not accessible by being in the remote being run on, and the psi game prevents the runner from accessing anything other than Adrian.

Eru: Similarly, Adrian protects archives, whereas Eru breaches R&D. In this case, when the psi game resolves so that only Adrian can be accessed, the runner is able to access/trash Adrian in game, but shouldn't be able to because Adrian is installed in Archives, but Eru makes the runner breach R&D. 

Expected behavior: in both cases the run should "fizzle" instead.

NBKelly:
This card has too many cursed interactions, but at least it's future-proofing for any similar worded cards in the future.

Closes #7675

### DIFF
--- a/src/clj/game/core/access.clj
+++ b/src/clj/game/core/access.clj
@@ -2,7 +2,7 @@
   (:require
     [game.core.agendas :refer [update-all-advancement-requirements update-all-agenda-points]]
     [game.core.board :refer [all-active]]
-    [game.core.card :refer [agenda? condition-counter? corp? get-agenda-points get-card get-zone in-discard? in-hand? in-scored? operation? rezzed?]]
+    [game.core.card :refer [agenda? condition-counter? corp? get-agenda-points get-card get-zone in-archives-root? in-deck? in-discard? in-hand? in-hq-root? in-remote-root? in-rd-root? in-scored? operation? rezzed?]]
     [game.core.card-defs :refer [card-def]]
     [game.core.cost-fns :refer [card-ability-cost trash-cost steal-cost]]
     [game.core.effects :refer [any-effects register-static-abilities register-lingering-effect sum-effects unregister-lingering-effects]]
@@ -560,13 +560,12 @@
                                  (pos? (+ max-access total-mod))
                                  true)
                       pos-total-cards? (pos? (+ total-cards-count total-mod))]
-
                   (cond
                     ;; Only 1 card to access
                     (and pos-max?
                          pos-total-cards?
                          only-card)
-                    (if (= (second (get-zone only-card)) (first server))
+                    (if (in-remote-root? only-card (first server))
                       (access-card state side eid only-card)
                       (effect-completed state side eid))
 
@@ -753,7 +752,7 @@
                     (and pos-max?
                          pos-total-cards?
                          only-card)
-                    (if (= (second (get-zone only-card)) :rd)
+                    (if (or (in-deck? only-card) (in-rd-root? only-card))
                       (access-card state side eid only-card)
                       (effect-completed state side eid))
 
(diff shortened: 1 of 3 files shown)


src/clj/game/core/access.clj
```diff
@@ -563,7 +566,9 @@
                     (and pos-max?
                          pos-total-cards?
                          only-card)
-                    (access-card state side eid only-card)
+                    (if (= (second (get-zone only-card)) (first server))
+                      (access-card state side eid only-card)
+                      (effect-completed state side eid))
```
NoahTheDuke, line 571: Are these and the similar lines necessary?

NBKelly: Yeah - if the only card you're allowed to access isn't actually on the server you're breaching (ie ash on HQ, beatriz on HQ to breach R&D, losing the ash trace), then we shouldn't be accessing that card (the ash) - since it's not already a candidate for this access.

NBKelly: wait, yes they are necessary, since this function is picking the actual card to access

NoahTheDuke: Okay, I see the point now.

src/clj/game/core/access.clj
```diff
@@ -748,7 +753,9 @@
                     (and pos-max?
                          pos-total-cards?
                          only-card)
-                    (access-card state side eid (first total-cards))
+                    (if (= (second (get-zone only-card)) :rd)
```
NoahTheDuke, line 756: This disallows choosing the card in root, which you can do if you're accessing a central. The list of candidates for centrals includes all cards in root and a set number of cards from the central itself. If a card is in R&D itself, the zone will be `[:deck]`.

NBKelly: Hmm. There aren't any effects that set the only card to access to be something that's not on the field, but it's probably still worth cleaning this up.

NoahTheDuke: I think changing it to be something like `(if (or (in-deck? only-card) (in-rd-root? only-card)) ...)` would be fine, and then the same for the other checks.

NoahTheDuke: (I also prefer that we use the card.cljc predicate functions instead of looking at the contents of `:zone`, cuz it's semantic and allows us to change it if we ever decide to.

NBKelly: Sick, did that. I added a helper in card.cljc for `in-remote-root?` too, to be consistent.

src/cljc/game/core/card.cljc
```diff
@@ -123,6 +123,13 @@
   (= (get-zone card) #?(:clj [:servers :rd :content]
                         :cljs ["servers" "rd" "content"])))
 
+(defn in-remote-root?
+  [card remote]
+  (let [remote-key (if (keyword? remote) remote (keyword remote))
+        remote-str (if (keyword? remote) (name remote) remote)]
+    (= (get-zone card) #?(:clj [:servers remote-key :content]
+                          :cljs ["servers" remote-str "content"]))))
```
NoahTheDuke, line 131:
```suggestion
  (let [remote #?(:clj remote
                  :cljs (if (keyword? remote) (name remote) remote))]
    (= (get-zone card) #?(:clj [:servers remote :content]
                          :cljs ["servers" remote "content"]))))
```

NoahTheDuke: There's never any time the server will have a string version of the zone, and it will be rare that we pass a keyword literal in the client.

PR #7683, opened 2024-08-27
Ice unit testing suite

### LINKED ISSUE #7682: Loot Box doesn't trash itself if the runner's stack is 0
I run with 0 stack against it and my opponent had to trash it manually. I don't know if this bug only appears when you're already at 0 stack or if it's enough to be at <3.

Replay: https://jinteki.net/replay/863ff9f6-bf9d-4edf-b16e-c1c3518d5a90
Screenshot: loot box bug

NBKelly:
I ported over the test suite I was using for ice on from playtest.

Here's my rationale for this:
>    Our unit tests for ice are mostly extremely bloated, and a lot more trouble to write than they should be.
>    Testing, for example, that vanilla ends the run should be as simple as:
>    `(do-game (fire-subroutine-test "Ice Wall" 0) (is (not (:run @state)))`
>    The fact that the corporation had to install vanilla, and then click for credits, and then have the runner run on the server vanilla is on, and then the corp needs to rez vanilla, is all essentially "noise" - they're things that happen in a game of netrunner, but they aren't what we should be actively trying to unit test when we want to know if vanilla ends the run.

If you scan through the unit tests for our ice, you'll actually notice that very few subroutines are tested, and it's primarily just the unique effects (on-rez, on-encounter) that are tested.

Even then, those are more bloated that they should be. Testing tollbooth's on-encounter ability, for example, should be as simple as this:
```clojure
(do-game 
  (run-and-encounter-ice-test "Toolbooth" {:runner {:credits 2}})
  (is (not (:run @state)) "Tollbooth ended the run"))
(do-game
  (run-and-encounter-ice-test "Toolbooth" {:runner {:credits 3}})
  (is (changed? [(:credit (get-runner)) 3]
        (click-prompt state :runner "Yes"))
      "Paid 3")
  (is (:run @state) "Tollbooth has not ended the run"))
```

I also added on the fire all subroutines, since that seems like a reasonable use case.

I put these in test-framework, rather than directly in the ice test file, because they will be useful for programs in the future as well (since they can have on-encounter abilities, or to directly get into specific ice encounters with specific programs).

The mini-framework supports the `:rig` key in it's options, which will automatically install whatever you put in it (if the program requires interaction to install, like Atman, then that's not supported), `:counters` to set the counters on the card to whatever you like, `threat` to specify the threat level, `disable` to rez the card in a disabled state (to avoid on-rez effects), and a `server` key to specify which server to put the ice on in particular (for positional ice, like bloom or afshar - default is HQ).

Also, this is a bit superflous, but if (ie for net damage subroutines) the only thing important is the count of cards in the hand (or deck, or whatever), then I felt like specifying the names for those cards was redundant, so I worked in a feature to treat `:deck X` as `:deck [(quantity replacement-card X)]`. It seems silly now, but it looks really clean in unit tests (trust me on this for later).

I'm not going to rewrite a bunch of our unit tests because it's not worth it, but I do want to use things like this for most of our future work.

Also, I fixed this interaction: Closes #7682, and I converted the happy path test to use fire-all-subs-test just as a proof of concept (and because it seemed over-engineered a little).

I also noticed that in test-framework, if you were to explicitly set your deck as `[]` (or empty), you would still receive 3 cards, because the result of `(transform (:deck ...))` was nil, so I added `:deck side` as an alternative case, which means you can specify an empty deck. I fixed the two unit tests that got caught out by this.

Then I noticed that (quantity x 0) actually gives `nil`, so `:deck [(quantity x 0)]` resolves to `:deck [nil]`, which was causing issues in exactly one unit test, so I updated that unit test.


test/clj/game/test_framework.clj
```diff
@@ -976,6 +979,126 @@
...
+         state (new-game players)
+         server (or server "HQ")
+         server-key (cond
+                      (= server "New remote") :remote1
+                      (= server "HQ") :hq
+                      (= server "R&D") :rd
+                      (= server "Archives") :archives)]
+     ;; adjust the threat level for threat: ... subs
+     (when threat
+       (do (game.core.change-vals/change
```
NoahTheDuke, line 1023: `when` has an implicit do, don't need it here.

test/clj/game/test_framework.clj
```diff
@@ -976,6 +979,126 @@
...
+       state))))
+
+(defn subroutine-test
+  ([card sub] (subroutine-test card sub nil))
+  ([card sub players] (subroutine-test card sub players nil))
+  ([card sub players {:keys [server] :as args}]
+   "Create a game, install an ice, encounter it, then fire a subroutine.
+    Optionally specify: player map, target server, any number or type of counters,
+    if the card should be disabled while rezzed, tags, threat level,
+    cards installed in the runner rig"
```
NoahTheDuke, line 1065: Docstrings go before the function bodies.

test/clj/game/test_framework.clj
```diff
@@ -976,6 +979,126 @@
...
+                (= :last sub) (dec (count (:subroutines (get-ice state server-key 0))))
+                :else sub)]
+     (card-subroutine state :corp ice sub)
+     state)))
+
+(defn fire-all-subs-test
+  ([card] (fire-all-subs-test card nil))
+  ([card players] (fire-all-subs-test card players nil))
+  ([card players {:keys [server] :as args}]
+   "Create a game, install an ice, encounter it, then fire all subroutines.
```
NoahTheDuke, line 1087: Same here.

PR #7684, opened 2024-08-27
Added ffg's unreleased identities

### LINKED ISSUE #4476: some extra casual identities for fun
For some fun casual play could we get the the two identities that were never released added?

https://ancur.fandom.com/wiki/The_Collective
https://ancur.fandom.com/wiki/Chronos_Protocol,_Haas-Bioroid.

### LINKED ISSUE #7660: Ob and Wage Workers interaction is misstimed
If Ob plays an operation such as extract 3rd click and fetches a Wage Workers, it should be active in time to see its trigger and give the Corp their click (assuming it is the 3rd operation) . This does not currently happen on Jnet.

NBKelly:
I thought it would be fun.

Note that these tests are gonna fail until netrunner-data is updated.

I added a PR in netrunner-data that inserts these, and they aren't legal for any formats except casual.
See: https://github.com/NoahTheDuke/netrunner-data/pull/54

Closes #4476

Additionally, I changed `play-ability` to be more aware of actions, and when an action is taken, it fires an event for pre-resolution and post-resolution (this was part of the collective).

Having done that, making wage workers behave right was pretty easy too, so I fixed that: Closes #7660

### DIFF
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -3047,13 +3047,16 @@
 
 (defcard "Wage Workers"
   (let [payoff {:msg "gain [Click]"
-                :effect (effect (gain-clicks 1))}]
+                :effect (effect (gain-clicks 1))}
+        relevant-keys (fn [context] {:cid (get-in context [:card :cid])
+                                     :idx (:ability-idx context)})]
     {:events [{:event :action-resolved
                :req (req (= :corp side))
                :async true
                :effect (req (let [similar-actions
                                   (event-count state side :action-resolved
-                                               (fn [[ctx]] (= ctx context)))]
+                                               (fn [[ctx]] (= (relevant-keys ctx)
+                                                              (relevant-keys context))))]
                               (continue-ability
                                 state side
                                 (when (= 3 similar-actions) payoff)
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -3047,18 +3047,17 @@
 
 (defcard "Wage Workers"
   (let [payoff {:msg "gain [Click]"
-                :effect (effect (gain-clicks 1))}
-        three-of (fn [state side cid idx]
-                   (= 3 (event-count state side :corp-spent-click
-                                     (fn [[context]]
-                                       (and (= cid (:action context))
-                                            (= idx (:ability-idx context)))))))]
-    {:events [{:event :corp-spent-click
+                :effect (effect (gain-clicks 1))}]
+    {:events [{:event :action-resolved
+               :req (req (= :corp side))
                :async true
-               :effect (req (let [{:keys [action ability-idx]} context]
-                              (if (three-of state side action ability-idx)
-                                (continue-ability state side payoff card nil)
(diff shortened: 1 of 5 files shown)


src/clj/game/cards/assets.clj
```diff
@@ -3047,18 +3047,17 @@
...
+    {:events [{:event :action-resolved
+               :req (req (= :corp side))
                :async true
-               :effect (req (let [{:keys [action ability-idx]} context]
-                              (if (three-of state side action ability-idx)
-                                (continue-ability state side payoff card nil)
-                                (effect-completed state side eid))))}]}))
+               :effect (req (let [similar-actions
+                                  (event-count state side :action-resolved
+                                               (fn [[ctx]] (= ctx context)))]
```
NoahTheDuke, line 3056: This will fail if we ever change the context object for this event. Better to check some subset or a specific key.

NoahTheDuke: Clojure loves open maps.

src/clj/game/cards/identities.clj
```diff
@@ -2109,6 +2121,27 @@
...
+  (let [gain-click-abi {:label "Manually gain [Click]"
+                        :once :per-turn
+                        :msg (msg "gain [Click]")
+                        :effect (req (gain-clicks state side 1))}]
+    {:events [{:event :action-resolved
+               :req (req (= :runner side))
+               :silent (req true)
+               :effect (req (let [current-queue (get-in card [:special :previous-actions])]
+                              (if (and (seq current-queue)
+                                       (= (first current-queue) context))
```
NoahTheDuke, line 2134: Same here.

src/clj/game/cards/identities.clj
```diff
@@ -491,6 +491,18 @@
...
+  {:events [{:event :damage
+             :req (req (= (:damage-type context) :brain))
+             :msg (msg "remove all copies of " (enumerate-str (map :title (:cards-trashed context))) ", everywhere, from the game")
+             :effect (req (doseq [c (:cards-trashed context)]
+                            (let [all-candidates (filter #(and (not= (:zone %) [:rfg])
+                                                               (runner? %)
+                                                               (= (:title %) (:title c)))
+                                                         (get-all-cards state))]
+                              (doseq [candidate all-candidates]
+                                (move state :runner candidate :rfg)))))}]})
```
NoahTheDuke, line 504: This can actually be one `doseq`:

```clojure
(defcard "Chronos Protocol: Haas-Bioroid"
  {:events [{:event :damage
             :req (req (= (:damage-type context) :brain))
             :msg (msg "remove all copies of " (enumerate-str (map :title (:cards-trashed context))) ", everywhere, from the game")
             :effect (req (doseq [c (:cards-trashed context)
                                  :let [all-candidates (filter #(and (not (in-rfg? %))
                                                                     (runner? %)
                                                                     (= (:title %) (:title c)))
                                                               (get-all-cards state))]
                                  candidate all-candidates]
                            (move state :runner candidate :rfg)))}]})
```

The reason to do it this way is that it produces faster bytecode.

NBKelly: neat

PR #7696, opened 2024-08-29
Hellion beta test correctness

### LINKED ISSUE #6177: Hellion Beta Test activating any time the runner trashes a corp card, rather than only when they trash an accessed card
The text on Hellion Beta Test allows it to be used only when the runner has trashed an accessed corp card in the previous turn. Currently, the corp is allowed to play Hellion Beta Test even if the trashed card was not accessed. Specifically, the trash effects from both Hippo and Stargate allow the corp to use Hellion Beta Test on the following turn, even if the runner trashed no accessed cards.

NBKelly:
Using a turn register for this seemed the most reasonable option.

Closes #6177

### DIFF
--- a/src/clj/game/core/access.clj
+++ b/src/clj/game/core/access.clj
@@ -145,6 +145,7 @@
                                         (when must-trash?
                                           (swap! state assoc-in [:run :did-access] true)))
                                       (swap! state assoc-in [:runner :register :trashed-card] true)
+                                      (swap! state assoc-in [:runner :register :trashed-accessed-card] true)
                                       (system-msg state side (str (:msg async-result) " to trash "
                                                                   (:title card) " from "
                                                                   (name-zone :corp (get-zone card))))
@@ -156,6 +157,8 @@
                           (let [ability-card (find-first #(same-card? % target) access-ab-cards)
                                 ability-eid (assoc eid :source ability-card :source-type :ability)
                                 ability (access-ab ability-card)]
+                            (when (:trash? ability true)
+                              (swap! state assoc-in [:runner :register :trashed-accessed-card] true))
                             (when (and (:breach @state)
                                        (:trash? ability true))
                               (swap! state assoc-in [:breach :did-trash] true))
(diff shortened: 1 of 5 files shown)


test/clj/game/cards/operations_test.clj
```diff
@@ -2246,6 +2246,20 @@
...
+               :runner {:deck ["Stargate"]}})
+    (take-credits state :corp)
+    (play-from-hand state :runner "Stargate")
+    (card-ability state :runner (get-program state 0) 0)
+    (run-continue state)
+    (click-prompt state :runner "Vanilla")
+    (take-credits state :corp)
+    (play-from-hand state :corp "Hellion Beta Test")
+    (is (is-hand? state :corp ["Hellion Beta Test"])
+        "Did not play because the restriction was not me")))
```
NoahTheDuke, line 2261:
```suggestion
        "Did not play because the restriction was not met")))
```

NoahTheDuke:
I think this should be checking the turn events for `:runner-trash` where `(:accessed context)`. Cuz right now, you're missing all of the `:access-ability` that trash the cards (such as Imp).

src/clj/game/core/access.clj
```diff
@@ -156,6 +157,8 @@
                           (let [ability-card (find-first #(same-card? % target) access-ab-cards)
                                 ability-eid (assoc eid :source ability-card :source-type :ability)
                                 ability (access-ab ability-card)]
+                            (when (:trash? ability true)
+                              (swap! state assoc-in [:runner :register :trashed-accessed-card] true))
```
NBKelly, line 161: This should catch runner trashing with access abilities

NBKelly: I'll add a unit test though, just to be sure

NoahTheDuke:
I'm sorry, let me be more direct: This should use the `turn-events` system instead of adding another entry to the register. If there are abilities that are trashing during access and not marking `{:access true}` in `trash`, then we gotta fix those.

NBKelly:
Fair enough. Looking through our code, it doesn't appear we actually preserve turn-events across turns (or I just missed it).
Should I update us to do that, and do away with all of our last-turn register stuff altogether?

NoahTheDuke:
nvm i'm dumb and this is good.

PR #7703, opened 2024-08-31
Gateway precon automation

### LINKED ISSUE #5720: Distinguish between System Gateway games
It should be pretty forward to make it so we can distinguish between types of System Gateway games:

- System Gateway [starter deck] (either specific starter decks or decks with 30/34 cards and the neutral identities)
- System Gateway [booster deck] (either specific booster deck, or cards from SG using neutral identities)
- System Gateway (the rest where whole SG card pool is legal

There seem to be some confusion especially when creating the games (title) and decks are valid so you dont discover it until after game is started :)

### LINKED ISSUE #7681: Divide the Gateway format into "Starter Decks" and "Gateway Constructed"
Right now beginners can hop into another beginner's System Gateway lobby and while one of them expects Starter or Booster decks the other expects to play Gateway Constructed. This could be solved by talking it out but most beginners don't know this, and this causes a lot of friction.

Ideally the "System Gateway" Lobby only allows for the starter decks to be picked and the "Gateway Constructed" lobby allows for all cards in SG but the two learning IDs.

NBKelly:
I made it so you can start games with the gateway precon decks (beginner/intermediate), or in constructed, when choosing the gateway format.

Closes #5720
Closes #7681

Here's a video (this is before I finished everything, so the text still has some typos):

[2024-08-31 19-36-09.webm](https://github.com/user-attachments/assets/03e641c2-6f10-475f-a38e-632a79a4a08d)

* When creating a gateway game, there's a pop-up radio-menu that lets you choose beginner, intermediate or constructed
* Constructed is what we already have
* Beginner and intermediate will remove the deck-picking functionality from the lobby. Instead, when a game is started, the players will load in with the starter deck for the selected side and complexity (beginner/intermediate)

Additionally, because it was bugging me:
* The format in the game-row (the thing you see in the lobby) now says "format", instead of default format. That's the one you're selecting.
* The one in your options page now says "default format", instead of format.
idk why these keys got mixed up, but it looks ugly to see "Default game format" take up half the game-row box on 15 different games. It makes the actual format chosen more difficult to spot in comparison.

### DIFF
--- a/src/clj/web/preconstructed.clj
+++ b/src/clj/web/preconstructed.clj
@@ -1,0 +1,96 @@
+(ns web.preconstructed)
+
+(def gateway-beginner-corp
+  {:format "system-gateway"
+   :identity {:title "The Syndicate: Profit over Principle" :side "Corp" :code 30077}
+   :name "System Gateway Starter Corp"
+   :cards [{:qty 3 :card "Offworld Office"}
+           {:qty 2 :card "Send a Message"}
+           {:qty 2 :card "Superconducting Hub"}
+           {:qty 2 :card "Nico Campaign"}
+           {:qty 2 :card "Regolith Mining License"}
+           {:qty 2 :card "Urtica Cipher"}
+           {:qty 2 :card "Government Subsidy"}
+           {:qty 3 :card "Hedge Fund"}
+           {:qty 2 :card "Seamless Launch"}
+           {:qty 1 :card "Manegarm Skunkworks"}
+           {:qty 2 :card "Brân 1.0"}
+           {:qty 3 :card "Palisade"}
+           {:qty 2 :card "Diviner"}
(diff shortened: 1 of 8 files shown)


NBKelly:
Oh, and admittedly, there's probably a ton of stuff in here I did *the wrong way*. This one was a real puzzle to figure out, and I probably jammed every other shape through the diamond shaped hole.

src/clj/web/preconstructed.clj
```diff
@@ -0,0 +1,96 @@
+(ns web.preconstructed)
```
NoahTheDuke, line 1: This is way better than the current solution of pulling from the database. We should put the existing preconstructed decks in here.

NoahTheDuke:
This looks great. I bet there's ways to generalize it, like ":deck-type" instead of "gateway-type", but we can do that when we need it.

PR #7709, opened 2024-09-03
Gatekeeper makes runner wait

### LINKED ISSUE #7708: Make the runner wait while Gatekeeper is resolving.

NBKelly:
I added a waiting prompt to gatekeeper and a unit test for it.

Then I added a `waiting?` function to the test framework, because I thought it should be straight-forward and look nice syntactically.

Then I updated the half-dozen different flavors we had of writing "am I in a waiting prompt" to all just call `(waiting? state side)`

Closes #7708


Larrea-Mon:
chefkiss

PR #7710, opened 2024-09-03
Feature: Precon deck format

NBKelly:
I refactored the gateway thing from the last patch, and expanded it a bit so it allows for the following:
* We can define matchups (a set of two decklists and the fluff associated with them)
* The frontend is able to pick up that one of these is defined, and list it as an option in the preconstructed format
* Right I've filled it with the matchups for all of the worlds events (2012-present)

### Why?
I think it's fun. It's a good way to preserve (some of) the history of this game, and it's a good way to organize games without worrying about deckbuilding or legality at all. It's also nice to step into the shoes of some champions or would-be champions and see how you perform in the same matchups (or if the side selection was reversed, the matchups that could have been).

I really liked when dota did this for the TI games a few years ago.

[worlds-precon-format.webm](https://github.com/user-attachments/assets/40c74652-7cd6-4323-af05-9323a7c884a8)


NBKelly:
OK, I think this is "complete" for now.

Here's another video:
[precon-mode-complete.webm](https://github.com/user-attachments/assets/550f3eb8-9b18-4ec2-b2f1-8db37f37e2fc)

I added a little more info, so you know what you're getting yourself into when you click on the lobby and when you create it.
decks_000
decks

src/cljc/jinteki/preconstructed.cljc
```diff
@@ -0,0 +1,1520 @@
...
+           {:qty 2 :card "Mayfly"}
+           {:qty 2 :card "Unity"}]})
+
+(def system-gateway-beginner
+  (matchup
+    [:lobby.gateway-format.beginner "Beginner"]
+    [:lobby.gateway-format.beginner "Beginner"]
+    [:lobby.gateway-format.beginner-info
+     "This lobby is using the System Gateway beginner decks for the Corporation and Runner. These decks are recommended for your first games. Games are played to 6 agenda points."]
+    [:lobby.gateway-format-beginner-ul "System Gateway - Beginner Teaching Decks"]
```
NoahTheDuke, line 123: None of these `-ul` are used anywhere or defined anywhere, and they're all duplicating other info. What's the purpose?

src/cljc/jinteki/preconstructed.cljc
```diff
@@ -0,0 +1,1520 @@
...
+           {:qty 1 :card "Verbal Plasticity"}
+           {:qty 2 :card "Carmen"}
+           {:qty 2 :card "Cleaver"}
+           {:qty 2 :card "Mayfly"}
+           {:qty 2 :card "Unity"}]})
+
+(def system-gateway-beginner
+  (matchup
+    [:lobby.gateway-format.beginner "Beginner"]
+    [:lobby.gateway-format.beginner "Beginner"]
```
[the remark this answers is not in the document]
NoahTheDuke: jk, these are calls to matchup

src/cljc/jinteki/preconstructed.cljc
NoahTheDuke: This and the others are also `:lobby.gateway-format-beginner-ul` instead of matching the other keys with `:lobby.gateway-format.beginner-ul`.

NoahTheDuke:
I'm merging cuz I want to use the precons in testing, but I think that the issues I mentioned should be handled at some point.

PR #7712, opened 2024-09-05
Minor cleanup of some ice

### LINKED ISSUE #7711: Hammer contains Typo + wording
Hammer's second subroutine reads "Choose a resource of piece of hardware". 
Also I don't know if jnet has a style guide but I think it should mention "trash" somewhere instead of just "choose". Just pure speculation on my part.

NBKelly:
We have a `gain-power-counter` function, so I made mamba and quicksand use it (I changed it to gain power counter so it was clear who was gaining the counter).

I made all the ice with plain "runner loses X" subroutines just call the "runner loses X subroutine" function, instead of redefining it 10 times.

I made blockchain and anvil point to the same "corp gains, runner loses" subroutine.

I added tests for quicksand and mamba because they had none.

And then I fixed the wording on Hammer as per #7711 

Closes #7711

### DIFF
--- a/src/clj/game/cards/ice.clj
+++ b/src/clj/game/cards/ice.clj
@@ -253,11 +253,12 @@
    :async true
    :effect (effect (gain-tags :corp eid n))})
 
-(def add-power-counter
+(def gain-power-counter
   "Places 1 power counter on a card."
   {:label "Place 1 power counter"
    :msg "place 1 power counter on itself"
-   :effect (effect (add-counter card :power 1))})
+   :async true
+   :effect (req (add-counter state side eid card :power 1 {:placed true}))})
 
 (defn trace-ability
   "Run a trace with specified base strength.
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/ice.clj
```diff
@@ -664,11 +673,7 @@
                                                               :value true
                                                               :duration :end-of-encounter}))}}})]
     {:on-encounter (encounter-ab)
-     :subroutines[{:label "Gain 1 [Credits], Runner loses 1 [Credits]"
-                   :msg "gain 1 [Credits] and force the Runner to lose 1 [Credits]"
-                   :async true
-                   :effect (req (wait-for (gain-credits state :corp 1)
-                                          (lose-credits state :runner eid 1)))}
+     :subroutines[(corps-gains-and-runner-loses-credits 1 1)
```
NoahTheDuke, line 676: spacing plz

NoahTheDuke, line 676:
```suggestion
     :subroutines [(corps-gains-and-runner-loses-credits 1 1)
```

PR #7715, opened 2024-09-07
dont duplicate cards trashed during checkpoints

NBKelly:
Uniques and consoles are trashed by the game during checkpoints.
* When they host cards, we first trash those cards
* When those cards get trashed, they each have a checkpoint
* Uniques and consoles are trashed by the game during those checkpoints
* We end up with X+1 copies of the host in the trash

So I'm suppressing the trash-events for the hosted cards until the same time as the host dies. I think under the CR, hosted cards get trashed *after* the host does (this is literally only relevant for the draft anarch id, you can't instantly recur your parasites), so we're at least a step closer to that.

### DIFF
--- a/src/clj/game/core/moving.clj
+++ b/src/clj/game/core/moving.clj
@@ -57,6 +57,7 @@
                                     :trash
                                     (update h :zone #(map to-keyword %))
                                     {:unpreventable true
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
nice find

PR #7721, opened 2024-09-09
Fenris and rebirth work for precon format

NBKelly:
I'm not sure if this is the right way to tackle this, I didn't think through these two cursed cards.

We keep old mwl info, but we don't keep old card pool info unfortunately, so we can't look through a history of that.

Happy to try out whatever else though.

### DIFF
--- a/src/clj/game/cards/events.clj
+++ b/src/clj/game/cards/events.clj
@@ -3132,12 +3132,17 @@
     :choices (req (let [is-draft-id? #(.startsWith (:code %) "00")
                         runner-identity (:identity runner)
                         format (:format @state)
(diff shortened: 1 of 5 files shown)


NoahTheDuke:
I don't think this is the best way to do it. Either introspect the current format somehow (include it on state?) or look at the `all-cards` atom or something else, but enumerating an ever increasing list of ids in multiple places is not the way to go.

NBKelly:
Fair. The good news: Fenris and Rebirth rotate when dawn comes out.
I think it's maybe best to just pull from `all-cards` though, since we don't have the saved cardpool data for the older rotations.

PR #7730, opened 2024-09-14
use stringer/strcat instead of str for all of our string composition

NBKelly:
I went through (almost) every backend class to do this.
Essentially, java is sloppy and slow with string operations using the basic libraries, and clojure inherits this. Stringer is a bit more optimized, so this is basically a "free" speedup (do note that the compile time will be a little longer, but we don't care too much about that).

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -61,6 +61,7 @@
                  [org.clojure/data.csv "1.0.0"]
                  [dev.weavejester/medley "1.8.0"]
                  [org.slf4j/slf4j-nop "1.7.32"]
+                 [stringer "0.4.1"]
                  [integrant "0.8.0"]
                  [cljc.java-time "0.1.18"]
                  [time-literals "0.1.5"]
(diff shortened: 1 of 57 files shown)


NoahTheDuke:
This is a pretty aggressive change. Do you have any reason to believe that java's native StringBuilder is causing us slowness?

NBKelly:
This change is admittedly based more on intuition than anything provable, I just figure string operations are actually sorta expensive and we use them a lot. At best this is going to be a super marginal improvement though, and if it's not worth that then no worries with rejecting this.

I figure since (looking at the server holdups) they're mostly a single thread with a super high stack depth thrashing, if we can minimize that a little bit, we're coming up ahead.

I could easily be wrong though, and this could be not a good idea after all :thinking:.

NoahTheDuke:
If you can point to something specific, like usage in the flame graph or an actual speed improvement in a performance test, then i'd be willing to try this out. otherwise, is an extra dependency for negligible improvements, when we could fix sending 50mb of lobby updates to 30 people every minute lol.

NBKelly:
fair :joy:

PR #7732, opened 2024-09-15
streamline broadcast to lobbies

NBKelly:
Marking this is a draft because:
1) I'm not 100% confident in this
2) I want to test it out on playtest for a week or two

So my two insights:
1) If we need to send stuff to 50 or so people, we're currently doing it all on one thread (we typically have a single thread maxed out)
2) people in a game only care about updates that involve them, and (for that little counter in the corner) if a game is created or destroyed

So I think I tackled both of those. Not sure if I did it right. It seems to work locally, but that's not a good measure of anything. I'll throw it up on the playtest server tonight.

### DIFF
--- a/src/clj/web/lobby.clj
+++ b/src/clj/web/lobby.clj
@@ -206,6 +206,25 @@
   []
   (filter #(app-state/receive-lobby-updates? %) (ws/connected-uids)))
 
+(defn already-in-game?
+  "Checks if a user with the given username is already in the game"
+  [{:keys [username]} lobby]
+  (some #(= username (get-in % [:user :username])) (get-players-and-spectators lobby)))
+
+(defn in-active-game?
+  "Checks if a user with the given username is in (any) started game, as a player or spec"
+  [user lobbies]
+  (some #(already-in-game? user %) (filter :started lobbies)))
+
+(defn users-not-in-game
+  "users not in an active game"
+  []
+  (let [user-cache (:users @app-state/app-state)
+        uids (lobby-update-uids)
+        users (map #(get user-cache %) uids)
+        lobbies (app-state/get-lobbies)]
+    (filter #(not (in-active-game? % lobbies)) users)))
+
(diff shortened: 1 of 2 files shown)


src/clj/web/lobby.clj
```diff
@@ -419,7 +438,8 @@
                                                       (handle-set-last-update (:gameid lobby) uid)))
             lobby? (get-in new-app-state [:lobbies (:gameid lobby)])]
         (send-lobby-state lobby?)
-        (broadcast-lobby-list)
+        ;;(broadcast-lobby-list) - not sure why we need to broadcast this to players outside
```
NoahTheDuke, line 441: we display the chosen ids on the lobby page. this should be a bespoke message tho, not a full send

src/clj/web/lobby.clj
```diff
@@ -219,7 +238,7 @@
    (let [lobbies (app-state/get-lobbies)]
      (doseq [[uid ev] (prepare-lobby-list lobbies users)]
        (when uid
-         (ws/chsk-send! uid ev))))))
+         (future (ws/chsk-send! uid ev)))))))
```
NoahTheDuke, line 241: this could blow up if there's many messages. maybe we should use pmap or a custom executor.

src/clj/web/lobby.clj
NBKelly: hmm - that's not until the game has actually started though, and that's covered by starting the game, right?
(it also breaks currently if somebody leaves and rejoins)

src/clj/web/lobby.clj
NBKelly: yeah, makes sense. I'll figure it out tonight.

src/clj/web/lobby.clj
NoahTheDuke: hah that's a good point.

NBKelly:
Obsoleted

PR #7744, opened 2024-09-18
effcom doesn't crash the lobby

### LINKED ISSUE #6713: EffComm crashes lobby when cards advanced
After taking Clicks from Efficiency Committee playing a card that placements advancements (Dedication Ceremony/Shipment from SanSan/Tennin/Kaguyo) cause the game to blackscreen.

Workaround:
play the operation, don't resolve it, then place advancements using /counter advancement n

### LINKED ISSUE #7734: Efficiency Commitee warning toast spam
After scoring and using EffComm in a draft game, the game almost broke by starting to send a continuous stream of warning toasts, even though I wasn't trying to advance anything. Undoing click/turn didn't help and in fact hindered the game's progress even further, because I now had to redo all my actions, and they were barely coming through, so I had to click every button like 50 times. 😃 
This continued into the runner's turn and then suddenly stopped when a run on the scoring remote was initiated.

In the attached video the EffComm toasts are combined with La Costa ones, but before I undid the turn it was just EffComm ones.

https://jinteki.net/replay/70336f01-79ea-4485-ae0b-98026b79b8a0

https://github.com/user-attachments/assets/35067786-9145-4eca-8693-5b2f8e055959

NBKelly:
Essentially:
* We create a toast 
* The frontend acks the toast
* When we process an action, we run checkpoint+cleanup
* this includes when we ack a toast...
* checkpoint+cleanup involves checking if we can advance cards (it runs through it at least twice)
* we just added two or more extra toasts

I believe this actually spikes the entire server, since it creates a very rapid dedicated stream of ws messages, so it (and potentially similar things) would explain a tiny slice of the lag complaints we get.

This is really quite cursed. Add to the fact that for some reason, the toasts don't actually flush on the corp side until the runner gets a toast most of the time (this has been ongoing for like 5 years too :joy:), and I can see how this stayed unresolved for so long.

For now, I've moved `can-advance?` to process-action, instead of basic, so it doesn't get queried every single server frame, only when you click to advance something.

Closes #7734
Closes #6713

### DIFF
--- a/src/clj/game/cards/agendas.clj
+++ b/src/clj/game/cards/agendas.clj
@@ -730,10 +730,7 @@
                 :cost [(->c :click 1) (->c :agenda 1)]
                 :effect (effect (gain-clicks 2)
                                 (register-turn-flag!
-                                  card :can-advance
-                                  (fn [state side card]
-                                    ((constantly false)
-                                     (toast state :corp "Cannot advance cards this turn due to Efficiency Committee." "warning")))))
+                                  card :can-advance (constantly false)))
                 :keep-menu-open :while-agenda-tokens-left
                 :msg "gain [Click][Click]"}]})
 
--- a/src/clj/game/cards/agendas.clj
+++ b/src/clj/game/cards/agendas.clj
@@ -730,7 +730,10 @@
                 :cost [(->c :click 1) (->c :agenda 1)]
                 :effect (effect (gain-clicks 2)
                                 (register-turn-flag!
-                                  card :can-advance (constantly false)))
+                                  card :can-advance
+                                  (fn [state side card]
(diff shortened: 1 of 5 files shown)


NBKelly:
little bit more digging - I believe I actually fixed (or at least, masked) the issue where prompts didn't flush on the corp side - it was happening when prompts of nil type were being passed to `toast`, which would `aget` a nil fn from `js/toastr`, and then try to execute that and silently fail.

NoahTheDuke:
I don't think can-advance? should toast. If we're relying on that somewhere, we should probably just toast manually, right? it's too useful a predicate and having a glaring side effect is bad, as we can see here

PR #7751, opened 2024-09-19
make matryoshka a cost

### LINKED ISSUE #6896: No "Match strength and break" for Matryoshka
As it is now, playing Matryoshka gets tedious, especially with Cezve.

E.g. to break an ICE with 3 strength and 1 subroutine, I have to click:
1. Match strength
2. Click Cezve to pay with it
3. Break X subroutines
4. Click the dropdown
5. Click "1"
6. Click Done
7. Click Cezve to pay with it
8. Click the subroutine to break

8 clicks that could have been 2. :')

I understand this is a minor issue, but it's impairing playtime a bit in the late game.

--

Thank you for such an awesome implementation of netrunner. I have had a blast, played 58 games in 11 days. Friendly community too. :)

### LINKED ISSUE #7617: Log: Matryoshka doesn't say how much cash was spent to activate its break ability
image
As you can see, the message is borked. You can work it out by counting how many subroutines were broken, but it interferes with the log and makes it hard to scan for credit spending at a glance.

NBKelly:
I was intending to do this ages ago and forgot about it.

~~This doesn't actually fix anything, just makes the label display nicely and I guess makes the code look nicer.~~

I figured out a nice way to handle x-costs within our auto-breaking system, and I was able to get lobisomem, matryoshka and utae to all work with it, so I guess this now Closes #7617 and Closes #6896

I added a unit test for utae (since I had to make our auto-break respect the `:once` key as well).

matryoshka-label
matry
lobi_000
lobi

### DIFF
--- a/src/clj/game/cards/programs.clj
+++ b/src/clj/game/cards/programs.clj
@@ -2099,31 +2099,18 @@
 
 (defcard "Matryoshka"
   (let [break-abi {:label "Break X subroutines"
-                   :cost [(->c :x-credits)]
-                   :break-cost [(->c :x-credits)]
-                   :req (req (let [hosted (:hosted (get-card state card))
-                                   valid (filter #(not (facedown? %)) hosted)
-                                   same-title (filter #(= (:title card) (:title %)) valid)]
-                               (and (active-encounter? state)
-                                    current-ice card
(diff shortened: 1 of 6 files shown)


NoahTheDuke:
Nice!

PR #7762, opened 2024-09-23
option to enable/disable auto-pass prio on rez

### LINKED ISSUE #7726: Game incorrectly goes to "[Corp] has no further actions" after Corp rezzes ice
When rezzing an ice in paid ability window 2b of the approach ice phase, the chat indicates that the Corp has no further action. However, the paid ability window should go on as normal.

An example where this is important is the interaction between Arruaceiras Crew and Border Control. In the approach ice phase, the Corp has the chance to rez and then trash BC to end the run, that is before the runner gets to trash it with Crew in the Encounter Ice phase.

Another example: Ari + Physarum Entangler + Mavirus.

NBKelly:
Just disabling the feature is bad, because 99.9% of the time it is just very convenient, so I made it an option you can toggle at instant speed in the ingame settings panel.

Closes #7726


NoahTheDuke:
fix the merge conflicts and you can merge it

PR #7775, opened 2024-09-25
Ganked raindrops interaction

### LINKED ISSUE #7773: Ganked + Raindrops interaction
I don't have a replay on hand, but what happened was:
Runner runs a server with Raindrops. The outermost ice fires but doesn't ETR due to Bankhar.
Runner accesses a Ganked! and encounters the same ice again, which now does ETR.
Raindrops gains counters and draws cards only from the first encounter sub activations.

NBKelly:
Following the rabbit hole to it's logical conclusion and trying to figure out why forced encounters specifically don't fire subroutine events before the run ends, but regular encounters do, was too hard.

Raindrops is the only card that interacts with subroutines like this, so I moved the event back a tiny bit to make sure it always gets caught.

Closes #7773

### DIFF
--- a/src/clj/game/core/ice.clj
+++ b/src/clj/game/core/ice.clj
@@ -276,10 +276,11 @@
                (update-current-encounter state :prevent-subroutine nil)
                (if prevent
                  (checkpoint state nil eid {:duration :subroutine-currently-resolving})
-                 ;; see if there are any effects which can prevent this subroutine
-                 (wait-for (resolve-ability state side (make-eid state eid) (:sub-effect sub) (get-card state ice) nil)
-                           (queue-event state :subroutine-fired {:sub sub :ice ice})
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
As you probably suspected, this fix is wrong, but I'm also not sure how it's wrong. I think given nothing else interacts here we can merge and move on, but i'll have to think about this a little bit more.

NBKelly:
yeah, I just spent hours trying to unravel the gordian knot of code in `runs.clj` and couldn't figure out a specific point I could change to fix it :joy:

NoahTheDuke:
i suspect it comes from how we have two eids going at once when a rund ends: the runs original eid and the player action eid when they click the button or whatever to perform an action to end the run. i still don't know how to fix  that

NBKelly:
Yeah - this is exclusive to forced encounters during the access phase, so it's particularly troublesome to figure out :weary:

PR #7796, opened 2024-10-08
Enforce conditions better

NBKelly:
* Stripped the trash-on-tagged and enforce-conditions stuff out of the card maps 
* stripped the trash-on-tagged stuff out of the engine entirely and redeffed it in terms of card events
* beckman/zona sul still pass all tests

I think this should debloat the gamestate a tiny bit, and also strips one repetitive function out of engine (from fake-checkpoint), which is nice


NoahTheDuke:
Can't use a simple trigger-event if you're gonna trash stuff in the abilities. things will get out of sync. gotta use trigger-event-sync or -simult out checkpoint.

PR #7812, opened 2024-10-10
Auto break standard cards

### LINKED ISSUE #7360: Boi-tatá cannot break subs anymore if a runner card was trashed that turn
My Boi seems to be bugged. Wouldn't let me pay to break subs, despite having enough strength due to turbine, and me having 2c (eggs) (though I should've only needed 1c)

Link to bug replay: https://jinteki.net/bug-report/292cb5a4-3000-4ae3-97ac-2bf440ef7746?b=0

Description:

NBKelly:
I went through all the standard-legal icebreakers and added `auto-icebreaker` functions for the ones that were missing.

* atman *just works*
* Pressure Spike required me adding a key so the `2:+9` is never considered for auto-breaking
* I rewrote boi-tata so that it actually works again. I updated `action.clj/ice.clj` to actually compute the pump costs, rather than just scry the key in the ability, in order to make this work. We can now apply the `cost-bonus` key to any ability to get scalable cost an abilities (maybe that will be useful later, idk).

I added automated tests for boi-tata too, so we don't break it later.

Additionally:
* I update auto-icebreaker to allow for cards that don't have strength (like botulus, or endurance)
* updated the following cards to use auto-icebreaker
  * D4v1d
  * Botulus
  * Poison Vial
  * Endurance
  * e3 Feedback Implants (I added the cond from posion vial too)

Didn't update quetzal because we don't respect the once-per-turn restriction in auto-icebreaker, and didn't add boomerang because apparently we can pay two trash costs and I don't have time to fix that.

Closes #7360

### DIFF
--- a/src/clj/game/cards/hardware.clj
+++ b/src/clj/game/cards/hardware.clj
@@ -32,7 +32,7 @@
                               lose-credits]]
    [game.core.hand-size :refer [hand-size runner-hand-size+]]
    [game.core.hosting :refer [host]]
-   [game.core.ice :refer [all-subs-broken? any-subs-broken? break-sub pump
+   [game.core.ice :refer [all-subs-broken? any-subs-broken? auto-icebreaker break-sub pump
                           reset-all-ice update-all-ice update-all-icebreakers
                           update-breaker-strength]]
    [game.core.installing :refer [runner-can-install? runner-can-pay-and-install? runner-install]]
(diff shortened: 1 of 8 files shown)


NBKelly:
auto-break-for-non-icebreakers
atman

NoahTheDuke:
you could fix the trash cost by changing :trash value to return the :cost/amount and then payable? checks if it's installed and the value is 1.

NBKelly:
I implemented that and boomerang now works as intended :100:.

So everything in standard that breaks ice should currently be convenient to use.

PR #7815, opened 2024-10-11
use latest java-time

NBKelly:
I got sick of seeing the warning about abs being shadowed in java-time, so I checked the package
* It got renamed in 2022 (see: https://clojars.org/cljc.java-time)
* Now we should be using this: https://clojars.org/com.widdindustries/cljc.java-time

I was a bit suspicious, but I double checked and it's
* the same person
* the same github account

It seems like an odd choice to muddle with clojars like that, but I've done worse.


NoahTheDuke:
yeah that seems right to me

PR #7816, opened 2024-10-11
set jvm to use more aggressive jit

NBKelly:
Not sure if this will make a difference, but it seems like our performance is typically pretty bad for ~1-2 days after a reset, and then it's butter smooth after that - so it might be worth trying this.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -87,6 +87,7 @@
                    :resource-paths ["target"]
                    :clean-targets ^{:protect false} ["target"]
                    :jvm-opts ["-Djdk.attach.allowAttachSelf"
+                              "-XX:TieredStopAtLevel=4"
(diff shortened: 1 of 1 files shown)


NBKelly:
actually, this will only apply in dev mode I think :thinking:. I'll fix it up a little later.

NoahTheDuke:
status?

NBKelly:
I put my extra five minutes of effort into it. We just need to add this to the startup args when running the jar, like:
`java -XX:-OmitStackTraceInFastThrow -XX:TieredStopAtLevel=4 -jar target/netrunner-standalone.jar`. I'm trying it out on the playtest server to makesure it doesn't somehow self-destruct anything.

PR #7818, opened 2024-10-12
add helper for trojans/hosting-req

NBKelly:
Essentially, we've written the exact same code (more than) a dozen times - so it might as well be a helper, since we're inevitably going to get more of them too.

Works like:
`(trojan cdef)` for almost all cards
or
`(trojan {:rezzed true} cdef)` for cards like egret or parasite


NoahTheDuke:
looks great

PR #7822, opened 2024-10-15
Issue 7799: Timestamps in game log

### LINKED ISSUE #7799: Feature Request: Have a timestamp next to log messages.
image
image

Lifted straight from pokemon showdown, this feature would allow the users to see what time the logs on a battle were printed.

heyshiloh:
Hello,

This PR contains an initial implementation of #7799 

 closes #7799 

Screenshot 2024-10-17 at 5 49 57 PM

**Description of change**
- Creates user account setting `log-timestamps`, and enables saving user preference from account settings page and in-game settings tab
- Uses selected option to conditionally wrap user and system generated log messages in `timestamp-wrapper` or `timestamp-wrapper-system`
- Adjusts new and existing CSS to ensure visual quality with and without this feature

**Methodology**
- Per discussion, timestamp has been added to the right of `username` and is formatted to contain the time without AM/PM
- Conditional div structure ensures that this feature does not implement unnecessary div/span nesting
- For the sake of consistency across modules, `js/Date.` was used to format. Previous implementations of this library were left with comments stating that this was not preferable, but it will be easier to adjust all implementations of `js/Date.` to a preferred library that it would to find all instances of date string formatting

**Local dev and testing environment**
IDE: `IntelliJ IDEA`
OS: `MacOS 13.2.1`
Browser: `Google Chrome`
Lein version readout: `Leiningen 2.11.2 on Java 21.0.4 OpenJDK 64-Bit Server VM`

**Next Steps and considerations**
- We could implement a toggle on the top right of the chat box to allow users to adjust this setting quickly during a game
- Should this be `off` by default? What is the process through which new features are rolled out?

This is my first PR to this repository and I would be interested in joining the slack channel to continue the conversation!

Thank you for your time

Shiloh

### DIFF
--- a/src/cljs/nr/gameboard/log.cljs
+++ b/src/cljs/nr/gameboard/log.cljs
@@ -182,7 +182,7 @@
   (if (get-in @app-state [:options :log-timestamps])
     (if (not= text "[hr]")
       [:div.timestamp-wrapper-system
-       [:span.timestamp "[" (.toLocaleTimeString (js/Date. timestamp)) "]"]
+       [:span.timestamp "[" (string/replace (.toLocaleTimeString (js/Date. timestamp)) #"\s\w*" "") "]"]
        (render-message (render-player-highlight text corp runner))
        ]
       (render-message (render-player-highlight text corp runner))
@@ -195,7 +195,7 @@
   (if (get-in @app-state [:options :log-timestamps])
     [:div.timestamp-wrapper
      [:div.username (:username user)]
-     [:div.timestamp "[" (.toLocaleTimeString (js/Date. timestamp)) "]"]
+     [:div.timestamp "[" (string/replace (.toLocaleTimeString (js/Date. timestamp)) #"\s\w*" "") "]"]
      ]
     [:div.username (:username user)]
     )
--- a/src/cljs/nr/gameboard/log.cljs
+++ b/src/cljs/nr/gameboard/log.cljs
@@ -8,7 +8,7 @@
     [nr.gameboard.actions :refer [send-command]]
     [nr.gameboard.card-preview :refer [card-preview-mouse-out
                                        card-preview-mouse-over zoom-channel]]
-    [nr.gameboard.state :refer [game-state not-spectator? all-users?]]
+    [nr.gameboard.state :refer [game-state not-spectator?]]
     [nr.translations :refer [tr]]
     [nr.utils :refer [influence-dot player-highlight-option-class
                       render-message render-player-highlight]]
--- a/src/cljs/nr/gameboard/log.cljs
+++ b/src/cljs/nr/gameboard/log.cljs
@@ -180,13 +180,7 @@
 
 (defn format-system-timestamp [timestamp text corp runner]
   (if (get-in @app-state [:options :log-timestamps])
-    (if (not= text "[hr]")
-      [:div.timestamp-wrapper-system
-       [:span.timestamp "[" (string/replace (.toLocaleTimeString (js/Date. timestamp)) #"\s\w*" "") "]"]
-       (render-message (render-player-highlight text corp runner))
-       ]
-      (render-message (render-player-highlight text corp runner))
-      )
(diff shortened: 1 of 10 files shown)


NBKelly:
Hi,  that's a pretty good start, but I do worry it makes the log a bit harder to read.

I think we could do several things:
* Add an option for displaying them (the relevant files are `src/cljs/nr/account.cljs` (your settings page), `src/cljs/nr/gameboard/options.cljs` (the settings tab in game) and `src/clj/we/auth.cljs` (setting the key in the profile as something that we're allowed to send to the frontend so we can read it).
* Once there's a profile key, you can wrap the timestamp like:
```Clojure
(when (get-in @app-state [:options :timestamps-enabled])
  ...)
```
* We could put them on the same line as the text (I think that screenshot Larrea posted is pretty nice - dimmer than the text you really care about, and without adding vertical space. I *think* if they're in the same div that should *just work*.

Don't be afraid to experiment with the css a little. If you find something that you think looks nice, there's a good chance others will think the same.

Noah probably has a few more ideas or opinions.

heyshiloh:
@NBKelly Thank you for your feedback!

What if we added the toggle as a little clock button or a clickable text option on the top right corner of the log? I think having the toggle in-game accessible would be best for its use case. Otherwise i will get it into account options and apply the feature flag as such

having the log above the text was a personal readability preference, the change to inline would be minimal and I will implement that if that is your preference!

NBKelly:
Hi, the settings tab in the panel is accessible in game, but a little clock in the top-right corner actually sounds very nice. Pick whatever you like (or do both, you can make them use the same keys) and find something that looks nice.

Inline is my preference though for the actual display though, it's just more compact so the busier gamestates dont get harder to parse.

heyshiloh:
Screenshot 2024-10-15 at 1 54 55 AM

timestamps inline, working on setting/toggle

src/cljs/nr/gameboard/log.cljs
```diff
@@ -1,20 +1,20 @@
...
+    [nr.gameboard.actions :refer [send-command]]
+    [nr.gameboard.card-preview :refer [card-preview-mouse-out
+                                       card-preview-mouse-over zoom-channel]]
+    [nr.gameboard.state :refer [game-state not-spectator?]]
+    [nr.translations :refer [tr]]
+    [nr.utils :refer [influence-dot player-highlight-option-class
+                      render-message render-player-highlight]]
+    [nr.ws :as ws]
+    [reagent.core :as r]
+    [reagent.dom :as rdom]))
```
heyshiloh, line 17: seems like IDE formatting I will handle this before merge

heyshiloh:
Screenshot 2024-10-15 at 9 32 38 AM

setting on/off

NoahTheDuke:
Could you move the timestamp to the right of the username? I think that would both look better (consistent username location with it on or off) and match how other applications work (like Slack).

heyshiloh:
Screenshot 2024-10-15 at 11 31 33 AM
83bcbdcc9e)
Screenshot 2024-10-15 at 11 32 19 AM


shored up css added options

NoahTheDuke:
Just the user messages.

heyshiloh:
Screenshot 2024-10-15 at 11 46 51 AM
@NoahTheDuke thoughts?

NoahTheDuke:
The inconsistency is a little annoying, but I'm not sure how to solve it without making each system message 2 lines (something like "[system] [timestamp]\ntest2 uses x to y!"), so it's probably fine.

heyshiloh:
@NoahTheDuke I'll take a look at what it would take to separate out the system logs. I agree that consistence would be nice to have

heyshiloh:
Screenshot 2024-10-15 at 1 12 20 PM

css final-ish

heyshiloh:
Screenshot 2024-10-17 at 5 48 46 PM
Screenshot 2024-10-17 at 5 49 57 PM

moved timestamp to the right, finalize css to ensure visual quality with and without setting

heyshiloh:
cleaned code updated PR description ready for review

PR #7826, opened 2024-10-17
Fix install and discard bad behaviour

### LINKED ISSUE #6893: Corp player cannot end turn due to discard even with proper hand size 5/5
Link to bug replay: https://jinteki.net/bug-report/f798f955-d98b-42cf-96cf-9e029575bec3?b=0

Description:

Corp player had 8 cards in hand.  Was told to discard to 5, but upon discarding to 5, was still told to discard.  Could not end turn even after discarding down to 4 cards.

### LINKED ISSUE #7824: Installing programs mid-run leads to broken game-state
You are playing Hoshiko. Currently, you are not in a run. you click the Unity in your hand to install it. Egads! you are over MU. no worries, jnet  simply stops you from playing it, you never even lost the credits or the click. You hit "done". You trash a Leech and try again, this time succesfully installing Cleaver.

You are playing Arissana. You make a run on a remote defended by a Wall of Static. You look at your hand and click on Arissana. A little menu pops up and you select Cleaver. Oh no! you don't have MU so you can't install Cleaver. You drag a Coalescence to the heap and click "done". The cleaver is not installed. Confused, you click on arissana to find out that she doesn't activate again. You have spent your one use of arissana, misplayed and you're going to have to /undo-click. This sucks.

This also happens with SMC and simulchip and any other sources of installing programs that are not the basic action.

NBKelly:
I did a few things here.

* First, you cannot drag cards during select prompts. This should prevent unwitting players from discarding from hq in the wrong way, or trashing programs in the wrong during install.
* Second, if you opt to trash some programs (but not enough), we can retry the install and ask again (I wish there was a trivial way to determine if enough MU was selected, but there are probably issues with it and effects)
* Third, you cannot attempt to trash the host of a card you are installing during installation (according to the rules team). I added a function in `hosting` to check if a card happens to be an ancestor of another card to facilitate this. This is nice, because doing that before actually locked up the game without presenting any errors, so we get a double-whammy here.

Closes #7824
Closes #6893 (by preventing the player from discarding like that)

### DIFF
--- a/src/clj/game/core/actions.clj
+++ b/src/clj/game/core/actions.clj
@@ -143,8 +143,10 @@
         from-str (card-str state c)
         s (if (#{"HQ" "R&D" "Archives"} server) :corp :runner)]
     ;; allow moving from play-area always, otherwise only when same side, and to valid zone
+    ;; here!
     (when (and (not= src server)
                (same-side? s (:side card))
+               (not= :select (get-in @state [side :prompt-state :prompt-type]))
                (or (= last-zone :play-area)
                    (same-side? side (:side card))))
       (let [move-card-to (partial move state s c)
--- a/src/clj/game/core/hosting.clj
+++ b/src/clj/game/core/hosting.clj
@@ -8,7 +8,7 @@
     [game.core.initializing :refer [card-init]]
     [game.core.memory :refer [init-mu-cost]]
     [game.core.update :refer [update! update-hosted!]]
-    [game.utils :refer [remove-once]]))
+    [game.utils :refer [remove-once same-card?]]))
 
 (defn remove-from-host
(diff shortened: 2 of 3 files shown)


src/clj/game/core/actions.clj
```diff
@@ -143,8 +143,10 @@
         from-str (card-str state c)
         s (if (#{"HQ" "R&D" "Archives"} server) :corp :runner)]
     ;; allow moving from play-area always, otherwise only when same side, and to valid zone
+    ;; here!
     (when (and (not= src server)
                (same-side? s (:side card))
+               (not= :select (get-in @state [side :prompt-state :prompt-type]))
```
NoahTheDuke, line 149: Good chance.

PR #7828, opened 2024-10-18
get user block options from the app state instead of the lobby

### LINKED ISSUE #7827: Games from users that have blocked you show up in the lobby
In the lobby, games from users that have blocked me show up as normal. If I try and join or spectate, the buttons do nothing, but it still shows up in the lobby, and I can see who they're playing and what IDs like normal. I don't believe chat messages show up, just games in the lobby.

NBKelly:
Turns out preprocessing the lobby list stripped the user options out before they could be checked for blocking.

We probably should have been checking options in the app-state (so they're up to date) anyway though.

Closes #7827

### DIFF
--- a/src/clj/web/lobby.clj
+++ b/src/clj/web/lobby.clj
@@ -209,7 +209,7 @@
        (select-non-nil-keys lobby-keys))))
 
 (defn get-blocked-list [user]
-  (->> user :options :blocked-users (map str/lower-case)))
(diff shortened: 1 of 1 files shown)


NBKelly:
Side note, I'm about 90% sure the username and the uid are the same. Is this ever not the case?

NoahTheDuke:
I have no idea, tbh. I don't remember the logic for determining uids.

PR #7833, opened 2024-10-20
7821: check multi subtypes card helper function

heyshiloh:


.gitignore
```diff
@@ -16,6 +16,7 @@ pom.xml
 .nrepl-history
 .rebel_readline_history
 .idea
+*.iml
```
NBKelly, line 19: If you want to adjust gitignore but not have it show up in commits, you can instead add things to `.git/info/exclude`, which is like gitignore but local to your copy of the project

heyshiloh: okay i will do this

NoahTheDuke: i'm fine with it being added here, there's no issue, i just wondered where it came from in the other pr.

NoahTheDuke:
Given it's unused, I don't think we need `has-all-subtypes?` but I'm not gonna pause the PR over it.

heyshiloh:
@NoahTheDuke it can be removed, but this at least documents code that provides the logic of interest presented in the issue. Let me know your thoughts if we should adjust it later

PR #7835, opened 2024-10-21
Worlds 2024 preconstructed lists

### LINKED ISSUE #5584: Feature request: Open decklists
For tournaments with open decklists, we could include another tab in the new menu above the game log (that's only visible in replay mode), where you can switch to see your opponents decklist during the game.

NBKelly:
Worlds is over, so I'm adding the precon decks for 2024 (just waiting on decklist names and player preference as to screen names vs. real names).

I also added a feature for displaying/checking open decklists.

Video is in slack, but here's what it looks like in the lobby.

open-decklists

Closes #5584

### DIFF
--- a/src/clj/game/core/diffs.clj
+++ b/src/clj/game/core/diffs.clj
@@ -425,6 +425,7 @@
    ;; :angel-arena-info
    :corp
    :corp-phase-12
+   :decklists
    :encounters
    :end-turn
    :gameid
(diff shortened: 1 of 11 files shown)


NBKelly:
ordered

I think I'm going to leave this here. I'm happy with this from a functional standpoint, but I really don't have the skills to make it look pretty. If somebody decides to do that at a later date, then :100:.

PR #7840, opened 2024-10-24
Rejig vs annicam

NBKelly:
Rejigging an aniccam previously wasn't getting a draw. Turns out we just missed out an async and defined it in a weird way to make up for that.

I also changed the `is-hand?/is-deck?/is-discard?` fn's to be error-wrapped macros using the `is'` function (this is still black magic to me), so if they're wrong it'll tell us where they're being called rather than where the `is-hand?/etc` function is.

### DIFF
--- a/src/clj/game/cards/events.clj
+++ b/src/clj/game/cards/events.clj
@@ -3201,7 +3201,7 @@
                  :choices {:card #(and (valid-target? %)
                                        (installed? %))}
                  :effect (req (move state side target :hand)
-                              (effect-completed state side (make-result eid (:cost target))))}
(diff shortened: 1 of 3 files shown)


NoahTheDuke:
normal `is` just prints directly when there's a failure. `is'` throws an exception when there's a failure. `error-wrapper` catches the thrown exception and then prints like normal `is`, but using the macro's location instead of the function's location. this lets us write normal functions for our helpers and then just use a little macro to expand them into the right calls in the tests. it's not ideal, but clojure.test is a bad test framework so we have to make due.

PR #7841, opened 2024-10-24
Async swaps on install

### LINKED ISSUE #7839: Project Yagi-Uda should use the Async swap
Midori should too I guess, and Daruma

NBKelly:
Added a unit test for yagi-uda, just relied on tests not breaking for daruma/midori.

Closes #7839


NoahTheDuke:
Is there ever a time we wouldn't want to async swap cards?

NBKelly:
If neither or both of them are installed, there aren't any events that matter (at least for now).
Sooner or later I'll see about making *all* movement async.

NoahTheDuke:
making all movement async would be... one hell of a pr. i've resisted it for years lol. the better move imo would be to change the engine to work on steps of some form and stop relying on callbacks (aka async). but that's also a lot of work... ;_;

PR #7850, opened 2024-10-30
Reveal hand when selecting cards from opponents hand

NBKelly:
I think this is a really nice QOL feature. I just have to root out all the cards it applies to.

Essentially, there's not much reason for us to to use lists to pick cards out of our opponents hand unless only a subset of the cards is known. It's both clunky (in our interface) and clunky (in our code).

Some of the cards/systems I can hit (these are all done and this pr is ready):
* Snoop
* Focus Group
* [Vera Ivanovna Shuyskaya](https://netrunnerdb.com/en/card/33115)
* Engram Flush
* Ibrahim Salem
* Chronos Protocol's damage effect

Demo:
with-revealed-hands.webm

### DIFF
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -20,7 +20,7 @@
    [game.core.checkpoint :refer [fake-checkpoint]]
    [game.core.damage :refer [damage damage-prevent]]
    [game.core.def-helpers :refer [corp-recur corp-rez-toast defcard
-                                  reorder-choice trash-on-empty get-x-fn]]
+                                  reorder-choice trash-on-empty get-x-fn with-revealed-hand]]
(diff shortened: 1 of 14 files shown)


NBKelly:
Realized I need a special safeguard for if the target adds cards to hand while things are revealed (ie steelskin draws cards, but there's also a buffer drive on the field and there's a prompt to resolve), so I'm adding that in.

NBKelly:
[steelskin-interaction.webm](https://github.com/user-attachments/assets/286a3747-baa9-4970-91a1-8112c0b2e806)

Interaction with steelskin - cards get flipped face-down when the runner draws, despite still resolving the consequences of the ability. I'm pretty happy with that.

NoahTheDuke:
still working on this or it's good to go?

NBKelly:
yeah this one's good to go

PR #7863, opened 2024-11-03
grip has an article

### LINKED ISSUE #7861: Dont drag a card into the zone it was from
Apparently you can drag a card from hand to hand which will log the card in the chat

NBKelly:
I changed the name-zone for Grip to instead yield "the Grip" so it would actually fit into sentences that used it, but apparently I needed to change it in gameboard and actions too.

Closes #7861

### DIFF
--- a/src/clj/game/core/actions.clj
+++ b/src/clj/game/core/actions.clj
@@ -142,10 +142,7 @@
         src (name-zone (:side c) (:zone c))
         from-str (card-str state c)
         s (if (#{"HQ" "R&D" "Archives"} server) :corp :runner)]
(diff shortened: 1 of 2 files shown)


src/clj/game/core/actions.clj
```diff
@@ -142,6 +142,8 @@
         src (name-zone (:side c) (:zone c))
         from-str (card-str state c)
         s (if (#{"HQ" "R&D" "Archives"} server) :corp :runner)]
+    (println src)
```
NoahTheDuke, line 145: 👀

PR #7922, opened 2025-01-23
async validation via context-free grammar

NBKelly:
I wrote a parser for our card data to try and pick out
* (some) things that are marked async but dont complete
* (some) things that complete, but are not marked async
This is added in as a unit test, and (for now) only covers the card data (I'll look at expanding it to some of the core files later maybe)

And I'm also fixing the bugs it picks up.
What I picked up so far:
* Things that were async but didn't complete: `SYNC BRE, Calibration Testing, Dummy Box, Uninstall, Helheim Servers` (I fixed all of these)
* Things that are async, but did not complete: `Brasilia Govt. Grid, Ash, Angelique, Adrian, The Back, Street Magic, Stim Dealer, PolOp, Dadiana, Climactic Showdown, Chatergee University, Utae, Umbrella, Persephone, ... and lots more - there were ~70 or so` - I fixed all those too
* For things that are async, the following is checked:
  *  The rightmost member of the `:effect` function should terminate (complete it's eid, or call a function that does), or continue to another ability (via continue-ability).
  * If the rightmost member is an `if, if-not, if-let`, then there should be two members which both terminate
  * If the rightmost member is a `when, when-not, when-let`, then we've made a mistake, because that can fail to terminate
  * All the outputs of `condp`, `cond`, `case` and `cond+` terminate, where they exist (doesn't check that an :else/fallback actually exists though - whether that's needed is unknowable by a grammar)
  * if the RHS of an effect is a def sourced from elsewhere, it's currently not validated. Maybe I'll solve that later.

This adds the instaparse package as a dev dependency - see: https://github.com/Engelberg/instaparse

Essentially this should help us pick out (most) async errors before shipping code in the future.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -65,6 +65,7 @@
                  [integrant "0.8.0"]
                  [com.widdindustries/cljc.java-time "0.1.21"]
                  [time-literals "0.1.5"]
+                 [instaparse "1.5.0"]
                  [metosin/reitit "0.5.18"]]
 
   :test-selectors {:default (fn [m] (not (:kaocha/pending m)))}
@@ -78,7 +79,8 @@
                                   [lambdaisland/kaocha "1.68.1059"]
                                   [thheller/shadow-cljs "2.16.8"]]
                    :plugins [[lein-eftest "0.6.0"]
-                             [cider/cider-nrepl "0.47.1"]]
+                             [cider/cider-nrepl "0.47.1"]
+                             ]
                    :eftest {:report eftest.report.pretty/report
                             :fail-fast? false}
                    :source-paths ["src/clj" "src/cljs" "src/cljc" "src/css"
--- a/project.clj
+++ b/project.clj
@@ -65,7 +65,6 @@
                  [integrant "0.8.0"]
                  [com.widdindustries/cljc.java-time "0.1.21"]
                  [time-literals "0.1.5"]
-                 [instaparse "1.5.0"]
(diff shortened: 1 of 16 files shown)


project.clj
```diff
@@ -65,6 +65,7 @@
                  [integrant "0.8.0"]
                  [com.widdindustries/cljc.java-time "0.1.21"]
                  [time-literals "0.1.5"]
+                 [instaparse "1.5.0"]
```
NoahTheDuke, line 68: this should be in dev dependencies

src/clj/game/cards/upgrades.clj
```diff
@@ -811,8 +811,7 @@
                 :req (req (and this-server
                                (pos? (count run-ices))
                                (pos? (count (:hand corp)))))
-                :async true
-                :cost [(->c :trash-from-hand 1)]
+                                :cost [(->c :trash-from-hand 1)]
```
NoahTheDuke, line 814: spacing

NBKelly:
All right, this is (if all tests pass) good to go :)

NBKelly:
I made it so when checking if something completes, it checks both sides of `if/if-not/if-let` conditionals, and all children of `cond/condp` conditionals. It also automatically fails on `when/when-not/when-let` conditionals, because that's a bad pattern.

And it picks out cancel-effects that aren't async, as well (picked up a couple more errors with that one).

*ok, now I'm done.*

PR #7933, opened 2025-01-30
Rewrite grammar properly

NBKelly:
I made the grammar actually (sort of) follow proper conventions, and now it requires less hacks works nicer. It also supports deffered functions (let, letfn) now.

Fixed about a dozen more cards with async issues too.

### DIFF
--- a/src/clj/game/cards/assets.clj
+++ b/src/clj/game/cards/assets.clj
@@ -304,6 +304,7 @@
 
 (defcard "B-1001"
   {:abilities [{:req (req (not this-server))
+                :async true
                 :cost [(->c :tag 1)]
                 :msg "end the run"
                 :label "End the run on another server"
--- a/src/clj/game/cards/hardware.clj
+++ b/src/clj/game/cards/hardware.clj
@@ -1253,6 +1253,7 @@
 
 (defcard "Lemuria Codecracker"
   {:abilities [{:action true
+                :async true
                 :cost [(->c :click 1) (->c :credit 1)]
                 :req (req (some #{:hq} (:successful-run runner-reg)))
                 :choices {:card installed?}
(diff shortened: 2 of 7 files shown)


NBKelly:
I extended this to cover the core files as well (and to read the files from the system, rather than using a hardcoded list, so it should be valid even when we add new mechanics).

I picked up the following errors in our core functions:
* command-score and command-trash weren't async
* the any-agenda-counter handler in costs wasn't async
* choosing a breach replacement, and being prevented from breaching in runs
* end-of-turn-discard apparently wasn't async

And I annotated a couple of files:
* deferring to custom payment abilities in pick-counters
* the effects in prompt.clj that are pulled from other files
* the ability map in sabotage-ability that defers the effect and cancel-effect to another def

Additionally, if the parser fails to parse a chunk of code, it'll print out the partial parse/a readable error, instead of just a stacktrace, so it'll be maintainable if we need to modify it later (say, for new language features down the line).

I think this is shippable in this state.

src/clj/game/core/prompts.clj
```diff
@@ -28,15 +28,15 @@
...
-         newitem {:eid eid
-                  :msg prompt
-                  :choices choices
-                  :effect f
-                  :card card
-                  :prompt-type (or prompt-type :other)
-                  :show-discard show-discard
-                  :cancel-effect cancel-effect
-                  :end-effect end-effect}]
+         newitem ^:ignore-async-check {:eid eid
```
NoahTheDuke, line 31: let's find a better way to format this so it doesn't add 20 characters indent.

NBKelly: better, or is it still a bit too weird?

NoahTheDuke: much better, thanks

PR #7934, opened 2025-01-31
Forced encounter fix

### LINKED ISSUE #7880: Sisyphus Protocol scored, re-encountering Konjin after Arruaceiras Crew trashes ICE bug
Hi, I forgot to use /bug during the game in which it occurred, but I will describe it here to the best of my ability. It is simple to reproduce.

Cards: **Arruaceiras Crew**, **Konjin**, **Sisyphus Protocol**, **Saisentan**

1. Situation: Corp has scored a **Sisyphus Protocol** one turn ago. During the next Runner's turn, the Runner runs Server 1. **Konjin** is in Server 1 Position 1. During the Psi-Game encounter ability, the Runner spends 0 credits, and the Corp spends 2 credits. Corp forces the Runner to encounter **Saisentan** in Server 2 Position 0. Runner uses the first ability of **Arruaceiras Crew** to tag themselves and reduce Saisentag to 0 strength. Runner uses the second ability of **Arruaceiras Crew** to trash **Arruaceiras Crew** and trash **Saisentan**.

2. Bug: After **Saisentan** is trashed, the Runner encounters **Konjin** again. There is no usual pop-up asking the Corp how they would like to pay for **Sisyphus Protocol**. Instead, **Konjin** is encountered and the Psi-Game encounter ability triggers again.

3. Outcome: Runner and Corp communicate, and both bet 0 credits on the second Konjin Psi-Game encounter ability. Runner then encounters **Tatu-Bola** in Server 1 Position 0. Runner breaks **Tatu-Bola** using **Audrey v2** and successfully breaches the server. Run completes without incident and there are no further bugs.

My Guess: Though **Sisyphus Protocol** is not explicitly part of this interaction, I think it is part of the bug. My guess is that Jnet encounters an issue when trying to resolve the interaction of re-encountering a trashed Sentry or Code Gate via **Konjin**. Regardless, the big point is that in no instance should **Konjin** be encountered again (except against **Rielle 'Kit' Peddler** when Konjin is made into a Code Gate).

Big fan of all your work on Jnet, thanks so much for keeping an integral part of the Netrunner community alive.

### LINKED ISSUE #7894: Runner re-approaches Konjin after nested encounter
This is a weird one. I was running a server protected by a single ice, which turned out to be Konjin. I lost the psi game and was sent to encounter a piece of ice on another server. After that encounter concluded, the game proceeded to approach konjin *again*.

We resolved it by just agreeing to both bid 0 the second time. After the second encounter, the run proceeded beyond Konjin as it should have done the first time.

Replay of the game: https://www.jinteki.net/replay/960e24cf-884f-4b86-bdfc-17b7602f3297 (The incident is on turn 9.)

Potentially relevant context: Hippo was installed, and behaved strangely as well. But I made a separate issue about that: https://github.com/mtgred/netrunner/issues/7895

### LINKED ISSUE #7895: Hippo incorrectly triggers on Konjin encounter with ice protecting another server
According to the Hippo rulings (https://netrunnerdb.com/en/card/21103), Hippo should *not* trigger when Konjin sends me to another server, even if the encountered ice is the outermost on that server, since Hippo only looks at the outermost ice on the *attacked* server. This happened to me while I was experiencing another piece of Konjin weirdness though (https://github.com/mtgred/netrunner/issues/7894).

I would have expected Hippo to trigger on Konjin once the nested encounter was over, and allowed me to trash it. But it didn't; it triggered on the ice on the other server, and then also made me approach Konjin an extra time, as described on the other issue.

NBKelly:
Resets the current ice after a forced encounter ends. I think it gets muddied up when the other ice is trashed otherwise somehow, but this seems to fix it.

I also made breaking ice track if it was on the attacked server in the event it sends out, which lets us enforce that restriction on hippo.

Closes #7895
Closes #7894
Closes #7880

### DIFF
--- a/src/clj/game/core/runs.clj
+++ b/src/clj/game/core/runs.clj
@@ -10,7 +10,7 @@
     [game.core.engine :refer [checkpoint end-of-phase-checkpoint register-pending-event pay queue-event resolve-ability trigger-event trigger-event-simult]]
     [game.core.flags :refer [can-run? cards-can-prevent? clear-run-register! get-prevent-list prevent-jack-out]]
     [game.core.gaining :refer [gain-credits]]
-    [game.core.ice :refer [active-ice? get-current-ice get-run-ices update-ice-strength reset-all-ice reset-all-subs! set-current-ice]]
+    [game.core.ice :refer [active-ice? break-subs-event-context get-current-ice get-run-ices update-ice-strength reset-all-ice reset-all-subs! set-current-ice]]
     [game.core.mark :refer [is-mark?]]
     [game.core.payment :refer [build-cost-string build-spend-msg can-pay? merge-costs ->c]]
     [game.core.prompts :refer [clear-run-prompts clear-wait-prompt show-run-prompts show-prompt show-wait-prompt]]
@@ -334,10 +334,15 @@
                                 (checkpoint state side eid)))))})
 
 (defn encounter-ice
+  ;; note: as far as I can tell, this deliberately leaves on open eid (the run eid).
+  ;; Attempting to change that breaks a very large number of tests, so I'm leaving this
+  ;; note here to remind me when I look at this later. -nbk, 2025
   [state side eid ice]
   (swap! state update :encounters conj {:eid eid
                                         :ice ice})
   (check-auto-no-action state)
+  ;; step 6.9.3a: The encounter begins. Conditions relating to the Runner encountering
+  ;; this ice are met (this is on-encounter effects, etc)
   (let [on-encounter (:on-encounter (card-def ice))
         applied-encounters (get-effects state nil :gain-encounter-ability ice)
         all-encounters (map #(preventable-encounter-abi % ice) (remove nil? (conj applied-encounters on-encounter)))]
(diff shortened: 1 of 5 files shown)


NBKelly:
Additionally, a "subroutines broken" event is fired when an ice without subroutines is encountered. According to the rules team, you're considered to have broken all subs when that happens. I annotated the steps for that function as well, because why not.

I also noticed that the encounter-ice function doesn't actually complete it's eid unless the encounter is meant to end. I don't know if that's an intentional part of the design, or an oversight that we've accidentally programmed around. Fixing specifically breaks all forced encounters.

If it actually is an accident, I'll see about fixing it later. For the most part, I really can't wrap my head around the control flow that's happening during a run :shrug:.

src/clj/game/core/runs.clj
```diff
@@ -349,8 +354,16 @@
...
-                (encounter-ends state side eid)))))
+              (if (should-end-encounter? state side ice)
+                (encounter-ends state side eid)
+                ;; step 6.9.3b: if there are no subroutines on the ice,
+                ;; the runner is considered to have broken all the subroutines on this ice.
+                ;; This should fire an event, so it can get picked up with cards like hippo
+                ;; or knifed.
+                (when-let [c-ice (get-current-ice state)]
+                  (when (and (same-card? c-ice ice) (zero? (count (:subroutines c-ice))))
+                    (wait-for (trigger-event-simult state side :subroutines-broken nil (break-subs-event-context state c-ice [] (get-in @state [:runner :basic-action-card]))))))))))
```
NoahTheDuke, line 365: no need for an empty `wait-for`, just call the `trigger-event-simult` directly with the eid.

PR #7947, opened 2025-02-05
Subroutines have labels

### LINKED ISSUE #7946: We should unit test that abilities and subroutines actually have labels
Seems like something doable, and I honestly wouldn't be surprised if there's a card or two lurking around with empty ability boxes somewhere.

NBKelly:
I extended the abilities-have-labels test to cover corp-abilities, runner-abilities and subroutines.

This caught out that minelayer and datapike had some unlabelled subroutines (whoops, good thing nobody played those cards).

Closes #7946

### DIFF
--- a/src/clj/game/cards/ice.clj
+++ b/src/clj/game/cards/ice.clj
@@ -1457,6 +1457,7 @@
 
 (defcard "Datapike"
   {:subroutines [{:async true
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
great fix, thanks.

PR #7949, opened 2025-02-06
offer to trash like cards on install

### LINKED ISSUE #2054: Allow corp to trash ice when installing new ice
This would be a nice enhancement: after the corp player chooses a server to install ice on, if there is ice on the server, list the ice and a "none" button so that the player can trash ice without having to drag it to the trash manually. You could label each ice button with <ice name> at position <position> so that it's clear which ice will be trashed if multiples of the same ice are on a server.

I'm not sure if this violates your "as few clicks as possible" principle but it better represents the rules and helps teach new players. You might also display the current cost to install ice on a server with one or more pieces at the top of the prompt.

### LINKED ISSUE #7948: Add a toggle for always trashing a program/trashing like cards on install
I think it will be useful. 
I'll probably just add checkbox at the top of the chat window that says "ask to trash like cards on install?", and sneakily updates a the game state in the background, then write it into the install functions for each side.

NBKelly:
Added a little checkbox you can tick for if you want to (have the option to) trash like cards on any ice, server or program install.

This works for the runner side (it's the same timing that you trash programs for MU), and it works for the corp side. I also made it so you can attempt to install something in a server that's too expensive so long as it's possible to trash things from that server to price it down enough, instead of just silently failing which is what we currently do.

There's a video in slack/

Closes #7948
Closes #2054 (woah that's an old issue)
and I think it hits the last one of #6497 (other than ncigs, which is going to go out the window in a few months)

### DIFF
--- a/src/clj/game/cards/basic.clj
+++ b/src/clj/game/cards/basic.clj
@@ -61,15 +61,17 @@
                                    (corp-can-pay-and-install?
                                      state side eid
                                      target-card server {:base-cost [(->c :click 1)]
-                                                    :action :corp-click-install
-                                                    :no-toast true})
+                                                         :ignore-ice-cost true
+                                                         :action :corp-click-install
+                                                         :no-toast true})
                                    (some
                                      (fn [server]
                                        (corp-can-pay-and-install?
                                          state side eid
                                          target-card server {:base-cost [(->c :click 1)]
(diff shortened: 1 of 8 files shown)


NBKelly:
I also took it out of `change` and just added a `set-property` hook in `process-actions`, which we can later use for other stuff if needed.

PR #7950, opened 2025-02-07
cursed assimilator interactions

### LINKED ISSUE #7943: Cards flipped face-up are never loaded, so the on-empty event should never fire
Relevant for Earthrise hotel, and other similar cards, as per the latest NSG rulings.

I think a realistic way to handle this would be for flipping a card faceup to add a 'was-flipped' key into the card's metadata, and for relevant cards to scry for that.

I'll go over these at some point, if nobody beats me to it.

NBKelly:
Closes #7943

The change that absolutely *nobody* asked for.

I also refactored juli to use the `:action` syntax rather than working backwards, and that fixed the one unit test that was failing.


NoahTheDuke:
Nice! That test is very silly but clever!

Issue #7957, opened 2025-02-12
Defer all the events that happen when we pay costs until the 'cost-paid' checkpoint


NBKelly:
This is sort of on a "to do eventually" list, but since all costs are paid simultaneously (I think) without checkpoints between them, we should (eventually) see if we can make this work.

PR #7958, opened 2025-02-12
DO NOT MERGE - queue cost events

NoahTheDuke:
for @NBKelly in #7957, merely for historical analysis.


NBKelly:
I finished/merged the other cost rework, so this should be good to close now. Thanks :)

PR #7964, opened 2025-02-18
Fixed Runner Quotes Against Weyland Consortium

Cluedrew:
I did this fix for my own quotes before, but I did some manual testing to confirm it and you do need to change:

> Weyland

To the formal name:

> Weyland Consortium

Now I only tested one runner identity (Esa) but the renaming should be identical. I also didn't do any reformatting, just the straight text substitution.


NoahTheDuke:
Thank you!

PR #7965, opened 2025-02-18
Corrected 2 Ashes Identity Quotes

Cluedrew:
I found an incorrect identity name in the Ashes, so I looked through the other identities in the cycle and it turns out another one had errors. I only tested, before and after the two I updated and just looked at the others with a critical eye.

The updated identities Hoshiko Shiro: Untold Protagonist (which confuses me because it is a very popular identity and had a completely different subtitle "Next Level Shut In") GameNET: Where Dreams are Real (I hear less about that and it was just bad casing).


NoahTheDuke:
Next Level Shut In was the original subtitle during development.

PR #7966, opened 2025-02-21
Issuas qol

### LINKED ISSUE #7962: Better Issuaq UI
When Issuaq has counters, could the agenda points UI for the corp be updated to say "2/6, 4/5" etc? (as appropriate for the number of counters)

NBKelly:
I made the card text agenda point reqs (issuaq, harmony) into static abilities, and the rules ones (draft format, etc) can stay as they are.

I also made it so that you can see changed agenda point requirements on the frontend (they only show up when it's not 7 points to win)

agenda-point-req

Closes #7962

### DIFF
--- a/src/clj/game/cards/identities.clj
+++ b/src/clj/game/cards/identities.clj
@@ -846,10 +846,8 @@
                       (conceal-hand state :runner)))}))
 
 (defcard "Harmony Medtech: Biomedical Pioneer"
-  {:effect (effect (lose :agenda-point-req 1)
-                   (lose :runner :agenda-point-req 1))
-   :leave-play (effect (gain :agenda-point-req 1)
-                       (gain :runner :agenda-point-req 1))})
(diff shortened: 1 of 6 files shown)


src/clj/game/cards/identities.clj
```diff
@@ -985,6 +983,9 @@
 (defcard "Issuaq Adaptics: Sustaining Diversity"
   {:effect (effect (lose :agenda-point-req (get-counters card :power)))
    :leave-play (effect (gain :agenda-point-req (get-counters card :power)))
+   :static-abilities [{:type :agenda-point-req
+                       :req (req (= :corp side))
+                       :value (req (* -1 (get-counters card :power)))}]
```
NoahTheDuke, line 988:
```suggestion
                       :value (req (- (get-counters card :power)))}]
```

Larrea-Mon:
According to GLC, the runner also gets this text, which they maybe shoulndn't. 
image

PR #7970, opened 2025-02-21
Use counters asynchronously (and a few related improvements) + costs paid checkpoint

### LINKED ISSUE #7957: Defer all the events that happen when we pay costs until the 'cost-paid' checkpoint

### LINKED ISSUE #7969: Make all our usages of counters async
I want to make all our usages of add-counter and add-prop async. This is (kind of) a pre-condition to making our cost events pend until the costs are paid and the cost paid checkpoint fires.

- [x] Update corp cards
- [x] Update runner cards
- [x] Update engine
- [x] Update props so that it's properly queuing events and firing them via checkpoint, but can optionally leave the events pending (ie, when managing costs)
- [x] Update all instances of taking credits from a card (or manually gaining them) to use the `take-credits` and `spend-credits` functions
- [ ] Make cards which `fire event and trash` on counter gain interact with prevention effects correctly
  - [ ] Resources
  - [ ] Hardware
  - [ ] Programs

NBKelly:
Closes #7969
Closes #7957

Currently:
* All cards use counters asyncronously
* ~~You can set the default autoresolve status of a card (or any other special keys) as part of the card data now (see bankroll for an example). IE, `:data {:special {:auto-place-credits true}}` will transfer the contents of that special map to the card when it is initialized.~~ I didn't need to do this, turns out I can just tack `:special` keys onto the cdef and it just works/is much cleaner. I added a few cards to this list *(cookbook, consume, bankroll, friday chip)*, and made sure it plays nice with cards like compile and the shaper ids that also manipulate that space.
* I updated the async validator to ensure that `:move-zone` fns were handled asynchronously
* I also added `add-counter` and `add-prop` to the list of fns that should be marked as async, so we can catch stuff in the future
* I removed the overloads for add-counter and add-prop that did not include an eid, since they were no longer being used anywhere
* I updated costs so that spending counters now directly calls `add-counter` with the checkpoint suppressed
* I updated the prompts that spend counters so that they make fake eids for it (maybe we should update those functions? that's a job for later)
* I updated pick_counters to actually trigger spend-counter-from-card events
* I added a couple of helper functions to `def-helpers`:
  * 'take-credits' - takes (up to) n counters from the target card and places them in your credit pool as if they were credits. Handles the events/them being simultaneous, etc
  * 'spend-credits' - the same thing, but fires a `:spent-credit-from-card` event (ie for paladin manual usage)
* I went through jamie's list of cards that can prevent themselves from being trashed, and checked/enforced all of these.
  * Crowdfunding, environmental testing, the trash-on-tag resources, jackpot, time bomb were modified (I only added a test for envtesting)
  * Everything else we were doing correctly anyway :100: 
  * I made it so that revealing, losing tags, gaining bad pub, and a few other things used checkpoints (and allowed them to be suppressed)
  * All of the cost-paid triggers happen after costs have been paid now (except `derez`, that's pending on derez itself being reworked later)

Somehow, I think I'm actually done with this/it's ready to go.


NBKelly:
Oh yeah, since adding counters will trigger a checkpoint (by default, anyway) now, there's a chance that a lot of cards can have their 'check win'/'update strength' functions or whatever they do after placing counters removed.

NBKelly:
I thought to make sure like costs are grouped (ie there's one trash event for Simulchip, when previously there were two: one for the hardware, and one for the program), and I was pleasantly surprised to find that queue-event actually handles that for us. That's make a some of the future playtesting stuff easier :100:.

PR #7971, opened 2025-02-22
[RFC] Use maps instead of strings for messages

csbisa:
This is a draft change for moving the message system to use maps instead of strings. The main intent is to enable localization of messages in the front-end by transforming those maps into strings, with some other benefits such as consistent templating for all messages. This is not a complete change -- I'm posting this to get some feedback on the approach and figure out what adaptions are needed in order to move forward with this change. (Please note the changes
are on a ~6 month old branch and will be rebased later.)

Today, the message that gets passed to the client is:
  `{:username <username> :text <string>}`
which has been changed to:
  `{:username <username> :side <:corp/:runner> :type <TYPE> :cost {COST ...} :effect {EFFECT ...} <other-args ...> (:raw-text <string>)}`
This is turned into a string in the user's language via `render-map`.

Type is more or less the main verb for the sentence ("use", "install", "rez", "steal", etc.), cost is self-explanatory, and effect is the result (i.e. anything that's not a cost). There's plenty of examples in `test/cljc/i18n/en_test.clj` ([en: Add test for rendering](https://github.com/mtgred/netrunner/commit/015974740c455ef4406d5c1612047945b02cbfc3)).

The changes are:
- Update the cljs side to accept maps and call out to `render-map` based on the language to get the string [say: Return text as a map](https://github.com/mtgred/netrunner/commit/f92f0f698b9b8497bea430e3fc646fdce67a0967)
- A bunch of commits updating core functionality (basic action card, costs, etc) to use maps
- Implement `render-map` for English [en: render-map support](https://github.com/mtgred/netrunner/commit/4ae6b55b03ffaf86dbf99e257421e0f678d20ca7)
- Update most SG card abilities to pass maps instead of strings [[WIP] SG cards](https://github.com/mtgred/netrunner/commit/5674f140b7afb8244f9f28478be20e9b5875c636)

I have a couple other changes (`render-map` for Japanese, more updates for NSG cards, etc.) but I'm excluding these for now to focus on the core changes.

As part of working through effects, I was poking around to see if it was possible to have effects generate messages instead of having to specify a :msg for every effect. More or less every effect should have a message (and we could just mark the ones that shouldn't as silent), so in theory it should be possible to coalesce and pass down effects -- instead of what you see here, which is more or less updating every single :msg in cards to follow an effect format.  I haven't been able to figure out if it's actually practical to do this, so I've just been updating cards with the new format for messages.

This change is also backwards compatible. With the current approach, generated messages are still the same even if `:msg` hasn't been updated to the new format. If a string in `{:msg ability}` ends up at `print-msg`, it just stuffs that into `:raw-text` which is later appended to the transformed string with the type and cost. The only exception I've found so far is that some messages incorporate the cost (which is now a map) into `:msg` directly, which requires wrapping all such usage to transform the cost map back into a string (look at [converting costs to strings for unconverted routines](https://github.com/mtgred/netrunner/commit/d4fe3747a9379a35fde5106b739521f6d69c7a07) for an example). Other than a few edge cases that I haven't gotten around to fixing, the tests are passing.

And now onto the downsides of the current implementation:
- People writing cards are going to have to figure out and remember how to write the messages. This is a lot more effort compared to just writing English. There'll need to some kind of tooling and/or documentation -- if anyone has ideas on a good approach for this I'm all ears.
- Lack of compile-time checks, especially in effects (since it's handcrafted for every card and not generated), means it's very easy to make errors that are only caught at runtime. Tests have caught some of these mistakes but the test suite isn't exhaustive, so I'd like to have something in place to make this less error-prone.

This does not cover:
- The other two major source of strings -- prompts and labels. These are separate undertakings.
- Card titles, which are still passed down as strings. While I'd like to make this a bit better (i.e. pass down card IDs), the current front-end i18n logic does a fairly good job of translating those strings so I haven't bothered with this.

Other questionable things you might see when you're looking over the changes -- I'm aware of the following (and open to suggestions):
- Using maps for costs/effects. While this is nice in terms of always enforcing a value is provided, it has a few flaws, such as not being able to order effects (so the print could come out with random ordering, i.e. 'end the run' not being the last effect) and overwriting values during merge for some cases (two separate effects with the same type should be summed, not overwritten). Using vectors might be better overall.
- Really messy keyword handling as seen in `src/cljc/i18n/en.cljc`. I'm not an expert but I'm assuming messages are converted to another format (JSON?) when going to the client and it ends up turning a some (but not all) keywords into strings.
- Having both map-msg and map-msg-apply is silly, I just need to bulk update everything to always provide a hashmap.

### DIFF
--- a/src/clj/game/core/installing.clj
+++ b/src/clj/game/core/installing.clj
@@ -23,7 +23,7 @@
     [game.core.say :refer [play-sfx system-msg implementation-msg]]
     [game.core.servers :refer [name-zone remote-num->name]]
     [game.core.state :refer [make-rid]]
-    [game.core.to-string :refer [card-str]]
+    [game.core.to-string :refer [card-str card-str-map]]
     [game.core.toasts :refer [toast]]
     [game.core.update :refer [update!]]
     [game.macros :refer [continue-ability effect req wait-for]]
@@ -100,7 +100,7 @@
     {:prompt (str "The " (:title prev-card) " in " server " will now be trashed.")
      :choices ["OK"]
      :async true
-     :effect (req (system-msg state :corp (str "trashes " (card-str state prev-card)))
+     :effect (req (system-msg state :corp {:type :trash :card (card-str-map state prev-card)})
                   (if (get-card state prev-card) ; make sure they didn't trash the card themselves
                     (trash state :corp eid prev-card {:keep-server-alive true})
                     (effect-completed state :corp eid)))}
@@ -136,9 +136,8 @@
 (defn- format-counters-msg
   [{:keys [advance-counter] :as counters}]
   ;; TODO - rewrite this if/when we support more counter types through installs
-  (if advance-counter
-    (str ", and place " (quantify advance-counter "Advancement counter") " on it")
-    ""))
+  (when advance-counter
+    {:place-counter [:adv advance-counter]}))
 
 (defn- corp-install-message
   "Prints the correct install message."
@@ -152,22 +151,24 @@
                             (:seen card)
                             (rezzed? card))
                       (:title card)
-                      (if (ice? card) "ice" "a card"))
-          server-name (if (= server "New remote")
-                        (str (remote-num->name (dec (:rid @state))) " (new remote)")
-                        server)
-          origin (if display-origin
-                   (str " from "
-                        (when origin-index (str " position " (inc origin-index) " of "))
-                        (name-zone :corp (:zone card)))
-                   "")
-          lhs (if install-source
-                (str (build-spend-msg-suffix cost-str "use") (:title install-source) " to install ")
-                (build-spend-msg-suffix cost-str "install"))]
-      (system-msg state side {:cost cost-str
-                              :raw-text (str lhs card-name origin
-                                             (if (ice? card) " protecting " " in the root of ") server-name
-                                             (format-counters-msg counters))})
+                      nil)
+          card-type (if (ice? card)
+                      :ice
+                      (when-not card-name :unknown))
+          server-info (if (= server "New remote")
+                        ;; TODO are we sure this info isn't somewhere already?
+                        {:server [:servers (keyword (str ":remote" (dec (:rid @state))))]
+                         :new-remote true}
+                        {:server (server->zone state server)})
+          origin (when display-origin
+                   (merge {:origin (:zone card)}
+                          (when origin-index {:origin-index (inc origin-index)})))]
+      (system-msg state side (merge {:type :install :cost cost-str
+                                     :card-type card-type :card card-name}
+                                    server-info
+                                    (when install-source {:install-source (:title install-source)})
+                                    origin
+                                    (format-counters-msg counters)))
       (when (and (= :face-up install-state)
                  (agenda? card))
         (implementation-msg state card)))))
@@ -176,7 +177,10 @@
 (defn corp-install-msg
   "Gets a message describing where a card has been installed from. Example: Interns."
   [card]
-  (str "install " (if (:seen card) (:title card) "an unseen card") " from " (name-zone :corp (:zone card))))
+  {:type :install
+   :card-type (if (:seen card) :known :unknown)
+   :card (when (:seen card) (:title card))
+   :origin (:zone card)})
 
 (defn reveal-if-unrezzed
   "Used to reveal a card if it cannot be rezzed when an instruction says to rez it
@@ -416,10 +420,9 @@
         cost-str (if (and hide-zero-cost (= cost-str {:credits 0})) nil cost-str)
         prepend-cost-str (get-in msg-keys [:include-cost-from-eid :latest-payment-str])
         discount-str (cond
-                       ignore-all-cost " (ignoring all costs)"
-                       ignore-install-cost " (ignoring it's install cost)"
-                       (and cost-bonus (pos? cost-bonus)) (str " (paying " cost-bonus " [Credits] more)")
-                       (and cost-bonus (neg? cost-bonus)) (str " (paying " (* -1 cost-bonus) " [Credits] less)")
+                       ignore-all-cost {:ignore-all-costs true}
+                       ignore-install-cost {:ignore-install-costs true}
+                       cost-bonus {:cost-bonus cost-bonus}
                        :else nil)
         card-name (if facedown
                     (if known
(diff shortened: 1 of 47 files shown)


NoahTheDuke:
This is an insane amount of work. I'm quite busy this weekend but will review on Monday!

src/cljc/i18n/defs.cljc
```diff
@@ -0,0 +1,26 @@
...
+(defmacro try-catchall
+  "A cross-platform variant of try-catch that catches all exceptions.
+   Does not (yet) support finally, and does not need or want an exception class."
+  [& body]
+  (let [try-body (butlast body)
+        [catch sym & catch-body :as catch-form] (last body)]
+    (assert (= catch 'catch))
+    (assert (symbol? sym))
+    (if (cljs-env? &env)
+      `(try ~@try-body (~'catch js/Object ~sym ~@catch-body))
```
NoahTheDuke, line 19:
```suggestion
      `(try ~@try-body (~'catch :default ~sym ~@catch-body))
```

src/cljc/i18n/en.cljc
```diff
@@ -805,3 +808,586 @@
...
+;; TODO this keyword logic is just silly
+(defn render-single-effect-force-check
+  [effect value side]
+  (let [effect (name effect)]
+    (if (ends-with? effect "-force")
+      (str "force the "
+           ;; This is inverted -- corp forcing effect means it's forcing runner to take the effect.
+           (if (= (keyword side) :corp) "Runner" "Corp")
+           " to "
+           ;; oh god
```
NoahTheDuke, line 1151: lmao

NoahTheDuke (commented):
small things to think about

src/cljc/i18n/en.cljc
NBKelly: Just as another note, in the current branch I've added an optional `:display-side` key that can be used to set the `msg` source without changing the side of the ability itself, and you can also just do `:msg :cost` to get something like `player pays [cost] to satisfy [card]`.

csbisa:
Just getting the rebase out of the way. No significant changes yet.

src/cljc/i18n/en.cljc
csbisa: Hmm, that's a good call-out. It might be possible to pass down both the side and display-side and then use that to figure out if "force" wording should be applied. It looks like only a few cards use this right now (Wildcat Strike, Public Trail), from which it seems something like `(when (and (= display-side :runner) (= side :corp)) <add force>)` would work. I'll check some more cards to see if that logic holds up.

src/clj/game/core/installing.clj
```diff
@@ -435,15 +435,10 @@
                         :else
                         (name-zone :runner (:previous-zone card))))
                  "")
-        pre-lhs (when (every? (complement empty?) [cost-str prepend-cost-str])
-                  (str prepend-cost-str ", and then "))
+        ;; currently loses ", and then" -- all costs are squashed into one
```
csbisa, line 436: This is fixable, but I want to check if people actually like the way this is. I'm sure there's a good reason for it, but it seems a bit contrived to do it this way instead of having two separate messages.

NoahTheDuke: can you give an example? what's wrong with the current message?

src/cljc/i18n/en.cljc
NBKelly: Yeah, that's the current usage because it's something I've ported over from playtest, but it will be more relevant for future cards (ie dawn/elevation, and the set coming after it). Anything which uses `cost-option` as part of the `choose-one-helper` will use this functionality.

For reference, wildcat strike would be `force the runner to x`, and public trail would be `pays x to satisfy y`

src/clj/game/core/installing.clj
NBKelly: So there are two reasons for this:
* I wanted to reduce clutter in the log. This is relevant for quite a few cards (ie SMC, tucana) - these used to be like 5 lines for one effect, now they're one concise and readable line
* I wanted to make it so we aren't rewriting the same display code with every install. So you simply do (install ... {:msg-keys {...}}) and let that handle the formatting. Sooner or later I'm going to add this system to rezzing, and a few other areas too. (currently, if an effect rezzes a card, you will get several lines, one for the cost, one for the discount, one for the card being rezzed, and one for whatever the ability map does - when there should just be one)

csbisa: This felt a bit complicated from the code POV, but if the reason is to reduce the number of logs, then that makes perfect sense. I'll fix this up so costs can be sequenced -- something like cost being a vector of map/vectors instead of a single map.

src/cljc/i18n/en.cljc
csbisa: Public Trail (and other cards that would use "satisfy" wording) are easy enough to deal with. For Wildcat Strike, I came up with some code to be able to handle it, but looking into Funhouse, it doesn't seem like the logic translates well.

Funhouse (corp card, corp display side, runner selection) encounter effect:
- force the runner to take 1 tag (Runner effect) on encountering it
- end the run (Corp effect)

and its subroutine (corp card, corp display side, runner selection):
- force the runner to pay 4c (Runner effect)
- give the runner 1 tag (Corp effect)

I can't figure out a way to make display-side work with this nicely, since in theory the display-side would need to be for each individual effect, not the overall ability. I'd like to know your thoughts on how this should (or should not) extend to ice abilities. In essence, I need a way to indicate that it's the other side paying for it, which is why I currently have this `<effect>-force` keyword logic (which I don't like, for the record).

NBKelly: hmm - so if the ability is formatted specifically as {:cost ... :msg :cost}, then regardless of side, it's `(active player) pays (cost) to satisfy (card)`. Is that helpful?

In any case, funhouse is done completely manually, and is very verbose, so that could simply be rewritten. I'll look at rewriting it a little later if you like.

NoahTheDuke:
Reading through this, I think it's a really great idea but the current execution of the i18n file is a no-go. The changes to our messaging system are solid and I think with some help will be the best way going forward. However, we can't have all of the logic as functions in the i18n/en.cljc file. That will be prohibitive for any external contributors, who now need to know both Clojure and the target language.

Ideally, we can write `:en {:rfg "removes {{card-title}} from the game" ...}` which can then be translated by a contributor to `:sp {:rfg "eliminar {{card-title}} del juego" ...}`.

I know you've done a ton of work here already, so I don't wanna saddle you with another massive set of changes. I'll take a crack at it when I get a moment and we can compare notes.

PR #7974, opened 2025-02-28
option to just skip/autoresolve everything

### LINKED ISSUE #4990: Rework trigger-event-simult
Right now, the interface gets very click-intensive, once there are multiple cards that trigger on the same event.
Example: Runner runs on a server with multiple cards that trigger `:encounter-ice` to ask whether to fire the effect (e.g. Devil Charm).
On every `:encounter-ice` trigger, `trigger-event-simult` will list all triggered cards. Assuming that the Runner does not want to resolve any of these cards during this run, they'll need to click on every card + "No" prompt for every piece of ice they encounter.

Proposed interface:
In the new interface, `trigger-event-simult` marks all cards with pending triggers with a blue border and opens a card-select prompt. The Runner can now click on any cards that should resolve. These are rewritten in a way to not ask a "Resolve Devil Charm Y/N" prompt anymore, but simply fire the effect. Once the Runner has resolved all cards they want to fire, they press "Done" to discard any triggers that are still pending.

### LINKED ISSUE #7563: Highlight cards that are selectable when a card selection prompt is active
I've thought about this a bit, and I've come to a few conclusions:
* When we initiate a card selection prompt, there's no reason to not pre-compute all the available cards for selection (filter every card not in the deck or an opponents hand)
* If we have the set of cards that are valid, we can actually send those to the frontend as part of the prompt state
* This means we can highlight all available choices on the frontend

downside:
* we're running out of distinct colors to use

### LINKED ISSUE #7984: Choose-replacement-ability for runs (breach replacement) is hard for duplicate cards
From ApexLui (james123lui):
> When there are 2+ Cataloguers installed, it's unintuitive to choose which one to use because the prompts say "breach R&D" or "Cataloguer". I guess "breach R&D" means denying replacement so the next one can trigger? It also doesn't have the "highlight frame" to indicate the source card of the ability. 

We should link the prompt button to the card like we do with other prompts

NBKelly:
We can mark events as `:skippable` or `:automatic`.

If something is **skippable**, it can be declined without causing any side-effects (most, but not all, optional prompts).
If something is **automatic**, it can be resolved in an arbitrary order without risking breaking the game (Honestly, this might just be everything - but I'm not entirely sure, so I'm being careful).

When resolving a queued-event (ie encounter, turn begins, turn ends, etc):
* If every handler is either **skippable** or **automatic**, offer the *pass priority* button in the list (I need a better name for this, that one is **not** it)
* When you click that, every skippable prompt is automatically skipped, and every automatic one resolves in an arbitrary order

Additionally, I made it so that you can:
* click on cards during choices prompts to select that card (if it's mapped to a choice)
* During prompts, all cards that are selectable are highlighted (Closes #7563)

Closes #4990

And also:
* the breach replacement prompts allow you to click on the cards, or hover the buttons to highlight the cards too

Closes #7984


NBKelly:
ok right now this allows you to:
* press done on any queued or triggered event prompt
* this will skip (almost) all things that can be skipped consequence free, and will fire everything else in sequence after that
Any I'm completely happy to take this part out:
* I added some functionality to sort the events, so similar things are batched and there's a basic bit of sequencing:
That full map is here:
```Clojure
(def automatic-priority
  {:pre-bypass 1
   :corp-damage 1
   :force-discard 1
   :lose-clicks 1
   :gain-clicks 2
   :drain-credits 4
   :bypass 4
   :lose-credits 4
   :pre-gain-credits 5 ;; bladderwort, etc
   :gain-credits 6
   :pre-draw-cards 7
   :draw-cards 8
   :post-draw-cards 9
   :pre-breach 9
   true 10
   :trace 11
   :corp-lose-tag 11
   :last 999 ;; pretty mary wants to trigger last always
   })
```

NBKelly:
If there are cards that are part of the prompt, you can click on those cards to click that button.
If a card happens to be two buttons, then :shrug:. Ideally all the selectable cards would glow, but somebody else will need to figure out how to do that for me.

src/cljs/nr/gameboard/board.cljs
```diff
@@ -170,10 +170,26 @@
...
+            ;; (js/console.log (str "Card:" clicked-card))
+            (when (= (:cid value) (:cid clicked-card)) uuid))
+          choices)))
+
 (defn handle-card-click [{:keys [type zone] :as card} shift-key-held]
   (let [side (:side @game-state)]
     (when (not-spectator?)
       (cond
+        ;; A prompt is open, and that card is among the prompt buttons
+        (prompt-button-from-card? card (get-in @game-state [side :prompt-state]))
```
NBKelly, line 190: I can move this into the let so I'm not computing it twice

PR #7982, opened 2025-03-01
swap sides command

NBKelly:
Added a command that allows you to swap sides with your opponent. This is primarily intended for teaching/learn to play, and also for debugging, but there's no reason players can't have fun with it either.

It requires both sides to consent, and you can just click "don't ask me again" if you like.


NBKelly:
oops, this one actually isn't ready to go yet - still a couple of wrinkles to iron out

PR #7983, opened 2025-03-02
NCIGS changes

NBKelly:
I made it so that instead of passing a fn in like:
```Clojure
{:effect ...
 :no-change-in-game-state (req ...)}
```

we instead pass in a map like
```Clojure
{:effect ...
 :no-change-in-game-state {:req (req ...)
                           :silent [truthy|nil]}}
```

I updated every ncigs reference in our code for this.

I'm going to go through and do ncigs changes on the other cards over the coming weeks.

Essentially the breakdown is as follows:
* You are allowed to play abilities that do nothing for the sole sake of burning currency
* Abilities that refer to the `ice being encountered` still require the ice to be encountered

- [x] Ice (make subroutines feel nicer to use)
- [X] Operations (done prior)
- [x] Identities
- [x] Assets
- [x] Upgrades
- [x] Agendas
- [x] Hardware
- [X] Events (done prior)
- [x] Resources
- [x] Programs (note the ice encountered rule)
- [x] Fix unit tests where card behaviour has changed (ie simulchip)

Along the way I've been fixing a few things:
* Data mine had a weird async issue, same for special offer
* Border control ability did not enforce it's condition
* The fairchildren ice wont bother you if you can't afford *either* option
* Watchtower shuffles if you cancel
* I made a few ice use our pre-defined subs instead of rewriting them
* Mamba can only be used during a run
* Hades Shard is automated now
* Ampere lives in it's alphabetical spot in the identities file now instead of down by N
* You can use SMC during signal jamming just to shuffle if you like

This has to wait for a month or two to go out, but it's "Done" for then.

### DIFF
--- a/src/clj/game/core/engine.clj
+++ b/src/clj/game/core/engine.clj
@@ -324,10 +324,12 @@
 
 (defn do-nothing
   "Does nothing (loudly)"
-  [state side eid ability card]
-  (when-not (get-in ability [:change-in-game-state :silent])
-    (system-msg state side (str "uses " (:title card) " to do nothing")))
-  (effect-completed state side eid))
+  ([state side eid ability card] (do-nothing state side eid ability card nil))
+  ([state side eid ability card payment-str]
+   (when-not (get-in ability [:change-in-game-state :silent])
+     (print-msg state side (assoc ability :msg "do nothing") card [] payment-str))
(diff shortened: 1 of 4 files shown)


NBKelly:
I ended up rolling this into the elevation changes

PR #7996, opened 2025-03-07
Clean Up in Quotes Code

Cluedrew:
When I was learning how to update the quotes file I noticed some weird things in the code that read it. So I have done a minor bit of clean-up, proving I can do some programming in Clojure.

- Added `load-quote-file` as a helper to absorb the repeated code in `load-quotes!` (and it also removes another bit of repeated work).
- Removed some extra lines of whitespace.

I have tested it and it seems to work and I have used the conventions as I have understood them from looking at the project.

### DIFF
--- a/src/clj/game/quotes.clj
+++ b/src/clj/game/quotes.clj
@@ -10,23 +10,20 @@
 
 (def identity-quotes (atom {}))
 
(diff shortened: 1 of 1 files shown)


NoahTheDuke:
looks good! thanks!

PR #8001, opened 2025-03-11
Modified Lobby Sort

### LINKED ISSUE #7587: New Games Shouldn't Appear At The Top
When a new game is created it should go below the most recently made games, so the person who's been waiting the longest is at the top, and also this would reduce the motion reducing the amount of people clicking to join a specific game and it moving as they click.

tylerjallen02:
This pull request is my proposed solution to issue #7587 

The goal of this solution is to make games that have been waiting longer more easily accessible, while also making newer games easier for spectators to access
 
This sorting solution splits the lobbies into 3 categories. Each group is individually sorted and combined in this order (top to bottom)
1. Open Lobbies - sorted with oldest on top - newest on bottom
2. Started Games Allowing Spectators - newest on top, oldest on bottom
3. Started Games Not Allowing Spectators - newest on top, oldest on bottom 

Let me know if this solution works, or if we want to take another approach to sorting lobbies. If accepted:

closes #7587

### DIFF
--- a/src/clj/web/lobby.clj
+++ b/src/clj/web/lobby.clj
@@ -194,13 +194,15 @@
    :spectatorhands
    :started
    :timer
-   :title])
+   :title
+   :old])
 
 (defn lobby-summary
   "Strips private server information from a game map, preparing to send the game to clients."
   ([lobby] (lobby-summary lobby nil))
   ([lobby participating?]
    (-> lobby
+       (assoc :old (> (count (or (:messages lobby) [])) 10)) 
        (update :password boolean)
        (update :players #(prepare-players lobby %))
        (update :spectators #(prepare-players lobby %))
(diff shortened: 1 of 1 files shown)


NBKelly:
That seems pretty reasonable, but I think there's one more thing worth adding - if there's been more than (say, a dozen) messages in a lobby, it's most likely that either:
1) both players are already in the lobby and chatting away
2) the lobby owner is afk, and a bunch of people have joined and then left the lobby
I think it's reasonable to move those games below the fresher ones.

tylerjallen02:
Would you recommend adding a criteria based on the number of messages, or after the game has existed for so long (like 10+ minutes)

NBKelly:
I think number of messages is probably reasonable - typically in an unattended lobby, somebody will join, see that other people have joined and left, then instantly leave - and these consume two messages each, and I think a fresh lobby starts with two messages?. Feel free to take whichever approach makes the most sense to you (and use whatever numbers you like). As it stands now, I think your change will already be a good improvement on what we have :+1:

tylerjallen02:
The most recent commit should put lobbies with more than 10 messages below other open lobbies. It should be ready to merge unless there are any other requested changes.

NoahTheDuke:
nah this si good :shipit:

PR #8006, opened 2025-03-13
Clean-Up of MU Ability Code

Cluedrew:
There was a bunch of redundancy in the `mu+`, `virus-mu+` and `caissa-mu+` functions. I added the calling convention other mu functions call `mu+` to its documentation and trimmed down the other functions so they only handle the cases where they are different from the `mu+`.

If a need for these cases comes up again, turning all three into wrappers around some private core might work best, forward their arguments, with the default mu type to it and they can handle all the complex logic. But for now this handles all the same cases with less code.


Cluedrew:
I did testing for the virus case. The caissa case is symmetric so I didn't test it separately.

PR #8011, opened 2025-03-14
hosting cards out of play calls leave-play fns

### LINKED ISSUE #8009: Bugged interaction between Malia and Cupellation
One of the members of my community server found a bugged jinteki.net interaction:
Image

This situation is a bit of a rules hole even in the comprehensive rules, so I've emailed rules@nullsignal.games about it. But the rules as intended is that when Malia is uninstalled and hosted on Cupellation, that cancels its ability and the printed text of the resource returns.

We discuss this interaction on GLC at length here: https://discord.com/channels/425105101451952128/425105101451952130/1350088182066184265

NBKelly:
I made it so we call `leave-play` fns when a card gets hosted out of play.

Closes #8009

### DIFF
--- a/src/clj/game/core/hosting.clj
+++ b/src/clj/game/core/hosting.clj
@@ -26,11 +26,11 @@
 (defn handle-card-is-uninstalled
   "If a card is hosted (uninstalled) from being installed and active, then call it's `leave-play` fn"
   [state side card {:keys [installed] :as target}]
-  (when-let [leave-play-fn (and (active? (get-card state target))
-                                (not installed)
-                                (installed? (get-card state target))
-                                (:leave-play (card-def target)))]
-    (leave-play-fn state (keyword (clojure.string/lower-case (:side target))) (make-eid state) target nil)))
+  (when-let [leave-play (:leave-play (card-def target))]
+    (when (and (not installed)
+               (installed? (get-card state target))
+               (active? (get-card state target)))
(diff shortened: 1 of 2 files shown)


src/clj/game/core/hosting.clj
```diff
@@ -23,11 +23,21 @@
   [card target]
   (when (and card target) (or (same-card? card target) (has-ancestor? (:host card) target))))
 
+(defn handle-card-is-uninstalled
+  "If a card is hosted (uninstalled) from being installed and active, then call it's `leave-play` fn"
+  [state side card {:keys [installed] :as target}]
+  (when-let [leave-play-fn (and (active? (get-card state target))
+                                (not installed)
+                                (installed? (get-card state target))
+                                (:leave-play (card-def target)))]
```
NoahTheDuke, line 32: move this to the front of the `and` cuz it's both worth checking otherwise without such an fn

NBKelly: hmm, I'm not quite sure what you mean.

NoahTheDuke: i mean that because `and` early exits, we should check that a :leave-play fn exists before we check if the card is active or installed or anything else. many cards will be active but only those with :leave-play fns matter.

NBKelly: oh I see, yeah that's fair. I was just trying to be cute with the `when-let` heh

PR #8012, opened 2025-03-15
Alt art grouping

### LINKED ISSUE #8004: Allowing for more than 1 alt art of the same card in the same folder
Currently, jnet picks up an alt art from a folder based on the file name (34034.png), and my understanding is that there is no way to upload multiple versions of the same card into the same folder.

It would be nice if there was a way to upload these versions and name them for example 34034a.png, 34034b.png etc so that they would be displayed alongside 34034.png in the card browser and would all be selectable.
This is required because some illustrators made multiple versions of the same card. In the current build the only workaround would be to offshore the other versions to different folders and that's not too kosher.

One of the prominent examples is Hippo in various stages of degradation, but there are other cases as well.

NBKelly:
The regex will probably need to be adjusted for alt versions of flip cards, but that's a bridge that can be crossed later.

The output on mongo looks like this, for a case with two versions of overclock in the same folder:
db-output_000

Closes #8004

### DIFF
--- a/src/clj/tasks/images.clj
+++ b/src/clj/tasks/images.clj
@@ -66,7 +66,7 @@
    (let [filename (.getName f)]
      (if (str/includes? filename "-")
        (add-flip-card-image db base-path lang resolution art-set filename)
-       (let [code (first (str/split filename #"\."))
+       (let [code (first (str/split filename #"[a-zA-Z]*\."))
              k (str/join "." ["images" (name lang) (name resolution) (name art-set)])
              prev-k-root (if (= :stock art-set) code (name art-set))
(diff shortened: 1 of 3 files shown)


NBKelly:
ok this was a nightmare to actually figure out, but I ended up with something that:
1) works
2) maintains compatability

Essentially, we're already storing arts on cards as 1-ary lists (I don't know why), so we can just store multiple things in those lists instead.

On the cardbrowser, we now seperate it by item in each list, rather than just by each list.

When you select an alt art, it now indexes `code [key idx]`, instead of `code key`.

When looking at a card art, we then check if `art` is sequential: if yes, we fetch the list of arts, then pick the art at the given idx.
If no, then we just do what we already do which is yield the first item from the list.

PR #8018, opened 2025-03-17
Updated defcard Macro for More Expressive Use

Cluedrew:
Argumented the `defcard` macro so the body/ability can be given in multiple parts. The doesn't reduce the amount of code you actually need to write for most card definitions, but allows for better organization, removing the need for nesting and ")" counting.

The macro now consists of the existing template section, that has very little macro logic in it but adds a lot of boiler plate code, and a new ability section, that does more computation to get the body. Because a similar standard macro exists, that part is pretty small.

The logical signature of the macro is `(defcard title transformers* body)`. If you don't use any transformers this does the exact same things as before. But you can put in transformers and they will wrap up the body. I think this makes some of the helpers a bit more readable. (See the diff for examples.)

I considered adding some more short-hands that you could use in the macro, but most could just be written as a function that you use as a transformer. Even something like a let binding can be used as a transformer. Only exception to that was trying to make the suffix of the body that is the map explicit, but in the end that doesn't save you much more than the `{}` and is pretty confusing. So I stuck with this minimal for, which also kept the implementation small enough it went inline pretty cleanly.

### DIFF
--- a/src/clj/game/cards/programs.clj
+++ b/src/clj/game/cards/programs.clj
@@ -1352,13 +1352,13 @@
                                 (reveal state side topcard)
                                 (system-msg state :runner (str "reveals " topcard
                                                                " from the top of R&D"))
-                                (continue-ability state side (force-draw topcard) card nil))))}}}])
-  {:events [{:event :successful-run
-             :req (req (= :rd (target-server context)))
-             :async true
-             :interactive (req true)
-             :waiting-prompt true
-             :effect (effect (continue-ability rvl card nil))}]})
+                                (continue-ability state side (force-draw topcard) card nil))))}}}]
+    {:events [{:event :successful-run
+               :req (req (= :rd (target-server context)))
+               :async true
+               :interactive (req true)
+               :waiting-prompt true
+               :effect (effect (continue-ability rvl card nil))}]}))
 
 (defcard "Euler"
   (auto-icebreaker {:abilities [(break-sub 0 1 "Code Gate" {:req (req (= :this-turn (installed? card)))})
--- a/src/clj/game/cards/programs.clj
+++ b/src/clj/game/cards/programs.clj
@@ -700,7 +700,7 @@
              :effect (effect (add-counter eid card :virus 1 nil))}]
    :abilities [(break-sub
                  [(->c :virus 1)] 1 "All"
-                {:req (req (same-card? current-ice (:host card)))})]})
+                 {:req (req (same-card? current-ice (:host card)))})]})
 
 (defcard "Brahman"
   (auto-icebreaker {:abilities [(break-sub 1 2 "All")
--- a/src/clj/game/cards/programs.clj
+++ b/src/clj/game/cards/programs.clj
@@ -691,16 +691,16 @@
                    (strength-pump 3 3 :end-of-encounter {:cost-bonus discount-fn})]})))
 
 (defcard "Botulus"
-  (auto-icebreaker
-    (trojan
-      {:implementation "[Erratum] Program: Virus - Trojan"
-       :data {:counter {:virus 1}}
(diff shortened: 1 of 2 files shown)


Cluedrew:
I believe I have a work around for the error, but it feels like a false positive.

src/clj/game/cards/programs.clj
```diff
@@ -1353,13 +1352,13 @@
...
                                 (system-msg state :runner (str "reveals " topcard
                                                                " from the top of R&D"))
-                                (continue-ability state side (force-draw topcard) card nil))))}}}]
-    {:events [{:event :successful-run
-               :req (req (= :rd (target-server context)))
-               :async true
-               :interactive (req true)
-               :waiting-prompt true
-               :effect (effect (continue-ability rvl card nil))}]}))
+                                (continue-ability state side (force-draw topcard) card nil))))}}}])
```
NoahTheDuke, line 1355: I'm not okay with this. Let bindings are syntactically traceable, tooling expects them to wrap the sub forms, and moving the sub forms to the left 2 spaces is not worth the cost of "doesn't look like normal clojure code".

NoahTheDuke (changes requested):
transformers are cool, wrecking lets is not.

src/clj/game/cards/programs.clj
Cluedrew: Done. I had already kind of decided against it but got distracted by the build failures. Not for any tooling reason I didn't know about, but because if the let is bigger than a few lines - I don't have an exact number - it just becomes hard to follow just on screen. In fact I discarded an idea for a built-in let for this reason, a short-hand let doesn't fix the issue that the value in the let is complex enough to break the flow of the transformers.

src/clj/game/core/def_helpers.clj
```diff
@@ -323,11 +323,18 @@
                                  (unregister-ev-callback)
                                  (effect-completed state side eid)))))})))
 
+; (defcard title transformers* body)
+; Define a card to be returned from card-def. The definition consists of the
+; title (a string), 0 or more transformers and a body (an expression, usually
+; a map). Each tranformer should be a symbol or a list, symbols are wrapped
+; in lists and then the last transformer has the body inserted at the end,
+; then the second last has the result inserted as its last item and so on.
 (defmacro defcard
```
NoahTheDuke, line 332:
```suggestion
(defmacro defcard
  "Define a card to be returned from card-def. The definition consists of the
  title (a string), 0 or more transformers and a body (an expression, usually
  a map). Each tranformer should be a symbol or a list, symbols are wrapped
  in lists and then the last transformer has the body inserted at the end,
  then the second last has the result inserted as its last item and so on."
  {:arglists '([title ability] [title transformers* ability])}
```

NoahTheDuke: Moves the comment into a proper docstring, and adds manual `:arglists` that show the two arities.

Cluedrew: Docstrings! Right.

I don't know about closure, but I did try to use a convention I saw elsewhere. `*` is 0 or more and `+` is one or more. So my intent was to get into the one generic signature.

NoahTheDuke: Clojure supports multiple arities, and the `defn` macro will process your fucntion bodies to add metadata for the "arglists":

```clojure
(defn foo
  ([one] one)
  ([one two] {:first one :second two}))
```

is saved as `([one] [one two])` in the var's metadata, and then tooling will display it like `foo [one] [one two]`.

When you're doing preprocessing in a macro, it can be helpful to write them yourself to be extra clear, especially given that identifiers in clojure can contain `*` and thus a reader might not know they can skip the transformers.

PR #8022, opened 2025-03-19
Derez QOL

NBKelly:
This does the following:
* makes derezzing async
* allows derezzing multiple cards at once, suppressing the event (when manually derezzing as the corp for example), and suppressing the checkpoint (ie simultaneous effects, costs)
* cleans up the code for all the cards that derez
* automatically generates text for cards being derezzed
* adds a little helper function for taking n tags (did this just because of how baklan was formatted), I used it in a few places where it made sense, and cleaned up a couple of cards (like joshua B) in doing so

This also hits one of the long-standing todos in #6729 (though I can probably close that after elevation, and make a new/updated list)

### DIFF
--- a/src/clj/game/core/commands.clj
+++ b/src/clj/game/core/commands.clj
@@ -402,7 +402,7 @@
      {:prompt "Choose a card to derez"
       :choices {:card #(rezzed? %)}
       :async true
-      :effect (effect (derez eid target))}
+      :effect (effect (derez eid target {:no-event true}))}
      nil nil)))
 
 (defn command-trash
--- a/src/clj/game/core/commands.clj
+++ b/src/clj/game/core/commands.clj
@@ -401,7 +401,8 @@
      state side
      {:prompt "Choose a card to derez"
       :choices {:card #(rezzed? %)}
-      :effect (effect (derez target))}
+      :async true
+      :effect (effect (derez eid target))}
      nil nil)))
 
 (defn command-trash
--- a/src/clj/game/core/rezzing.clj
+++ b/src/clj/game/core/rezzing.clj
@@ -144,16 +144,21 @@
 
 ;; TODO: make async
 (defn- derez-message
-  [state side eid cards {:keys [source-card] :as msg-keys}]
+  ;; note:
+  ;;  source-card - the card that's derezzing (optional)
+  ;;  and-then - text to append to the end of the message (ie derezz x` and trash itself`)
+  ;;              I suggest only using this if the rhs thing is part of the same instruction.
+  ;;  include-cost-from-eid [eid] - include the last payment str from the eid as if it was for this
+  [state side eid cards {:keys [source-card and-then] :as msg-keys}]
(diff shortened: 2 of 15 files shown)


src/clj/game/core/rezzing.clj
```diff
@@ -143,23 +143,50 @@
...
+                 ;; this is currently only for lycian fixing subtypes on derez
+                 ;; should happen even if the card is disabled - nbk
+                 (resolve-ability state :corp derez-effect (get-card state c) cdef))
+               (when-let [derezzed-events (:derezzed-events cdef)]
+                 (register-events state :corp c (map #(assoc % :condition :derezzed) derezzed-events))))
+             (unregister-static-abilities state :corp c))
+           (update-disabled-cards state)
+           (when-not no-event
+             (queue-event state :derez {:cards cards
+                                        :side side}))
```
NoahTheDuke, line 186: in which cases would we not want to queue an event?

NoahTheDuke: no other handlers support this and none of your code in this PR uses it, so I don't think we should add it.

NBKelly: oops, I was meant to flag that as true for the derez command and the derez button on the gameboard (these are for fixing mistakes, and shouldn't fire events)

src/clj/game/core/rezzing.clj
```diff
@@ -143,23 +143,50 @@
...
+(defn- derez-message
+  ;; note:
+  ;;  source-card - the card that's derezzing (optional)
+  ;;  and-then - text to append to the end of the message (ie derezz x` and trash itself`)
+  ;;              I suggest only using this if the rhs thing is part of the same instruction.
+  ;;  include-cost-from-eid [eid] - include the last payment str from the eid as if it was for this
+  [state side eid cards {:keys [source-card and-then] :as msg-keys}]
+  (let [card-strs (enumerate-str (map #(card-str state % {:visible true}) cards))
+        prepend-cost-str (get-in msg-keys [:include-cost-from-eid :latest-payment-str])
+        title (or (:title source-card) (:printed-title source-card))]
```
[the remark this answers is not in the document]
NBKelly: true, it should just do that :joy:

src/clj/game/core/commands.clj
```diff
@@ -401,7 +401,8 @@
      state side
      {:prompt "Choose a card to derez"
       :choices {:card #(rezzed? %)}
-      :effect (effect (derez target))}
+      :async true
+      :effect (effect (derez eid target {:no-event true}))}
```
NoahTheDuke, line 405: this and the other derez command should probably also not trigger checkpoints, right?

NoahTheDuke:
this looks great, a very welcome change

src/clj/game/core/commands.clj
NBKelly: hmm, I think the checkpoint is probably fine (if only because it will force game values to be correct and force any lingering effects to be processed)

PR #8027, opened 2025-03-21
draft ids reworked/cleaned up

NBKelly:
I cleaned up the draft ids a bit.

My plan is to (sooner or later) go through all the old cards and clean them up/make them consistent with our modern implementations.

I added a `shuffle-cards-into-deck!` fn in `shuffling` that takes as input:
1) state
2) acting-side
3) source card
4) side having their cards shuffled
5) `& targets`

It:
* Handles the log printout, grouping cards by their zone, and displaying if the player shuffled, or was forced to shuffle, etc
* Handles the discard being zone-locked
* handles the cards not existing anymore, or already being in the deck
* Forces a shuffle even when there are no cards to get shuffled back (ie with gatekeeper)

I'll try and use this in the future when I clean up some other cards so we can avoid writing the same stuff over and over.

### DIFF
--- a/src/clj/game/cards/identities.clj
+++ b/src/clj/game/cards/identities.clj
@@ -52,7 +52,7 @@
    [game.core.say :refer [system-msg]]
    [game.core.servers :refer [central->name is-central? is-remote? name-zone
                               target-server zone->name]]
-   [game.core.shuffling :refer [shuffle! shuffle-into-deck]]
+   [game.core.shuffling :refer [shuffle! shuffle-into-deck shuffle-cards-into-deck!]]
    [game.core.tags :refer [gain-tags lose-tags]]
    [game.core.to-string :refer [card-str]]
    [game.core.toasts :refer [toast]]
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/identities.clj
```diff
@@ -2106,17 +2108,21 @@
...
-             :player :corp
-             :show-discard true
-             :msg (msg "shuffle " (if (:seen target) (:title target) "a card")
-                       " into R&D")
-             :effect (effect (move :corp target :deck)
-                             (shuffle! :corp :deck))}]})
+                            (has-most-faction? state :corp "Haas-Bioroid")))
+             :async true
+             :effect (req (if (empty? (:discard corp))
+                            (do (shuffle-cards-into-deck! state :corp card :card [])
```
NoahTheDuke, line 2114: `:card` is the target side having their deck shuffled?

src/clj/game/core/shuffling.clj
```diff
@@ -21,6 +23,30 @@
     (swap! state update-in [:stats side :shuffle-count] (fnil + 0) 1)
     (swap! state update-in [side kw] shuffle)))
 
+(defn shuffle-cards-into-deck!
+  "Shuffles a given set of cards into the deck. Will print out what's happened. Will always shuffle."
+  [state from-side card shuffle-side & targets]
```
NoahTheDuke, line 28: Can this take `targets`, not `& targets`? Just force everyone to put things in a list or vector?

src/clj/game/core/shuffling.clj
```diff
@@ -21,6 +23,30 @@
...
+  (let [targets (set (keep #(get-card state %) (flatten targets)))
+        targets (filter #(if (= (:zone %) [:discard]) (not (zone-locked? state shuffle-side :discard)) true) targets)
+        lhs (str " uses " (:title card)
+                 (when-not (= from-side shuffle-side)
+                   (str " to force the " (clojure.string/capitalize (name shuffle-side))))
```
NoahTheDuke, line 33:
```suggestion
                   (str " to force the " (str/capitalize (name shuffle-side))))
```

lol i expect there's a require line for this but if not: `[clojure.string :as str]` plz

src/clj/game/cards/identities.clj
NBKelly: whoops

PR #8043, opened 2025-03-30
Tr fixes

### LINKED ISSUE #8039: Odd text in Run menu
See attached image.

Image

### LINKED ISSUE #8040: Unknown Game Prompt (server-13) instead of the server name.
"Unknown Game Prompt (server-13)"
Image

Link to bug replay: https://jinteki.net/bug-report/dbf679da-fe38-4fa4-ac27-503602aa814b?b=0

NBKelly:
* If there's one spectator, it will say "1 Spectator" instead of just "Spectator"
* The server labels have been fixed for the install and run menus (when click run, or click a card in hand)

Closes #8040
Closes #8039

### DIFF
--- a/resources/public/i18n/en.ftl
+++ b/resources/public/i18n/en.ftl
@@ -599,7 +599,7 @@ game_show-decklists = Show/Hide decklists
 game_shuffle = Shuffle
 
 game_spec-count = {$cnt ->
-    [one] Spectator
+    [one] {$cnt} Spectator
     *[other] {$cnt} Spectators
 }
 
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
i think this means we shouldn't have one big selector for the prompts. where are these even coming from?

NBKelly:
They come from the backend (generate-runnable-servers, generate-installable-servers) - which gives the ordered list of servers which can be installed in/ran at

NoahTheDuke:
test paths weren't updated so the tr changes here are broken, which i suspect is why people are getting messed up strings currently.

PR #8045, opened 2025-03-30
Adding quotes for Liberation and Borealis Cycles

BlueOakData:
I am finally getting around to writing, at least a default quote, for the remaining IDs from Liberation and Borealis. These are pulled from the actual ID's flavor text. Ideally I will build on this in a future PR with some more unique quotes.

### DIFF
--- a/data/quotes-corp.edn
+++ b/data/quotes-corp.edn
@@ -1,4 +1,9 @@
-{"Acme Consulting: The Truth You Need"
+{"A Teia: IP Recovery"
+ {"Default"
+  ["Caveant Fugitivi."]}
+ "Acme Consulting: The Truth You Need"
+ {"Default"
+  ["Affordable, Effective, and Uncompromising."]}
  {"Anarch"
   ["Some grandstanding vigilante thinks they can threaten us? Our report will conclude all damage was self-inflicted. Case closed."]
   "Criminal"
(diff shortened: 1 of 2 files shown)


data/quotes-corp.edn
```diff
@@ -1,4 +1,9 @@
-{"Acme Consulting: The Truth You Need"
+{"A Teia: IP Recovery"
```
BlueOakData, line 1: Not sure if my formatting is correct here on this first line. Also some lines had the quote on the same line as the key (so "Default" [...] vs "Default"\n [...]"

data/quotes-runner.edn
```diff
@@ -369,11 +378,11 @@
   "Weyland Consortium"
   ["No matter how much they build, they'll never bury their secrets deep enough."]}
  "Nyusha \"Sable\" Sintashta: Symphonic Prodigy"
- {"Default" ["Every day, my mission is simple. Acquire credits or starve."]
+ {"Default" ["Flaws hold both beauty and opportunity."]
```
Cluedrew, line 381: Why the change here? Sable already had a unique default quote, so why switch back to the quote from the card?

BlueOakData: I thought the flavor-text derived quote was better than the one I had previously written. If y'all disagree and prefer the previous, I can revert the change.

NoahTheDuke: the square brackets indicate this is an array, meaning it can hold multiple quotes. each quote is surrounded by double-quotation marks and separated by spaces and/or linebreaks: ["quote 1" "quote 2" "quote 3" ...]

we're saying use both

BlueOakData: Oh..... my bad. Still getting used to Clojure and misunderstood what y'all meant.

BlueOakData:
Any other changes on this PR I can make?

PR #8046, opened 2025-03-30
use active player instead of side that triggered event

NBKelly:
I'm fairly certain we've hardcoded sides into dozens of event handlers to deal with this bug lmao

### DIFF
--- a/src/clj/game/cards/agendas.clj
+++ b/src/clj/game/cards/agendas.clj
@@ -158,7 +158,6 @@
   {:events
    [{:event :run
      :req (req (first-event? state side :run))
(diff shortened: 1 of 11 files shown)


NoahTheDuke:
omgggggggggg. some day when i have infinite free time, i want to remove the `side` parameter and track it all through the eid, but that's a nightmare of work, tbh.

NBKelly:
ok I stripped the hardcoded sides out of *almost* every event map, and then I noticed trigger-event-sync was still doing stuff wrong, so I made that fire in order of active-side -> opposing side

PR #8069, opened 2025-04-12
Card Back options

### LINKED ISSUE #8051: Allow for custom card backs
I think we can allow for custom card backs in the same way we do for custom backgrounds. Further, I think we could introduce a small set of features in this regard:

* [ ] Allow setting custom card-backs for each side
* [ ] Allow for setting an rgb color as your card back too (alternatively, we just have a small range of color options to pick from)
* [ ] Add a small range of official/tasteful designs as "sleeves" you can choose for each side
* [ ] Have a toggle allowing you to see the card back/sleeve your opponent picked (do not allow for showing your custom links to your opponent though, since that's a security breach), or always using your own choices for each side

I think this could be a nice flavorful addition to the user experience. We could even add special sets of sleeves that we can assign to users as tournament prizes and stuff like that.

NBKelly:
Refactors card backs to be on a side by side basis, allow adding more backs, allow only displaying your choices (ie I dont care what my opponent things looks cool), and allows for assigning card backs to individual players (ie as tournament prizes).

Right now:
* The card backs are distinct by side, some some can be runner only, or corp only, or both (ie ffg/nsg applies to both sides)
* You can choose to 1) display your opponents back 2) display your choices 3) only display nsg backs for your opponents 4) only display ffg backs for your opponents

Admin:
* I added a new route, `/prizes`
* Moderators and admins can access it.
* It allows you to load a player, adjust the prizes they have access to, then update the player with those prizes
* For the sake of demonstration, I put one piece of art by me in there, available only on the runner side

Because I was editing the set of options we're saving anyway, I made it so the chat log timestamps option *actually* gets saved too.

Closes #8051


src/clj/game/core/diffs.clj
```diff
@@ -366,11 +366,14 @@
...
    :background
    :card-resolution
+   :corp-card-sleeve
+   :runner-card-sleeve
    :language
    :pronouns
    :show-alt-art])
 
 (defn options-summary [options]
+  (println options)
```
NoahTheDuke, line 376: 👀

src/clj/web/prizes.clj
```diff
@@ -0,0 +1,49 @@
...
+   [web.mongodb :refer [find-maps-case-insensitive]]
+   [web.stats :refer [fetch-elapsed]]
+   [web.utils :refer [response]]
+   [web.ws :as ws]))
+
+(defn auth [_]
+  (response 200 {:message "ok"}))
+
+(defmethod ws/-msg-handler :prizes/load-user
+  prizes--fetch-user
```
NoahTheDuke, line 22:
```suggestion
  prizes--load-user
```

NoahTheDuke:
what's the migration path to keep people from having to select their preferred card back again?

NBKelly:
hmm - I didn't think that far enough ahead for data migration to be honest.

Everyone should get placed on the nsg frames by default, and I think it's easy enough to switch back to the ffg ones.

Anyway I noticed the sample prize/proof of concept I picked was too big, so I switched to one of my arts from april fools last year.

card-backs

NoahTheDuke:
I'd prefer to have some method of migration, so if you can add support through loading user data and or cookies, that would be killer. Like, changing the loading data to allow :card-backs and then save as the appropriate new key.

NoahTheDuke:
actually, that's fine don't worry about it.

NBKelly:
I updated this so at compile time (shadow-cljs build) it packs in whatever data is in `edn/card-backs.edn`, and merges that with the base options (ffg/nsg card backs).

Next step is to staple a lein task to update those.

NBKelly:
and it's done. There's now a task, `lein update-prizes` that will update the prize data and load up the images based off of this repo: https://github.com/NBKelly/netrunner-prizes/tree/main
I've added you to that repo as well.

src/clj/tasks/update_prizes.clj
```diff
@@ -0,0 +1,87 @@
...
+(def cli-options
+  [["-l" "--local PATH" "Path to fetch card edn from"
+    :validate [#(.exists (io/file (str % "/data/card-backs.edn")))
+               "Could not find local data file"]]
+   ["-i" "--card-images" "Fetch card images from the prizes repo (default)"
+    :id :card-images
+    :default true]
+   ["-j" "--no-card-images" "Do not fetch card images from the prizes repo"
+    :id :card-images
+    :parse-fn not]])
```
NoahTheDuke, line 74: This can just be "--[no-]card-images" without the -j one

NBKelly: I just used the same args as fetch because they're automatic for me now :joy:

Issue #8078, opened 2025-04-18
"Offer to trash like cards" Showing corp cards being installed to the runner


bimblybojang:
game 1

![Image](https://github.com/user-attachments/assets/053c56ae-7757-4ffc-93f1-0588162a6816)

![Image](https://github.com/user-attachments/assets/13103e39-d6b5-4393-99d4-da940469c4db)

https://www.jinteki.net/replay/84e8df07-924d-4e1b-ad51-a7d563f92dc4

[TEST.json](https://github.com/user-attachments/files/19812359/TEST.json)

game 2

![Image](https://github.com/user-attachments/assets/dae9a57a-88ee-4d0c-971b-9beffaf514e9)

![Image](https://github.com/user-attachments/assets/4612017d-4e1b-414e-9812-a133e35bb7f8)

![Image](https://github.com/user-attachments/assets/2aecdd6b-3399-45df-926d-1a82d7cc3b5b)

[testgame2.json](https://github.com/user-attachments/files/19812421/testgame2.json)

When the corp has "offer to trash like cards" active, the runner is able to see all cards being installed in the root of or protecting a server. In game 1 the runner visibly sees the ice in the positions on the remote, and in game 2, the runner is able to see ice in centrals and on the root of a remote. When the runner installs programs with the same option active, however, there is no such information displayed to the corp. This bug is easily reproducible across multiple versions.

version for game 1 : MacOs Firefox 137.0 (aarch64)
version for game 2: Windows 10 Firefox 137.0.2 (64-bit)

PR #8079, opened 2025-04-20
Elevation

NBKelly:
There's a lot to unpack here. I'm not going to comment on the cards. Here are the features and bugfixes that got rolled in:

### FEATURES/MAJOR CHANGES
* I updated all the cards in the game for the NCIGS changes. I did this before the "things in a run during run" rule got added, so there are probably a few ways to waste resources that shouldn't be there, but for the most part this should be good
* I updated the way runs are initiated for the run timing change. Runs now start n the initiation phase, and players may do whatever they like there before proceeding to approach/approach-server
* I added functionality for *the mechanic on Petty Cash*
* Added a system for the **Dividends** mechanic, and a helper to all the agendas for the old and the new way of doing it *(beale works the same, dividends-beale would scale from sansan)*
* Updated the cost system so cards like *Biawak* can work
* The 'do-nothing'/`ncigs` system got an overhaul. This means a lot of ice that can't do anything wont bother you to press useless buttons.
* Support for `:on-host` and `:in-rfg` as a locations for events and static abilities
* Support for `on-forfeit` events (this needs and auto-sets the `:in-rfg` key to work)
* The prevention system needed a touch more work so it would play nice with **shred**
* I had to update about 1200 unit tests to deal with the run timing change
* We can apply specific sounds to rezzing specific cards now, and also suppress the rez sound for rezzing certain cards. **Flyswatter** does this is it doesn't overlap the purge and the rez sounds.
* The turn-ends phase has a checkpoint when discarding cards. Also, discarding has a `waiting-prompt` so people can't fire abilities while you try to discard cards
* The `lein fetch` command is now capable of using the `-r --repo` and `-b --branch` arguments to designate the repo (default is `NoahTheDuke/netrunner-data`) and branch (default is `master`) to pick the data from when running a fetch
* As per the startup policy, startup decks can have 4 big agendas now
* Cards are only displayed as `new` while you're not selecting a card via a prompt. 
* I added the new step into the run timing diagram
* I updated the test framework for the run timing change. Tests that play events that make runs that aren't done through `play-run-event` basically have to rip a continue more in their code to not brick

------------------------
### BUGFIXES/QOL STUFF
* Added a `this-card-is-run-source` def into macros, so we can easily check if a card is the run source. This should let us fix anything relating to doppelganger (eventually) 
* Expends now interact with undo-click
* Corp-abilities on runner cards now properly display costs and card names
* Added helpers in `def-helpers` for:
  * draw-abi `(draw-abi x) -> "side draws x cards"
  * draw-loud `(draw-loud state side eid card n) -> draws a card loudly`
  * run-server-ability `(run-server-ability server) -> runs a server`
  * run-any-server-ability `runs any server (choice)`
  * run-remote-server-ability `runs any remote server`
  * run-central-server-ability `runs any central server`
  * run-server-from-choices-ability `(run-server-from-choices-ability [choices]) -> runs a server from among the choices`
  * in-hand*? `(in-hand*? [state card]) -> is the card playable as if it was in hand`
  * all-cards-in-hand* `(all-cards-in-hand* state side) -> all cards playable as if they were in the hand for a side)`
  * corp-install-up-to-n-cards `[n args]`
  * gain-credits-ability `[x]`
  * drain-credits `(drain-credits draining-side victim-side qty (opt: multiplier) (opt: tags-gained) -> for use with siphon-like effects)`
  * trash-on-purge `for our virus programs`
* Runs keep track of how many subroutines fired during them (required for **phoenix**)
* Cost messages for installing cards are auto-generated now, so we not write the same stuff over and over
* I added a few more helpers into the test framework
  * `is-deck-stacked? [state side expected-deck]` Is the deck stacked as we expect it to be? If you pick 3 cards out of a 30 card deck, that's still valid.
  * `provides-mu [cname x]` Does a card `cname` trivially provide `x` memory when installed by the runner.
* The trash-like-cards option now no longer leaks info. Whoops! Closes https://github.com/mtgred/netrunner/pull/8078


NoahTheDuke:
i didn't read any of the card changed or test changes, but the engine changes all looked good to me.

PR #8102, opened 2025-04-25
shadow-cljs actually rebuilds card-backs on file update

NBKelly:
This makes the macro yield a result which should be pinned to the contents of the data file. This forces shadow-cljs to actually reload the macro when the edn file changes.

### DIFF
--- a/src/clj/tasks/update_prizes.clj
+++ b/src/clj/tasks/update_prizes.clj
@@ -50,6 +50,10 @@
                         (with-open [w (io/output-stream card-path)]
                           (.write w body)))
                   (println "Error downloading art for card-back: " title error))))))
+        ;; this forces shadow-cljs to recompile the macro which depends on this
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
this won't work. it will return the sorted map, not the quoted form

NBKelly:
hmm - it's working locally I think. The macro defines the form, then emits the value (which shadow-cljs checks against when recompiling).
Before the changes, it was yielding nil as a value every time.

NoahTheDuke:
the macro returns a form which is placed into the form being read and evaluated. the returned form is a def which is evaluated to a var containing the merged data. i'm not sure what you're doing to get a nil.

NBKelly:
oh, actually you're write, I was getting tricked by some stale files once again.

NBKelly:
ok, new approach was just to touch the `prizes.cljc` file whenever prizes are fetched.
It feels like a hack, but every other alternative I looked at was 10x hackier, so :man_shrugging:

PR #8107, opened 2025-04-26
convert nrdb images to png format

### LINKED ISSUE #8100: Rename card image files from png to jpg
In the switch to NetrunnerDB v1 to v2, cards started getting downloaded in JPG format, but were still save with a ".png" extension.  We should switch to ".jpg" for consistency.

Note that double-sided cards in `resources/public/img/cards/en/default/stock/` are actually in PNG format and need to be converted to JPG or re-downloaded.

NBKelly:
this makes it so we convert the images we download to png rather than just save jpgs and pngs

(all the alt-arts, high def images, and foreign images we have are pngs already, and we historically used pngs before nrdb switched to jpg, so this is the path of least resistance I think)

Closes #8100

### DIFF
--- a/src/clj/tasks/nrdb.clj
+++ b/src/clj/tasks/nrdb.clj
@@ -13,7 +13,9 @@
     [tasks.utils :refer [replace-collection]]
     [tasks.images :refer [add-images]]
     [tasks.index :refer [create-indexes]]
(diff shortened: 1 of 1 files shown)


NoahTheDuke:
is this conversion lossless?

NBKelly:
that's... a very good question that I didn't even think about.

PR #8117, opened 2025-04-27
fix card backs for new accounts

NBKelly:
I was setting the wrong key for new accounts I think, so I corrected that, and added fallbacks for anyone who managed to save their card backs as an invalid value.

### DIFF
--- a/src/cljc/jinteki/card_backs.cljc
+++ b/src/cljc/jinteki/card_backs.cljc
@@ -8,7 +8,13 @@
                    :file "ffg"}
    :nsg-card-back {:description "The current Null Signal Games card backs."
                    :name "NSG Card Backs"
-                   :file "nsg"}})
+                   :file "nsg"}
+   ;; fallbacks, just in case
+   "" {:file "nsg"}
+   nil {:file "nsg"}
+   :ffg {:file "ffg"}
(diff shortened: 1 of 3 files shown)


src/cljc/jinteki/card_backs.cljc
```diff
@@ -8,7 +8,13 @@
                    :file "ffg"}
    :nsg-card-back {:description "The current Null Signal Games card backs."
                    :name "NSG Card Backs"
-                   :file "nsg"}})
+                   :file "nsg"}
+   ;; fallbacks, just in case
+   "" {:file "nsg"}
+   nil {:file "nsg"}
+   :ffg {:file "ffg"}
+   :nsg {:file "nsg"}
```
NoahTheDuke, line 16: Instead of assuming a fixed set of incorrect values, the function that calls/uses this should have a default value. i don't know how it works so i'm not sure where to put that, but it should look something like `(or (card-back-foo ...) "nsg")`.

NBKelly: Yeah, that's fair. I added a fallback to where we use cardbacks in the gameboard, but I kept the :nsg and :ffg keys because I *know* I got those wrong the first time, so a handful of accounts do actually have those keys. (whoops)

PR #8133, opened 2025-04-30
Update project-agenda Uses

Cluedrew:
The new `project-agenda`[`-helper`] function is a great example of a function that would work as a defcard transformer, unlike the similar but differently structured `agenda-counter` helper. At least one of the uses was not positioned in a way for a simple transformation and I left those alone. This should make the functions that were refactored to use the function more like their

I also did a bit of clean-up on the function itself:
- Removed the `-helper` suffix because it seemed like noise. If `project-agenda` is too brief then something like `add-project-on-score` might be more descriptive.
- Made the function private. It was not used elsewhere and it have very little reason to be, as it is mostly a
- Removed the unused args alias from inside the function.
- Reorder the math for the number of functions, reducing duplication and taking out an extra operation.

### DIFF
--- a/src/clj/game/cards/agendas.clj
+++ b/src/clj/game/cards/agendas.clj
@@ -84,7 +84,7 @@
                          :req (req (has-subtype? target subtype))
                          :value 1}]}))
 
-(defn project-agenda-helper
+(defn- project-agenda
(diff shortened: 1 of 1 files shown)


NBKelly:
neat :)

PR #8144, opened 2025-05-04
adding a "core" format that is System Gateway + Elevation

danlangford:
i looked at https://github.com/mtgred/netrunner/pull/8137 to determine what maybe needs to be done. im only about 75% confident in the translations. i did use ChatGPT to try to research what translated term FFG used when marketing other "core sets" to those regions.

I also noticed that some snapshot and snapshot-plus blurbs were still in help.cljs so I removed those while I was in there

Data PR here: https://github.com/NoahTheDuke/netrunner-data/pull/71

### DIFF
--- a/resources/public/i18n/en.ftl
+++ b/resources/public/i18n/en.ftl
@@ -379,6 +379,7 @@ format_name = {$format ->
     [casual] Casual
     [classic] Classic
     [core-experience] Core Experience
+    [core] Core
     [eternal] Eternal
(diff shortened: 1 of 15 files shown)


butzopower:
I left the snapshot blurbs in there as I wasn't sure if it was still useful to have the documentation on the format but also it is probably confusing if it's not a playable format. It's pretty discoverable with all the various links to `https://nullsignal.games/players/supported-formats/` anyways so makes sense to repurpose.

PR #8159, opened 2025-05-17
correct browser language hints

NBKelly:
This will filter things like: 
`en-US -> en`
`fr-CA -> fr`
`zh (of any sort) -> zh-simp`
`unsupported langauge -> en`

I tried playing around with my browser spoofing the locale and it seems to work.


NoahTheDuke:
ugh thank you

PR #8166, opened 2025-05-18
Load i18n ftl files from uberjar or local directory

xdg:
This PR amends ftl loading to support loading from an uberjar as well as the filesystem.  This allows packaging and serving ftl files from within a docker-ized deployment such as in `docker/Dockerfile.prod`, just as we do with CSS and other static assets.  (Except card images which are not bundled in the uberjar.)

I have tested this both locally with `lein run` and via `docker compose` on a private server.

As I've written this with the aid of AI tools and am not very fluent in Clojure, please do consider this change thoughtfully before merging.

### DIFF
--- a/src/cljc/jinteki/i18n.cljc
+++ b/src/cljc/jinteki/i18n.cljc
@@ -19,14 +19,16 @@
 #?(:clj
    (defn load-dictionary!
      [dir]
-     (let [langs (->> (io/file dir)
-                      (file-seq)
-                      (filter #(.isFile ^java.io.File %))
-                      (filter #(str/ends-with? (str %) ".ftl"))
-                      (map (fn [^java.io.File f]
-                             (let [n (str/replace (.getName f) ".ftl" "")
-                                   content (slurp f)]
-                               [n content]))))
+     ;; List of supported language files (based on resources/public/i18n directory)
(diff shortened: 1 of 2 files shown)


src/cljc/jinteki/i18n.cljc
```diff
@@ -19,14 +19,17 @@
...
+     (let [languages ["en" "fr" "ja" "ko" "la-pig" "pl" "pt" "ru" "zh-simp"]
+           ;; Try loading each language directly as a resource (works in both jar and filesystem)
+           langs (->> languages
+                      (mapv (fn [lang]
+                              (let [res-path (str dir "/" lang ".ftl")
+                                    ;; Try to load from classpath (including jar)
+                                    res (io/resource res-path)]
+                                (when res
+                                  [lang (slurp res)]))))
+                      (remove nil?))
```
NoahTheDuke, line 32:
```suggestion
           langs (->> languages
                      (keep (fn [lang]
                              (let [res-path (str dir "/" lang ".ftl")
                                    ;; Try to load from classpath (including jar)
                                    res (io/resource res-path)]
                                (when res
                                  [lang (slurp res)])))))
```

NoahTheDuke:
good catch about this.

Issue #8042, opened 2025-03-29
Bug using Physarum on Ravanna


Quarknator08:
Link to bug replay: https://jinteki.net/bug-report/769b1e20-ab2c-4ada-9cc8-cc389b7b5675?b=0

Description:

Basic run with a Physarum on Ravanna at R&D. When clicking the pay 2 to bypass, bug is reported.

PR #8179, opened 2025-05-20
Select prompts index by eid

### LINKED ISSUE #7255: I couldn't spend Paladin Poemu credits or choose cards in the heap for Steve Cambridge's ability
Link to bug replay: https://jinteki.net/bug-report/d2d9e3df-24c5-4ffd-ba79-6003b0f8eb5f?b=0

Description:

On turn 14 the corp plays retribution. After this action I was prompted to begin my turn, despite the corp only spending 1 click. Their second action to gain a credit happens before I'ce started my turn, and at that point we need to manually give the corp a credit as they final action. After this, I could not use credits hosted on Paladin Poemu, or choose cards from the heap for Steve Cambridge

### LINKED ISSUE #7830: Anansi won't le me draw a card if I encounter it with sysifus
As per the title: I encountered anansi a second time and let it fire. After clicking yes for the subroutine that lets the runner draw for 2 credits the game defaulted to me not drawing.

Link to the replay in question: https://jinteki.net/replay/caf6b448-fa17-44e8-8165-b5c679ccd901

### LINKED ISSUE #7985: Runner cannot discard cards
In the middle of the game I could not discard any card in discard phase playing as a runner. Previous game with the same person was okay. No previous issues experienced with Jinteki before.

Refreshing the game or reloading the browser did not solve the issue even make it worse - cursor icon has changed from "target" to regular.

Workaround was to:

1) Use /discard #n 
2) Use /undo-click
3) Repeat the click
4) End of turn (only appeared in this sequence, just using step 1) still no "End of turn" was visible and stuck on discarding)

Replay below:

https://www.jinteki.net/replay/61267d70-9f6f-47cb-9148-4c991cf93f73?n=82&d=0

### LINKED ISSUE #8163: Add a `clear-stuck-eids` command
Sometimes, either through repeated requests from a client getting stacked on the same eid, or a player spamming commands or something silly, we get into a state where each command is linked to the wrong eid, and this can actually lead to unsalvageable games (ie when you click your cohort trigger, the card gets `/rfg`'d from a previous `/rfg` command.

I want to add a "we really just need this game to work" command to at least be able to salvage these games, that just clears out every eid attached to a prompt or pending ability or something.

NBKelly:
Closes #8163 (by removing the breaking condition, it wasn't eids, it was always pulling the topmost prompt, which was sometimes wrong)

I'm 99% sure this was the cause of #7985, so Closes #7985

I'm 99% sure this was the cause of #7830, so closes #7830

I'm 99% sure this was also the cause of #7255, so closes #7255

There's a chance it could have been the cause of the following issues:
#8042

### DIFF
--- a/src/clj/game/core/actions.clj
+++ b/src/clj/game/core/actions.clj
@@ -19,7 +19,7 @@
     [game.core.play-instants :refer [play-instant]]
     [game.core.expend :refer [expend expendable?]]
     [game.core.prompt-state :refer [remove-from-prompt-queue]]
-    [game.core.prompts :refer [resolve-select]]
+    [game.core.prompts :refer [resolve-select first-prompt-by-eid]]
     [game.core.props :refer [add-counter add-prop set-prop]]
     [game.core.runs :refer [continue get-runnable-zones]]
     [game.core.say :refer [play-sfx system-msg implementation-msg]]
@@ -220,7 +220,7 @@
 (defn resolve-prompt
   "Resolves a prompt by invoking its effect function with the selected target of the prompt.
   Triggered by a selection of a prompt choice button in the UI."
-  [state side {:keys [choice] :as args}]
+  [state side {:keys [choice eid] :as args}]
   (let [prompt (first (get-in @state [side :prompt]))
         effect (:effect prompt)
         card (get-card state (:card prompt))
@@ -281,7 +281,7 @@
 (defn select
   "Attempt to select the given card to satisfy the current select prompt. Calls resolve-select
   if the max number of cards has been selected."
-  [state side {:keys [card shift-key-held]}]
+  [state side {:keys [card shift-key-held eid]}]
   (let [target (get-card state card)
         prompt (first (get-in @state [side :selected]))
         ability (:ability prompt)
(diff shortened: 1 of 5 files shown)


NBKelly:
actually, I think there's still one more section to hit

NoahTheDuke:
I feel like there's a better underlying abstraction here, but it's hard to know what exactly. Having to manually write `:eid (get-eid ...)` on every click handler is hell lol. I thought I put uuids on every prompt already but those aren't sent to the client (maybe they should be). Can a given eid be tied to multiple prompts? I think so but I'm not sure.

Maybe there could be an equivalent to the "click-card" test helper that correctly finds and sends the appropriate message? So each handler just says `(click-card {:choice "poop"})` and then it sends the right stuff.

NBKelly:
I think the uuids are on the buttons themselves (which one you click, for cases with identical names/cards).
I don't think it's possible for an eid to be tied to multiple prompts unless something very cursed is going, just since `wait-for` will be making a new eid every time. There is an exception I think when manually creating waiting prompts though. In any case, the worst case scenario for this is that we do what we did before, which is only touch the first prompt.

Issue #7889, opened 2024-11-30
Liberation cycle sysops shouldn't move to their own servers.


Larrea-Mon:
Spotted this with Holoman but maybe it applies to the others too, i haven't done the testing.
Right now they can use their ability to move themselves to their current server. This is an issue only with Isaac Liberdade who has an ability relating to this, but also I think the menu to move the card should have a "cancel" option, and hide the "current server" option.

PR #8185, opened 2025-05-21
Prevent mobile sysops from moving to its own server

csbisa:
Exclude the current server from options to prevent mobile sysops (i.e. Isaac Liberdade) from triggering itself in the same server.

Fixes https://github.com/mtgred/netrunner/issues/7889

### DIFF
--- a/src/clj/game/cards/upgrades.clj
+++ b/src/clj/game/cards/upgrades.clj
@@ -7,7 +7,7 @@
                              steal-cost-bonus]]
    [game.core.bad-publicity :refer [lose-bad-publicity]]
    [game.core.board :refer [all-active-installed all-installed all-installed-corp card->server
-                            get-remotes server->zone server-list]]
+                            get-remotes server->zone server-list server-list-exclude]]
    [game.core.card :refer [agenda? asset? can-be-advanced?
                            corp-installable-type? corp? get-card get-counters get-zone
                            has-subtype? ice? in-discard? in-hand? installed? operation? program? resource? rezzed?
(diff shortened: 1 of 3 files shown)


NBKelly:
Neat, that's a pretty elegant fix. I wonder if it's worth stapling it onto a few other cards? 

I think relevant cards would be:
* Ablative Barrier  
* Georgia Emelyov 
* Mind Game
* Bloom
* Sandstorm
* Tributary

csbisa:
Here's the current functionality:
- remove current card's server from installable-servers
ablative: `this (zone->name (second (get-zone card)))` `:choices (req (remove #(= this %) (installable-servers state nice)))`
trib: `:choices (req (remove #(= this %) (installable-servers state nice)))`
bloom: `:choices (req (remove #(= this %) (installable-servers state nice)))`
- remove current run's server from all servers
mind game: `:choices (req (remove #{(-> @state :run :server central->name)} servers))`
sand storm: `:choices (req (remove #{(zone->name (:server (:run @state)))} (cancellable servers)))`
- remove current card's from all servers
georgia emelyov: `:choices (server-list state)`
mobile sysops (i.e. this PR)

Looks like georgia and the mobile sysops should also switch to using installable-servers although I don't think there's any functional difference today. So I might just delete `server-list(-exclude)` and modify installable-servers instead. BTW, do you have any suggestions on how to make the new function more clojurey? `server-list-exclude` feels kind of iffy. Maybe either pass in a `:exclude` kw or tweak it a bit further to take card instead of zone since that's the only use-case today.

NoahTheDuke:
there's a couple ways to do it, so depends on your preferences. two functions, a function with two arities, a function that takes an optional map. when i'm at my computer again, i can write them out for you

PR #8187, opened 2025-05-21
Prevent `:last-update` from being updated when leaving/joining lobby

tylerjallen02:
Prevents lobby inactive "timer" from being reset whenever a player leaves or joins a lobby. This should prevent AFK games from being kept alive when players unknowingly try to join (and leave) AFK games. 

Timer should still be reset for other triggers such as sending a message.


NoahTheDuke:
Nice fix!

PR #8192, opened 2025-05-22
update core set cards

### LINKED ISSUE #8017: Free Lunch did not log the use of its power counters
Link to bug replay: https://jinteki.net/bug-report/dc79df66-0779-4fc2-9d38-9ffa0c091cc0?b=0

Description: when the GAMEnet corp used the counters on Free Lunch, they did not log. I haven't repeated this yet so idk if it's a general problem with Free Lunch or a niche GAMEnet bug.

NBKelly:
Just trying to bring some of the older impls in line with what our engine is capable of now.

Features here:
* Agendas
  * ABT no longer leaks information if you have no valid targets
  * Nisei MkII can only be used during a run
  * Priority requisition looks nicer
*  Assets
   * Byte! and Snare! no longer waste time if the corp can't pay
   * The PAD Campaign type cards now all use a single defn, instead of us rewriting the same thing 5 times
* Ice
   * the `power-counter-ability` function got simplified - turns out it was trying to transcribe a `:message` key from abilities into a `:msg` key, so this fixes #8017
* Identities
  * PE is properly interactive
  * Noise is interactive (matters for sabotage interactions)
  * Kate no longer clutters the log telling you she discounted your stuff by 1 (no other card does this)
* Operations
   * I added a feature to the `x-credits` cost that allows you to specify a `:maximum` key, which can be either a number or a 5-fn. That's the max you're allowed to spend
   * I added the ability for `play-instant` to check for a `:base-play-cost` key, which will be used instead of the `play-cost` that `cost_fns` spits out, so we can handle cards like `psychographics` more naturally. ~~(TODO - make sure this factors in play cost increases, like donut taganes)~~
   * I added a `tutor-abi` fn into `def-helpers`. For now I marked aggressive negotiation as using it. We have so many tutor effects in the game I'm shocked we haven't done this already.
   * I simplified a couple more operations
* Programs
   * Djinn uses the tutor abi 
* Hardware
  * No changes
* Resources 
  * Wyldside should be a tiny bit more natural to use
* Events
  * Modded and career fair use `runner-can-pay-and-install?`
  * A few cards are basically just calling a helper and that's it (diesel, easy mark)
  * Infiltration looks nicer
  * Inside job now plays nicely with doppelganger
  * Special Order uses the `tutor-abi` helper

Bonus:
* I updated Ad Blitz, Brute Force Hack, recruiting trip and Bribery to make use of the `:base-play-cost` feature.
* IP Enforcement can't do it, because it needs the tag cost to match the credit cost - it could probably just use another cost handler, but that's gross
* Brute force hack isn't as clean as it could be, because prompts strip the eid, so the prompt cannot refer to the costs in the eid.


NBKelly:
I cleaned up the code a bit more, added support for handling cost modifiers for `:x-credit` costs, and make the eids for choices prompts keep relevant info for calculation.

PR #8208, opened 2025-05-28
Add card name auto-completions

### LINKED ISSUE #8024: Autocomplete card titles when writing commands in the game log
When we do a /summon or /replace-id command, we need to get card titles perfect. We already autocomplete the command for you, I wonder if it would be possible to do the same for the title of the card you are typing?

strongoose:
Add card name completions to chat commands which require a card name.

Closes #8024

### DIFF
--- a/src/cljs/nr/gameboard/log.cljs
+++ b/src/cljs/nr/gameboard/log.cljs
@@ -167,7 +167,6 @@
     :corp   (= (:side card-info) "Corp")
     :runner (= (:side card-info) "Runner")))
 
-; TODO refactor into complete-id, move completion logic into generic function and add the replacement bits in here
 (defn complete-cardname [state full-input card-input]
   (let [cardnames (->> @all-cards
                        (filter filter-side)
--- a/src/cljs/nr/gameboard/log.cljs
+++ b/src/cljs/nr/gameboard/log.cljs
@@ -90,14 +90,13 @@
             next-index
             (+ score (or next-index 0))))))))
 
-(defn find-command-matches
-  ([input commands]
-   (when (= "/" (first input))
-     (->> commands
+(defn find-matches
+  ([input potential-matches]
+     (->> potential-matches
        (map (fn [target] {:match target :score (fuzzy-match-score input target)}))
        (filter :score)
        (sort-by :score)
-       (map :match)))))
+       (map :match))))
 
 (defn show-command-menu? [s]
   (seq (:command-matches s)))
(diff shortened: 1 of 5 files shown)


src/cljs/nr/gameboard/log.cljs
```diff
@@ -88,19 +77,23 @@
...
-      (if (not (seq rest-input))
-        score
-        (let [next-index (string/index-of target (first rest-input) (inc target-index))]
-          (recur
-            (first rest-input)
-            (rest rest-input)
-            next-index
-            (+ score (or next-index 0))))))))
+  ;; scoring is case insensitive
+  ;; TODO - scoring for similar letters (e.g. Şifr =~ sifr, 1337 =~ leet)
```
NBKelly, line 81: I think we can use java.text.normalizer to deal with non-latin characters. No clue what the smart way to do the numbers is though.

strongoose: oh wait, **java**.text.normalizer? We're in ClojureScript land here 😅

strongoose: there is a [js equivalent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize) though which I think I can make work

strongoose:
Thanks for the comments both - have addressed them and added the implementation for `/replace-id` (plus a small QoL tweak to filter cards by side). Removing WIP as I think those are the two commands that need completion for now.

src/cljs/nr/gameboard/log.cljs
```diff
@@ -74,95 +75,150 @@
...
-(defn command-menu-key-down-handler
+(defn has-args? [completion]
+  (some? (get-in command-info-map [completion :has-args])))
+
+(defn autosend? [completion]
+  ;; Commands with arguments do not autosend
+  (if (and (is-command? completion) (has-args? completion))
+    false
+    ;; Other completion types (commands with no args, card completions) do autosend
+    true))
```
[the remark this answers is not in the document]
NoahTheDuke: jk this is fine.

src/cljs/nr/gameboard/log.cljs
```diff
@@ -74,95 +75,150 @@
...
-                       (not (get-in command-info-map [command :has-args])))
+            (when (and (= key "Enter") (autosend? completion))
               (send-msg state))))
         ;; else
         nil))))
 
+(defn complete-command [state input]
+  (swap! state assoc :completions
+         (->> (find-matches commands input)
+              (map (fn [match] {:completion-text match :display-text (get-in command-info-map [match :usage])})))))
```
NoahTheDuke, line 163:
```suggestion
              (mapv (fn [match] {:completion-text match :display-text (get-in command-info-map [match :usage])})))))
```
`map` is lazy, we want it to be done immediately.

Issue #8212, opened 2025-05-31
Bug Report- IP Enforcement


MirrorCubeSquare:
Cards Involved: IP Enforcement & Stoke the Embers
Description: If using IP Enforcement to took back Stoke The Ember, triggered ability of Stoke the Embers will Trigger twice instead of once
not sure about whether still having tags after IP Enforcement matters because both times the bug happens  it does have

NBKelly:
Do you mean that it prompted you twice to use stoke?
Or that stoke ended up with 2 agenda counters on it (which is correct, because ip enforcement placed a counter too)?

MirrorCubeSquare:
yes it prompted me twice (ending 3 agenda points if i choose yes for both)
===========
when  i test played myself just now , i found i cant replicate the bug either, maybe trigger condition so complicate?
but i remember the same bug happend twice during normal plays(making both me and my opponent surprise,and even undo click to double check)

PR #8215, opened 2025-05-31
test for stoke vs ip enforcement

NBKelly:
I couldn't reproduce #8212 in a unit test


NBKelly:
I *think* I have *maybe* found the cause of this - inspecting the state, I noticed that the prompt queues were typed as lazyseqs, which can potentially cause issues in cases where state is copied or changed ie:
1) concurrent edits to the same atom (we get half way through a state update, then throw it away because the update conflicted with another update)
2) when we reset the state with undo-click or undo-turn
Again, this is something rather cursed, but I haven't been able to reliably reproduce it, even writing unit tests that fully simulate big games :weary:

PR #8229, opened 2025-06-06
Allow /save-replay during prompts

NoahTheDuke:
`/save-replay` should be allowed at all times, cuz it's not related to the game itself.


NBKelly:
good catch, I missed that one

PR #8233, opened 2025-06-11
bigger picture does not double-print it's outcome anymore

NBKelly:
Fixes the double-printing here:
bigger-picture_001


jdeng89:
bad bot

PR #8234, opened 2025-06-11
Make giving tags use an ability

NBKelly:
I moved the `give-tags` ability from `ice.clj` into `def_helpers.clj`, and made all the non-ice card files use it.

This change should be invisible I think, I just got sick of writing the same code like 50 different times.

### DIFF
--- a/src/clj/game/core/def_helpers.clj
+++ b/src/clj/game/core/def_helpers.clj
@@ -187,7 +187,7 @@
    (resolve-ability state side eid (draw-abi n args) card nil)))
 
 (defn give-tags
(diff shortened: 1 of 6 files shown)


src/clj/game/core/def_helpers.clj
```diff
@@ -186,6 +186,15 @@
   ([state side eid card n args]
    (resolve-ability state side eid (draw-abi n args) card nil)))
 
+(defn give-tags
+  "Basic give runner n tags subroutine."
```
NoahTheDuke, line 190:
```suggestion
  "Basic give runner n tags ability."
```

PR #8246, opened 2025-06-16
Synapse vs hannah

### LINKED ISSUE #8245: bug report -interaction between hannah and Synapse Global
Description:
When runner is at 0 tag, and use hannah for additional click(which is legal)
the Synapse Global would trigger

though in reality since runner have 0 tag to remove, the ability won't trigger

=======possible fix=======
maybe code adjustment of hannah, to not trigger"lose tags" event when runner have 0 tag?

NBKelly:
Adjusts our tag loss event to constrain the tags lost to the number of tags there are to lose.

Additionally, adjusts amanuensis, valentina, and synapse to check that there is actually a positive number of tags removed.

Closes #8245

### DIFF
--- a/src/clj/game/core/tags.clj
+++ b/src/clj/game/core/tags.clj
@@ -73,11 +73,12 @@
   ([state side eid n {:keys [suppress-checkpoint] :as args}]
    (if (= n :all)
      (lose-tags state side eid (get-in @state [:runner :tag :base]))
-     (do (swap! state update-in [:stats :runner :lose :tag] (fnil + 0) n)
-         (deduct state :runner [:tag {:base n}])
-         (update-tag-status state)
-         (queue-event state :runner-lose-tag {:amount n
-                                              :side side})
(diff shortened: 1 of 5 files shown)


src/clj/game/core/tags.clj
```diff
@@ -73,11 +73,12 @@
...
      (lose-tags state side eid (get-in @state [:runner :tag :base]))
-     (do (swap! state update-in [:stats :runner :lose :tag] (fnil + 0) n)
-         (deduct state :runner [:tag {:base n}])
-         (update-tag-status state)
-         (queue-event state :runner-lose-tag {:amount n
-                                              :side side})
-         (if suppress-checkpoint
-           (effect-completed state nil eid)
-           (checkpoint state eid))))))
+     (do (let [n (min n (get-in @state [:runner :tag :base]))]
```
NoahTheDuke, line 76: `let` has an implicit `do`, so you don't need to wrap this in a `do`.

PR #8261, opened 2025-06-22
Installs during access are not candidates

### LINKED ISSUE #8181: "Poétrï Luxury Brands" can be a definite AMBUSH attack if the agenda is stolen
Poétrï Luxury Brands: while resolving a remote server's agenda stole process with All the Rage's ability, Corp installed ambush (byte!) on the server Runner was accessing and it started accessing continuously as is. (Runner has no choice).

As stated in URL(*), there is a nasty rule that says “if more cards are added to a route while breaching that route, the runner can deny access to those new cards” (whereas if more cards are added in the central server, they cannot be denied), so that rule prohibits forced access!

(*)https://nullsignal.games/blog/ashes-remaster-rules-update/

[Poétrï Luxury Brands: All the Rage · NetrunnerDB](https://netrunnerdb.com/en/card/35036)

This is a somewhat difficult part of the rule, so please let me know if I am mistaken.

Thanks to all who always make jinteki.net better.

NBKelly:
Closes #8181

I split `access-card` into `access-card` and `access-card-continue`, and added logic for if the card was installed during the active breach (ie for loki ansel, or poetri nonsense).

Then I made it so that if you access a card that was installed during the current breach, the game will ask you for consent.

### DIFF
--- a/src/clj/game/cards/identities.clj
+++ b/src/clj/game/cards/identities.clj
@@ -2089,6 +2089,7 @@
                :async true
                :prompt "Install a non-agenda from HQ?"
                :change-in-game-state {:silent true :req (req (seq (:hand corp)))}
+               :waiting-prompt true
                :choices {:card (every-pred corp? in-hand? (complement agenda?) (complement operation?))}
                :effect (req (corp-install state side eid target nil {:msg-keys {:install-source card}}))}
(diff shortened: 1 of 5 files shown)


NoahTheDuke:
good change!

PR #8263, opened 2025-06-22
Mystic Maemi random trash cost gives more info

NBKelly:
Also just made the cost in general more informative on the runner side.

I'll fix the test a little later when I figure out what the heck is wrong here...
test


NoahTheDuke:
it uses regex so you gotta escape the parenthesis i think

PR #8278, opened 2025-07-05
Make run helper more convenient

### LINKED ISSUE #8270: Eru effect does not persist after being trashed by Hammer mid-run
Eru's successful run effect is attached to the run, and should still apply when trashed mid-run (the Threat 3 text should not apply regardless of threat however). As is, Eru will not apply and the runner breaches Archives.

NBKelly:
I adjusted the helpers for `run-server-ability` and `run-any-server-ability` so they can:
1) Apply events before making the run (and make them persist if the card dies)
2) Play a noise if the ability was an action

Closes #8270


NoahTheDuke:
this is really nice

PR #8286, opened 2025-07-11
Timbre stratify logs

### LINKED ISSUE #8285: Update our logging to split data into relevant files
Write now, we just pipe everything into a single file (usually `clojure.log`).
We could reroute everything through timbre (which we already use) and intelligently pipe stuff into say, 
* `clojure.log` for routine stuff
* `exceptions.log` for exceptions.
* `mod-actions.log` for mod actions (they currently appear in our clojure.log)

Maybe there's one or two other things that should be there, idk

Additionally:
* exceptions should be timestamped where possible
* I should make the `continue pressed at wrong time` and `no such prompt type` exceptions print out the last of the chat log too

NBKelly:
* All (just about) of our logging now goes through timbre
* The routine stuff goes in clojure.log
* All exceptions go into exceptions.log, and also go to stdout (so you can actually debug stuff)
* All mod actions go to mod-actions.log
* These logs are not erased between restarts
* The run exceptions and other ones that didn't spit out relevant info now do

Additionally
* The exceptions and routine logs get cycled between each restart
* the mod actions one can probably just never get deleted, it's fine

Outside of this repo:
* Update how we run jinteki to accomodate this
* make the exceptions and routine logs available for inspection
* make the mod-actions log explicitly not available for inspection (except maybe with credentials? idk)

Closes #8285

### DIFF
--- a/src/clj/game/core/actions.clj
+++ b/src/clj/game/core/actions.clj
@@ -22,13 +22,14 @@
     [game.core.prompts :refer [resolve-select first-prompt-by-eid first-selection-by-eid]]
     [game.core.props :refer [add-counter add-prop set-prop]]
     [game.core.runs :refer [continue get-runnable-zones]]
-    [game.core.say :refer [play-sfx system-msg implementation-msg]]
+    [game.core.say :refer [play-sfx system-msg implementation-msg n-last-logs]]
     [game.core.servers :refer [name-zone zones->sorted-names]]
     [game.core.to-string :refer [card-str]]
     [game.core.toasts :refer [toast]]
     [game.core.update :refer [update!]]
     [game.macros :refer [continue-ability req wait-for]]
-    [game.utils :refer [dissoc-in quantify remove-once same-card? same-side? server-cards to-keyword]]))
+    [game.utils :refer [dissoc-in quantify remove-once same-card? same-side? server-cards to-keyword]]
+    [taoensso.timbre :as timbre]))
 
 (defn- update-click-state
   "Update :click-states to hold latest 4 moments before performing actions."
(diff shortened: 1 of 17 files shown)


src/clj/game/core/actions.clj
```diff
@@ -206,9 +207,9 @@
 
 (defn- prompt-error
   [context prompt prompt-args]
-  (.println *err* (with-out-str (print-stack-trace (Exception. (str "Error " context)))))
-  (.println *err* (str "Prompt: " prompt))
-  (.println *err* (str "Prompt args: " prompt-args)))
+  (timbre/error (str (with-out-str (print-stack-trace (Exception. (str "Error " context))))
```
NoahTheDuke, line 210: let's use a real exception here instead of printing it

src/clj/game/core/runs.clj
```diff
@@ -526,18 +527,21 @@
 (defmethod start-next-phase :default
   [state _ _]
   (when-not (= :success (:phase (:run @state)))
-    (.println *err* (with-out-str
-                      (print-stack-trace
-                        (Exception. "no phase found and not accessing cards")
-                        2500)))))
+    (timbre/error (str "start-next-phase :default: \n"
```
NoahTheDuke, line 530: everywhere it prints a stack trace, just pass the exception to timbre.

src/clj/web/stats.clj
```diff
@@ -113,11 +114,11 @@
   (doseq [player original-players]
     (if (:side player)
       (inc-game-stats db (get-in player [:user :_id]) (game-record-start player))
-      (println "NULL start player side in stats for gameid" gameid)))
+      (timbre/error (str "NULL start player side in stats for gameid" gameid))))
   (doseq [player ending-players]
     (if (:side player)
       (inc-game-stats db (get-in player [:user :_id]) (game-record-end state player))
-      (println "NULL end player side in stats for gameid" gameid))))
+      (timbre/error (str "NULL end player side in stats for gameid" gameid)))))
```
NoahTheDuke, line 121: this and the others will result in "NULL end player ... for gameid12345" because `println` adds spaces between args and `str` doesn't.

src/clj/web/ws.clj
```diff
@@ -26,7 +26,7 @@
               ajax-post-fn ajax-get-or-ws-handshake-fn private]} chsk-server
       conns_ (:conns_ private)]
   (defonce handshake-handler (fn [& args] (try (apply ajax-get-or-ws-handshake-fn args)
-                                               (catch Exception _ (println "Caught an error in the handshake handler")))))
+                                               (catch Exception _ (timbre/error "Caught an error in the handshake handler")))))
```
NoahTheDuke, line 29:
```suggestion
                                               (catch Exception ex (timbre/error ex "Caught an error in the handshake handler")))))
```

PR #8288, opened 2025-07-12
Fix get-image-path to always return an array

xdg:
Function previously returned either a string or array based on the images field value. This inconsistency caused issues in calling code that expected a uniform return type.

This should fix flip identity images in historical replays saved when images were strings, not an arrays.

The changes that made code expect an array and the database store an array were done in 29f91b0bb and befb48262, respectively.

### DIFF
--- a/src/cljs/nr/utils.cljs
+++ b/src/cljs/nr/utils.cljs
@@ -381,23 +381,28 @@
       (reset! scroll-top-atom h))))
 
 (defn get-image-path
-  "Image search priority: Resolution > Language > Art"
+  "Image search priority: Resolution > Language > Art. Always returns an array."
   ([images lang res art] (get-image-path images lang res art 0))
   ([images lang res art depth]
    (when (< depth 4)
-     (or (get-in images [lang res art])
-         ;; try to downgrade the res first
-         (and (not= res :default)
-              (get-image-path images lang :default art (inc depth)))
-         ;; try to downgrade the language second
-         (and (not= lang :en)
-              (get-image-path images :en res art (inc depth)))
-         ;; try to downgrade the art last
-         (and (not= art :stock)
-              (get-image-path images lang res :stock (inc depth)))
-         ;; fallback if on 0 depth only
-         (and (zero? depth)
-              "img/missing.png")))))
+     (let [result (or (get-in images [lang res art])
+                      ;; try to downgrade the res first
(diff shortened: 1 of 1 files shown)


xdg:
Sorry -- I think this is a good defensive fix, but after deploying to a private server and testing, it doesn't seem to fix replays.  Maybe the replay image code is different?  I'll keep looking.

src/cljs/nr/utils.cljs
```diff
@@ -381,23 +381,28 @@
...
+                      (and (not= art :stock)
+                           (get-image-path images lang res :stock (inc depth)))
+                      ;; fallback if on 0 depth only
+                      (and (zero? depth)
+                           "img/missing.png"))]
+       (if (sequential? result)
+         result
+         (if result
+           [result]
+           nil))))))
```
NoahTheDuke, line 405: This should be (when result (if (sequential? result) result [result]))

NBKelly:
replays actually have the image urls hardcoded in them (in the same way that the gamestate does)
if you examine a replay file, you'll see what I mean

there's two issues with older replays:
1) the images aren't arrays (that's easily fixable with any text editor)
2) there are pointers to the override images we had that we removed, fixing these is more annoying to do

xdg:
I'd like to be able to fix up replays out of the database on the fly, but I haven't been able to track down where the replay images are getting wired into the UI.

NoahTheDuke (approved):

NoahTheDuke:
i just saw that you used claude to author some commits here (and in other prs it seems). i do not like or agree with LLM-generated code at all, and will not be merging this PR as i cannot ethically or morally accept it. i'll see about fixing this bug myself.

xdg:
I think I've been very transparent that I've used LLMs to be able to contribute in a (relatively-obscure) language that I am less fluent in than the ones I work in every day.  I apologize for offending your sensibilities and encourage you to update CONTRIBUTING.md to make your position clear.

NoahTheDuke:
I sat on it over the weekend, and realized that I have overstepped my responsibilities. My personal feelings on LLMs are not those of the jnet team, nor of minh specifically, and it's not my place as merely the current "most senior" contributor to demand others conform to my feelings. I'm sorry I said what I said, I'm thankful for the work you've done in the past, and I welcome your contributions in the future.

PR #8289, opened 2025-07-12
Update Chinese and English localization files; Updating the Label can display localized card title; fix ingame-settings localized display

klingeling:


src/cljs/nr/gameboard/board.cljs
```diff
@@ -753,7 +753,7 @@
            side facedown card-target icon new ghost runner-abilities subroutines seen
            subtype-target corp-abilities flashback-fake-in-hand flashback-playable]
     :as card} flipped disable-click]
-  (let [title (get-title card)]
+  (let [title (or (-> (get @all-cards (get-title card)) :localized :title) (get-title card))]
```
NoahTheDuke, line 756: should this just be in get-title?

klingeling: (tr-data :title (get @all-cards (:title card))) can work

PR #8318, opened 2025-08-05
Track source card for rfg

NBKelly:
I added this under macros so I can expand it later (ie, with `trash-can`).

Since continue-ability (and resolve-ability in general) seem to refresh the reference to the card, abstracting some of these is awkward, so I added a macro to make it a bit easier.

Makes it so spin and moon pool log correctly - get-card doesn't work because the card is no longer there (it was rfg'd), so we can track the source via the eid and costs paid.


NoahTheDuke:
could you pass the moved card to :paid/targets? i think we do that in other cases

PR #8348, opened 2025-09-04
false lead can be queued to prompt after the runner turn begins trigg…

### LINKED ISSUE #8338: Queue false lead to fire on (post) runner turn start
We can set up a dummy event for this. This will just be a shortcut so you can avoid click racing people.

NBKelly:
…ers have fired

Just some QOL to reduce the friction when using false lead.

Closes #8338

### DIFF
--- a/src/clj/game/core/turns.clj
+++ b/src/clj/game/core/turns.clj
@@ -50,8 +50,9 @@
                              (trigger-event-simult state side eid :corp-mandatory-draw nil nil)))
                (effect-completed state nil eid))
              (swap! state dissoc (if (= side :corp) :corp-phase-12 :runner-phase-12))
-             (when (= side :corp)
-               (update-all-advancement-requirements state)))))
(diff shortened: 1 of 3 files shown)


src/clj/game/core/turns.clj
```diff
@@ -50,8 +50,9 @@
                              (trigger-event-simult state side eid :corp-mandatory-draw nil nil)))
                (effect-completed state nil eid))
              (swap! state dissoc (if (= side :corp) :corp-phase-12 :runner-phase-12))
-             (when (= side :corp)
-               (update-all-advancement-requirements state)))))
+             (wait-for (trigger-event-simult state side (if (= side :corp) :post-corp-turn-begins :post-runner-turn-begins) nil nil)
+                       (when (= side :corp)
+                         (update-all-advancement-requirements state))))))
```
NoahTheDuke, line 55: Need an `effect-completed` after the `when`.

PR #8355, opened 2025-09-10
To announcement functionality

### LINKED ISSUE #4046: Enchancement request: adding timer to games
Hello!

I was wondering if it's possible to have timed games on Jinteki? How I envision it would be:
- As an optional setting when making games (similarly as password or spectator) have an option to set a timer.
- Any timeframe could be set (could have an upper limit or could even use tournament standards of 45, 60, 65), not sure what's easier to do.
- Time would be shown between the number of spectators and leave game, so something like: 3 spectators | 21:13 left | Leave game (might not even need the "left" text)
- Would enforce tournament rules once time is over: current player finishes turn, other player gets a turn, then more points win, tie on a tie.
- On the question of "who counts as current player" once the End Turn button is pressed it's the next players turn.

Mentioning @NoahTheDuke becaus he told me to assign it to him :smile:

### LINKED ISSUE #7454: Give admin accounts broadcast access for tournament rooms
Basically, if we can give admin accounts a way to post something like "5 minutes in round" to every tournament room at once, that would be sick.

NBKelly:
I finally decided to tackle a user friendly tournament helper.

Essentially:
* This is functionality to help manage rounds
* You can set the round time, time until start, end message, rules reminder, etc
* You can manage time extensions, and exclude tables from the ongoing timer if you like (ie some asynch is playing at the same time, you don't want to bother those players)
* players get notified on any alerts given, and when the round ends
* players that join after you set up the thing get registered, active lobbies beforehand get registered, only applies to comp lobby
* Round end time (and time extensions) get displayed to users in the gameboard
* summary of actions is displayed in the top of the TO area

WIP example:
alerts
manager

Closes #4046 (the feature already exists in that you can time games, but this will be a definitive close - note that strict enforcement is bad because TO's have discretion over their time policies)
Closes #7454

### DIFF
--- a/resources/public/i18n/en.ftl
+++ b/resources/public/i18n/en.ftl
@@ -504,6 +504,7 @@ game_gain-credit = Gain Credit
 
 game_game-start = Game start: {$timestamp}
 
+
 game_grip = Grip
 
 game_heap = Heap ({$cnt})
@@ -585,6 +586,10 @@ game_rez-all = Rez All
 game_rfg = Removed from the game
 
 game_rnd = R&D
+
+game_round-end = Round end: {$timestamp}
+
+game_round-extension = (Includes {$minutes}m time extension)
 
 game_run = Run
 
--- a/src/clj/web/app_state.clj
+++ b/src/clj/web/app_state.clj
@@ -9,6 +9,7 @@
 (defonce app-state
   (atom {:lobbies {}
          :lobby-updates {}
+         :tournament nil
          :users {}}))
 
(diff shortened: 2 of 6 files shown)


NoahTheDuke:
rip the tournament tab, you would have loved this feature :sob:

NBKelly:
> rip the tournament tab, you would have loved this feature 😭

Yeah, I remember the attempt, but sadly we can't pin features to cobra/aesops and have them also be reliable :weary:

PR #8372, opened 2025-09-28
Add multi-deck deletion feature

xdg:
This commit implements bulk deck deletion functionality in the deck builder:
- Add cleanup mode with deck selection checkboxes and selection controls
- Implement atomic bulk deletion with progress feedback and error handling
- Add confirmation modal with proper pluralization across 9 languages
- Discourages navigation during deletion to avoid data loss

Notes:
- This code was AI-assisted, but had several rounds of manual and AI review, focusing on adhering to Clojure style in the codebase.
- Translations were AI-generated and spot checked manually against Google Translate.

### DIFF
--- a/resources/public/i18n/en.ftl
+++ b/resources/public/i18n/en.ftl
@@ -138,9 +138,18 @@ deck-builder_card-name = Card name
 
 deck-builder_clear-stats = Clear Stats
 
+deck-builder_cleanup-decks = Cleanup decks...
+
+deck-builder_cannot-be-undone = This cannot be undone.
+
 deck-builder_completed = Completed: {$completed}
 
 deck-builder_confirm-delete = Confirm Delete
+
+deck-builder_confirm-delete-multiple = Delete {$cnt ->
+    [one] {$cnt} selected deck
+    *[other] {$cnt} selected decks
+}?
 
 deck-builder_copy = Copy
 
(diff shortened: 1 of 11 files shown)


jdeng89:
There's a few oddities that probably should be addressed, including some changes to spaces/tabs which seem unnecessary.

Additionally can you opine on how this was tested?

NBKelly:
The translations are probably ok, but I think this should really just be a single db query.

xdg:
I tested this manually creating and deleting decks, including selecting, changing filter, selecting/unselecting, etc.

I'll explore doing a deleteMany.  Any pointers on how that's best accomplished?

I'll poke at the whitespace changes.  I have something that strips trailing spaces but I can probably unwind that if needed.

xdg:
I attempted to fix clojure whitespace/indents with narrowly targeted use of `cljfmt`.  I kept the edit that removed a trailing space in `deckbuilder.styl`.

Still looking into bulk deletes, but I have to figure out how to add an API endpoint for that as it doesn't fit the REST model for individual decks.

NBKelly:
Add an endpoint in `src/clj/web/api.clj`, and link it to a handler similar to the ones in `decks.clj`. If you feed it a list of deck-ids, then you can do a deletemany for all decks which belong to user id with a deck id included in the list.

xdg:
Implemented bulk deletion.  Did some local testing, including using a filter, selecting, resetting filter, etc.  Everything seemed to work as expected.

Definitely would appreciate more testing.

As mentioned in Slack, deleting a deck that is in use in a game can result in a phantom deck that goes away on refresh, but this also happens with single deck deletion so I've made no attempt to prevent that in this commit.

xdg:
@NBKelly could you please take a look when you have a chance?

NBKelly:
Sure - I probably wont have time to take a proper look until at least the coming weekend though.

Sent with [Proton Mail](https://proton.me/mail/home) secure email.

PR #8381, opened 2025-10-01
Add pagination support to history endpoint via skip query parameter

LLBlumire:
Currently it is not possible to paginate the stats page. This would lay initial groundwork to make this possible by enabling a skip query parameter on the internal API. Custom UI can obviously be built into JNET to paginate the state page later, but for now this enables power users to grab their data.

```
GET /stats/history?skip=100
```

Would allow reading the second page of history, for example.

I am not set up for JNET development so this is entirely untested, but I think the change is simple and straightforward enough.

### DIFF
--- a/src/clj/web/stats.clj
+++ b/src/clj/web/stats.clj
@@ -214,21 +214,24 @@
 
 (defn history
   [{db :system/db
-    {username :username :as user} :user}]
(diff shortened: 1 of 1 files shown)


NoahTheDuke (approved):
Thanks!

PR #8385, opened 2025-10-05
Add admin toggle to pause game creation for maintenance

xdg:
Adds ability for admins to pause new game/lobby creation to allow draining active games before maintenance windows. Includes admin UI toggle, backend state management with persistence, and real-time broadcast of pause status to all connected users.  In addition to disabling buttons on the front-end, back-end handlers also check for paused game creation and reject creating lobbies or starting games when creation is paused.

Disclaimer: this was written with the assistance of Claude Code.  As always, I have reviewed the code thoroughly to the best of my ability and have prompted Claude to review for idiomatic Clojure that matches existing Clojure style within the repo.

### DIFF
--- a/resources/dev.edn
+++ b/resources/dev.edn
@@ -36,6 +36,8 @@
            :web/email #ig/ref :web/email}
  :web/banned-msg {:initial "Account Banned"
                   :mongo #ig/ref :mongodb/connection}
+ :web/pause-game-creation {:initial false
+                           :mongo #ig/ref :mongodb/connection}
  :web/i18n nil
  :frontend/version {:initial "1"
                     :mongo #ig/ref :mongodb/connection}
--- a/resources/public/i18n/en.ftl
+++ b/resources/public/i18n/en.ftl
@@ -775,6 +775,8 @@ lobby_corp-perspective = Corp Perspective
 
 lobby_create = Create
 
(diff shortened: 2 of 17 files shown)


NoahTheDuke:
this is a great idea. i think the solution in the pr is overengineered. i'm going to think on the smaller way to achieve this, let us sit on this for a minute.

xdg:
Give it a try, though.  IMO, the UX feels really nice.  As soon as the check box is clicked, all the "New Game", "Create", "Start" buttons go red (so it doesn't just look like some prerequisite wasn't met) and then the tooltip gives info about why. Even though admins can always announce, users who join after are visually hinted that things are blocked.

NBKelly:
I think preventing games from being created is incorrect - if people want to play games in the 30/60 minute window before a shutdown, and risk them being cut short, they should be free to do that.

xdg:
That's a matter of when you click the button.  Easy to give a 30 minute warning and then click the button at T-15 minutes.

I'm personally interested in this for when I run a private server for an event like the Showdowns.  I'd like to be able to have the site up for people to log in, upload decks, etc. before opening it up for game play.  And afterwards, I'd like to let people look at replays, etc. without allowing them to start games.

xdg:
Closed in favor of #8392

PR #8386, opened 2025-10-06
translations: Add card language option

csbisa:
The new card language option controls the language used for card data (which currently influences card data displayed in the card browser and card names displayed during games). The current language option now controls the language for the UI rather than all language settings.

This allows users to select different languages for the UI and card names, which is particularly useful if localized names are not known and/or commonly used.


NoahTheDuke:
The code looks good. Do we support card image or card text data in other languages?

PR #8391, opened 2025-10-13
Expose translation keys on the frontend + fixes

NBKelly:
I want it to be possible to work on translations without needing to actually read clojure at all.

To this end, I'm trying to make it so all (or most) of our translated content is tagged with a key on the front-end so you can check keys from your browser by just using inspect element.

Done so far:
* Translated the about page, added in media redirects for anyone I could find, removed the card impl status spreadsheet (it was about 4 years out of date at least)
* We can pass tr-vecs to `tr-non-game-toast`.
* Made the account page tag all of the translations used.
* Done account page
* Done auth menu (when you click on your username)
* Done cardbrowser as much as it is feasible
* Done chat, deckbuilder, etc - everything outside of gameboard

TODO:
* The help page is not translation friendly at all, but it needs to be touched up more than the scope of this pr, so that's left for later

In Parallel:
https://github.com/mtgred/netrunner/wiki/How-to-help-out-with-translations

Browser Extension:
https://greasyfork.org/en/scripts/552721-translation-helper
https://github.com/NBKelly/jinteki-translation-helper

### DIFF
--- a/src/cljs/nr/utils.cljs
+++ b/src/cljs/nr/utils.cljs
@@ -343,10 +343,14 @@
     "log-player-highlight-red-blue"))
 
 (defn cond-button
-  [text cond f]
-  (if cond
-    [:button {:on-click f :key text} text]
-    [:button.disabled {:key text} text]))
+  "Conditional button component. Renders enabled/disabled button based on condition.
+  Accepts optional attributes map to merge into button element."
+  ([text cond f]
+   (cond-button text cond f nil))
+  ([text cond f attrs]
+   (if cond
+     [:button (merge {:on-click f :key text} attrs) text]
+     [:button.disabled (merge {:key text} attrs) text])))
 
 (defn checkbox-button [on-text off-text on-cond f]
   (if on-cond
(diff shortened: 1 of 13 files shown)


src/cljs/nr/utils.cljs
```diff
@@ -111,6 +112,16 @@
...
      (f msg))))
 
+(defn tr-non-game-toast
+  "Display a toast warning with the specified translation vector."
+  ([tr-vec toast-type] (tr-non-game-toast tr-vec toast-type nil))
+  ([tr-vec toast-type options] (tr-non-game-toast tr-vec nil toast-type options))
+  ([tr-vec tr-params toast-type options]
+   (set! (.-options js/toastr) (toastr-options options))
+   (let [f (aget js/toastr toast-type)
+         msg (rdom-server/render-to-string (tr-span tr-vec tr-params))]
```
NoahTheDuke: Should we be using `tr-span` here? I know that toasts need to be translated too, but seems... ill-advised to be rendering hiccup to string like this. maybe not? hard to say.

PR #8392, opened 2025-10-15
Add admin toggle to pause game creation for maintenance

NoahTheDuke:
Supersedes #8385

### DIFF
--- a/src/clj/web/app_state.clj
+++ b/src/clj/web/app_state.clj
@@ -10,6 +10,7 @@
   (atom {:lobbies {}
          :lobby-updates {}
          :tournament nil
+         :allow-game-creation true
          :users {}}))
 
 (defonce lobby-subs-timeout-hours 1)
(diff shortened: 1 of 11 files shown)


src/clj/web/app_state.clj
```diff
@@ -10,6 +10,7 @@
   (atom {:lobbies {}
          :lobby-updates {}
          :tournament nil
+         :allow-game-creation true
```
xdg: minor: As a general rule, I prefer defaults to begin false/empty because they are less surprising.  Here, `pause-game-creation` would invert this so that "false" is the desired state.  It inverts the logic check, but aligns the state to the button on the admin page, simplifying that.

xdg (commented):
I see how this is simpler.  I have one minor though about inverting the state boolean.

src/clj/web/app_state.clj
NoahTheDuke: I generally like states to be "positive", cuz it makes thinking through the logic easier ("is game creation allowed?" vs "is game creation not disallowed?"), but it's used in very few places so I think you're correct here. I'll make the change.

PR #8393, opened 2025-10-18
Vantage point spoilers from worlds

NBKelly:
Just collating what's been spoiled, since the set is still at least 2.5 months out.

Will make another pr in netrunner-data too.


NBKelly:
These are all on netrunnerdb/the netrunner-data repo, so they might as well go live to the client.

PR #8420, opened 2025-11-13
Giest has real unit tests

### LINKED ISSUE #8412: Simulchip + Geist engine issues
When Simulchip is used with the Geist engine, all triggers seem to be off.
Geist triggers twice instead of once. Tech Traders trigger twice instead of once. Wasteland doesn't trigger at all, and its trigger is lost for this turn.

Image

### LINKED ISSUE #8421: Review all trash cost abilities for giest
From my other PR:
> Chances are there are a lot of instances where we should be using trash-can and instead just trash the card, and a lot of instances where we shouldn't be using trash-can but do. I'll probably review those separately later in another PR.

I'm going to do a review of all the runner side trash abilities and ensure they all work as intended, at some point in the future.

NBKelly:
I made the following changes:
* In unit tests, `card-ability ...` can now refer to the card by name (the same way we do with click-card).
* A costs-paid event is fired when costs are paid
* giest uses this event
* tech trader uses this event
* I added a useful unit test for giest
And then:
* I had to fix a unit test for street peddler, so I made it actually pay
* I decided to change over the trash-icon to a fake-cost array like I planned to years ago (this supports any arbitrary fake cost anywhere we query regular costs, and builds into labels).

~~Chances are there are a lot of instances where we should be using trash-can and instead just trash the card, and a lot of instances where we shouldn't be using trash-can but do. I'll probably review those separately later in another PR.~~

Closes #8412
Closes #8421

### DIFF
--- a/src/clj/game/core/engine.clj
+++ b/src/clj/game/core/engine.clj
@@ -1299,7 +1299,8 @@
       (complete-with-result state side eid nil)
       (wait-for (pay-next state side (make-eid state eid) costs card [])
                 (let [payment-result async-result]
-                  (queue-event state :costs-paid {:payment async-result})
+                  (queue-event state :costs-paid {:side side
+                                                  :payment async-result})
                   (wait-for (checkpoint state nil (make-eid state eid) nil)
                             (complete-with-result
                               state side eid
--- a/src/clj/game/cards/identities.clj
+++ b/src/clj/game/cards/identities.clj
@@ -385,7 +385,8 @@
   {:events [{:event :costs-paid
              :async true
              :interactive (req true)
-             :req (req (->> context :payment (map :paid/type) (some #{:trash-can})))
+             :req (req (and (= :runner (:side context))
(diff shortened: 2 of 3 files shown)


NBKelly:
It actually turned out to only be five or so cards that needed their trash-cost signatures fixed, so I did that.

I added another trash-cost (trash-self), which is identical to trash can except it does not have the trash-can signature, so we can use that in nested costs (ie fransofia ward) without borking cards like giest and the back.

src/clj/game/core/engine.clj
```diff
@@ -1299,6 +1299,7 @@
       (complete-with-result state side eid nil)
       (wait-for (pay-next state side (make-eid state eid) costs card [])
                 (let [payment-result async-result]
+                  (queue-event state :costs-paid {:payment async-result})
```
NoahTheDuke: oh that's hot

NBKelly:
I got annoyed again, so you can use the card name for the rez command too.

Made sure that giest and tech trader respect the calling side (and other things can too).

Issue #7940, opened 2025-02-02
Enhancing the Alt arts switch in the Settings


xiaat:
Asking for something a bit in advance. I think it would make sense to rewrite this setting to be a selection of 1 out of 3 options:
1. Enable all alt arts
2. Enable only official alt arts
3. Disable all alt arts

This would likely involve assigning additional values to the alt art folders or something.

NBKelly:
I think there's one more option I would like *(this would be a checkbox which is only selectable when opt 1 or 2 are selected)*:
4. Enable alt arts for my side only

It's easy to recognize a couple of alt arts you picked yourself, but recognizing the hundreds that other players are using actually puts me off using them altogether 😂

xiaat:
Great idea!

PR #8425, opened 2025-11-15
Zoomed card uses base art option

### LINKED ISSUE #6334: Feature request: Rotate promo ICE with landscape text
Inspired by the Midnight Sun booster pack containing two full art ice ([Hakarl](https://netrunnerdb.com/en/card/32004) and [Anemone](https://netrunnerdb.com/en/card/32005)) but also because my opponent today complained about me using the alt art Border Control 😅 

It would be pretty useful if any cards with text oriented this way were automatically displayed in landscape rather than the normal portrait mode, especially because the text on the promos is comparatively small already.

### LINKED ISSUE #8376: Option to have the card pop-up default to standard art
I think having an option to have the card pop-up on the top right default to standard art, even when you're hovering over alt-arts, is probably a useful accessibility feature.

Would also tackle #6334.

NBKelly:
Allows the user to set an option to have the zoomed card always use the base art, even when alt arts are enabled.

preview-base-art

Closes #8376

Closes #6334 by giving the user a way to not have to deal with it (there's no way we keep a list of arbitrary card arts that need rotating)

### DIFF
--- a/src/cljs/nr/gameboard/board.cljs
+++ b/src/cljs/nr/gameboard/board.cljs
@@ -50,23 +50,26 @@
     @corp-prompt-state
     @runner-prompt-state))
 
-(defn- image-url [{:keys [side code] :as card}]
-  (let [lang (get-in @app-state [:options :card-language] "en")
-        res (get-in @app-state [:options :card-resolution] "default")
-        special-user (get-in @game-state [(keyword (lower-case side)) :user :special])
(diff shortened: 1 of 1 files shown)


xiaat:
Not sure that it necessarily closes https://github.com/mtgred/netrunner/issues/7940 but it's a step in that direction.

PR #8433, opened 2025-11-21
Compile works with hosted

### LINKED ISSUE #8432: Compile didn't bottom its installed program at the end of the run for some reason
https://jinteki.net/replay/781a0c80-93bd-4aee-abf6-dae889317744

See last runner turn. There were a lot of things that could possibly have interacted with the compile trigger - being hosted on leprechaun, brahman adding something to the top, nisei token - and I don't know which it could have been. But anyway here's the replay.

NBKelly:
Closes #8432


NoahTheDuke:
Good change. Probably something that needs to happen across the entire card pool, but that's a ton of work.

PR #8434, opened 2025-11-21
Touch up identities

NBKelly:
I'm going through and touching up all the identities to be in line with modern implementations.

* Alice can no longer brick if you sabotage the only card in HQ during her trigger
* Blue sun now queues like a normal event
* Edward kim now has a prompt so he's not so jarring (myself and other players frequently found themselves confused when playing him)
  * I moved the registry hacking here to trash-cards itself
  * Only two cards in the game seem to require these: Aumakua and hellion beta test
  * Maybe we can fix that soon?
* Turns out esa was sabotaging anyway when you declined the draw, so I fixed that
* I added a `scry` function in def-helpers which does what it says. I used it for epiphany and chryseopian skimming for now.
* Kabonesa wu got cleaned up a bit, and labels the cards that she is going to mulch at the end of the turn too, and does a different event for each card (and the events are labelled)


NoahTheDuke:
looks good, left some questions

NBKelly:
Reading back through the ids, I noticed that I missed tennin and seidr. I'm going to go back and do those too.

PR #8437, opened 2025-11-22
fix #8436.

### LINKED ISSUE #8436: Popup window of the corp's deck instantly closes When clicking 'Show' to looks at the deck.
When players click 'R&D' and then click 'Show' to looks at their decks, the popup window for decks will show up but then instantly closes.

I think I already found what the problem is, and I'll commit a pull request to fix this issue.

Image

eric03742:
fix #8436 

I think the problem is that `deck-view` mistakenly adds `div.panel.blue-shade.menu` as the child node of `div.blue-shade.deck`. So when players click `Show` button of that menu, they also trigger the `on-click` event of `div.blue-shade.deck`, causing the opening popup window hides itself back.


NoahTheDuke:
oh perfect. thanks for finding this

PR #8447, opened 2025-11-28
Fix MU display not updating when programs installed

### LINKED ISSUE #8445: MU tracker does not update
The MU tracker only updates when the page is refreshed. Similar to the deck size bug there was recently.

xdg:
Fixes #8445. The display-memory function is a Reagent Form-2 component where the inner render function wasn't receiving updated memory parameters, causing the MU count to remain stale when programs were installed or uninstalled.

Moved the let bindings inside the inner render function so it receives and processes updated memory values on each render.

This fix is identical in nature to fee21af52 which fixed deck count rendering with the same root cause.


NoahTheDuke:
you can remove the fn entirely because it's not closing over any state

PR #8451, opened 2025-12-02
Option to flag games with descriptions

NBKelly:
Allows you to tag games with descriptions:
* Learning the game
* Default (nothing)
* Casual play
* Competitive play
* Looking to play against meta decks (what these are is an exercise to the player)

Ideally, this should let players be better aware of what they're in for when they make or join a game.

game-tags

game-tags_000


NoahTheDuke:
this is cool! and so easy

PR #8471, opened 2025-12-25
fix: scoring->stealing on runner win copy

### LINKED ISSUE #8367: Runner winning message incorrectly refers to "scoring" instead of "stealing"
When the Runner wins the game by stealing their 7th point, the end screen stat window says they win by scoring agenda points instead of stealing. This is a pretty common misuse of scoring instead of stealing, but there's no reason to cement it in the players' mind. Those are different things according to rules and some effects trigger on one and not the other, so we don't need to further feed the confusion.

filipesilva:
Fix #8367


filipesilva:
Runner copy:
CleanShot 2025-12-25 at 17 01 37@2x

Corp copy stays the same:
CleanShot 2025-12-25 at 17 05 39@2x

PR #8472, opened 2025-12-25
fix: handle no deck after edit

### LINKED ISSUE #8053: If you cancel editing a deck, and have no saved decks, that's a black screen
I assume we try to load a non-existent deck for you.

filipesilva:
Fix #8053

### DIFF
--- a/src/cljs/nr/deckbuilder.cljs
+++ b/src/cljs/nr/deckbuilder.cljs
@@ -239,7 +239,9 @@
         selected (or (:old-deck @s) (first decks))]
     (end-edit s)
     (load-decks decks)
-    (when selected (put! select-channel selected))))
+    (if selected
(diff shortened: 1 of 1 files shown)


filipesilva:
This isn't actually a black screen anymore. But it still leaves the UI in a bad state without the fix.

For example, if you do a new corp deck then cancel with no decks, the new deck that was never saved stays on the right side:

CleanShot 2025-12-25 at 17 32 12@2x

Then if you make a new runner deck and cancel, you'll actually see that corp deck as if it was there on the list:

CleanShot 2025-12-25 at 17 33 23@2x

If you try to delete it, you get a 409, and if you reload the page it isn't there.

The fix in this PR handles all these failure modes by setting the state :deck to nil.

PR #8486, opened 2026-01-07
Touch up events

### LINKED ISSUE #8483: Turn labels into static abilities
Had a thought that we could turn card labels into static abilities (at least, in some cases), and that could handle the cleanup for us when the source card or effect leaves play instead of needing to manually unlabel cards.

Additionally, if we link the labels to the source card, then we could see about having a syntax highlight where if you hover a card it potentially highlights the cards it's applying labels to. (that would be really helpful if there were multiple triests, for example)

NBKelly:
I simplified a few events and then got sidetracked, so:

* This changes icons to be static abilities instead of manually adding/removing them
* This also allows for multiple icons to be on a card at once
* If you hover an icon, the source card is outlined in bold red if possible

Closes #8483


NBKelly:
There's still a bit to do, and I think some of the other prs added an icon or two, so I'll need to go over this again the hard way before we do a patch.

PR #8487, opened 2026-01-10
Update bin/scripts and use them in the README docker section

filipesilva:
This PR updates the scripts in `bin/` to be executable, and uses them in the README for all the docker related operations.

The `bin/database-drop` script is new though. There was no instructions for it in the README, but I needed to do it to clear some junk user accounts I created, so I thought I'd add it as a counterpart to `bin/database-seed`.


NoahTheDuke:
nevermind, that's under the docker section.

PR #8490, opened 2026-01-10
fix: show card preview on button hover too

### LINKED ISSUE #8488: Card preview shows on "Choose a program" text hover instead of button hover
When you're selecting a card to install (e.g. via self-modifying code), you don't get a preview when the cursor is over the button itself. Instead, the preview is triggered by the cursor being over the text.

https://github.com/user-attachments/assets/429b5dc4-621c-4a7b-9eb2-dd439da52aa4

I expected that if I could click to select it, I should get the preview.

filipesilva:
This PR changes the preview to work on buttons, given the only child of that button would have triggered a preview.

I don't feel super confident this is the right approach, but `nr.utils/card-patterns-impl` is used in a lot of places and it felt risky to edit it.

Fix #8488

### DIFF
--- a/src/cljs/nr/gameboard/card_preview.cljs
+++ b/src/cljs/nr/gameboard/card_preview.cljs
@@ -4,9 +4,15 @@
 
 (defonce zoom-channel (chan))
 
+(defn- safe-get-attribute [target attribute]
+  (when (.-getAttribute target)
+    (.getAttribute target attribute)))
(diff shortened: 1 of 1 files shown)


filipesilva:
Demo 

https://github.com/user-attachments/assets/13e126c3-cfff-4db5-be42-c7bd19ba37c5

PR #8491, opened 2026-01-12
Quick draft/Drafting Format

NBKelly:
This is a little draft-like format for jinteki.net.

Players take turns picking cards to assemble a decklist, and then play a game with the chosen cards.

I'll make a slightly more detailed summary later and post a video in slack.

I also added a feature for allowing new formats to not be automatically hidden - we have a list of "new" formats, and any time you load formats, you check against 1) this list 2) which of these new formats you have seen.
If they are different, then the difference gets saved into your visible formats and the set you have seen is updated.
What this means is exactly once per format added to this list, you'll have your visible formats silently updated to include that item.

I did this because we had issues with april fools a few years ago, chimera, throwback, etc where they were very hard to discover without already knowing about them because existing accounts did not have them added to their list of visible formats.

### DIFF
--- a/src/clj/game/core/quick_draft.clj
+++ b/src/clj/game/core/quick_draft.clj
@@ -145,7 +145,7 @@
     "Surveillance Network Key 2"
     })
 
-;; How are we doing this shit?
+;; How are we doing this?
 ;; CORP:
 ;;  We must have agendas.
 ;;  We need to draft agendas.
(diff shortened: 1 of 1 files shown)


NBKelly:
Full explanation is here for reference: https://github.com/mtgred/netrunner/wiki/Drafting-on-Jinteki.net-%E2%80%90-Quick-Draft-Format

PR #8498, opened 2026-01-17
set ws keepalive to 5 seconds

NBKelly:
This sets the WS keepalive to 5 seconds, instead of the default 50 seconds (25000 x 2).

I know it says 2500ms, and not 5000ms - for info on that, see here: https://github.com/taoensso/sente/issues/455.

My local testing has the keepalives at about a millisecond away from 5 seconds.

## But why?

Supposedly this will help SEA and LATTAM players get less disconnects from their hostile networks.

### DIFF
--- a/src/clj/web/ws.clj
+++ b/src/clj/web/ws.clj
@@ -19,7 +19,8 @@
 
 (let [chsk-server (sente/make-channel-socket-server!
                     (get-sch-adapter)
-                    {:user-id-fn (fn [ring-req]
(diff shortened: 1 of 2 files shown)


DenyExistence:
Thank you so much! I'll forward this to my friend, and report back what we find

NBKelly:
Do note that this wont be active until our next patch :)

DenyExistence:
Roger! We'll try out disabling websockets in the meantime then. Thanks for all the help :D

PR #8502, opened 2026-01-19
Trashing logs

### LINKED ISSUE #8489: Thrashing ice leaves no game log
When a corp trashes a piece of ICE, there's no entry in the game log:

https://github.com/user-attachments/assets/b86991ce-ec00-4513-bee7-7148f6447492

It would be nice to have a log to keep track of replaced ice and even when the trashed ice had stuff on it, like trojans.

NBKelly:
Closes #8489


NoahTheDuke:
are those flaky tests? that's ominous lol

NBKelly:
Maybe, I haven't seen that one fail before :joy: 
Though actually, it might be because I applied the `:unpreventable true` tag to it, and that test might be using the trash button for triggering sac con

NoahTheDuke:
Excellent, thank you.

PR #8509, opened 2026-01-23
haze and cycle name update

furrygoober:
haze and cycle name update


NoahTheDuke:
Closing, as this doesn't look like it's intended for us.

PR #8514, opened 2026-01-28
Liberation Cycle Quotes

Cluedrew:
I added a default and faction quotes for every identity in the liberation cycle. There should be a non-filler quote in every one of those slots now.

Some already had a quote there. I skipped some of those but added others either because I had an idea or I just forgot that there already was one there. At least Mercury's default slot got multiple quotes and some identity match-ups also got some quotes.

I've shown these to a few people but if anyone catches any mistakes last minute I can try and fix them.

(Also fixed Epiphany in the ordering.)


Cluedrew:
Oh yeah, I didn't do any manual testing, but I think new tests check for most of the common errors now.

Issue #8497, opened 2026-01-17
Imported deck note uses wrong URL


filipesilva:
If you import https://netrunnerdb.com/en/decklist/d8888676-f12e-4fa1-a531-9284a06a9ad0/tread-loudly-5th-at-worlds-2025, you will get this note on the deck:
```
imported from https://netrunnerdb.com/en/92324
```

This url doesn't work though, instead it should be either https://netrunnerdb.com/en/decklist/92324 or https://netrunnerdb.com/en/decklist/d8888676-f12e-4fa1-a531-9284a06a9ad0

I can fix this, just wanted to make an issue.

Dupe of https://github.com/mtgred/netrunner/issues/8496 that I closed by accident then didn't have perms to reopen.

PR #8515, opened 2026-01-28
fix: imported nrdb deck url

filipesilva:
Fix https://github.com/mtgred/netrunner/issues/8497

### DIFF
--- a/src/clj/web/nrdb.clj
+++ b/src/clj/web/nrdb.clj
@@ -7,7 +7,7 @@
             [taoensso.timbre :as timbre]))
 
 (def nrdb-decklist-url "https://netrunnerdb.com/api/2.0/public/")
-(def nrdb-readable-url "https://netrunnerdb.com/en/decklist/")
+(def nrdb-base-url "https://netrunnerdb.com/en/")
 
 (def private-endpoint "deck/")
 (def public-endpoint "decklist/")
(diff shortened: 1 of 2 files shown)


NBKelly:
I think this only works if the decklist was public (published and unpublished lists have different endpoints), but it's still a step closer to being correct.

filipesilva:
So how do those unpublished ones work? I opened clicked on `My Decks`, then the `View` button on one of my decklists. Then tried opening that URL in incognito and got a `I tried very hard, but I couldn't find what you're looking for.` message. Do I have to do something to get a URL that is public but not published? The URL I tried is below.

https://netrunnerdb.com/en/deck/view/251826ae-d882-4c44-b544-8fdf7f8b6ebf

NBKelly:
Yeah, it's a setting in your profile.
IE: https://netrunnerdb.com/en/deck/view/95db310b-79e7-4bf0-8e4e-e643cb7f6c95
I think the file being worked on figures out which endpoint is being used as you go, so you can probably work from that.

filipesilva:
@NBKelly ok now it accounts for both:

CleanShot 2026-02-15 at 18 04 09@2x
CleanShot 2026-02-15 at 18 04 20@2x

I tried to not touch the rest of the names of things in that namespace, so the endpoint type is kept as `:private` instead of something new like `:unpublished`.

PR #8531, opened 2026-02-13
Fix player panels not swapping when changing view during replay

matthewlehew:
It's been throwing me off that if you swap between runner/corp view in a replay, all elements (including scored agendas) swap placement _except_ for the player info and stats panels. This fixes that by adding React keys to the `stats-view` component.


matthewlehew:
Created issue #8535 to document the problem.

NoahTheDuke:
Thanks!

PR #8532, opened 2026-02-14
Vantage Point

### LINKED ISSUE #5279: Feature Request - Have Bacterial Reprogramming allow/redo all 3 modes.
Many times myself and others have accidentally draw 7 cards from this card. /Undo-click works when it's stolen off a single access or scored, but it gets awkward when it needs to be undone in the middle of a multiaccess on HQ or RnD.

NBKelly:
This should cover all the rest of the VP cards and mechanics in order as they become public.

I'm going to add notes for mechanics and stuff here as they come.

Also unless I specify at the bottom of the post, I've added unit tests for every card.

* Melies-U has four faces (like biotech) and there are enough triggers that I felt it was worthwhile actually leaving the comments in. It also needs an entry in our image handling stuff.
* Bad Publicity change: There's an evil hack I'm using for how the button interacts, and I need to fix that sooner or later and just have an endpoint and a frontend UI hook for it.
* We could make the bad pub change optional for eternal or casual games. There's one single critical point in the code, so it's very easily doable.
* Editorial Division is just having illicit as an option as well for the search hook for now. Whatever other details can be figured out later.
* A couple of tests/cards rely on cancel maps from the other PR, I'll merge this weekend for that
* Startup is going to 3x 3-pointers max again
* Flagship requires `access.clj` to track the both the random-access-limit being throttled to zero, and installed cards being removed from candidacy for access (at least, on centrals).
* For myoshu, we need to be able to work out of a scored agenda was installed this turn. The score event now also includes the card being scored (before modifications) as `:card-scored`.
* There's a new option for prompts, `:show-opponent-discard`, which shows the opponents discard (like `:show-discard` does for us). Ansel 2.0 uses this, and a few other cards can be updated for it.
* I added support for computing the costs of cards as if they were going to be hosted (ie, dhegder) - so you can pick them as targets for install abilities if you're on the credit threshold for those discounts now.
* For lotus haze, it is not legal to even attempt:
  * Moving a region on top of another region
  * Moving a card in a way where it violates it's zoning restriction (ie ZATO City Grid: Remove server only)
  * But the classic FFG cards which only had **install restrictions** can be freely moved
  * So I separated out `install-req` into `install-req` and `legal-zones` in order to enforce this
  * I wanted to get this right because it's one of the few ways to game loss yourself playing online despite being an experienced player
* There's support for effects with prevent you losing credits, and effects which prevent you from spending credits **from your credit pool** (other sources are fine)
* I reworked bacterial programming to allow completely starting over. This should prevent the occasional game loss in eternal games. (this is in here because it's the same code as cultivate). Closes #5279
* Man in the Middle adds support for `cannot-forfeit` clauses on agendas
* Sipa means that the `pass-ice` event needs information about the position of the ice (was it outermost), and if all subs were broken

### TODO:
* I have no unit tests for Knowledge Seeker
* Update other cards for `show-opponent-discard`. Hailstorm, ark lockdown, etc.

See Also: https://github.com/NoahTheDuke/netrunner-data/pull/81


src/clj/game/cards/identities.clj
```diff
@@ -755,6 +755,21 @@
...
+(defcard "Editorial Division: Ad Nihilum"
+  {:events [{:event :corp-gain-bad-publicity
+             :optional {:req (req (let [valid-ctx? (fn [[ctx]] (pos? (:amount ctx)))]
+                                    (and (valid-ctx? targets)
+                                         (first-event? state side :corp-gain-bad-publicity valid-ctx?))))
+                        :prompt "Search for a card?"
+                        :waiting-prompt true
+                        :yes-ability {:prompt "Choose a card"
+                                      :msg (msg "add " (:title target) " to HQ from R&D") ;; TODO - once the illicit->liability change is through, we can adjust this (or just leave it)
+                                      :choices (req (cancellable (filter #(has-any-subtype? % ["Illicit" "Black Ops" "Gray Ops" "Liability"]) (filter (complement agenda?) (:deck corp))) :sorted))
```
NBKelly: Illicit is here as a placeholder for now

src/clj/game/core/prompts.clj
```diff
@@ -191,27 +205,31 @@
                             (str " " (pluralize "target" min-choices))
                             " a target"))
                         " for " (:title card)))
-                    (if all ["Hide"] ["Done"])
+                    (concat (when (:offer-bad-pub? ability) [(str "Bad Publicity (" (:offer-bad-pub? ability) " available)")])
```
NBKelly: This is an evil hack and should be removed/replaced with a frontend hook

NBKelly:
~~We'll need to update how we do damage: previously, damage always had the runner trash the cards as per the ffg manual.~~
~~Now, the cards being trashed is being attributed to the source of the damage (this was an oversight for saisentan/au co, and is being silently ruled for this outcome now).~~

~~This means we can simplify a few cards.~~ this is done

I also want to add a `card-trashed` event that fires alongside `corp-trash/runner-trash/game-trash` just to simplify some cards, but that's a job for later maybe.

test/clj/game/cards/upgrades_test.clj
```diff
@@ -1589,6 +1589,57 @@
         (take-credits state :runner)
         (is (= (+ 3 total-corp-credits) (:credit (get-corp))) "Corp does not gain any extra c with agenda")))))
 
+(letfn [(setup-state []
```
NoahTheDuke: instead of this, you could use `before-each`.

PR #8540, opened 2026-02-17
Added support for cancel maps, cleaning up cancel effects

### LINKED ISSUE #8534: Add a `cancel-msg` key to ability maps
This can save having cancel-effects which manually construct messages.

NBKelly:
I added support for the `cancel-msg` key in ability maps, so you can just use that instead of using an effect to system-msg to reconstruct the text to say something.

And I'm also cleaning up a lot of the `declines to use ...` effects.
* We should be logging things that happen
* Most (but not all) of the `declines to use` messages are noise
* In *most* cases, the combined absence of information and use of waiting prompts are sufficient

Specific case in point is the federal fundraising decline prompt, which I've had multiple people say (skim) reading it they thought the ability was actually being activated.

* We should keep prompts for things that give information - ie, the corp used project atlas to look at the deck before deciding "wait, nevermind"
* I think they're kind of acceptable for multi-stage effects, ie moon pool. For something like elivigar bifurcation, it's 100% noise

Cards changed:
* A lot of cards had noisy declines stripped out, I'm just listing the cards with quantifiable changes in this region
* I replaced ability maps which just shuffle the deck with the new `shuffle-my-deck!` ability wrapper in the `shuffle` class
* ABT has an actual useful log when you cancel, saying how many cards were trashed
* Eminent domain now shuffles if you fail to find a card to rez
* Next Big Thing correctly shuffles if you choose zero cards to go back
* Yagi uda no longer spends and unspends counters as you cancel
* Remote enforcement correctly shuffles if you fail to find an ice
* Executive boot camp shuffles on a fail to find
* Same for executive search firm, levy university, mumbad city hall, shannon claire
* Accelerated diagnostics properly logs when it trashes from the top of the deck
* Moon pool shuffles even if you reveal zero cards
* Recruiting trip correctly shuffles if you cancel your search before X number of cards are picked
* Malapert Data Vault correctly shuffles if you fail to find a non-agenda
* Compile correctly shuffles if you search and fail to find a program
* Meeting of minds correctly shuffles and continues if you opt to search and then fail to find
* Mutual favor lets you correctly fail to find
* Kabonesa wu now lets you fail to find

I also updated the async checker to account for this.

Closes #8534

### DIFF
--- a/src/clj/game/cards/events.clj
+++ b/src/clj/game/cards/events.clj
@@ -861,7 +861,8 @@
                                                              (can-pay? state side eid card nil
                                                                        [(->c :credit (rez-cost state side %))]))}
                                        :effect (effect (rez :corp eid target))
-                                       :cancel {:effect (effect (register-run-flag!
+                                       :cancel {:async true
+                                                :effect (effect (register-run-flag!
                                                                   card
                                                                   :can-rez
                                                                   (fn [state _side card]
--- a/src/clj/game/cards/operations.clj
+++ b/src/clj/game/cards/operations.clj
@@ -177,6 +177,7 @@
                                     (ad state eid card remaining starting-shuffle-count)
                                     card nil))))
                  :cancel {:msg (msg "trash " (quantify (count remaining-cards) "card") " from the top of R&D")
+                          :async true
                           :effect (req (trash-cards state side eid remaining-cards {:unpreventable true :cause-card card}))}}
                 :else
                 {:prompt "There are no playable cards"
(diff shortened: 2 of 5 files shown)


NoahTheDuke:
I know you just did all this work, but what if it was `:cancel {...}` instead of `:cancel-msg` and `:cancel-effect`? That way, we wouldn't have to bother with annoying `effect-completed` calls in synchronous cancel abilities.

NBKelly:
Hmm, yeah I was thinking making the cancel effect an ability itself might be cleaner. I think I'll do that.

PR #8543, opened 2026-02-21
strip image list and other useless info from engine cards

NBKelly:
This appears to work fine locally, but I'm not sure what it's going to do on an actual server, so I'm going to try it out on playtest tomorrow.

This should cut a big chunk of data out of our diffs so we don't need to send it over and over.


NBKelly:
Nobody on playtest reported any issues with this, so I think it's fine.

PR #8544, opened 2026-02-21
Update card-backs compilation warning message

jwarwick:
Also removed redundant .gitignore paths.


NoahTheDuke:
welcome back! are we losing valuable info by not including the full exception?

jwarwick:
I don't think so, since it's just reading the edn file. I assume if you get a message other than "file not found" you can grep for where the exception is happening easily enough.

PR #8548, opened 2026-02-24
Fixed alt art selection in cardbrowser, chat, and deckbuilder

### LINKED ISSUE #8281: Selected alt art indication isn't working
When selecting an alt art in the card browser, even though the choice is recorded somewhere internally and consequently displayed during the game, the orange frame in the card browser itself doesn't move, and the buttons continue indicating I selected a non-alt-art version of the card.

There's also an adjacent case: after selecting an alt art of Biotic Labor, the card browser now thinks I don't have an art selected at all: the orange frame is no longer present, and clicking on any of them displays the Select Art button.

Image

### LINKED ISSUE #8455: Art Selection not working
I started selecting alt arts for my decks, when I click the "Select art" button I get a green check that says it worked, but similar to Issue 8281 https://github.com/mtgred/netrunner/issues/8281 The UI still shows the old art highlighted. Unlike that issue, when I add the card to a deck I do NOT see the new art. If I use the alt-art selection in the profile settings tab that will cause alt arts to be visible in the deck builder, but I lose the ability to individually pick art for a card with multiple alternates.

Things I have tried:
Resetting all alts to default in Settings Tab.
Using 'Update Profile" Button
Refreshing Browser.
Chrome & Firefox

Before Clicking "Select Art":
Image

Select Art Confirmation, note the wrong art is still highlighted.
Image

After selecting an alt from 'Cards' tab:
Image

After selecting the World Championship 2015 Alt Set from the 'Settings' tab:
Image

Related but maybe should be it's own issue
If you select World Champ 2017 in Settings, in the cards tab both Biotic Labors are highlighted:
Image

This causes an issue in the deck builder where the image is missing for Biotic Labor:
Image

### LINKED ISSUE #8469: on the cards screen it does not correctly show what alt arts are equipped.
On the Cards screen if you select a alternative art that you would like to use instead of the original one it does not show the yellow outlines on the alt. Just shows a checkmarck and "updated art" in the top right corner. 

Everything shows up correctly in games and the art is actually selected. Just missing what you have selected is bit annoying when you want to try bling out your decks.

jwarwick:
Using a vector to store alt arts was causing the alt art selection and display to break in most cases. It's a little tricky, since I don't want to break any existing alt preferences saved on the prod server, so this is a bit of a workaround to attempt to handle all the cases. 

Did not modify any of the selection logic or the board display logic, so things shouldn't be any worse than they already are.

Separately it appears the prod server has two copies of `System Update (World Champion 2017)` in the db, which also breaks selection in a different way. (see: Clot) 

Fixes #8469 
Fixes #8281 
Fixes #8455

### DIFF
--- a/src/cljs/nr/cardbrowser.cljs
+++ b/src/cljs/nr/cardbrowser.cljs
@@ -190,11 +190,13 @@
   ([card allow-all-users]
    (let [lang (get-in @app-state [:options :card-language] "en")
          res (get-in @app-state [:options :card-resolution] "default")
-         art (if (show-alt-art? allow-all-users)
-               (get-in @app-state [:options :alt-arts (keyword (:code card))] "stock")
-               "stock")
+         alt-art (if (show-alt-art? allow-all-users)
+                   (get-in @app-state [:options :alt-arts (keyword (:code card))] "stock")
+                   "stock")
+         art (if (sequential? alt-art) (first alt-art) alt-art)
+         art-index (if (sequential? alt-art) (second alt-art) 0)
          images (image-or-face card)]
-     (get-image-path images (keyword lang) (keyword res) (keyword art)))))
+     (nth (get-image-path images (keyword lang) (keyword res) (keyword art)) art-index))))
 
 (defn- base-image-url
   "The default card image. Displays an alternate image if the card is specified as one."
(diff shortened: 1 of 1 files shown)


NBKelly:
Thanks for doing this. This is one of the areas where I was really dreading figuring things out :joy:

PR #8549, opened 2026-02-25
Only import single card art for an art package

jwarwick:
For a given directory, only take a single card image. Use the highest quality image format found. Should address issues with the World Champs 2017 alt arts on prod.


NBKelly:
hmm - were you able to test this with any of the flip ids? I can send you a few to try out with if not.

jwarwick:
Hmm, think the flip ids were handled in a separate part of the code, but don't recall if I verified. I'll check.

jwarwick:
Gave this a quick test with some jpg and png flips ids and it appears to work correctly.

NBKelly:
Sick, thanks for checking :)

PR #8557, opened 2026-02-28
Added Vantage Point Quotes

Cluedrew:
Added the default quotes for all Vantage Point ids, the full set for Hiram and Editorial Division (Hiram also got an identity quote for the Editorial Division).

This will probably fail the tests until the Vantage Point identities are added. But then it should all pass, I've checked the names as best I can manually.


NBKelly:
Tests pass on my end, and the quotes look good to me. Nice job :+1:

PR #8573, opened 2026-03-02
fix: fallback to :printed-title on preview

### LINKED ISSUE #8547: Jeitinho Pop-up Showing Wrong Card
When hovering over Jeitinho in the Scored Area, the card that pops up in the top-right of the screen is Nebula Talent Managment:

Image

### LINKED ISSUE #8562: Mousing over a Myoshu in the corp's score area shows Melies U instead
Title. Score a Myoshu, mouse over it in the corp's score area, and the card shown in the large card pane in the upper right is Melies U instead.

filipesilva:
Cards added to the score area as agenda via `convert-to-agenda` don't
have a title, but they have a printed title.

Fix #8562
Fix #8547

### DIFF
--- a/src/cljs/nr/gameboard/card_preview.cljs
+++ b/src/cljs/nr/gameboard/card_preview.cljs
@@ -17,7 +17,7 @@
 
 (defn put-game-card-in-channel
   [card channel]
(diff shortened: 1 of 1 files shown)


filipesilva:
Working preview:

CleanShot 2026-03-02 at 23 25 52@2x

Issue #8591, opened 2026-03-07
[EDITME] Please give a short description of your bug here


Cyber-Shark:
Link to bug replay: https://jinteki.net/bug-report/6623f4c7-c5da-4c9c-a3eb-2fefadef0b79?b=0

Description:

The Tungsten Tailor triggeres on encountering tithe, which wasnt broken

filipesilva:
I tried to reproduce but wasn't able to. Anything you can think of that could have contributed to this that didn't show in the replay?

Edit: NBKelly tipped me to what the repro was

PR #8598, opened 2026-03-07
fix: tungsten tailor doesn't trigger when no subs broken

filipesilva:
Fix https://github.com/mtgred/netrunner/issues/8591


NoahTheDuke:
i feel like i had it that way because of a rule in the CR but who can remember. if the tests pass, we're probably fine.

PR #8600, opened 2026-03-08
Migrate the :breach-server event.

Cluedrew:
Following some instructions given to be my NoahTheDuke in chat I have migrated the `:breach-server` event to use the new context map system. This was not the most interesting example of it (with only one argument) but the pattern works.

Although not required the implementation of Mwanza City Grid made this change harder, but obscuring which maps where connected to which events, so I inlined those assoc calls to make it more obvious. It also saved some lines of code.


NoahTheDuke:
Thanks for doing this! every bit helps.

PR #8609, opened 2026-03-12
Migrate :reveal-spent-credits Event

Cluedrew:
The `:reveal-spent-credits` event now passes a map instead of a list. This one is actually a good use of the map, with multiple values not all of which are always passed.


NoahTheDuke:
oh this is good, very nice.

PR #8611, opened 2026-03-13
Support meta-data when copying in a decklist

### LINKED ISSUE #5775: Read the IDs when pasted into the text parser of the deck builder
Right now when you paste the whole decklist you still need to manually click the drop down and select the list.

Imo not a high priority, just filling out the issue list.

jwarwick:
Added support for setting the identity, title, and notes of a deck in the decklist editor.
```
;; identity: Andromeda
;; identity-code: 36009
;; title: My world's winning deck
;; notes: probably not
```
You only need to specify one of `identity` or `identity-code`. If you use both, whichever one is last will take precedence.

Will make a PR against NRDB to export this info in the Jinetki copyable format.

Fixes #5775

### DIFF
--- a/src/cljs/nr/deckbuilder.cljs
+++ b/src/cljs/nr/deckbuilder.cljs
@@ -1121,9 +1121,9 @@
          (:display-name card)]))]])
 
 (defn- lookup-identity-by-code
-  "Look up an identity card by card code. Returns nil if not found or not an identity."
-  [code]
-  (first (filter #(and (= (:code %) code) (= (:type %) "Identity"))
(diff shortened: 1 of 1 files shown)


jwarwick:
NRDB PR here: https://github.com/Null-Signal-Games/netrunnerdb/pull/990

We currently ignore any input that doesn't look like card info, so these don't need to be coordinated before merging.

NoahTheDuke:
i wish we had cljs testing of some kind, i'd love to see this have a test or two. i feel like it'll need some tweaking once it's live

PR #8621, opened 2026-03-17
Add ability to Melies U to display chosen flip identity

jwarwick:
Present menu on identity and generates a corp-only toast with the selected identity.


NBKelly:
Neat, thanks.

PR #8641, opened 2026-03-21
Side specific logs

### LINKED ISSUE #8628: Expand the msg/log system to allow for side specific messaging
I think it would be quite neat if side-specific messaging was possible.
This would be useful for things that we currently need to communicate in toasts or prompts (ie which cards are secretly revealed or chosen, cards accessed from R&D, etc).
I *think* it can be worked into our differ system without any performance hit, since it is still a list of single entries.

NBKelly:
Builds on #8630

This splits the logs into three seperate logs.
Replays and spectators only see the public logs.

The corp and runner only see their side specific logs (along with side-specific spectators).

system-msg can take a side, or list of sides, as things to print to.
```Clojure
:corp   - corp side
:public - public only
:runner - runner only
:all    - all of the above
```
You can also do `multi-msg state side [map]`. `:public` backfills into empty `:corp` or `:runner`.

`:msg` maps can optional be a map of `{:public msg :corp msg :runner msg}`. Public backfills into empty corp or runner.

Some examples:
access-logging
say-example

Closes #8628

My goal for this PR is that this works for basically everything in standard (excepting costs right now).

I've gone through the following so far to take advantage of this:
* Vantage Point
* Elevation
* General installing, a few helpers, sabotage
* Borealis
* TAI
* Gateway
* Ashes


NoahTheDuke:
I'm finally reading through this and I think the vast majority of it is good, a bunch of good changes. We've really struggled with this stuff for a while, and I like this approach.

However, the way it's tracked on `state` has me feeling nervous/worried. We're duplicating some messages up to 3 times. Long-running games can be _really_ verbose, and I worry about having 3x the number of strings stored on state.

What do you think of instead storing maps in the `:log` and when we send diffs to the client, each side's log is updated with something like `(update :log (fn [log] (keep #(when-let [entry (or (:runner %) (:public %))] entry) log)))`?

Then for `:all`/`:public` entries, we just have a map with `{:public "..."}`, and for side specific entries, we have `{:runner "..." :public "..."}`, very much like you have it right now when calling `say` (but with the message functions already called). This won't save us much space when there's 3 separate messages, but it'll help a lot when there's public messages.

I'd love to hear your thoughts on this.

NBKelly:
hmm - so I think the worst case is ~60-100mb across 30-50 games, but that is a lot of to be adding that we don't need to be, yeah.

I think the map approach will be good, and will still be 1000x faster than diffing was, yeah.
Something like:
`(vec (keep #(or (side %) (:public %)) ...))` on both sides, and then it's still invisible to the client and the replay system.

I'll give it a go over easter.

NoahTheDuke:
The log diffs should be based on the `:logs` itself, right? We'd only be sending the filtered logs to the clients after the "new" logs are grabbed.

NBKelly:
yeah - currently I haven't needed to make any changes to the frontend or replays because it ends up invisible to both of them

NBKelly:
Alright, I was able to update the log system so that the messages are maps of {:public .. :corp .. :runner ..} as a single list.
It turns out that even diffing three seperate entries (corp/runner/public), the time taken is on the order of 0.1ms, instead of the 3-5ms it used to be.

NoahTheDuke:
That's incredible, thank you. Merge when ready, we can fix whatever bugs have snuck in after we get some debugging in.

PR #8653, opened 2026-03-28
Handle html tags in card text

### LINKED ISSUE #8309: cards has escaped html tags in text description of card
Image

I am a new user, so i give all the information which is relevant:

I scrolled to the bottom of the cards page.
I am using Firefox 141.0 (64-Bit) on Windows 11. 

I could separately produce it in incognito mode of Firefox and in Edge  138.0.3351.95 by scrolling to this card without any selectors

Image

I have some add-ons installed in both browsers. In Edge i only have Tampermonkey installed and it says no active user script for that site. I would be suprised if that addon affected things.

I have a shitty internet connection.

I got the error "Image corrupt or truncated. 10067.jpg" in the firefox console. Seems unrealted but might be a seperate bug?

This occurred today and also impacts other cards then one i've shown. Unsure if all card cards are impacted and have not tested this in other browsers.

### LINKED ISSUE #8645: Render <strong> tag in card description
When a card has a `<strong>` tag in its description, it should be rendered with a bold string like NetrunnerDB does. Example:

Deck Builder:
Image

[NetrunnerDB](https://netrunnerdb.com/en/card/11066):
Image

jwarwick:
Handle `<strong>`, `<em>`, and `<ul><li>` tags in the card text displays in the card browser and the in-game card zoom window (when the `display zoomed card as text` option is selected).

image

image

image

Fixes #8645 
Fixes #8309

### DIFF
--- a/src/cljs/nr/cardbrowser.cljs
+++ b/src/cljs/nr/cardbrowser.cljs
@@ -13,9 +13,9 @@
    [nr.local-storage :as ls]
    [nr.translations :refer [clean-input tr tr-data tr-faction tr-format tr-set
                             tr-side tr-type]]
-   [nr.utils :refer [banned-span buildable-format->slug deck-points-card-span faction-icon
-                     get-image-path image-or-face influence-dots
-                     non-game-toast render-icons restricted-span rotated-span
+   [nr.utils :refer [banned-span deck-points-card-span faction-icon
+                     buildable-format->slug get-image-path image-or-face influence-dots
+                     non-game-toast render-safe-html restricted-span rotated-span
                      set-scroll-top slug->buildable-format store-scroll-top
                      tr-non-game-toast]]
    [reagent.core :as r]))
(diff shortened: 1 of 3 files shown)


NoahTheDuke:
I guess we don't need to worry about ZA̡͊͠͝LGΌ ISͮ̂҉̯͈͕̹̘̱ TO͇̹̺ͅƝ̴ȳ̳ TH̘Ë͖́̉ ͠P̯͍̭O̚​N̐Y̡ H̸̡̪̯ͨ͊̽̅̾̎Ȩ̬̩̾͛ͪ̈́̀́͘ ̶̧̨̱̹̭̯ͧ̾ͬC̷̙̲̝͖ͭ̏ͥͮ͟Oͮ͏̮̪̝͍M̲̖͊̒ͪͩͬ̚̚͜Ȇ̴̟̟͙̞ͩ͌͝S̨̥̫͎̭ͯ̿̔̀ͅ ([source](https://stackoverflow.com/a/1732454/3023252)) because our inputs (card text and possible tags) are constrained?

jwarwick:
Two fold: we only read the data from netrunner-data so we have some control over what arrives and this code only supports limited tags (strong, em, ul). Definitely don't want to support all html tags, that's just asking for trouble.

PR #8660, opened 2026-03-30
Deregister user synchronously on websocket close

xdg:
This commit move user deregistration from the lobby thread to a synchronous update when the websocket closes, mirroring the synchronous registration when a websocket opens.

It also removes the dead code `send-lobby-list`, as there is no socket to receive the message.

This should eliminate a data race between deregistration of an old socket and registration of a new socket that could, in theory, be responsible for reports that blocked users appear in the sent lobby list.


NoahTheDuke:
Good find!

PR #8671, opened 2026-04-05
Fix hosting programs on unrezzed ice

### LINKED ISSUE #6988: Stacked Cards - Multiple trojans
If you have the `Stacked Cards` setting enabled, When installing multiple trojans on an unrezzed ice the UI gets confused where things should be. This only affects the runner UI
image

When the ice is rezzed the cards unstack and it looks ok
image

If `Stacked Cards` setting is not enabled there is no problem
image

jwarwick:
Issue has good images of the failure. Type was getting stripped out for the runner, so checking if it's in the ice zone to make image stacking work correctly.

Fixes #6988

### DIFF
--- a/src/cljs/nr/gameboard/board.cljs
+++ b/src/cljs/nr/gameboard/board.cljs
@@ -818,7 +818,8 @@
 
      (when (pos? (count hosted))
        [:div.hosted
-        (if (and (not (ice? card))
+        (if (and (not (or (ice? card)
+                          (= "ices" (last (:zone card)))))
                  @as-stacked-cards)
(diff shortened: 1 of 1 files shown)


NoahTheDuke:
Why is the type getting stripped? Shouldn't that be fixed instead?

NBKelly:
I think it's so you can't tell asset from upgrade from agenda.
Potentially, we should be preserving type for facedowns if and only if they are installed ice.

NoahTheDuke:
Oh yes, if this is for opponents, that's secret information. Thanks, I forgot.

PR #8677, opened 2026-04-05
Handle chat scroll state across tabs

### LINKED ISSUE #8274: Chat tab shows oldest messages on opening
When I open https://jinteki.net/chat , it defaults to showing the oldest messages of #general. As soon as I input any amount of scroll, click on any channel including #general, or even select a different browser tab and then return, it snaps to the newest messages (bottom of the chat). I suspect it should just always snap to the newest ones.

jwarwick:
Store chat scroll location and message counts when leaving the chat page. Returning to the chat page should restore the previousscroll location, previous channel, and give you a notification about new messages.

New message notification is also displayed when scrolled up on the chat page.

image

Fixes #8274


NBKelly:
neat, that looks pretty good

PR #8679, opened 2026-04-06
Metrics for end of game

### LINKED ISSUE #8582: Feature Request: player engagement metrics
_This is something that I have messaged @NBKelly about, but I thought I should record it here_

As a Standard Format Curator, I would like to be able to graph Jinteki.net metrics such as

- Daily Active Users
- Peak Concurrent Users
- Total Number of Games played per day
- New user sign-ups

So that we can track player engagement with the standard format (over time)

NBKelly:
This adds a couple of functions to gather some anonymous data (fully compliant with the GDPR).

Essentially what we're gathering is:
1) Number of players who play, number of games played, number of new accounts made -> daily
2) An index of cards, their playrate, the faction and id they're played in, and a tiny metric for scoring those cards -> just compound the data until we get restarted, but log it daily

These are stored in `~/jinteki/analytics/...`, and will not be publicly accessible.

Why?
1) Closes #8582. This is both legal and ethical. No personally identifiable information is used whatsoever.
2) A couple of users wanted this for a research project, and I think it's both legal and ethical to provide in this case. The data is all mixed in such a way that it's impossible to identify any information about any player from it.

My metric for scoring:
* 3.5 points for a win
* If you lose, your agenda point total - 3.5

### DIFF
--- a/src/clj/game/core/winning.clj
+++ b/src/clj/game/core/winning.clj
@@ -26,6 +26,7 @@
                            :loser (other-side side)
                            :winning-user (get-in state [side :user :username])
                            :losing-user (get-in state [(other-side side) :user :username])
+                           :losing-score (get-in state [(other-side side) :agenda-point])
                            :reason reason
                            :end-time now
                            :winning-deck-id (get-in state [side :deck-id])
--- a/src/clj/web/analytics.clj
+++ b/src/clj/web/analytics.clj
@@ -1,0 +1,62 @@
+(ns web.analytics
+  (:require
+   [clojure.set :as set]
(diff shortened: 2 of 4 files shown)


Cluedrew:
Am I correct in that identifying information and details are not just inaccessible but completely erased? I have heard stories of people being able to reverse engineer "private concealed" data from data bases that were supposed to hide it, although I don't know enough to say anything definitive about it.

NBKelly:
Yeah, they are completely erased. No store is made in any database of any identifying information, or saved to disk. It's as anonymous as we can get (having a temporary state, like the lobby list).

PR #8682, opened 2026-04-09
Hansei Review's credits and trash should be separate instructions

cls:
Based on some discussion in GLC, I think #6218 actually introduced a regression, in that [Hansei Review](https://netrunnerdb.com/en/card/30048)'s instructions _should_ be separate and so the Corp _should_ be given credits before choosing a card to trash from HQ - and for example [The Zwicky Group](https://netrunnerdb.com/en/card/35069) should trigger between these.

I've never used Clojure, so this was an experience, but hopefully it's not too wrong.

### DIFF
--- a/src/clj/game/cards/operations.clj
+++ b/src/clj/game/cards/operations.clj
@@ -1389,7 +1389,7 @@
     :effect (req (wait-for (gain-credits state :corp 10)
                            (continue-ability
                              state :corp
-                             (if (seq (:hand corp))
+                             (when (seq (:hand corp))
                                {:prompt "Choose a card in HQ to trash"
                                 :choices {:max 1
                                           :all true
@@ -1397,9 +1397,8 @@
                                 :msg {:public "trash a card from HQ"
                                       :corp (msg "trash " (card-str state target {:maybe-visible true}) " from HQ")}
                                 :async true
-                                :effect (trash-cards state side targets {:cause-card card})}
-                               (effect-completed state side eid))
-                              card nil)))}})
+                                :effect (req (trash-cards state side eid targets {:cause-card card}))})
+                             card nil)))}})
 
 (defcard "Hard-Hitting News"
   {:on-play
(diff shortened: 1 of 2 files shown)


src/clj/game/cards/operations.clj
```diff
@@ -1385,21 +1385,21 @@
...
+                             (if (seq (:hand corp))
+                               {:prompt "Choose a card in HQ to trash"
+                                :choices {:max 1
+                                          :all true
+                                          :card #(and (corp? %) (in-hand? %))}
+                                :msg {:public "trash a card from HQ"
+                                      :corp (msg "trash " (card-str state target {:maybe-visible true}) " from HQ")}
+                                :async true
+                                :effect (trash-cards state side targets {:cause-card card})}
+                               (effect-completed state side eid))
```
NoahTheDuke: Instead of this branch, change the `if` to `(when (seq (:hand corp))`

src/clj/game/cards/operations.clj
```diff
@@ -1385,21 +1385,21 @@
...
+                             state :corp
+                             (if (seq (:hand corp))
+                               {:prompt "Choose a card in HQ to trash"
+                                :choices {:max 1
+                                          :all true
+                                          :card #(and (corp? %) (in-hand? %))}
+                                :msg {:public "trash a card from HQ"
+                                      :corp (msg "trash " (card-str state target {:maybe-visible true}) " from HQ")}
+                                :async true
+                                :effect (trash-cards state side targets {:cause-card card})}
```
NoahTheDuke:
```suggestion
                                :effect (req (trash-cards state side eid targets {:cause-card card}))}
```

NoahTheDuke:
Good first attempt! Holler if you want to contribute more and join the dev slack server.

cls:
Thanks! Unfortunately I couldn't get the tests to run locally - it would just consume a ton of CPU and RAM seemingly without effect - so I was unable to really get into it and add a new test case for this and so on. Sorry!

NoahTheDuke:
i'm not sure why the tests didn't run for you, but they pass here so thank you!

PR #8685, opened 2026-04-11
Migrated :pre-trash-interrupt Event

Cluedrew:
Changed the `:pre-trash-interrupt` event to use a context map. Ended up going with `:trashed-cards` for the context key, even though the cards haven't been trashed yet.

I spent longer than usual staring at this one so it got some extra clean-up. Generally just removing extra bits of complexity that had a simpler form.

### DIFF
--- a/src/clj/game/core/moving.clj
+++ b/src/clj/game/core/moving.clj
@@ -368,7 +368,7 @@
      (effect-completed state side eid)
      (wait-for (resolve-trash-prevention state side cards args)
                (let [trashlist (:remaining async-result)
-                     trashed-cards (map :card trashlist)
+                     trashed-cards (not-empty (mapv :card trashlist))
                      _ (update-current-ice-to-trash state trashed-cards)]
                  (wait-for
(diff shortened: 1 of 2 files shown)


src/clj/game/core/moving.clj
```diff
@@ -368,9 +368,10 @@
      (effect-completed state side eid)
      (wait-for (resolve-trash-prevention state side cards args)
                (let [trashlist (:remaining async-result)
-                     _ (update-current-ice-to-trash state (map :card trashlist))]
+                     trashed-cards (map :card trashlist)
```
NoahTheDuke:
```suggestion
                     trashed-cards (not-empty (mapv :card trashlist))
```

NoahTheDuke (approved):

src/clj/game/core/moving.clj
Cluedrew: What is this for?

Cluedrew: I asked and it is nothing more complex than getting a strict and directly testable result.

PR #8686, opened 2026-04-11
stoke remove wait

NBKelly:
This doesn't fix the issue (#8678), but it makes it ~5% less obvious.

Short of adding a waiting prompt to every single corp install, there's no real fix for this.


NoahTheDuke:
huh that's not the worst idea. why don't we have a wait prompt during every install?

PR #8690, opened 2026-04-12
Restrict hidden deck info in api. Filter logs in api.

jwarwick:
The API `/deck` endpoint was always showing the entire deck, restricted to only show deck order if that setting is enabled.
Filter API  `/log` endpoint to only show public and side-specific messages.


NBKelly:
hmm - is it actually possible to engineer a scenario where either side would not have knowledge the full set of cards remaining in their deck (even temporarily)? This is still good though.

PR #8711, opened 2026-04-25
Start new game from replay

### LINKED ISSUE #7707: Feature Request: Continue from here
Adding some kind of "Recreate game state from here." button on replays would be cool.

lostgeek:
https://github.com/user-attachments/assets/29fa24f2-0440-4fe2-b1ee-c70363a74307

This PR implements a new feature for replays, where you can start a new game from a given replay state. This cannot be done without losses, as replays only capture client-diffs and therefore cannot preserve game state regarding stateful information (i.e. Oppo Research needing to know, whether the Runner has completed a run last turn, or Simulchip needing to know whether a program has been trashed this turn).

Nevertheless, we are able to use `card-init` to restore most functionality and this feature should be very usable for recreating board states for analysis.

Closes https://github.com/mtgred/netrunner/issues/7707


NBKelly:
I think this is good for public testing

PR #8714, opened 2026-04-27
Migrated :candidates-determined Event

Cluedrew:
I migrated the `:candidates-determined` event. The `:hq` flag is wrapped `:breached-server` even though it is not needed. Future proofing I guess.

I also asked about how the event functions will change when this migration is done. I simplified the context form to make sure replacing `& targets` with `context` will work. Turns out nothing actually uses the transformation (and if something does need to use it, go through `target` instead. So we just get to strip out some unneeded code.

### DIFF
--- a/src/clj/game/macros.clj
+++ b/src/clj/game/macros.clj
@@ -22,12 +22,7 @@
                         (contains? t :value))
                  (:value t)
                  t))
-      context (let [t (first targets)]
-                (if (and (map? t)
-                         (contains? t :uuid)
-                         (contains? t :value))
-                  (:value t)
-                  t))
+      context (first targets)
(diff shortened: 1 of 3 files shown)


src/clj/game/macros.clj
```diff
@@ -22,12 +22,7 @@
                         (contains? t :value))
                  (:value t)
                  t))
-      context (let [t (first targets)]
-                (if (and (map? t)
-                         (contains? t :uuid)
-                         (contains? t :value))
-                  (:value t)
-                  t))
+      context (first targets)
```
NoahTheDuke, line 25: yeah i think this is a holdover from me copying the `target` logic over

PR #8720, opened 2026-05-01
Great macro shuffle: req, and effect

NoahTheDuke:
Originally, `effect` was a way to write the abilities/effects of card abilities in a concise way. Because everything took `state` and `side`, it was easy to trim those and have a list of actions that happened one after another. With the advent of "async" abilities and the growing complexity of our abilities, it stopped being useful a long time ago but I only got around to removing it a couple days ago (in #8718).

With that done, the question became "is `req` still a good name for the base macro?" After discussion with the team, we determined that it was not, and we should in fact match `:effect` with `effect` and `:req` with `req`. That has resulted in this behemoth.

In short, `effect` is now what `req` used to be, a macro that expands to a function taking `state`, `side`, `eid`, `card`, and `targets`. It binds the appropriate variables but otherwise doesn't do anything else. And `req` is now a macro relying on `effect` that wraps all forms in an implicit `and` to make it easier/more readable to check conditions. 90% of uses of the `:req` function were a single check or a handful of checks that were wrapped in an `and` already, so stripping those was fairly simple. The more complex cases can continue to rely on a `let` as `(and (let [...] (and ...)))` is a no-op.

I did no indentation fixes so as we edit these files, we'll probably see some ancillary changes which I think is better than updating the indentation of ~50 files.

It will take some getting used to but I believe in us. A better world is possible, we just have to rely on `rewrite-clj`.


Cluedrew:
Woot! I know it is a lot of changes but as a newer who found the `effect`/`req` divide to be a bit confusing I am happy to see this new version.

Issue #8729, opened 2026-05-15
KPI reveales ice installed


Gruntownie:
- replay :  https://jinteki.net/replay/8addb042-76e3-4703-b81c-aac0b6b2420b
- turn 21 3rd click

odsywrenn:
I think this is related to the option to "trash like cards" being turned on, when installing ice on a server that already has ice. Here is another replay from a month ago where a Greasing the Palm install on turn 9 click 1 revealed the Drafter being installed to the Runner. At the time of the game, it was visible to the Runner while making the decision, but during the replay, it's only visible (in "Runner view") if you step past the decision and then step backwards: https://www.jinteki.net/replay/c8f64cb8-de1c-4ae5-b4d4-2ee7e2e8582f

najmam:
Could reproduce this bug on the latest master, and the following indeed need to be true for it to happen:
- corp has "trash like cards" enabled
- the destination server has ice installed

The runner gets shown the ice that had been selected for installation, during the time it takes for the corp to select whether they trash already installed ice.

The bug is also reproduced when using Greasing the Palm instead of KPI.

<img width="1920" height="902" alt="Image" src="https://github.com/user-attachments/assets/6cf7952a-1638-46ba-9769-13e6f0cc697f" />

NoahTheDuke:
Good detective work, thanks for putting that together.

PR #8731, opened 2026-05-17
Avoid leaking selected ice to runner when trash-like-cards is enabled

najmam:
This PR is not ready for merging, it just makes progress on issue #8729 by reproducing it in a failing test.

One way to make the test pass is to set `:waiting-prompt false` instead of `true` in [game.core.installing/corp-install-pay](https://github.com/mtgred/netrunner/blob/0a80cb0856cd43da768159bc3c6db8bac2c0b60a/src/clj/game/core/installing.clj#L400), but I'm not sure it's the right fix, could be too heavy-handed. I didn't work out the consequences enough to assess that.

Also note that the added test is filed under the existing one for Key Performance Indicators, however the bug is not specific to that card. Don't know if the test would be better placed elsewhere.

Not sure I'll have time to continue working on this so feel free to pick it up.

### DIFF
--- a/test/clj/game/cards/operations_test.clj
+++ b/test/clj/game/cards/operations_test.clj
@@ -2731,6 +2731,27 @@
       (is (no-prompt? state :corp) "No lingering prompts")
       (is (no-prompt? state :runner) "No lingering prompts"))))
 
+(deftest key-performance-indicators-issue-8729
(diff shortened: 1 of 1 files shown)


najmam:
Issue solved by https://github.com/mtgred/netrunner/pull/8740

PR #8745, opened 2026-05-27
Ice swap vs beta build

### LINKED ISSUE #8737: Beta build + pichacao +sipa bugged
Link to bug replay: https://jinteki.net/bug-report/bf38a709-4cde-4220-9434-c0421c9a380f?b=0

Description:

I used beta build to install pichacao from stack onto diviner (the only ICE on RnD) and made a run on rnd, broke diviner with buzzsaw and used sipa to swap it with unknown HQ ICE. After run ended pichacao should've been added to the top of the stack but it stayed on the board

NBKelly:
Closes #8737

We needed to call get-card on the card to bounce.

But then it turns out calling `get-card` on the old reference wouldn't return a result because it was looking up the host in the index of the old hosted reference and not the new one.

I kind of wonder why we don't have a better system for finding cards. Surely a map in state of zones like `{:corp {:cids {"aaaaaaaaa" [:servers :hq :ices] ...}}}` could take a lot of awkward code out of the code base (I get that it will add more too).

### DIFF
--- a/src/clj/game/cards/events.clj
+++ b/src/clj/game/cards/events.clj
@@ -307,17 +307,17 @@
                            :effect (effect (runner-install state side eid target {:ignore-all-cost :true :msg-keys {:display-origin true :source-card card}}))}
                           card nil)]
                        (continue-ability state side
-                         (run-any-server-ability
-                           {:events [{:event :run-ends
-                                      :unregister-once-resolved true
-                                      :duration :end-of-run
-                                      :interactive (effect true)
-                                      :automatic :last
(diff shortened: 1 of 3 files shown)


NoahTheDuke:
There's an argument to be made that we don't need to track any zones at all, but instead put all cards into a vector or map, and track where they are on the card object directly. And then when you say "what's in the play-area", you just say `(filter in-play-area? all-cards)`.

It would solve a ton of issues, but it would require a complete rewrite of a ton of code

PR #8747, opened 2026-05-28
Fix card-def Error

Cluedrew:
The two versions of the error message in `card-def` were not consistent. So I changed it so they had to be.

I was going to do this with a larger fix that didn't work out, so here is this tiny change on its own.


NoahTheDuke:
lol thanks, i don't even know why we're passing the message into `:msg` but i won't question it right now

PR #8748, opened 2026-05-29
i18n game log/messages

NoahTheDuke:
this is another take at @csbisa 's monumental #7971 which stalled as we were unable to get it all finalized.

csbisa had a really good starting point but the number of technical details and vast array of functions to render the text was a barrier to any translation progress, so my attempt here is smaller: use fluent files to build templates, and then have a mechanical process to combine those templates into our desired messages.

i started clean slate without looking at 7971, but i plan on going back and plucking what i can from it because it has some good features.


resources/public/i18n/en/messages.ftl
```diff
@@ -0,0 +1,839 @@
...
+    [one] 1 advancement counter from {$card-str}
+    *[other] {$count} advancement counters from {$card-str}
+}
+
+place-virus-counters = place {$count ->
+    [one] 1 virus counter on {$title}
+    *[other] {$count} virus counters on {$title}
+}
+
+charge-card = trash {$card-str} {$count ->
```
csbisa: trash -> charge

PR #8759, opened 2026-06-20
Daniella fix

### LINKED ISSUE #8754: Bug for Daniela Jorge Inácio & René “Loup” Arcemont: Party Animal
Image
Daniela should fire before Loup as it is an additional cost. and Loup triggers off the trash. Costs should be paid first.

Additionally, I can't see what was sent to the bottom by Daniella and in what order, this should be information the runner has. This should be secret information in the chat log.

(You can't always deduce it from your hand when you have Loup because you might start with cards A B C in hand, and after you see you have card A and B in hand. but does that mean you trashed A & C and drew an extra copy of A or trashed B and C and drew an extra copy of B?) and currently you cant deduce the order.

You can see the bug here:
what jnet does is first Loup draws an extra Steelskin, then Daniela bottoms Steelskin and Chisel.
I only know this because I only have 2 paladins in the deck.
So there are 3 issues:
1. This is the wrong order to resolve the effects
2. Unclear what cards were bottomed by Daniela.
3. Unclear the order of the cards on the bottom (which could matter especially if the stack was empty.)
[CyberShark's game.json](https://github.com/user-attachments/files/28864974/CyberShark.s.game.1.json)

NBKelly:
This adds support for additional trash costs.

Daniella now applies one of these. This also adds support for hidden info in payment messages.

Closes #8754

### DIFF
--- a/src/clj/game/core/costs.clj
+++ b/src/clj/game/core/costs.clj
@@ -1011,14 +1011,14 @@
                                          (str " (" (enumerate-cards async-result :sorted) ")")))
                    :paid/type :trash-entire-hand
                    :paid/value (count async-result)
-                   :paid/targets async-result})
+                   :paid/targets async-result}
                   {:paid/msg (str "trashes all (" (count async-result) ") cards in [their] grip"
                                   (if (= :runner side) "[their] grip" "HQ")
                                   (when (seq async-result)
                                     (str " (" (enumerate-cards async-result :sorted) ")")))
                    :paid/type :trash-entire-hand
                    :paid/value (count async-result)
-                   :paid/targets async-result}))))
+                   :paid/targets async-result})))))
 
 ;; TrashHardwareFromHand
 (defmethod value :trash-hardware-from-hand [cost] (:cost/amount cost))
--- a/src/clj/game/core/costs.clj
+++ b/src/clj/game/core/costs.clj
@@ -25,6 +25,17 @@
    [game.utils :refer [enumerate-cards enumerate-str quantify same-card?]]))
 
 (defn- can-forfeit? [card] (not (get-in card [:flags :cannot-forfeit])))
+(defn- side-paid-msg [side] (keyword (str "paid/" (name side) "-msg")))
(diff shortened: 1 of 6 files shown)


src/clj/game/core/costs.clj
```diff
@@ -25,6 +25,17 @@
    [game.utils :refer [enumerate-cards enumerate-str quantify same-card?]]))
 
 (defn- can-forfeit? [card] (not (get-in card [:flags :cannot-forfeit])))
+(defn- side-paid-msg [side] (keyword (str "paid/" (name side) "-msg")))
```
NoahTheDuke, line 28: One of my goals with namespacing the keywords is to improve searching, so this should be explicit (even tho it's not "DRY"):

```clojure
(if (= :corp side) :paid/corp-msg :paid/runner-msg)
```

NoahTheDuke:
hmm actually, on second thought, i don't think we need the `:paid/corp-msg` stuff, my work in #8748 should cover that already. i really gotta finish that up, to stop blocking these other PRs.

NoahTheDuke:
which is to say, the daniella fix is good, the engine changes shouldn't be necessary, but don't do anything just yet.

NBKelly:
Your changes will allow for payment messages (potentially) having hidden info?

That's the last block for side-specific messages before I can go through the old card pool and fix up all the old cards, and what I'm hoping to get here.

NoahTheDuke:
sort of. the work i did converts all of the costs to use the new message maps, which includes the ability for side-dependent rendering, and i think we can put them there.

src/clj/game/core/costs.clj
```diff
@@ -25,6 +25,17 @@
...
+(defn- side-zone [side zone]
```
Cluedrew, line 29: There isn't a helper for this? It seems like there should be. (Yes, this is just a clean-up thing.)

NoahTheDuke:
i've opened this PR locally and i'm making it work with #8748 and then i'll merge it locally and push, so you don't have to do anything extra.

PR #8767, opened 2026-06-27
feat: default decks and game options

### LINKED ISSUE #8614: Automatic save gameplay
Hello,

Would it be possible to create an option for "my deafault lobby settings"? As in I can make a blueprint where my lobby always starts with (for example) a custom name, save game on, a set password etc?

If not, would it be at least possible to add a setting in the account setings to make games I play be automatically saved?

filipesilva:
Adds a `Select Default Deck` button to casual game deck selection when you have a default deck for that side+format. 

CleanShot 2026-06-27 at 18 34 07@2x

Defaults decks show a `★` before their name. You can have one default deck for each side+format, and are set in the deck details screen with the `Set Default` button for constructed formats.

CleanShot 2026-06-27 at 18 35 29@2x

You can auto-select the default decks on game create/join/swap sides with a new setting in the new `Game Settings` section, along with the the existing game format, and new game description, password, and replay defaults.

CleanShot 2026-06-29 at 22 42 17@2x

One last tweak is that the password field will be focused when it shows up. This will make it a bit faster to add or edit the password on game creation and on join.

Fix https://github.com/mtgred/netrunner/issues/8614

### DIFF
--- a/src/clj/web/lobby.clj
+++ b/src/clj/web/lobby.clj
@@ -10,7 +10,7 @@
     [crypto.password.bcrypt :as bcrypt]
     [game.core :as core]
     [game.utils :refer [server-card]]
-    [jinteki.utils :refer [select-non-nil-keys side-from-str superuser? to?]]
+    [jinteki.utils :refer [constructed-game? select-non-nil-keys side-from-str superuser? to?]]
     [jinteki.preconstructed :refer [all-matchups]]
     [jinteki.validator :as validator]
     [medley.core :refer [find-first]]
@@ -545,7 +545,7 @@
 
 (defn auto-select-decks
   [db lobby]
-  (if (or (:started lobby) (not= "casual" (:room lobby)))
+  (if (or (:started lobby) (not= "casual" (:room lobby)) (not (constructed-game? lobby)))
     lobby
     (update lobby :players
             (partial mapv #(if (:deck %) % (maybe-auto-select-deck db lobby %))))))
--- a/src/clj/web/lobby.clj
+++ b/src/clj/web/lobby.clj
@@ -25,6 +25,8 @@
 
 (read-write/print-time-literals-clj!)
 
+(declare auto-select-decks)
+
 (defonce telemetry-buckets (agent {}))
 (defn log-delay!
   [timestamp id]
@@ -345,8 +347,9 @@
   (update lobby :messages conj message))
 
 (defn try-create-lobby
-  [uid user ?data]
+  [db uid user ?data]
   (let [lobby (-> (create-new-lobby {:uid uid :user user :options ?data})
+                  (->> (auto-select-decks db))
                   (send-message
                     (core/make-system-message (str (:username user) " has created the game."))))
         new-app-state (swap! app-state/app-state update :lobbies
@@ -359,7 +362,7 @@
 
 (defmethod ws/-msg-handler :lobby/create
   lobby--create
-  [{{user :user} :ring-req
+  [{{db :system/db user :user} :ring-req
     uid :uid
     ?data :?data
     id :id
@@ -368,7 +371,7 @@
     (if (:block-game-creation @app-state/app-state)
       (ws/chsk-send! uid [:lobby/toast {:message :lobby_creation-paused
                                         :type "error"}])
-      (try-create-lobby uid user ?data))
+      (try-create-lobby db uid user ?data))
     (log-delay! timestamp id)))
 
 (defn clear-lobby-state [uid]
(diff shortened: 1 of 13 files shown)


src/cljs/nr/pending_game.cljs
```diff
@@ -35,26 +35,25 @@
                   (tr-non-game-toast [:lobby_select-error "Cannot select that deck"] "error")))
   (reagent-modals/close-modal!))
 
+(defn deck-valid-for-lobby?
```
filipesilva, line 37: This just raises the deck filters into a single one, so it can be reused later in `default-deck-for-lobby`.

src/clj/web/lobby.clj
```diff
@@ -526,6 +529,27 @@
...
+(defn maybe-auto-select-deck
+  [db lobby player]
+  (let [options (:options (app-state/get-user (:uid player)))
+        deck (when-let [deck-id (auto-select-deck-id options (:format lobby) (:side player))]
+               (process-deck (find-deck-for-user db deck-id (:user player))))]
+    (if (and deck (valid-deck-for-lobby? lobby deck))
+      (assoc player :deck deck)
+      player)))
+
+(defn auto-select-decks
```
filipesilva, line 553: This fn is the reason why a lot of the other fns in this namespace now take a `db` as a first param. Deck selection needs a db to look up decks, and create/join/swap can now select a deck.

It's a bit wonky how a lot of a do call sites do `(->> (auto-select-decks db))` because they are in the middle of a `->` macro, but db is usually the first param pretty much everywhere so I didn't want to change that.

src/cljc/jinteki/utils.cljc
```diff
@@ -50,6 +50,14 @@
   (or (get-in @state [:runner :tag :is-tagged])
       (pos? (count-tags state))))
 
+(defn constructed-format?
```
filipesilva, line 53: Some fns that were only used in the FE are now also used in the BE, so I moved them here.

src/cljs/nr/account.cljs
```diff
@@ -378,14 +379,37 @@
...
-           [:select.format
-            {:value (or (:default-format @s) "standard")
-             :on-change #(swap! s assoc :default-format (.. % -target -value))}
-            (doall
-             (for [[k v] slug->format]
-               ^{:key k}
-               [:option {:value k :data-i18n-key k} (tr-format v)]))]]
+           [tr-element :h3 [:settings_lobby-settings "Game Settings"]]
+           [:div
+            [tr-element :h4 [:lobby_default-game-format "Default game format"]]
```
filipesilva, line 384: This existing translation key is a bit different than the rest of the settings. It's `:lobby_` instead of `:settings_` like most of the others on this page. I left it as is, but used `:settings_` for the rest of the section.

src/cljs/nr/game_row.cljs
```diff
@@ -214,7 +214,7 @@
    :pending-game_new-player "Looking To: Learn the game"})
 
 (defn description-span [description]
-  (when (and description (not= description :new-game_default))
+  (when (and description (not= description "new-game_default"))
```
filipesilva, line 217: This was a bug on live atm. If you go to create a game, select one of the other descriptions, then go back to the default, you'd get a dash with nothing else. That's because the argument to this fn is a string, not a kw.

Image

src/cljs/nr/new_game.cljs
```diff
@@ -177,11 +177,11 @@
   [:<>
    [:p
     [:label
-     [:input {:type "checkbox" :checked (:private @options)
+     [:input {:type "checkbox" :checked (:protected @options)
```
filipesilva, line 180: This was a bug as far as I can tell. Options didn't have a `:private` key.

Cluedrew:
You should probably get some other people's input on this, but it seems to me that the "default deck" marker should go with the format the deck is for, because the default is by format and also because names are a bit more freely form.

From
`* Pow- Pow- Power Charges         Standard Legal`
To something like:
`Pow- Pow- Power Charges         * Standard Legal`

filipesilva:
@Cluedrew for comparison:

CleanShot 2026-07-03 at 16 23 56@2x

CleanShot 2026-07-03 at 16 37 47@2x

No strong opinion on my side. Do you like the changed version better?

Cluedrew:
Yup, I didn't like the disruption in the deck names and that fixes that.

src/cljs/nr/pending_game.cljs
```diff
@@ -35,26 +34,25 @@
...
+  [deck {:keys [format singleton]} side]
+  (and (= side (get-in deck [:identity :side]))
+       (or (not singleton) (singleton-deck? deck))
+       (or (= "casual" format)
+           (get-in deck [:status (keyword format) :legal]
+                   (get-in (trusted-deck-status (assoc deck :format format))
+                           [(keyword format) :legal]
+                           false)))))
```
NoahTheDuke: Instead of nested get defaults, can you just move these to the `or`? `(or ... (get-in deck [:status ...]) (get-in (trusted-deck-status ...) [...]) false)`

src/cljs/nr/utils.cljs
```diff
@@ -404,6 +404,12 @@
     [:button.on {:on-click f :key on-text} on-text]
     [:button.off {:on-click f :key off-text} off-text]))
 
+(defn focus-on-mount
```
NoahTheDuke, line 407: this is clever!

src/cljs/nr/pending_game.cljs
filipesilva: Yeap, done now

PR #8770, opened 2026-06-28
fix: hosting triggers uninstalled too

### LINKED ISSUE #8752: Luana Campos doesn't return the bad pub if it becomes hosted on cupellation
Link to bug replay: https://jinteki.net/bug-report/fa06a73c-7298-4517-b892-fcaecf45454e?b=0

Description:

used cupellation to host luana on it. As luana becomes uninstalled she should give back the hosted bad pub counters but that doesn't happen

filipesilva:
Fix #8752

### DIFF
--- a/src/clj/game/core/hosting.clj
+++ b/src/clj/game/core/hosting.clj
@@ -1,5 +1,6 @@
 (ns game.core.hosting
   (:require
+    [clojure.string :as str]
     [game.core.card :refer [active? assoc-host-zones corp? get-card program? installed? rezzed? runner?]]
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
I'm blown away this hasn't come up before. Thanks for the fix

filipesilva:
@NoahTheDuke that's because there's only 2 cards with `:uninstall` keys: `Luana Campos` and `Muertos Gang Member`.

NoahTheDuke:
oh of course. netrunner is too damn complicated lol

PR #8776, opened 2026-06-30
feat: card state colors

### LINKED ISSUE #8724: Colorblind Mode
1. I often struggle with the green and orange 'halos' around cards when selecting multiple cards (like during discard).

2. I also struggle to count influence in a runner Deck because Anarch and Shaper cards look similar. Maybe the ability to add a faction watermark or border to cards?

Issue 1 impacts me way more often and significantly than issue 2.

filipesilva:
Adds a new `Card state colors` setting, with a preset for colorblind-safe colors, and custom colors.

CleanShot 2026-06-30 at 17 28 49@2x

CleanShot 2026-06-30 at 17 28 59@2x

CleanShot 2026-06-30 at 17 27 59@2x

Fix https://github.com/mtgred/netrunner/issues/8724


src/css/account.styl
```diff
@@ -40,3 +40,54 @@
 
   .reset-all
       margin-top: 10px
+
+.card-colors-preview
```
filipesilva, line 1: This file is a bit weird. As far as I can tell, the `#account` selector at the top doesn't match anything, and so the styles in it are not used. I can remove them if you want.

src/css/gameboard.styl
```diff
@@ -1,5 +1,14 @@
...
 @require "palette"
+
+// unquote needed to support runtime css vars
+glow(prop, pct)
+    unquote('color-mix(in srgb, var(' + prop + ') ' + pct + '%, transparent)')
+glow-wash(prop, pct)
+    unquote('color-mix(in srgb, color-mix(in srgb, var(' + prop + ') 20%, white) ' + pct + '%, transparent)')
+glow-spin(prop, pct)
+    unquote('hsl(from var(' + prop + ') calc(h - 15) s l / ' + pct + '%)')
+
```
filipesilva, line 11: These are the runtime equivalents of the Stylus color functions the card-state glows used. The stylus color functions were compile-time transformations, but to use the css variables I needed runtime functions instead. The unquote was needed because stylus doesn't parse the `color-mix` and `hsl` fns.

Here's what the new fns replace:
- `glow`: `alpha(c, pct%)`
- `glow-wash`: `lighten(alpha(c, pct%), 80%)` (lighten 80% = 20% color + 80% white)
- `glow-spin`: `spin(alpha(c, pct%), -15deg)`

src/css/account.styl
NoahTheDuke: Our stylus files are messed up. A year or two back, Eevee tried to move us to pure css (cuz we're not using anything in stylus that isn't supported in all modern browsers already), but ran out of steam.

If you find that some files are broken or unused, you're welcome to delete/reorder/do whatever you need.

NoahTheDuke:
Why is it called "poison"? is that a recent netrunner term I don't know about?

src/css/account.styl
filipesilva: Ok will keep an eye out. Meanwhile removed this one.

filipesilva:
I'm a recent player so I wouldn't know if it's existed for a while. But my understanding is that putting stuff in archives that'd bad for the runner is called "poisoning the archives". You can see it [here](https://netrunnerdb.com/en/card/36038):
> Where this card shines, though, is in protecting your Archives by "poisoning" them.

That's also how NBKelly referred to the state in https://github.com/mtgred/netrunner/issues/8724#issuecomment-4754234881.

NoahTheDuke:
Looks like you've got a conflict, but i'm good to merge once it's resolved.

filipesilva:
@NoahTheDuke thanks for the review 🙏 conflict should be sorted now, it was just both changes to settings.

PR #8783, opened 2026-07-09
inner leftpane fixes

filipesilva:
This PR adds fixes for 2 things. First is how some button panels, like self-modifying code and malapert data vault, could break the board layout vertically by pushing it down. Now they will get scroll bars instead:

CleanShot 2026-07-10 at 00 09 35@2x

CleanShot 2026-07-10 at 00 09 58@2x

The second fix is for the spacing between the boxes in that button panel column. You can see in the first image that there was no spacing between game start, last revealed, and the prompt. In the second image there is a bit of spacing, like in the left-most column.

### DIFF
--- a/src/css/gameboard.styl
+++ b/src/css/gameboard.styl
@@ -675,10 +675,12 @@ glow-spin(prop, pct)
         display-flex()
         flex-direction(column)
         justify-content(space-between)
+        max-height: 100%
 
         .inner-leftpane
             flex(1)
             display-flex()
+            min-height: 0
 
             .left-inner-leftpane
                 width: 201px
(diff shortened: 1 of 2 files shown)


src/css/gameboard.styl
```diff
@@ -737,25 +741,33 @@ glow-spin(prop, pct)
...
-                                height: 234.833px
+                                max-width: 100%
+                                height: auto
+                                aspect-ratio: unquote("168 / 234.833")
 
                                 >div:first-child
                                     display:inline
-                                    width: 168px !important
-                                    height: 234.833px
+                                    width: 100% !important
```
NoahTheDuke: still need !important?

filipesilva: I'm not quite sure why it was there to begin with, so I just left it. Tried removing it and I don't really see a difference:

CleanShot 2026-07-10 at 13 09 15@2x
CleanShot 2026-07-10 at 13 09 52@2x

I also tried removing it from master directly and see no difference, so I guess it's a leftover of something that's gone now? (images below are before and after removal)

CleanShot 2026-07-10 at 13 12 54@2x
CleanShot 2026-07-10 at 13 13 36@2x

Pushed a commit that removes the `!important`.

NoahTheDuke: i think we have a ton of leftover `!important` uses because folks weren't super familiar with css.

PR #8785, opened 2026-07-11
Zoom controls & command button

filipesilva:
This PR contains four changes aimed at improving accessibility when not using a keyboard, with the ulterior goal of improving the tablet/mobile experience. 

There's one big caveat that I should mention in advance. On iOS, when you add a page to the home screen so it works as a standalone app, and you open it in landscape mode, it will get a deadzone along the top and bottom of the screen. This deadzone is roughly the height of the top navbar, and bottom chat buttons. This means that taps there won't register at all until you pinch to zoom until they are big enough to not be affected by the deadzone. I wasn't able to find a good way around this, but this is still the best way to play on iOS.

The first change is adding zoom controls. These work similarly to browser zoom via cmd/ctrl =/-/0, but are independent from it since it's a css zoom instead. They are persisted per device, not per account.

CleanShot 2026-07-12 at 00 53 43@2x

Note that these buttons have slightly different padding than other elements in the bar. Even though they look like they are just the `-`/`+` characters, they actually take up all the space from `100%` to the edge. This is to make them easier to click/tap, especially when you're doing it repeatedly.

CleanShot 2026-07-12 at 00 55 39@2x

The second change is a dedicated command button with the most common commands. It works the same as typing `/` then selecting from the list. Commands that require no parameters are ran immediately (e.g. undo-click). Commands that require parameters will just get input on the chat box (e.g. `counter type n`).

CleanShot 2026-07-12 at 00 54 22@2x

The third change is a dedicated messages button, with configurable messages. It works like the command button:

CleanShot 2026-07-17 at 21 28 48@2x

The message list can be customized in account settings:

CleanShot 2026-07-17 at 21 29 47@2x

The fourth change is fully centering the game start/end screen. Without this change, it would often end up with the keep/mull buttons outside the screen, and you'd need to zoom out a lot to use them. This is how it looks on my phone:

Screenshot 2026-07-12 at 00 51 09

### DIFF
--- a/src/cljs/nr/gameboard/log.cljs
+++ b/src/cljs/nr/gameboard/log.cljs
@@ -40,13 +40,17 @@
                  ^{:key i}
                  [:span " " influence-dot " "]))]])))
 
+(defn send-text [text]
+  (when (and (not (:replay @game-state))
+             (seq text))
+    (reset! should-scroll {:update false :send-msg true})
+    (ws/ws-send! [:game/say {:gameid (current-gameid app-state)
+                             :msg text}])))
+
 (defn send-msg [s]
   (let [text (:msg @s)]
-    (when (and (not (:replay @game-state))
-               (seq text))
-      (reset! should-scroll {:update false :send-msg true})
-      (ws/ws-send! [:game/say {:gameid (current-gameid app-state)
-                               :msg text}])
+    (when (seq text)
+      (send-text text)
       (swap! s assoc :msg ""))))
 
 (defn send-typing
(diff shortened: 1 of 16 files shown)


src/cljs/nr/gameboard/log.cljs
```diff
@@ -216,10 +223,24 @@
                   [:span {:on-mouse-over #(swap! state assoc :completion-highlight i)
                           :on-click #(do
                             (fill-completion state completion-text)
-                            (.focus @!input-ref))}
+                            (if (autosend? completion-text)
```
filipesilva: This now matches the behaviour on using the keyboard to select a command suggestion. Without it, clicking on something that should be auto-sent, like `undo-click`, would just put it in the input box instead.

PR #8787, opened 2026-07-12
feat: zoom last-played-or-rezzed

### LINKED ISSUE #8500: Add option for card preview to pin the most recently played/rezzed card
From aksu:
> would be peak if Jnet had a option to keep the most recent card rezzed / played on the zoomed in view

filipesilva:
Adds a new option to zoom on the last (operation/event/runner install/face up corp install) or rezzed card. It works the same as if you mouse-over the card. This option is accessible in both account settings and in-game settings, like `Keep zoomed cards on screen`.

This is useful to direct both player and spectator attention to what's currently happening, and especially useful when streaming.

Fix #8500


src/cljs/nr/gameboard/actions.cljs
```diff
@@ -50,8 +51,14 @@
 (defn handle-diff! [{:keys [gameid diff]}]
   (when (= gameid (str (current-gameid app-state)))
     (let [old-sequence (:sequence @game-state)
+          old-last-played-or-rezzed (:last-played-or-rezzed @game-state)
           patch (differ/patch @last-state diff)]
       (reset! game-state patch)
+      (let [new-last-played-or-rezzed (:last-played-or-rezzed @game-state)]
+        (when (and new-last-played-or-rezzed
+                   (not= old-last-played-or-rezzed new-last-played-or-rezzed)
```
filipesilva, line 62: This is the trick. Since `:last-played-or-rezzed` includes an `eid`, it's possible to distinguish repeated cards played, and to have a notion of when it was played. So the state stays pure, we don't get double triggers on state changes, and still work with playing/rezzing the same card twice.

I tried some other approaches that didn't involve the `eid`, but they needed to bend over backwards to keep track of time, or broke weirdly with pinning.

src/clj/game/core/revealing.clj
```diff
@@ -18,6 +19,16 @@
...
 
+(defn set-last-played-or-rezzed
+  "Tracks the most recently played, rezzed, or turned-faceup public card so
+  clients can zoom it when the :zoom-last-played-or-rezzed setting is on.
+  Includes current eid so playing the same card twice is distinguishable."
+  [state card]
+  (when (and card (is-public? card (other-side (to-keyword (:side card)))))
+    (swap! state assoc :last-played-or-rezzed
+           {:card (select-keys card [:title :printed-title :code :side])
+            :eid (:eid @state)})))
```
NoahTheDuke, line 30: to be clear, this is the number, right?

NoahTheDuke:
i like this

src/clj/game/core/revealing.clj
filipesilva: This is a monotonically increasing number, so it's always different. Starts at 0 for a new game state and is increased by a lot of stuff via `make-eid`.

PR #8788, opened 2026-07-12
fix: don't drop :type from prompt card

### LINKED ISSUE #8784: Tiny card back when paying access costs
If you have to pay access costs to access, e.g. vs gagarin, you see a tiny card back, followed by full card front.

Image

Image

filipesilva:
Its used in the prompt div to check it's not the basic action:
```
;; src/cljs/nr/gameboard/board.cljs:1802
     (when (and card (not= "Basic Action" (:type card)))
```

Fix #8784

### DIFF
--- a/src/clj/game/core/diffs.clj
+++ b/src/clj/game/core/diffs.clj
@@ -251,7 +251,7 @@
   (when same-side?
     (-> prompt
         (update :eid #(when (:eid %) (select-keys % [:eid])))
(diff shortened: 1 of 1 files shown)


filipesilva:
After this fix, the prompt in the issue looks like this, while after paying you still see the card as normal:

Screenshot 2026-07-12 at 21 03 59

PR #8805, opened 2026-07-24
Use :edn packer in dev

filipesilva:
This way we can use edn packer on dev for inspectability, and still use msgpack on prod.

### DIFF
--- a/src/clj/web/pages.clj
+++ b/src/clj/web/pages.clj
@@ -11,7 +11,7 @@
 
 (defn index-page
   ([request] (index-page request nil nil))
-  ([{user :user server-mode :system/server-mode} og replay-id]
+  ([{user :user server-mode :system/server-mode ws-config :system/ws-config} og replay-id]
    (html-response
      200
      (hiccup/html5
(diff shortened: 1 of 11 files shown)


src/clj/web/pages.clj
```diff
@@ -48,7 +48,8 @@
         (hiccup/include-js "https://maxcdn.bootstrapcdn.com/bootstrap/3.2.0/js/bootstrap.min.js")
         (hiccup/include-js "/lib/js/toastr.min.js")
         [:script {:type "text/javascript"}
-         (str "var user=" (json/generate-string user) ";")]
+         (str "var user=" (json/generate-string user) ";"
+              "var ws_config=" (json/generate-string (or ws-config {})) ";")]
```
filipesilva, line 52: The FE sente config needs to agree with the BE one. So here we write the config to a window var, then pick it up in `src/cljs/nr/ws.cljs`.

src/cljs/nr/ws.cljs
```diff
@@ -24,7 +27,9 @@
           ?csrf-token
           {:type (if (get-in @app-state [:options :disable-websockets]) :ajax :auto)
            :ws-kalive-ms 2500 ;; note - supposedly this will help SEA/LATAM connections
-           :packer (msgpack/get-packer)
+           :packer (if (= (:packer ws-config) "edn")
```
filipesilva, line 30: `"edn"` instead of `:edn` like in the BE because kws get printed down to strings in json.

src/clj/web/ws.clj
```diff
@@ -51,15 +33,68 @@
...
+    (reset! server_ (assoc server
+                           :buffer buffer
+                           :buffer-size buffer-size))))
+
+(defn stop-server! []
+  (when-let [{:keys [buffer]} @server_]
+    (close! buffer))
+  (reset! server_ nil))
+
+(defn handshake-handler [& args]
```
filipesilva, line 68: These fns are just handling the case where server is not initialized yet.

NoahTheDuke:
i've been thinking about this and unless we wanna make all of our `ws/chsk-send!` calls annoying, it's probably not getting much better than this. i'm worried a little bit about the dangling `go` channels when resetting, but idk how to avoid that.

PR #8813, opened 2026-07-28
fix: don't store cards on local storage

### LINKED ISSUE #8809: Deck builder stuck on `Loading deck collection...` on Safari
Originally reported on `jnet-nrdb-cobra` by `ossa`.

It looks like this:

Image

Seems to be Safari specific, doesn't happen on Chrome for the same account. I also see a `QuotaExceededError: The quota has been exceeded.` on the console, perhaps related.

filipesilva:
We can get the same behaviour using cache headers.

Fix https://github.com/mtgred/netrunner/issues/8809

### DIFF
--- a/src/clj/web/utils.clj
+++ b/src/clj/web/utils.clj
@@ -25,6 +25,22 @@
 (defn html-response [status-code msg]
   (resp/status (resp/content-type (resp/response msg) "text/html") status-code))
 
+(defn cached-response
+  "Let the browser cache a response in its HTTP cache, keyed on etag.
+  Cache-Control no-cache means the browser must revalidate with If-None-Match
+  before using the cached copy, and a 304 response does not need to send a body.
+  See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cache-Control#up-to-date_contents_always
+  and https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/If-None-Match"
+  [request etag body-fn]
+  (let [etag (str "\"" etag "\"")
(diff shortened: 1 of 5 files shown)


NoahTheDuke:
Should we remove the `"/data/cards/version"` endpoint too? Clean up the api a little?

src/clj/web/utils.clj
```diff
@@ -25,6 +25,22 @@
...
 
+(defn cached-response
+  "Let the browser cache a response in its HTTP cache, keyed on etag.
+  Cache-Control no-cache means the browser must revalidate with If-None-Match
+  before using the cached copy, and a 304 response does not need to send a body.
+  See https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cache-Control#up-to-date_contents_always
+  and https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/If-None-Match"
+  [request etag body-fn]
+  (let [etag (str "\"" etag "\"")
+        req-etag (get-in request [:headers "if-none-match"])
```
NoahTheDuke, line 36: is ring smart enough to not convert the headers to kebab-case keywords? that's great news. i don't mess with them too much to know for certain.

NoahTheDuke:
looking good

src/clj/web/utils.clj
filipesilva: TBH I never know, it's always deep inside some middleware chain. What I can say here is that I verified the header gets sent back, is identified as matching, and we reply with a 304
CleanShot 2026-07-28 at 15 04 56@2x

PR #8815, opened 2026-07-28
fix: validate custom chat messages length

### LINKED ISSUE #8811: Custom messages limit enforced only on FE
So if _someone_ were to edit the req, they could jam a bigger message there.

Image

filipesilva:
Fix https://github.com/mtgred/netrunner/issues/8811

### DIFF
--- a/src/cljc/jinteki/settings.cljc
+++ b/src/cljc/jinteki/settings.cljc
@@ -46,6 +46,8 @@
 (def default-chat-messages
   ["Good luck & have fun!" "Thinking..." "Good game!" "Thank you for the game!" "Yes" "No"])
 
+(def chat-message-max-length 100)
+
 ;; Validation combinators
 (defn- validate-coll-of
   "Returns a validator that checks if value is a collection of items matching pred"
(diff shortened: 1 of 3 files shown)


src/cljc/jinteki/settings.cljc
```diff
@@ -107,8 +109,12 @@
   "Validates default-decks is a map of side -> {format -> deck-id-string}"
   (validate-map-of keyword? (validate-map-of keyword? string?)))
 
-(def validate-chat-messages
-  (validate-coll-of string? vector?))
+(defn validate-chat-messages
+  [value]
+  (and (vector? value)
+       (<= (count value) (count default-chat-messages))
```
NoahTheDuke, line 115: what's this for?

filipesilva: So it's always 6 messages total. It's not a super important check tbh, I've added/removed it a few times already from my dev branches.

But in view of this particular issue, in the same way you could craft a large single message, you could instead do a large amount of smaller messages and achieve the same result, as far as inserting data that's too big goes. It wouldn't be as bad insofar as instant chat messages go though.

PR #8820, opened 2026-08-01
rfct: scroll inner-left-leftpane instead of just prompts

filipesilva:
Should help when there's a lot of extra panels.

### DIFF
--- a/src/css/gameboard.styl
+++ b/src/css/gameboard.styl
@@ -707,6 +707,12 @@ glow-spin(prop, pct)
                 flex-direction(column)
                 row-gap: 5px
 
+                // scroll instead of letting other panels crowd out the prompt,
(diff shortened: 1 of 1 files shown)


filipesilva:
Here's an example of how things are currently on master after https://github.com/mtgred/netrunner/pull/8783. If you are in a game state with a couple of panels already, you'll get a tiny scroll box for the cultivate:
CleanShot 2026-08-01 at 23 48 07@2x

With this change, you get a scroll bar on the whole thing instead:
CleanShot 2026-08-01 at 23 49 27@2x
CleanShot 2026-08-01 at 23 49 56@2x

No scrollbar when there's nothing to scroll:
CleanShot 2026-08-02 at 00 01 03@2x

NoahTheDuke:
oh this is nice

PR #8828, opened 2026-08-14
Fix sometimes broken scrolling on firefox

### LINKED ISSUE #8824: Chat Scrolling Broken By "Scroll to Bottom"
Sometimes chat gets in a state where I cannot scroll up with mouse wheel OR by dragging the scroll bar, it causes chat to flicker instead of scrolling.
it seems to be the appearance of the button causes the cat window to resize and reset the scroll to the bottom.

For some users they cannot scroll up in chat at all, sometimes you get lucky and it 'just works' for one frame before occuring again.

https://github.com/user-attachments/assets/649a68c2-9b40-47f3-8619-a6c59825aaa5

filipesilva:
Best reviewed commit by commit.

This was a weird one. I was able to reproduce the flickering effect on firefox with about 300 messages only when scrolling by click hold on the scroll bar and dragging. It seemed to be due to how the button appears under the message log, and thus moves it up, messing with the logic to show it.

I wasn't able to reproduce being unable to scroll up at all, but I was able to trigger a snapping behavior where the scroll bar snaps to the bottom when scrolling down. It's due to the interaction between the new statekeeping for the scroll button and the existing behavior where you should scroll to the bottom when getting a new message via the `:component-did-update` hook. Before the button was added, this only trigger when getting new messages. But then it also started triggering when scrolling because `scrolled-away-from-end?` changed. So scrolling itself could trigger a scroll to bottom.

With these changes the button is now just a floating down arrow (so it doesn't get added in a way that changes the message log size), auto-scroll being based on message counts (so scrolling can't trigger auto-scroll), and I simplified the logic around auto-scrolling down with the notion of being pinned to the bottom or not.

This is how it looks:

CleanShot 2026-08-14 at 21 01 41@2x

CleanShot 2026-08-14 at 21 01 36@2x

Fix https://github.com/mtgred/netrunner/issues/8824

### DIFF
--- a/src/cljs/nr/chat.cljs
+++ b/src/cljs/nr/chat.cljs
@@ -15,7 +15,8 @@
    [nr.news :refer [news]]
    [nr.translations :refer [tr tr-element tr-pronouns tr-span]]
    [nr.utils :refer [non-game-toast render-message format-date-time
-                     day-word-with-time-formatter tr-non-game-toast]]
+                     day-word-with-time-formatter tr-non-game-toast
+                     scroll-to-bottom!]]
    [nr.ws :as ws]
    [reagent.core :as r]))
 
@@ -119,8 +120,7 @@
                                    :msg       text
                                    :username  (:username user)
                                    :emailhash (:emailhash user)}])
-          (let [msg-list (:message-list @chat-state)]
-            (set! (.-scrollTop msg-list) (+ (.-scrollHeight msg-list) 500)))
+          (scroll-to-bottom! (:message-list @chat-state))
           (swap! s assoc :msg "")
           (.focus input))))))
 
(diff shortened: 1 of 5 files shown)


src/cljs/nr/chat.cljs
```diff
@@ -119,8 +120,7 @@
                                    :msg       text
                                    :username  (:username user)
                                    :emailhash (:emailhash user)}])
-          (let [msg-list (:message-list @chat-state)]
-            (set! (.-scrollTop msg-list) (+ (.-scrollHeight msg-list) 500)))
```
filipesilva, line 123: I _think_ this `+500` wasn't doing anything? Maybe it's needed for some case I don't understand. But didn't seem to be needed from click testing.

src/css/gameboard.styl
```diff
@@ -1182,13 +1182,18 @@ glow-spin(prop, pct)
                         div
                             animation: fadein 0.5s
 
+                    .log-scroll-to-bottom-wrapper
```
filipesilva, line 1185: Needed to hold the absolutely positioned scroll-to-bottom button while the message log scrolls.

src/cljs/nr/chat.cljs
NoahTheDuke: i think it's vestigial. i added it to attempt to not snap to the bottom when you were scrolled to the bottom or very close to it but it wasn't great.

PR #8831, opened 2026-08-15
Update i18n\zh-simp.ftl

klingeling:


resources/public/i18n/en.ftl
```diff
@@ -1644,7 +1645,7 @@ settings_card-colors-selectable = Selectable / playable
 
 settings_card-colors-selected = Selected
 
-settings_card-iamge = Card Image
+settings_card-image = Card Image
```
NoahTheDuke, line 1648: omg thank you lol

PR #8835, opened 2026-08-18
Move :last-update to a standalone ConcurrentHashMap

NoahTheDuke:
We've been updating the `app-state` atom every single time _anyone_ touches anything in a lobby or game. I think this has been driving contention on the atom, so this is the first step towards pulling out some of the more contentious pieces into separate, better pieces that we can update or check without fighting everyone else. ConcurrentHashMap is fast and built to handle loads like this, so it's ideal.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -18,7 +18,7 @@
                  :init-ns web.dev
                  :init (go)}
 
-  :dependencies [[org.clojure/clojure "1.12.4"]
+  :dependencies [[org.clojure/clojure "1.12.5"]
                  [org.clojure/clojurescript "1.11.132"
                   :exclusions [org.clojure/google-closure-library
                                org.clojure/data.json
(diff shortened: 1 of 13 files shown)


project.clj
```diff
@@ -29,13 +29,13 @@
...
-                 [ring/ring-core "1.9.4"]
-                 [ring/ring-devel "1.9.4" :exclusions [org.clojure/java.classpath]]
-                 [ring/ring-anti-forgery "1.3.0"]
+                 [com.taoensso/tufte "3.1.0"]
+                 [ring/ring-core "1.15.5"]
+                 [ring/ring-devel "1.15.5"]
+                 [ring/ring-anti-forgery "1.4.0"]
                  [ring/ring-json "0.5.1"]
                  [ring-cors "0.1.13"]
-                 [compojure "1.6.2"]
```
NoahTheDuke, line 37: we haven't used compojure in multiple years

project.clj
```diff
@@ -49,15 +49,14 @@
                  [http-kit "2.7.0"]
                  [com.draines/postal "2.0.5"]
                  [throttler "1.0.1"]
-                 [clj-http "3.12.3"]
```
NoahTheDuke, line 52: we haven't used clj-http in multiple years

src/clj/web/lobby.clj
```diff
@@ -438,12 +438,6 @@
   (or (player? uid lobby)
       (spectator? uid lobby)))
 
-(defn handle-set-last-update [lobbies gameid uid]
```
filipesilva, line 441: There's a couple of `lobby/handle-set-last-update` still in `~/repos/personal/netrunner/src/clj/web/angel_arena.clj`.

src/clj/web/game.clj
```diff
@@ -420,12 +411,13 @@
     {:keys [gameid msg]} :?data
     id :id
     timestamp :timestamp}]
-  (let [new-app-state (swap! app-state/app-state update :lobbies lobby/handle-set-last-update gameid uid)
+  (let [new-app-state (app-state/get-lobby gameid)
```
filipesilva: This doesn't look right, `app-state/get-lobby` doesn't return a full app-state I think?

src/clj/web/lobby.clj
NoahTheDuke: angel arena is bitrotted and currently unused. i don't think any code references it or calls into it. we could clean it up or delete it, but i haven't spent time updating these calls

src/clj/web/game.clj
NoahTheDuke: shit good call

src/clj/web/lobby.clj
filipesilva: Removing it in https://github.com/mtgred/netrunner/pull/8839

NoahTheDuke:
thinking about it more, we should probably set up redis and use it for this and other similar stuff (caching of which lobbies are occupied and by who, etc).

PR #8838, opened 2026-08-18
fix: runner replay pruning

filipesilva:
I think `(get-in @state [:corp :runner])` is wrong and always nil here, should be `(get-in @state [:runner :user]))`.


NoahTheDuke:
omg, we're probably holding gigs of replays we should delete. good catch

PR #8839, opened 2026-08-19
rfct: remove angel-arena code

filipesilva:
It's bitrotten and not used anymore.


NoahTheDuke:
killer idea, but the community didn't like it so much... rip in pepperonis

PR #8840, opened 2026-08-19
Optimize end-of-game queries

filipesilva:
There's two things that happen at the end of a game that can be pretty heavy on mongo.

First is deleting old replays. There's no index for fields, so if a user has a lot of games the query will scan them all. I added indexes to this query to make fetch less docs.

Second is the game history. This is also a pretty heavy, since mongo will still read the documents from disk even though the query has `{:replay 0 :log 0}`. This isn't a big problem since the stats page isn't that common to visit. But it's also fetched at the game end, for no real purpose since going to the stats page fetches it again. I removed the game end fetch. 

Note that deployments will need to call `lein create-indexes` to create the new indexes.

### DIFF
--- a/src/cljs/nr/stats.cljs
+++ b/src/cljs/nr/stats.cljs
@@ -32,8 +32,7 @@
 
 (defmethod ws/event-msg-handler :stats/update [{{:keys [userstats deck-id deckstats]} :?data}]
   (swap! app-state assoc :stats userstats)
-  (update-deck-stats deck-id deckstats)
-  (fetch-game-history))
+  (update-deck-stats deck-id deckstats))
 
 (defn- set-replay-shared [state gameid shared?]
   (swap! state (fn [s]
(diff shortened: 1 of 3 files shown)


src/cljs/nr/stats.cljs
```diff
@@ -32,8 +32,7 @@
 
 (defmethod ws/event-msg-handler :stats/update [{{:keys [userstats deck-id deckstats]} :?data}]
   (swap! app-state assoc :stats userstats)
-  (update-deck-stats deck-id deckstats)
-  (fetch-game-history))
```
NoahTheDuke, line 36: don't need this?

filipesilva: As far as I can tell no, because the stats route handler just fetches it again anyway.

```
(defn stats []
  (let [stats (r/cursor app-state [:stats])
        list-scroll-top (atom 0)
        log-scroll-top (atom 0)]
    (fetch-game-history)
    (fn []
      [:div.page-container
       [:div.stats-bg]
       [:div.lobby.panel.blue-shade
        [left-panel state stats]
        [right-panel state list-scroll-top log-scroll-top]]])))
```

NoahTheDuke: wtf is that component...

PR #8844, opened 2026-08-22
Clean up logging

NoahTheDuke:
Initialize in system so we can reset with `(restart)`, allow the log directories to be set in `dev.edn`/`prod.edn`. Some other small clean ups.

### DIFF
--- a/resources/dev.edn
+++ b/resources/dev.edn
@@ -52,4 +52,6 @@
                     :mongo #ig/ref :mongodb/connection}
  :sente/router {:ws #ig/ref :web/ws}
  :game/quotes nil
+ :jinteki/cards-version {:initial "1"
+                         :mongo #ig/ref :mongodb/connection}
  :jinteki/cards {:mongo #ig/ref :mongodb/connection}}
--- a/resources/dev.edn
+++ b/resources/dev.edn
@@ -4,6 +4,16 @@
                       :name "netrunner"
                       ; Monger requires a db name in the URI string, be sure to specify
                       :connection-string #or [#env MONGO_CONNECTION_URI nil]}
+ :logging/timbre {:appenders
+                  ;; set appender args:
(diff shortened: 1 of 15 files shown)


NoahTheDuke:
i've rebased this on top of #8852

src/clj/web/system.clj
```diff
@@ -155,6 +151,13 @@
                (transient {}))
               (persistent!))))
 
+(defmethod ig/init-key :jinteki/cards-version [_ {initial :initial
+                                                  {:keys [db]} :mongo}]
+  (if-let [version (:cards-version (mc/find-one-as-map db "config" nil))]
+    (reset! frontend-version version)
```
filipesilva: I think this reset should be `cards-version` instead, and the below too.

resources/dev.edn
```diff
@@ -4,6 +4,16 @@
                       :name "netrunner"
                       ; Monger requires a db name in the URI string, be sure to specify
                       :connection-string #or [#env MONGO_CONNECTION_URI nil]}
+ :logging/timbre {:appenders
+                  ;; set appender args:
+                  ;; `:enabled?` sets whether an appender should run
+                  ;; `:log-path` defaults to `~/logs/jinteki-*.log`
+                  ;;
+                  ;; the `:println` appender should be disabled in prod
+                  {:println {:enabled? false}
```
filipesilva: This one confuses me a bit. It says that the println should be disabled in prod, but it's disabled in dev too. I expected it to be enabled in dev, but not in prod or something.

filipesilva (approved):

resources/dev.edn
NoahTheDuke: ope thank you

src/clj/web/system.clj
NoahTheDuke: gdi lol

PR #8850, opened 2026-08-29
fix: read insts as cljc.java-time.instant

filipesilva:
Followup to https://github.com/mtgred/netrunner/pull/8847

### DIFF
--- a/src/cljc/jinteki/msgpack_ext.cljc
+++ b/src/cljc/jinteki/msgpack_ext.cljc
@@ -5,7 +5,7 @@
    #?@(:cljs [[cljc.java-time.instant :as inst]
               [cljs.reader :as reader]
               [java.time :refer [LocalDateTime]]
-              ;; loads taoensso.msgpack.impl to override -1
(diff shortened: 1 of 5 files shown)


NoahTheDuke:
this is probably correct. `-1` is strange, is that part of the msgpack documentation?

filipesilva:
I used the same number as sente's msgpack implementation for `java.time.Instant`:
https://github.com/taoensso/sente/blob/1442b8b9becf9a94a159f0f15f30ad3bda61ea80/src/taoensso/msgpack/impl.clj#L456-L464

I didn't find documentation there though. Sente's code is adapted from https://github.com/rosejn/msgpack-cljc, and the documentation there uses `extend-msgpack` in the example to add support for `java.time.Instant` instead of the `extend-packable` in sente.

I tried to only extend the unpack instead of the pack+unpack, but in hindsight that makes it harder to reference against the source. Pushed a commit that makes it more explicit and easier to trace back to the source.

PR #8852, opened 2026-08-30
use web.game/update-and-send-diffs! for all test actions

NoahTheDuke:
this necessarily increases the execution time of tests, but it helps validate that we're actually adhering to the web.game apis, and checks diff generation/history saving in every functional test.


src/clj/game/core/diffs.clj
```diff
@@ -296,34 +296,42 @@
    :agenda-point
    :agenda-point-req])
 
+(defn identity-summary [card state side]
+  (card-summary card state side))
+
+(defn basic-summary [card state side]
```
filipesilva: These two look the same. Maybe you wanted to do something else here but then forgot?

src/clj/game/core/servers.clj
```diff
@@ -67,7 +67,7 @@
 
 (defn zones->sorted-names
   [zones]
-  (->> zones (sort-by zone->sort-key) (map zone->name)))
+  (->> zones (sort-by (memoize zone->sort-key)) (mapv zone->name)))
```
filipesilva: Does this work? I think this is making a new memoized fn each time. I'm used to doing it like this:

```
(def zone->sort-key-memo (memoize zone->sort-key))

(defn zones->sorted-names
  [zones]
  (->> zones (sort-by zone->sort-key-memo) (mapv zone->name)))
```

filipesilva (approved):

src/clj/game/core/diffs.clj
NoahTheDuke: oops. let me revert the whole file, this is unrelated to the intended pr

src/clj/game/core/servers.clj
NoahTheDuke: i was thinking that for a given set of zones, we only want to memoize for that sorting, but on second thought, we won't ever have more than... 20 remote servers across all games, right? even if we get up to 100,000 remotes (someone fucks around), that's only 100,003 entries, and they'll apply globally.

okay, you've convinced me.

Issue #8862, opened 2026-09-12
Bug Report: improper player name rendering under certain circumstances


MirrorCubeSquare:
If a player's name includes card name(s)
under certain circumstances, jinteki will render card-name part as link to card, rather then display them as-is
this cause some funny display when display language is non-english

Scene include:
1) when that player spectates a game
2) when players swap sides and enter the game
3) when one player leave the game , and other player/spectator refreshes the page
  
minimal recreation for all scenes(suppose all players/specs have buggy nickname)
1) player A, B enter the lobby, C spectates
2) A and B swap sides
3) enter the game , A and C leave the lobby
4) B refresh the page

Sample Screenshot(card name is simplified chinese):
<img width="271" height="40" alt="Image" src="https://github.com/user-attachments/assets/198dab0a-7db9-4e1f-bf2d-2528fc2dd55d" />

MirrorCubeSquare:
i have written a fix for the bug, and tested on Chinese clone of jinteki
i'll make a pr asap

PR #8863, opened 2026-09-12
Keep usernames literal in system messages

MirrorCubeSquare:
Usernames containing card names can be parsed as card links or translated card text in system messages. For example, `MirrorCubeSquare` can have its `Mirror` prefix replaced in spectator notifications or repeated names in side-swap messages. Icon and special-message tokens inside a username can also be interpreted as formatting.

Keep usernames separate from renderable text using structured system-message parts. Only message text goes through card, icon, special-code, and pronoun processing. Apply this to lobby events, side swaps, commands, game notifications, and saved game logs. Retain the flattened `:text` field and the existing rendering fallback for older messages and replays.

Preserve player highlighting and timestamps, including after a player leaves. Explicitly mark spectator names before removing them from the lobby so their join/leave messages remain plain and have no player timestamp.

Validation on current `master` (`d94d18cdc`):
- Backend: `game.core.say-test`, `web.lobby-test`, and `web.game-test` — 13 tests, 279 assertions, no failures or errors.
- Browser: `nr.utils-test` — 5 tests, 117 assertions, all passing in Chromium. Covers translated card names, icon/special tokens, HTML escaping, timestamps, spectator styling, departed players, and legacy messages.
- Shadow CLJS `release` of `:app` — successful, zero compiler warnings.
- `git diff --check` — clean.

### DIFF
--- a/src/clj/game/core.clj
+++ b/src/clj/game/core.clj
@@ -714,11 +714,13 @@
    indicate-action
    make-message
    make-system-message
+   make-system-message-parts
    play-sfx
    say
    system-msg
    implementation-msg
-   system-say])
+   system-say
+   system-say-parts])
 
 (expose-vars
   [game.core.servers
--- a/src/clj/game/core/process_actions.clj
+++ b/src/clj/game/core/process_actions.clj
@@ -16,7 +16,7 @@
    [game.core.rezzing :refer [derez rez]]
    [game.core.runs :refer [check-for-empty-server continue handle-end-run
                            jack-out start-next-phase toggle-auto-no-action]]
-   [game.core.say :refer [indicate-action say system-msg system-say]]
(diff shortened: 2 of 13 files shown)


MirrorCubeSquare:
This is a bugfix for issue #8862 
After fix it shoud be like the following screen shot (this is taken from my testbench server)
under current version 
MirrorCubeSquare will be rendered as"镜子CubeSquare"
and F2PDrafter will be rendered as "免付费内容建筑师"
微信图片_20260906090919_167_153
微信图片_20260906090825_166_153

NoahTheDuke:
this is a great idea and i've considered something similar for a while (like card names or symbols). i'm gonna wait to review/merge this until i've merged my i18n message PR, because these both touch a lot of the same code paths.

PR #8866, opened 2026-09-18
Use buttons for credit prompts

NBKelly:
Adds the ability to use buttons for credit source selection prompts.
This should make Biawak more intuitive, but without taking away agency.

The shift-key functionality is still preserved.

I want to refactor this to not need to add info to the prompt state though.


NBKelly:
I did the refactor, this now only touches the frontend and doesn't bloat the network (but adds a little bit of crunch on the prompt only)

PR #8867, opened 2026-09-19
reorder EH to be N(ice)EH

NBKelly:
The "pay 3" will be at the same position on both prompts.


filipesilva (approved):
This came up a few times at Nottingham lol!

